PR #106, opened 2016-07-06
More complex uses of Need* artitecture need to guard against redundant calls with the same source

jeffschwMSFT:
As I explored more complex uses of NeedProcesses and NeedLoadedDotNetRuntimes I found that the check for when to add callbacks was too aggressive.  I built a NeedThreadPool* extension for CAP and was receiving mulitiple JIT callbacks because for a given source callbacks were added mulitple times.  For more complex scenarios NeedLoadedDotNetRuntimes may be called multiple times witht the same source.

Tested with the more complex case and PerfView scenarios.

Note this case is not hit in PerfView, but as more complex external scenarios are built it is an important scenario.


vancem:
LGTM

PR #114, opened 2016-08-29
Make it possible to modify payloads of events with ETWTraceEventSource

daklimek:
This can be useful if you need to remove PII from events.

There was code in ETWReloggerTraceEventSource that already did this, so I uncommented it. There were some issues with code in that were easily fixed:
-The temporary buffer (m_scratchBuffer) would need to expand sometimes and the bytes weren't copied from the old to the new copy so garbage was getting sent in the payload.
-curBlobPtr wasn't getting set correctly after writing a string to the buffer so payloads written after that were getting written to incorrect locations.
-SetPayload didn't know how to serialize Guids and Enums which I needed so I added support for these two types.

For testing, I added a unit test that relogs through a small 2KB ETL trace with 1 event, replaces the full path of a payload with the file name in the generated etl file. It then verifies the full path was actually removed. It takes 500ms to run.


daklimek:
We now don't currently need this feature. Closing for now.

PR #118, opened 2016-09-19
Fixed links

shchahrykovich:


vancem:
Thanks for finding this.  I did a quick scan to see if there were other places where SupportDlls was used and you seemed to have found the only instances.   Thanks

PR #120, opened 2016-09-20
Show exception handling call stacks as a seperate activity

weng5e:
Show exception handling call stacks as a seperate activity.
This can help users notice the time spent in the exception handling logic.


vancem:
LGTM

PR #122, opened 2016-09-21
Added BGC events handling

Maoni0:
@swgillespie @adityamandaleeka @vancem this is mostly for processing the revisited pages so we can use it in our perf analysis


vancem:
FYI @jeffschwMSFT

PR #123, opened 2016-09-21
Refactor of the ETLStackBrowse\ZL - 

lukasz-pyrzyk:
typos, redundant code, auto properties, null-propagation operator, coding style etc


vancem:
Thank you for your contribution.    As mentioned in the contribution guide the bigger the change the more careful we need to be about it.    The problem is that this that many non-semantic changes (formatting) have been combined with more interesting changes (coding style etc).   I would like these to be separated.   Frist submit a new pull request that JUST does the automatic reformatting.  I will approve that immediately.   Then you can merge your current pull request with that and we can see what else was done without the noise of the other uninteresting changes.

On a larger note, the whole ETLStackBrowse subdirectory is likely to be discarded (It has been on the chopping block (along with the ZIP directory) for years now, only its low priority is saving it).   The ETLSTackBrowse code dates to a time when PerfView's own ETW parsing was suspect and we needed a way of looking at trace data from XPERF's CSV files.    It has not been useful for years.   Thus any investment here is dubious,   Thus I want to minimize the time I spend on any changes here. 

Alternatively, we can just close the pull request, and either do nothing, or actually rip out the code (which is probably the better answer).

lukasz-pyrzyk:
Sorry, unfortunately i dont have enough time to split changes to the smaller pieces. Let's close this pr

PR #125, opened 2016-09-26
Added NoWait option

shchahrykovich:
Added command line argument for closing PerfView immediately after the execution

### DIFF
--- a/src/PerfView/App.cs
+++ b/src/PerfView/App.cs
@@ -83,8 +83,8 @@ namespace PerfView
 
             // If we created a new console (collect command), prompt before closing it so the user has a chance to look at it.
             if (newConsoleCreated)
-            {
-                if (newConsoleCreated)
+            {                
+                if (!CommandLineArgs.NoWait)
                 {
                     Console.WriteLine("Press enter to close window.");
                     Console.ReadLine();
--- a/src/PerfView/CommandLineArgs.cs
+++ b/src/PerfView/CommandLineArgs.cs
@@ -183,6 +183,7 @@ namespace PerfView
 
         // Collect options
         public bool NoGui;
+        public bool NoWait;
         public int CollectMultiple;     // Collect several instances (incrementing the file name)
 
         // Mark options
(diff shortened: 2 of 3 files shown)


vancem:
Can you describe the scenario for this no-wait option?    Generally speaking you should be using the /logFile=XXXX option (which does not launch a console or therefore wait to make sure you have seen the console.    

Vance

shchahrykovich:
For some reason /LogFile option does create a new console window and I have to press enter at the end.
I might be wrong, but the code [here](https://github.com/Microsoft/perfview/blob/master/src/PerfView/App.cs#L124) sets NoGui to true.

vancem:
/LogFile is not intended to create a new console window (I am assuming you don't want one, which is why you want it to auto-close).     If you use /logFile PerfView should not accept any input and should complete.  Is that what happens.   

What is the scenario you actually want to enable?

shchahrykovich:
`PerfView.exe collect test.etl -Merge:True -Zip:True -NoNGenRundown -AcceptEULA -LogFile:test.txt` creates a new window and shows 'press enter to close the window'. PerfView version: 1.9.0.0

My goal is to use PerfView in automated scenarios. Right now PerfView always asks for a confirmation("press enter to close the window") after execution.

Is there a way how I can get rid of ["press enter to close the window"](https://github.com/Microsoft/perfview/blob/master/src/PerfView/App.cs#L89) message after execution?

vancem:
The normal expectation for the scenario (as mentioned in the docs on automating PerfView collection), is to  either

1. Use the /MaxCollectionSec:XX possibly combined with other trigger mechanism (e.g. /StopOn*) to avoid the need to tell PerfView when to stop (if it does not need input, it will not make the console).

OR

2. Use the start and stop operators (thus two commands) rather than the collect operator.    

See 'Automating Collection' in the users guide for more.

Does that solve your problem?    Generating the console is something you pretty much DON'T want when you automate.  it only exists so that for systems that don't support the GUI you can still do collection 'by hand'.

shchahrykovich:
Looks like I'm the only one who has this problem :) So it doesn't make sense to include this change in the main code base.

PR #129, opened 2016-10-13
Update the MemoryProcessMemInfoTraceData class so it can parse V2 of the event

daklimek:
Microsoft-Windows-Kernel-Memory/MemoryProcessMemInfo was showing bogus payloads because the additional payloads were added to the MemoryProcessMemInfoValues structure in Version2 and PerfView only knows how to parse the payloads from Version1. This has been happening for at least several months but we need these events for a metric we want to report in our perf tests so I'd like to get this fixed. This change makes it so TraceEvent can parse events from either version 1 or version 2.


daklimek:
@vancem, please let me know what you think.

vancem:
LGTM

Issue #108, opened 2016-08-03
GcStats failure to open on certain traces


swgillespie:
I have a trace on my machine that fails to open GcStats, failing with this stack trace:

```
Started: Opening GCStats
Completed: Opening GCStats   (Elapsed Time: 2.541 sec)
Exception Occurred: System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Stats.GcStats.PrintCondemnedReasonsToHtml(TraceGC gc) in E:\GitHub\perfview\src\PerfView\GcStats.cs:line 900
   at Stats.GcStats.PrintEventCondemnedReasonsTable(TextWriter writer, TraceProcess stats, TraceLoadedDotNetRuntime runtime, Int32 start, Predicate`1 filter) in E:\GitHub\perfview\src\PerfView\GcStats.cs:line 805
   at Stats.GcStats.ToHtml(TextWriter writer, TraceProcess stats, TraceLoadedDotNetRuntime runtime, String fileName, Boolean doServerGCReport) in E:\GitHub\perfview\src\PerfView\GcStats.cs:line 177
   at Stats.ClrStats.ToHtml(TextWriter writer, List`1 perProc, String fileName, String title, ReportType type, Boolean justBody, Boolean doServerGCReport) in E:\GitHub\perfview\src\PerfView\ClrStats.cs:line 77
   at PerfView.PerfViewGCStats.WriteHtmlBody(TraceLog dataFile, TextWriter writer, String fileName, TextWriter log) in E:\GitHub\perfview\src\PerfView\PerfViewData.cs:line 2045
   at PerfView.PerfViewHtmlReport.<>c__DisplayClass16_0.<Open>b__0() in E:\GitHub\perfview\src\PerfView\PerfViewData.cs:line 952
   at PerfView.StatusBar.<>c__DisplayClass19_0.<StartWork>b__0(Object <state>) in E:\GitHub\perfview\src\PerfView\GuiUtilities\StatusBar\StatusBar.xaml.cs:line 216
An exceptional condition occurred, see log for details.
```

I've checked this repros on PerfView built from master and doesn't repro on the PerfView currently hosted at `\\clrmain\tools\PerfView.exe`. I doubt I'm allowed to go posting the trace that I have that causes this to repro but I'd be happy to share it privately. I can take a look at this a little later in the week to see if it's an easy fix.

swgillespie:
On a debug build:

```
---- DEBUG ASSERTION FAILED ----
[...]

PR #130, opened 2016-10-14
Fix for #108 and similar code path

jeffschwMSFT:
Two similar 'used before initialized' crashes.  IsNotCompacting fails if GlobalHistory event was not seen for this GC - checked for null.  GetCondemnedReasons fails if the length of the CondemnedReasons does not include details even for the first heap - added a check to exit gracefully.

Fix for #108


vancem:
LGTM

PR #131, opened 2016-10-14
Add strongly typed TraceEvent for Microsoft-Windows-Kernel-Memory/MemInfo event

daklimek:
This PR adds an event handler for the the Microsoft-Windows-Kernel-Memory/MemInfo event. Without this, the only way I was able to successfully listen to this event was to subscribe to the Kernel.All event and look at only events with a particular TaskName (which is not cheap since a string needs to be allocated for every kernel event).

### DIFF
--- a/src/TraceEvent/Parsers/KernelTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/KernelTraceEventParser.cs
@@ -2048,6 +2048,22 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
             }
         }
 
+        /// <summary>
+        /// Rasied every 0.5s with memory metrics of the current machine.
+        /// </summary>
+        public event Action<MemInfoTraceData> MemoryMemInfo
+        {
+            add
+            {
+                // action, eventid, taskid, taskName, taskGuid, opcode, opcodeName, providerGuid, providerName
+                source.RegisterEventTemplate(new MemInfoTraceData(value, 1, 1, "MemInfo", Guid.Empty, 0, "", MemoryProviderGuid, "Microsoft-Windows-Kernel-Memory", State));
+            }
+            remove
+            {
+                source.UnregisterEventTemplate(value, 1, MemoryProviderGuid);
+            }
+        }
+
         // TODO Added by hand without proper body decode.  
         public event Action<EmptyTraceData> MemoryPFMappedSectionCreate
         {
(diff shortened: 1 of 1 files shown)


daklimek:
@vancem, please let me know what you think.

src/TraceEvent/Parsers/KernelTraceEventParser.cs
```diff
@@ -2048,6 +2048,22 @@ public string FileIDToFileName(Address fileKey)
             }
         }
 
+        /// <summary>
+        /// Rasied every 0.5s with memory metrics of the current machine.
```
data-nerd: typo: Raised

src/TraceEvent/Parsers/KernelTraceEventParser.cs
```diff
@@ -2976,71 +2992,72 @@ protected internal override void EnumerateTemplates(Func<string, string, EventFi
...
-                templates[180] = new VirtualAllocTraceData(null, 0xFFFF, 0, "VirtualMem", VirtualAllocTaskGuid, 99, "Free", ProviderGuid, ProviderName, null);
-                templates[181] = new DispatcherReadyThreadTraceData(null, 0xFFFF, 0, "Dispatcher", ReadyThreadTaskGuid, 50, "ReadyThread", ProviderGuid, ProviderName, null);
-                templates[182] = new StringTraceData(null, 0xFFFF, 0, "PerfInfo", PerfInfoTaskGuid, 34, "Mark", ProviderGuid, ProviderName, false);
-                templates[183] = new BuildInfoTraceData(null, 0xFFFF, 0, "SysConfig", SysConfigTaskGuid, 32, "BuildInfo", ProviderGuid, ProviderName);
-                templates[184] = new VolumeMappingTraceData(null, 0xFFFF, 0, "SysConfig", SysConfigTaskGuid, 35, "VolumeMapping", ProviderGuid, ProviderName);
-                templates[185] = new StringTraceData(null, 0xFFFF, 0, "SysConfig", SysConfigTaskGuid, 34, "UnknownVolume", ProviderGuid, ProviderName, true);
+                templates[127] = new MemInfoTraceData(null, 1, 1, "MemInfo", Guid.Empty, 0, "", MemoryProviderGuid, "Microsoft-Windows-Kernel-Memory", State);
```
vancem: I am OK with you putting this in the right place, but in general I would just put the new item at the end to avoid all the renumbering.   The order does not matter.

vancem:
David,  I am assuming that the new event is per-machine (that is the page counts are for the entire machine).   

Does this event get parsed by the RegisteredTraceEventParser?   For kernel events it can go either way.   

Anyway I am generally OK with having strongly typed templates, especially for kernel events (which are classic ETW which as you have discovered are a pain, if someone does not provide a template).

src/TraceEvent/Parsers/KernelTraceEventParser.cs
daklimek: OK. I'll remember to that if I need to make a change like this again.

src/TraceEvent/Parsers/KernelTraceEventParser.cs
daklimek: Next time I make a PR, I'll go and clean this up.

daklimek:
This event is indeed raised per machine (not per process). 

I've never used RegisteredTraceEventParser before, but I'll give that a try now and reply here with what I find.

daklimek:
It actually does work so I could have done this instead.

            using (var session = new ETWTraceEventSource(new [] { traceName }))
            {
                session.Dynamic.AddCallbackForProviderEvent("Microsoft-Windows-Kernel-Memory", "MemInfo", delegate (TraceEvent e)
                {
                    Console.WriteLine(e);
                });
                session.Process();
            }

PR #140, opened 2016-11-13
Improve readability of GCStats and other views

adityamandaleeka:
As part of some GC work I've been doing lately, I've been using the GCStats view quite a lot. While using it, I noticed a few things that could be improved about how PerfView generates its HTML views and thought I'd send a PR to address them. I'm not a web/UI dev, so it's pretty minimal stuff and there's nothing fancy.

This change adds some very basic styling to the generated views like GCStats, Processes, and the others that rely on table-heavy HTML.

Main changes:
- Table data is now in a monospaced font, which makes visually parsing numeric values much easier.
- As you hover over a table cell, the entire row gets highlighted. This is really helpful for horizontal tracking (especially when you're looking at a screen full of numbers representing each GC in the GCStats view).
- The unsightly detached (default) borders are all collapsed.
- Some miscellaneous typos were fixed.

I kept the CSS inline so as to avoid adding any complexity to how the file is generated, packaged, or reasoned about.

### DIFF
--- a/src/PerfView/GcStats.cs
+++ b/src/PerfView/GcStats.cs
@@ -26,7 +26,7 @@ namespace Stats
     {
         public static void ToHtml(TextWriter writer, TraceProcess stats, TraceLoadedDotNetRuntime runtime,  string fileName, bool doServerGCReport = false)
         {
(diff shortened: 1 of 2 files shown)


adityamandaleeka:
@vancem Please take a look and let me know what you think.

vancem:
I am OK with this.  I merged it.

adityamandaleeka:
Thanks Vance. I expect to have another PR soon with some more improvements.

PR #144, opened 2016-11-18
Add trigger to stop collection on long BGC final pause

adityamandaleeka:
This change adds a command line option to stop collection when a background GC is encountered whose final pause takes longer than the specified number of milliseconds.

@Maoni0 PTAL

### DIFF
--- a/src/PerfView/CommandLineArgs.cs
+++ b/src/PerfView/CommandLineArgs.cs
@@ -246,7 +246,8 @@ namespace PerfView
             parser.DefineOptionalQualifier("StopOnGCSuspendOverMSec", ref StopOnGCSuspendOverMSec,
                 "Trigger a stop of a collect command if there is a .NET Garbage Collection (GC) where suspending for the GC took over the given number of MSec.");
             parser.DefineOptionalQualifier("StopOnBGCFinalPauseOverMsec", ref StopOnBGCFinalPauseOverMsec,
-               "Trigger a stop of a collect command if there is a background .NET Garbage Collection (GC) whose final pause is longer than the given number of MSec.");
+               "Trigger a stop of a collect command if there is a background .NET Garbage Collection (GC) whose final pause is longer than the given number of MSec. To work correctly, "+
+               "this requires that heap survival and movement tracking is not enabled.");
             parser.DefineOptionalQualifier("StopOnAppFabricOverMsec", ref StopOnAppFabricOverMsec,
                 "Trigger a stop of a collect command if there is a AppFabric request is longer than the given number of MSec.");
 
--- a/src/PerfView/CommandLineArgs.cs
+++ b/src/PerfView/CommandLineArgs.cs
@@ -79,7 +79,7 @@ namespace PerfView
 #endif
 
         // view options
-        public string Process;              // A process name to focus on.  
+        public string Process;              // A process name to focus on.
 
         // run Options 
         public string[] CommandAndArgs;     // This is broken up into words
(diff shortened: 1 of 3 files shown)


Maoni0:
LGTM!

vancem:
@adityamandaleeka Just so you know the /StopOnEtwEvent is very powerful, and can often be used to achieve what you are after without any code (just a very long command line).   I have not tried it for this particular example, but it seems like it should work (see StopOnEtwEvent in the help for details).   

I would be interested in seeing if this actually works.  If it does (or can be made to work) we have to decide whether this form of /StopOn is going to be useful to enough people (we could put it as an example in the help to raise its visibility).   If we believe only people on this team will care, maybe just having a batch script that does the proper /StopOnEtwEvent is a better approach.

Let me know how the experiment with /StopOnEtwEvent goes.  In the end if we think it is worth having this extra /StopOn* command line I don't have a big problem with it, but my inclination is to try to have a view 'generic' and 'popular' /StopOn* capabilities and use those.

Maoni0:
@vancem this works - Aditya already tried and verified that it works. I actually didn't how you would customize /StopOnEtwEvent to do this - in the past you've added things like /StopOnGCOverMSec so I figured if that needed its own command then this one would too. We actually wanted to do something more sophisticated but as long as you collect the trace without the heapmovements events we are fine with the implementation Aditya did. 

If we want to be perfectly accurate, what we really needed was to detect the SuspendEEStart/RestartEEStop event pair fired from the BGC threads (so we needed to recognize BGC threads to begin with), and ignore the 2nd pair fired from the same BGC to accommodate when people are using the default "PerfView collect" command. But we settled for just detecting the SuspendEEStart/RestartEEStop event pair with the specific reason field because if you specifically don't collect the heapmovement events you don't have the 2nd pair in a BGC. And there are other cases SuspendEE can be called for this reason but it's rare. So my questions are, is what Aditya did achievable with a customized /StopOnEtwEvent cmdline? and what about if we want to do what I described above? Is that also achievable?

adityamandaleeka:
I actually started typing a comment on another machine and was interrupted before I hit 'submit', but after reading @vancem's comment, I did find a set of parameters that can be passed to `/StopOnEtwEvent` (`StartStopId`, `TriggerMsec`, etc.) that in theory should do what this PR does in its current form. I have not successfully used it though (when I tried running with it, it failed to detect any of these events, but I suspect there may be a minor syntax error or something causing that).

However, I think what @Maoni0 described above about ignoring the 2nd Suspend/Restart from the same BGC is too complicated to do using the knobs for `StopOnEtwEvent`. @vancem can correct me if I'm wrong.

vancem:
Doing selection based on non-constant value (like what Maoni describes above) is not possible using /StopOnEtwEvent.    /StopOnGCOverMSec is there because it predates /StopOnEtwEvent AND it seems reasonably common that people might wish to do this.   It is not so clear people care about more specialized GC triggers.   

I don't really mind that much if we have /StopOnBGCFinalPauseOverMsec but if /StopOnEtwEvent is good enough I prefer that unless you think others are likely to use it.   If you find that /StopOnEtwEvent does not work (since Aditya did not get it working), just say so and I will merge it.

adityamandaleeka:
It looks like the problem was that specifying the full name of the `StopEvent` (`Microsoft-Windows-DotNETRuntime/GC/RestartEEStop`) doesn't work. Changing it to `GC/RestartEEStop` works.

I'm okay with closing this PR and opening one when we want to do something more complicated that requires more functionality. On the other hand, it may be better to do this as a separate switch now so that if/when we change how it works under the hood, partners/customers who use it can continue to use the same command line args. Not sure how big of a deal that is though...

@Maoni0 what do you think? I'm fine with either option.

vancem:
Using /StopOnEtwEvent is always ugly and fragile.   It is meant for 'ad hoc' things where it is for a rare investigation, or when only a small number of people (who can share a script) will ever need it. 

So it sounds like we have command line (which we can wrap into a script) that works (in the simple-minded case).  If we really believe we will want to tell more than one or two others to capture traces in this way, then we should check in the new qualifier.   Otherwise I think we should not.  Just let me know...

Maoni0:
I do think we would definitely use this with more than one or two customers so it would be nice to have a switch instead of the long cmdline. 

Also how much work would it be too add the logic to skip the 2nd pair of SuspendEE/RestartEE?

vancem:
I am merging what we have here (since I am out next week).   If Aditya can make the improvement that Maoni mentions above, it can be done with another PR.

adityamandaleeka:
Thanks @vancem

PR #146, opened 2016-11-23
EventCounter data visualization

cshung:
This change create a new context menu option on the EventWindow to allow EventCounter data to be visualized as a graph.

Currently, it shows only a plot of mean against time, we can potentially add more visualization for max/min/standard deviation data.

The charting capability is currently using SparrowChart from https://sparrowtoolkit.codeplex.com/, unfortunately the project is current discontinued, if we have better charting control for WPF we should definitely switch.

The option would show up like this:
image

The generated plot would look like this.
image

Comments welcomed!

### DIFF
--- a/src/PerfView/EventViewer/EventCounterWindow.xaml
+++ b/src/PerfView/EventViewer/EventCounterWindow.xaml
@@ -1,0 +1,21 @@
+﻿<Window x:Class="PerfView.EventCounterWindow"
+        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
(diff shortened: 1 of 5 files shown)


mattwarren:
> Comments welcomed!

As an avid user of PerfView, I think this is a **great** idea!!

vancem:
I realize I have let this languish.   We talked back in November and I have concerns about adding new dependencies, and doing the UI at least 'well enough' (a graph with NO interactivity, begs for more).    I am assuming at this point that there is no urgency to this PR at this point in time (having waited 6 months another week or two is OK.   I will see about tinkering/thinking about this in the next week or so and if two weeks go by it will be time for me to put up or shut up and give you guidance about how to proceed.

cshung:
Thanks for resurrecting the pull request - I am looking forward to it.

davidfowl:
@vancem we need this

cshung:
Closing the pull request as obsolete. The work to redo this is tracked by the issue [#839](https://github.com/Microsoft/perfview/issues/839).

PR #148, opened 2016-12-06
Make help text editors readonly

adityamandaleeka:
I noticed that "Command Line Help" and "User Command Help" open in editable controls, which is not useful. This change fixes that and also removes a file path that (I think) was accidentally included in a comment.


adityamandaleeka:
@vancem @swgillespie

swgillespie:
👍

PR #158, opened 2017-02-15
Added missing ALPC stack trace settings

zodiacon:
Added the missing ALPC event guids and types in SetStackTraceIds


vancem:
LGTM

PR #168, opened 2017-03-18
Fixed bug in setting ServerGcHeapHistories switchSpan durations relative to GC pause duration

ASethi77:
# Purpose
In the original `TraceLoadedDotNetRuntime` code for setting up ETW event callbacks, `_gc.GCEnd()` is called when a `GC/Stop` ETW event is observed. This function sets the last `GcWorkSpan` duration based on the GC pause time for a single GC. However, calling `_gc.GCEnd()` in the `GC/Stop` callback handler is incorrect because `GC/Stop` events are fired *before* the GC pause duration is known/set.

# Impact
In the case where CSwitch events were captured during a GC, the last `GcWorkSpan` stored in a `ServerGcHistory`'s `SwitchSpans` will have a negative duration.

# Fix
Move the call to `_gc.GCEnd()` from the `GC/Stop` callback to the `GC/RestartEEStop` callback, which actually sets `_gc.PauseDurationMSec` and is fired after `GC/Stop`.

# Testing
Prior to making my change, I ran the tests mentioned under "Testing and Contributing tests" in CONTRIBUTING. With a fresh clone of PerfView, I find that the test suite results in 20 tests passed, 6 tests skipped, and one test, `ETW_GeneralParsing_Basic` fails. I re-built and ran the test suite after adding in my change and got the same test results. I'm happy to add more info or screenshots of the test failure if needed.


vancem:
@ASethi77 - Can you comment on end user behavior before and after the change (e.g. in example X, before the change statistic Y say Z but it should be Z`.   The change fixes this to be Z`.  If you have an actual ETL file that you used for unit testing, please include it.  

@Maoni0  @jeffschwMSFT, can you take a look?   

Vance

Maoni0:
yes this is fine. GCEnd would only do something if it's a blocking GC. and since in GCRestartEEStop it's only called in blocking GC case it would have the same behavior except giving you the updated pause time.

jeffschwMSFT:
Thanks @ASethi77 for identifying this as an issue.  The challenge with moving the end marker to Restart is that Restart is not last event for concurrent GCs (foreground or background GCs).  Here is an example event stream for a background gc:
Event Name                                       	  Time MSec
Microsoft-Windows-DotNETRuntime/GC/Start         	937,636.014
Microsoft-Windows-DotNETRuntime/GC/RestartEEStart	937,642.167
Microsoft-Windows-DotNETRuntime/GC/RestartEEStop 	937,642.173
Microsoft-Windows-DotNETRuntime/GC/SuspendEEStart	937,675.351
Microsoft-Windows-DotNETRuntime/GC/SuspendEEStop 	937,675.361
Microsoft-Windows-DotNETRuntime/GC/RestartEEStart	937,677.708
Microsoft-Windows-DotNETRuntime/GC/RestartEEStop 	937,677.713
Microsoft-Windows-DotNETRuntime/GC/Stop          	937,691.248
Microsoft-Windows-DotNETRuntime/GC/HeapStats     	937,691.253

An avenue worth pursuing is to explore TraceGC.OnEnd.  The logic present in TraceGC.GCEnd may just be in the wrong place.  The advantage of this approach is that OnEnd is in all the right places to catch the true end of GCs (either Restart or HeapStats).  Thanks!

Maoni0:
@jeffschwMSFT , RestartEE isn't the last event for background GCs but as I mentioned above what GCEnd does is only applicable to blocking GCs and RestartEE is the last event for them. Do you see a problem with this?

Maoni0:
we could rename GCEnd to something like BlockingGCEnd to make it more obvious.

jeffschwMSFT:
@Maoni0, no issue - I submitted my comment before seeing yours. :)  I like your suggestion of BlockingGCEnd and the placement.

Maoni0:
@jeffschwMSFT ok! let's do that then - since it used to be called in an event that's the true end you could have done stuff related to that GC's end whether it's background GC or not but since now it's in RestartEE we want to make clear that it's only for blocking GCs. @ASethi77 could you rename this please?

jeffschwMSFT:
Thanks @ASethi77 lgtm.  It is great to have this fixed.

PR #176, opened 2017-05-05
Update to Visual Studio 2017

sharwell:
* Update solution and documentation to reference Visual Studio 2017 exclusively
* Fix a variety of build errors and simplify build configurations


src/EtwClrProfiler/CorProfilerTracer.cpp
```diff
@@ -119,7 +119,7 @@ EXTERN_C void __stdcall TailcallMethod(FunctionID functionID)
 #if defined(_M_IX86)
 // see http://msdn.microsoft.com/en-us/library/4ks26t93.aspx  for inline assembly.   Not supported on X64.   
 
-void __declspec( naked ) __stdcall EnterMethodNaked(FunctionID funcID)
```
vancem: Why did this have to be removed?    I am surprised.

src/PerfViewExtensions/GlobalSrc/Global.csproj
```diff
@@ -11,24 +11,6 @@
     <AssemblyName>Global</AssemblyName>
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
-    <!-- Determine if we are building 'next to' the main PerfView sources -->
```
vancem: This funky logic has a purpose.   Global.csproj is actually mean to support adding extensions to PerfView.   In perfView go to Help -> Extending PerfView for the experience.   Bascially perfView can set up a SLN that allows you to build things that extend perfView.   This Global.csproj is the template for this (We build it here just just so we are less likely to break the experience).   In 'Normal' use this project refers to the PerfVIew.exe that is placed in the 'installed location', but at build time it refers to where the build put it.  

In short, unless you want to revamp that, leave this alone.   You should also test the Extention experience to insure we did not break it.

src/TraceEvent/Ctf/CtfTracing.Tests/CtfTracing.Tests.csproj
```diff
@@ -36,19 +36,8 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
+    <Reference Include="Microsoft.VisualStudio.QualityTools.UnitTestFramework" />
   </ItemGroup>
-  <Choose>
```
vancem: I am assuming that the tests still run and pass.   Also I need this to work in 2015 too.

vancem:
First, thanks for the contribution.   There is a lot to like about the PR.    Here is my feedback.

The most important feedback is that while I certainly want perfView to be buildable from VS2017 (it is sad that it does not 'just work'), I need it to also continue to be buildable from VS2015 (at least for a while)    I just attempted to this this and it did not work for me.  I got errors 

[errorMsg.txt](https://github.com/Microsoft/perfview/files/981238/errorMsg.txt)
 
I have not looked carefully at it.   It may have something to do with the removal of the __stdcall prefix in the edit.  

Building from VS2015 will be the experience for many users so it needs to 'just work'    Note also I would like the experience for VS2017 users to also be good in the sense that they are not forced to use upgrade the C++ libraries (which it does typically).    What we want is to use the VC toolset that current invokcation of VS likes.   Thus you need to dynamically decide which to use based on your VS version.  You should search for <PlatformToolset>v140</PlatformToolset> for some logic that does this and add the next VS  version to it (it would be nice if VS just did this, but it doesn't).  

 You ripped out the editbin command for PerfView.exe and HeapDump.exe.   I would love to get rid of this turd, but it was making PerfView.exe  able to use 4GB in a 32 bit process.   As https://github.com/dotnet/roslyn/issues/9209 indicates the CSC will eventually support this but we are not there yet, (and I need it to work 'downlevel').    Worse in VS2017 the location of this tools is not really deterministic, making finding EditBin ugly.      I think the best answer is to write a trivial tool that does this (it literally is a bit flip, and I already have the PE header stuff in C#)  I can do this, but it will conflict with what you have done.    

I am happy to avoid needing corflags. (since VS supports 32bitPerferred)  

It turns out that ETWClrProfilerX64.vcproj and ETWClrProfilerX86.vcproj are supposed to be clones of one another (just building with a differnet arch.   It is useful if they are kept that way so that we can diff them to find fixes that should have gone into both.   If you want I would be more than happy if you factored all the common stuff which was imported by two stub files.   That just has not been done.  If you don't do that, I would like the diffs between these two to be kept to a minimum.   

The TraceEventCore project is desired, but I agree it needs to upgrade .NET Core 2.0 an I am OK with removing it for now.   

 There seems to be many files that have 'differences' that are only whitespace.   I tried

src/EtwClrProfiler/CorProfilerTracer.cpp
sharwell: I am not sure. The X86 release build would not build without it.

src/TraceEvent/Ctf/CtfTracing.Tests/CtfTracing.Tests.csproj
sharwell: > Also I need this to work in 2015 too.

Removing support for 2015 was an intentional design change. It allows the code to be cleaned up and use new language features moving forward. I believe this is not problematic for the following reasons:

1. Visual Studio 2017 is freely available for contributors
2. Visual Studio 2015 and 2017 have very different approaches to .NET Core projects which are not compatible
3. Visual Studio 2017 introduces C# 7, some features of which have proven useful for improving application performance and overall code quality in various situations

src/PerfViewExtensions/GlobalSrc/Global.csproj
sharwell: Can you clarify the purpose of the extension mechanism? Is this something just used for demos, or is it / has it been used in the past to provide core functionality?

sharwell:
> You ripped out the editbin command for PerfView.exe and HeapDump.exe.

Starting with Visual Studio 2015, the specific combination of AnyCPU with the Prefer32Bit flag sets the application as large address aware. [Reference](http://stackoverflow.com/questions/34905703/compiling-c-sharp-with-any-cpu-sets-application-can-handle-large-2gb-addresse). By changing the build from explicitly X86 to this combination, the need for a post build action was eliminated.

> It turns out that ETWClrProfilerX64.vcproj and ETWClrProfilerX86.vcproj are supposed to be clones of one another

I implemented this change in order to eliminate the possibility of confusion for contributors. There is now only one possible way to build each project, and that way is the right way. A side benefit of this change is it can reveal subtle differences between X86 and X64 builds that previously went unnoticed. For example:

* X86 uses `<PreprocessToFile>false</PreprocessToFile>` while X64 does not
* X86 uses the target name **EtwClrProfiler** while X64 uses **ETWClrProfiler**

Also keep in mind that these files essentially never change. I do not believe you will find the change problematic in the long run, and I certainly have not in cases where I use this pattern.

src/EtwClrProfiler/CorProfilerTracer.cpp
sharwell: ➡️  I think I figured this out such that no change is required.

vancem:
> Starting with Visual Studio 2015, the specific combination of AnyCPU with the Prefer32Bit flag sets the application as large address aware (Reference). 

This is good to know.  It is good to eliminate the need for using EditBin.

src/PerfViewExtensions/GlobalSrc/Global.csproj
vancem: The sensibility feature was used to built extensions to PerfView by the VS Performance team.     Like most features it is hard to know who else is using it (But I have referred several people to it).  

I think non-trivial take-backs have to be done with a more careful process.   At the least we have to give people notice, and we should ideally have its replacement in hand.   The change that makes the break should not be mixed in with other things if that can be avoided.  

But first we have to understand why we are doing it.    You mention that it is getting in your way, but don't mention how.   

My recommendation is that you assume that we do need to keep it for now, and fix this pull request accordingly.      If you feel strongly that it is getting in the way, open an issue on that alone and we start a discussion on the pros and cons.   That discussion would probably also include what we think is a reasonable replacement (some part of which would probably need to be done before removing the old support).    But to avoid bogging down this issue on the sensibility issue (which make take a while to resolve),   I would simply avoid breaking it in this change.

sharwell:
I should be able to update this late today with a much better approach to the C++ build. Sorry I couldn't get it earlier but wanted to give you a heads-up.

sharwell:
On the extensibility issue, I understand. I just wanted to get a quick feel for where it fits in the scheme of things (i.e. if no one was committed to it before then I wouldn't need to explain all the details, but now I have better context to do so) :smile:

src/TraceEvent/Ctf/CtfTracing.Tests/CtfTracing.Tests.csproj
vancem: > Removing support for 2015 was an intentional design change.

VS 2017 is really quite new (it has been out only since march).   Dropping compatibility for any previous version of VS should not be done lightly.     Your second point about about .NET Core is interesting and may be a good enough reason, but what is also true is that perfView itself (as opposed building a .NET Core compatible TraceEvent), does not require .NET Core.   Making .NET Core projects require 2017 is OK.   

In general whenever there is a non-trivial take away, a more formal process is needed.   It is one thing to add support for 2017, it is another to remove support for 2015.    To do the later, we need to warn people, and probably provide useful guidance on workarounds (e.g. a label of the last place in the code base  where VS2015 worked).  

Now if it was HARD to make 2017 work without dropping support for 2015, then we probably would warn people and make the switch, but from what I can tell, it is NOT that hard (The main issue I ran into when I tried it was the EDITBIN, which you point out we can simply remove).    

I realize that unless we set the language version to #6, reasonably quickly C#-7 constructs will be introduced into the code base that effectively breaks VS2015 users (although maybe not, I don't know if we would end up shipping VS2015 updates which bring C# 7 to VS2015).   But still, I feel there is value in keeping build system and other 'non-code' changes compatible (which allows people to fork and selectively cherry-pick more effectively).   

So how to proceed:    First I would like to understand if we can support BOTH without too much trouble.  If this takes less than a couple hours, I would like to do that.    

I also need to understand your intent about C#-7.    Normally it is pretty easy to avoid new features for a while and that is what we do.   That would allow this PR to proceed more easily.  

If you intend to immediately use C#-7 features, or if supporting both is hard then we need to warn people in the readme/markdown, and create a commit tag for VS2015 users and add guidance in markdown for using that tag.

src/EtwClrProfiler/CorProfilerTracer.cpp
sharwell: @vancem I can no longer reproduce the specific case where this failure occurred. It's back to using the code as originally written in this file. Very strange.

src/PerfViewExtensions/GlobalSrc/Global.csproj
sharwell: Thank you for this explanation. I believe we could improve this scenario in the future in several ways, but totally agree that it would best be done as a separate and very intentional change. The original behavior has been restored for the extensibility case while preserving the new behavior (project references) for building **PerfView.sln**.

src/TraceEvent/Ctf/CtfTracing.Tests/CtfTracing.Tests.csproj
sharwell: > First I would like to understand if we can support BOTH without too much trouble. If this takes less than a couple hours, I would like to do that.

It is easy for me to account for both. I'd like the opportunity during the normal work week (not taking up a weekend) to have a brief phone call (say 15 minutes) and make my case for the change as I proposed, and if I do not prove to be convincing in your mind then we'll go with a hybrid approach. :smile:

sharwell:
@vancem Thanks for the meeting today! As we discussed: I added back support for Visual Studio 2015 in the build files, but kept the messaging and links directing users to Visual Studio 2017. We can now work towards a new stable release and sometime after that we can talk about the benefits C# 7 might be able to provide moving forward. 👍

vancem:
LGTM 

Thanks for doing this.   This kind of cleanup is not trivial, and is often a thankless task.   I appreciate it.

PR #179, opened 2017-05-07
Sort columns in descending order when they are first clicked

### LINKED ISSUE #174: Sort order should be descending by default
### Steps to reproduce

1. Switch to **By Name** view, with the sort order set to any column except **Inc**
2. Click **Inc**

### Expected results

Sort order is now **Inc** (descending).

### Actual results

Sort order is now **Inc** (ascending).

sharwell:
Fixes #174


vancem:
LGTM

PR #180, opened 2017-05-07
Preserve column sort order after symbol lookup

### LINKED ISSUE #173: Sort order should not change in By Name view after symbol lookup
### Steps to reproduce

1. Switch to **By Name** tab
2. Set grouping to **[no grouping]**
3. Set sort order to **Inc** (descending)
4. Right click on any run and click **Lookup Symbols**

### Expected result

Sort order remains as **Inc** (descending).

### Actual result

Sort order switches to **Exc** (descending).

sharwell:
Fixes #173


vancem:
LGTM

Issue #147, opened 2016-12-06
ConvertMarkTimes doesn't fix MarkOverflow time


adityamandaleeka:
When reading in the mark events in the analysis code (`TraceManagedProcess`), we initially set the time for each type of marking to the relative timestamp of the ETW event itself (in milliseconds since the start of the trace), and then when we get to the end event for the corresponding GC, we replace those timestamps with the actual amount of time each type of marking took.

We handle this properly for most of the types (handles, FQ, etc.) but we do nothing with the MarkOverflow type, so it still shows the time that the event occurred.

PR #181, opened 2017-05-08
various bug fixes

Maoni0:
enable LOHWait analysis to look at waits due to allocating too much
enable mark times to recognize overflow marking event
check for cases where we are simply missing a PerHeapHistory event


Maoni0:
@jeffschwMSFT @vancem @adityamandaleeka PTAL

Maoni0:
#147

PR #183, opened 2017-05-08
Gather more details about .NET Versions for managed processes

jeffschwMSFT:
Actionable .NET Runtime versions are key to analysis.  Prior to .NET 4.? the RuntimeStart event did not include the qfe version (revision).  As a result many of the runtime versions are not actionable.  This change monitors KernelTraceControl/ImageID and KernelTraceControl/ImageID/FileVersion to capture the build time and version for clr.dll/mscorwks.dll.

### DIFF
--- a/src/PerfView/GcStats.cs
+++ b/src/PerfView/GcStats.cs
@@ -32,7 +32,9 @@ namespace Stats
                 writer.WriteLine("<LI><Font size=3 color=\"red\">Warning: Did not recognize the V4.0 GC Information events.  Falling back to V2.0 behavior.</font></LI>");
             if (!string.IsNullOrEmpty(stats.CommandLine))
                 writer.WriteLine("<LI>CommandLine: {0}</LI>", stats.CommandLine);
-            writer.WriteLine("<LI>Runtime Version: {0}</LI>", runtime.RuntimeVersion != null ? runtime.RuntimeVersion : "<Unknown Runtime Version>");
+            var runtimeBuiltTime = "";
+            if (runtime.RuntimeBuiltTime != default(DateTime)) runtimeBuiltTime = string.Format(" (built on {0})", runtime.RuntimeBuiltTime);
+            writer.WriteLine("<LI>Runtime Version: {0}{1}</LI>", runtime.RuntimeVersion ?? "&lt;Unknown Runtime Version&gt;", runtimeBuiltTime);
             writer.WriteLine("<LI>CLR Startup Flags: {0}</LI>", runtime.StartupFlags.ToString());
             writer.WriteLine("<LI>Total CPU Time: {0:n0} msec</LI>", stats.CPUMSec);
             writer.WriteLine("<LI>Total GC CPU Time: {0:n0} msec</LI>", runtime.GC.Stats().TotalCpuMSec);
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -121,6 +121,11 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
         /// Returns the .NET startup flags
         /// </summary>
         public StartupFlags StartupFlags { get; internal set; }
+        /// <summary>
+        /// Date and time of when the runtime was built
+        /// This is useful when a more detailed version is not present
+        /// </summary>
+        public DateTime RuntimeBuiltTime { get; internal set; }
 
         /// <summary>
         /// Garbage Collector (GC) specific details about this process
@@ -170,6 +175,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
         {
             runtimeVersion = new Version(0, 0, 0, 0);
             StartupFlags = StartupFlags.None;
+            RuntimeBuiltTime = default(DateTime);
         }
 
         /// <summary>
@@ -233,7 +239,9 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                TraceProcess process = data.Process();
                TraceLoadedDotNetRuntime mang = currentManagedProcess(data);
 
-               mang.runtimeVersion = new Version(data.VMMajorVersion, data.VMMinorVersion, data.VMBuildNumber, data.VMQfeNumber);
+               // replace the current runtimeversion if it is currently not set, or this version has information including revision (eg. qfe number)
+               if (mang.runtimeVersion.Major == 0 || data.VMQfeNumber > 0)
+                   mang.runtimeVersion = new Version(data.VMMajorVersion, data.VMMinorVersion, data.VMBuildNumber, data.VMQfeNumber);
                mang.StartupFlags = data.StartupFlags;
                // proxy for bitness, given we don't have a traceevent to pass through
                process.Is64Bit = (data.RuntimeDllPath.ToLower().Contains("framework64"));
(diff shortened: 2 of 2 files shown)


jeffschwMSFT:
@vancem @Maoni0

Maoni0:
would be nice to make a change to the GCCollectOnly commandline to have it include "/KernelEvents=Process+ImageLoad" by default.

src/PerfView/GcStats.cs
```diff
@@ -32,7 +32,9 @@ public static void ToHtml(TextWriter writer, TraceProcess stats, TraceLoadedDotN
                 writer.WriteLine("<LI><Font size=3 color=\"red\">Warning: Did not recognize the V4.0 GC Information events.  Falling back to V2.0 behavior.</font></LI>");
             if (!string.IsNullOrEmpty(stats.CommandLine))
                 writer.WriteLine("<LI>CommandLine: {0}</LI>", stats.CommandLine);
-            writer.WriteLine("<LI>Runtime Version: {0}</LI>", runtime.RuntimeVersion != null ? runtime.RuntimeVersion : "<Unknown Runtime Version>");
+            var runtimeBuiltTime = "";
+            if (runtime.RuntimeBuiltTime != default(DateTime)) runtimeBuiltTime = string.Format(" (built on {0})", runtime.RuntimeBuiltTime);
+            writer.WriteLine("<LI>Runtime Version: {0}{1}</LI>", runtime.RuntimeVersion != null ? runtime.RuntimeVersion : "<Unknown Runtime Version>", runtimeBuiltTime);
```
sharwell: :bulb: `runtime.RuntimeVersion ?? "<Unknown Runtime Version>"`

❗️ It looks like the output is HTML. Need to escape the `<` and `>` in the fallback string, or use a different delimiter.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -121,6 +121,11 @@ public class TraceLoadedDotNetRuntime
         /// Returns the .NET startup flags
         /// </summary>
         public StartupFlags StartupFlags { get; internal set; }
+        /// <summary>
+        /// Date and time of when the runtime was built
+        /// This is useful when a more detailed version is not present
+        /// </summary>
+        public DateTime RuntimeBuiltTime { get; internal set; }
```
sharwell: ❓ Is there a reason not to use `DateTime?` in order to clean up the code and keep it clear that this may or may not have a value?

❓ Is there a reason not to use `DateTimeOffset` instead of `DateTime`? It is a generally more reliable type for representing points in time in an application.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -244,6 +251,36 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
...
+                    // replace the current runtimeVersion if there is not good revision information
+                    if (mang.runtimeVersion.Major == 0 || mang.runtimeVersion.Revision == 0) mang.runtimeVersion = new Version(version.Major, version.Minor, version.Build, version.Revision);
+                    if (mang.RuntimeBuiltTime == default(DateTime)) mang.RuntimeBuiltTime = data.BuildTime;
+                }
+            };
+            symbolParser.ImageID += delegate (ImageIDTraceData data)
+            {
+                TraceProcess process = data.Process();
+
+                if (!string.IsNullOrWhiteSpace(data.OriginalFileName) &&
```
sharwell: :bulb: Use `string.Equals(data.OriginalFileName, "clr.dll", StringComparison.OrdinalIgnoreCase)`. It eliminates the need for the null/empty/whitespace check, and also for this case eliminates the need to have both `&&` and `||` in the same expression.

:memo: This occurs in another case above as well.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -244,6 +251,36 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
...
+                    (data.OriginalFileName.Equals("clr.dll", StringComparison.OrdinalIgnoreCase) || data.OriginalFileName.Equals("mscorwks.dll", StringComparison.OrdinalIgnoreCase))
+                    )
+                {
+                    TraceLoadedDotNetRuntime mang = currentManagedProcess(data); // this will create a mang instance for this process
```
sharwell: :bulb: Prefer to always place new comments on the line preceding code. In an open source project, contributors will inevitably use different editors over time (or the same editor with different extensions) and they all tend to treat end-of-line comments differently.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -244,6 +251,36 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
...
+            symbolParser.ImageIDFileVersion += delegate (FileVersionTraceData data)
+            {
+                TraceProcess process = data.Process();
+
+                if (!string.IsNullOrWhiteSpace(data.OrigFileName) &&
+                    (data.OrigFileName.Equals("clr.dll", StringComparison.OrdinalIgnoreCase) || data.OrigFileName.Equals("mscorwks.dll", StringComparison.OrdinalIgnoreCase))
+                    )
+                {
+                    TraceLoadedDotNetRuntime mang = currentManagedProcess(data); // this will create a mang instance for this process
+                    var version = Version.Parse(data.ProductVersion);
```
sharwell: ❓ Is it possible (now or ever) for `data.ProductVersion` to not be a true `Version`? It seems like we should be using `Version.TryParse` here instead to be sure. For example, what happens if a non-.NET process on the machine happens to have a file called **clr.dll** or **mscorwks.dll**?

jeffschwMSFT:
@Maoni0 I think we should open a separate issue to track updating the default.  I hesitate doing it without more exploration since GCCollectOnly contains *only* in the name and I am not sure the volume associated with the process+imageload events.
@sharwell Thanks for the feedback, most I can accommodate.  The reason for a non-nullable DateTime is merely to align with other DateTime's within TraceEvent.  Nullable is rarely used (in fact the only public use I noticed was on TraceProcess for exitcode).

src/PerfView/GcStats.cs
jeffschwMSFT: RuntimeVersion is initially set to  'V 0.0.0.0', I am not sure unknown runtime version is any better.

src/TraceEvent/Computers/TraceManagedProcess.cs
jeffschwMSFT: Non-nullable to align with other DateTime types within TraceEvent.  It is a DateTime merely since that is the underlying type present in the ETL file.

src/TraceEvent/Computers/TraceManagedProcess.cs
jeffschwMSFT: Good point.

src/TraceEvent/Computers/TraceManagedProcess.cs
jeffschwMSFT: Good point.

src/TraceEvent/Computers/TraceManagedProcess.cs
jeffschwMSFT: That is a great consideration, fixed.

src/PerfView/GcStats.cs
sharwell: @jeffschwMSFT Sorry about the confusion. The way I first read this, it looked like the following expression was added by you:

```csharp
runtime.RuntimeVersion != null ? runtime.RuntimeVersion : "<Unknown Runtime Version>"
```

My suggestion (:bulb:) gives a shorter expression which is equivalent to it, while the ❗️ points out that if the fallback string is ever used then the output is invalid. Even though you didn't add this code originally, it would be good to correct at least the latter issue while you're on the same line as it by changing `<`&rarr;`&lt;` and `>`&rarr;`&gt;`. Otherwise I'll file a separate issue for this.

src/TraceEvent/Computers/TraceManagedProcess.cs
sharwell: Thanks for the explanation. Due to the number of times I've seen these two items cause actual problems, I'm still going to leave a suggestion. With that said, I understand that this could pretty easily be a "I wish it could be this way" type of situation where the current form ends up being what we're stuck with for some reason or another.

:bulb: Consider using `DateTimeOffset?` for this property as an API guard against certain possible confusion about its meaning/use throughout the portions of the code where it appears.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -244,6 +252,35 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
             clrRundownParser.RuntimeStart += doAtRuntimeStart;
             source.Clr.RuntimeStart += doAtRuntimeStart;
 
+            var symbolParser = new SymbolTraceEventParser(source);
+            symbolParser.ImageIDFileVersion += delegate (FileVersionTraceData data)
+            {
+                TraceProcess process = data.Process();
+
+                if (string.Equals(data.OrigFileName, "clr.dll", StringComparison.OrdinalIgnoreCase) || string.Equals(data.OrigFileName, "mscorwks.dll", StringComparison.OrdinalIgnoreCase) || string.Equals(data.OrigFileName, "coreclr.dll", StringComparison.OrdinalIgnoreCase))
```
sharwell: :bulb: Duplicated expression; consider using a helper method.

src/PerfView/GcStats.cs
jeffschwMSFT: It makes more sense now, I did not even notice that portion ;)  It frankly is dead code, but as you suggested I am on that line so I will make your suggested update.  Testing now.

src/TraceEvent/Computers/TraceManagedProcess.cs
vancem: It turns out that null values are actually not as good as default values.   Typically user logic simply uses the value.  If it can be null, it throws an exception, and generally it is better just to use the 'obvious empty value'.    For just this reason we generally return empty strings rather than null strings.      You can argue about the merits of this, but for better or worse it is the convention of this library.

vancem:
I am OK with the change as is.  I do agree with @sharwell  that it would be good to have a routine that means 'IsDotNetRuntimeDll'.   If anyone wants to fix that please do so.

vancem:
Note I have made GCCollectOnly collect Image information in PR #187.     I think the tradeoff there is good (minor size increase, but the version information can be really important).

Issue #153, opened 2017-01-17
.NET 3.5 support for TraceEvent library


NN---:
Currently TraceEvent cannot be compiled for .NET 3.5.

Missing parts:
1. Concurrent collections, NuGet package can be used for this: https://www.nuget.org/packages/System.Threading.dll/
2. System.Tuple class, There is NuGet package for this too https://www.nuget.org/packages/NetLegacySupport.Tuple/
3. Missing framework methods such as string.IsNullOrWhitespace can be emulated with simple wrapper class.

I can help with PRs if you need any help there.

vancem:
Support for older environments tends to create tradeoffs (forcing the code to conform to old conventions etc).    TraceEvent has support V3.5 runtimes in the past and I am open to continuing this IF IT IS NOT TOO CONSTRAINING.    My experience with the past V3.5 support was not too bad, but it is getting harder to do (use of newer features start to become more and more 'indispensable').   I do not wish constrain forward looking updates much at all.

Ultimately I think what we will end up with is a V3.5 BRANCH.   Maybe some of those changes can be integrated into master without pain,but the rest will simply live in the other branch.    Yes there is pain associated with picking up updates from the master branch into the V3.5 branch, but I think that is the best we can do.

NN---:
I understand that 3.5 is a problem, but it is still officially supported by MS.

sharwell:
@NN Can you provide a bit more information about what a .NET 3.5 build of this application would help you accomplish?

NN---:
I am using the library as ETW consumer. More functionality is not needed right now.
The only important part for me is to have continuous bugfixes.

sharwell:
@NN Thanks for the additional information! I'm new to looking at this project so I wasn't yet clear on the difference between the TraceEvent library and the PerfView application. (I'm sure it made sense to @vancem the whole time, but you did help me.)

NN---:
No problems.
Perhaps you can split repositories to develop library and UI independently and use NuGet reference.
With new VS2017 ProjectReference feature it is very easy to maintain.

NN---:
Btw for .NET 3.5 there is TaskParralellsLibrary NuGet workaround to provide Concurrent collections and Tasks.

brianrob:
This issue has been open for some time, and things have only become more tied to newer releases.  As such, we are not going to provide support for .NET 3.5.  Closing this issue.

PR #185, opened 2017-05-09
Trivial fixes

northtyphoon:
1. Make TraceLoggingEventId to implement IDisposable.
2. Optimize some codes in TraceLoggingEventId for perf.
3. Fix typos.

### DIFF
--- a/src/TraceEvent/TraceLoggingEventId.cs
+++ b/src/TraceEvent/TraceLoggingEventId.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
 
@@ -10,19 +11,20 @@ namespace Microsoft.Diagnostics.Tracing
     /// <summary>
     /// TraceLoggingEvnetId is a class that manages assigning event IDs (small 64k numbers)
     /// to TraceLogging Style events (which don't have them).  Because TraceEvent uses EventIDs
-    /// so fundamentally this difficiency is very problematic.   
+    /// so fundamentally this deficiency is very problematic.   
     /// 
     /// Arguably this should have been done by the ETW system itself.  
     /// 
     /// You use it by calling TestForTraceLoggingEventAndFixupIfNeeded on eventRecords.  
     /// You also have to explicitly call 'Dispose' when you are done with this class.  
     /// </summary>
-    internal unsafe struct TraceLoggingEventId
+    internal unsafe struct TraceLoggingEventId : IDisposable
     {
         /// <summary>
         /// Checks to see if eventRecord has TraceLogging meta data associated with it (EVENT_HEADER_EXT_TYPE_EVENT_SCHEMA_TL)
         /// and if so updates EventHeader.Id to be an event ID unique to that provider/opcode/meta-data blob. 
         /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void TestForTraceLoggingEventAndFixupIfNeeded(TraceEventNativeMethods.EVENT_RECORD* eventRecord)
         {
             // This method is designed to be inlined and thus have very low overhead for the non-tracelogging case.   
@@ -38,8 +40,8 @@ namespace Microsoft.Diagnostics.Tracing
             if (m_traceLoggingEventMap == null)
                 return;
 
-            foreach (var key in m_traceLoggingEventMap.Keys)
-                Marshal.FreeHGlobal((IntPtr)key.Provider);
+            foreach (var kvp in m_traceLoggingEventMap)
+                Marshal.FreeHGlobal((IntPtr)kvp.Key.Provider);
             m_traceLoggingEventMap = null;
             m_nextTraceLoggingIDForProvider = null;
         }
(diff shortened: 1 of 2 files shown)


src/TraceEvent/TraceLoggingEventId.cs
```diff
@@ -10,19 +11,20 @@ namespace Microsoft.Diagnostics.Tracing
...
     /// You also have to explicitly call 'Dispose' when you are done with this class.  
     /// </summary>
-    internal unsafe struct TraceLoggingEventId
+    internal unsafe struct TraceLoggingEventId : IDisposable
     {
         /// <summary>
         /// Checks to see if eventRecord has TraceLogging meta data associated with it (EVENT_HEADER_EXT_TYPE_EVENT_SCHEMA_TL)
         /// and if so updates EventHeader.Id to be an event ID unique to that provider/opcode/meta-data blob. 
         /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
sharwell: :question: Were you encountering situations where this method was not already getting inlined? I would prefer to omit this change in light of requests like #153 unless there is a demonstrated problem with the current approach.

src/TraceEvent/TraceLoggingEventId.cs
```diff
@@ -10,19 +11,20 @@ namespace Microsoft.Diagnostics.Tracing
     /// <summary>
     /// TraceLoggingEvnetId is a class that manages assigning event IDs (small 64k numbers)
     /// to TraceLogging Style events (which don't have them).  Because TraceEvent uses EventIDs
-    /// so fundamentally this difficiency is very problematic.   
+    /// so fundamentally this deficiency is very problematic.
```
sharwell: :bulb: In the future, it helps to place all formatting-only changes and spelling corrections in their own commit, and preferably in their own independent pull request. The primary advantages this offers are:

1. It's easier to expedite spelling corrections so they don't get tied up in code review
2. It makes the code review process less complex for the actual code changes (for example, this pull request would contain a one-line change in one file, and nothing else)

The scope in this pull request is fairly small so I don't see it causing a problem, but it adds up pretty quickly so it's a good thing to keep in mind for the future. 👍

src/TraceEvent/TraceLoggingEventId.cs
```diff
@@ -38,8 +40,8 @@ public void Dispose()
             if (m_traceLoggingEventMap == null)
                 return;
 
-            foreach (var key in m_traceLoggingEventMap.Keys)
-                Marshal.FreeHGlobal((IntPtr)key.Provider);
+            foreach (var kvp in m_traceLoggingEventMap)
+                Marshal.FreeHGlobal((IntPtr)kvp.Key.Provider);
```
sharwell: :memo: For those unaware, this change eliminates an unnecessary allocation during the `Dispose` call. One of the things I'm enjoying about C# 7 (which we aren't currently using on this project) is the ability to gain the efficiency of this form without losing the clarity of the original code by writing this:

```csharp
foreach (var (key, _) in m_traceLoggingEventMap)
  Marshal.FreeHGlobal((IntPtr)key.Provider);
```

This form applies a deconstruction to the `KeyValuePair<TKey, TValue>` in the declaration of `key` and an unused placeholder `_`.

northtyphoon: Thanks for sharing the C# 7 tips.

src/TraceEvent/TraceLoggingEventId.cs
northtyphoon: I agree. As you said, the scope of this PR is very small, so I combined the changes. I will follow the guidance in the future. 
Can you also add the guidance to https://github.com/Microsoft/perfview/blob/master/CONTRIBUTING.md

src/TraceEvent/TraceLoggingEventId.cs
pharring: @sharwell Pedantic: You can't deconstruct a KeyValuePair like that. You'll need a custom Deconstruct method.

sharwell: @pharring It depends on where things are running. It's already added to coreclr in dotnet/coreclr#8560 (not sure what release version that corresponds to). For Roslyn I added the `Deconstruct` extension method when I used the feature in dotnet/roslyn#19126.

sharwell:
@northtyphoon Just making sure you saw my question on the use of `[MethodImpl]` above. :smile:

src/TraceEvent/TraceLoggingEventId.cs
northtyphoon: I didn't try to check if the method is inlined or not in runtime. Since I see @vancem has changed TraceEvent lib to target 4.5 and therefore I made the change.
For 3.5 support, do we consider to create a release branch and only do bug fix as needed?

sharwell: @northtyphoon Currently the library does not support .NET 3.5, and there is no concrete plan to do so. My suggestion here is to only include this change if it is needed, e.g. if:

1. PerfView revealed this method in traces (a clear indication that not only is it not being inlined, but it's heavily used)
2. You saw this method in disassembled JIT code and observed that it wasn't inlined, and knew separately that it's heavily used so it should be
3. Some other indicator suggesting that the attribute provides measurable value

For me, the bar here is low. I just don't want to include it if no one will notice that it was included, because it makes things a tiny bit more complicated if we ever decide to go back and address the specific request I linked to.

src/TraceEvent/TraceLoggingEventId.cs
sharwell: @northtyphoon It seems to be covered well enough already. I mostly assume that people don't read that, or do a cursory look at best. I don't let it bother me when deviations happen and I'm always willing to help if someone isn't sure how to break things up. :smile:

vancem:
LGTM

PR #187, opened 2017-05-09
Add Image Events to /GCCollectOnly

vancem:
It is useful for /GCCollectOnly to have version information on the runtime.  We can get this
by simply turning on the kernel events for Image loads.   This does not increase the size of the file
significanctly.


Maoni0:
that's great! Thanks @vancem

PR #192, opened 2017-05-10
Improve the performance of opening the GCStats window

sharwell:
This change adds a local cache from `TraceProcess` to `TraceLoadedDotNetRuntime`, resulting in a dramatic reduction in the time it takes to open the **GCStats** window for large traces.

:memo: It wasn't clear exactly how you prefer items like this to be implemented. However, the performance difference from this change is somewhere between 3:1 and 10:1 (saving dozens of seconds to minutes), so I'm particularly interested in finding a good way to get it included.

### DIFF
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -290,14 +290,27 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                 var clrRundown = new ClrRundownTraceEventParser(source);
                 clrRundown.RuntimeStart += doAtRuntimeStart;
                 source.Clr.RuntimeStart += doAtRuntimeStart;
+                Dictionary<TraceProcess, TraceLoadedDotNetRuntime> processRuntimes = null;
 
                 CircularBuffer<ThreadWorkSpan> RecentThreadSwitches = new CircularBuffer<ThreadWorkSpan>(1000);
                 source.Kernel.ThreadCSwitch += delegate (CSwitchTraceData data)
                 {
+                    if (processRuntimes == null)
+                    {
+                        // lazy initialization
+                        processRuntimes = new Dictionary<TraceProcess, TraceLoadedDotNetRuntime>();
+                        foreach (var proc in source.Processes())
+                        {
+                            var runtime = proc.LoadedDotNetRuntime();
+                            if (runtime != null)
+                                processRuntimes[proc] = runtime;
+                        }
+                    }
+
                     RecentThreadSwitches.Add(new ThreadWorkSpan(data));
                     TraceProcess tmpProc = data.Process();
-                    TraceLoadedDotNetRuntime mang = tmpProc.LoadedDotNetRuntime();
-                    if (mang != null)
+                    TraceLoadedDotNetRuntime mang;
+                    if (processRuntimes.TryGetValue(tmpProc, out mang))
                     {
                         mang.GC.m_stats.ThreadId2Priority[data.NewThreadID] = data.NewThreadPriority;
                         int heapIndex = mang.GC.m_stats.IsServerGCThread(data.ThreadID);
@@ -307,10 +320,10 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                         }
                     }
 
-                    foreach (var proc in source.Processes())
+                    foreach (var pair in processRuntimes)
                     {
-                        mang = proc.LoadedDotNetRuntime();
-                        if (mang == null) continue;
+                        var proc = pair.Key;
+                        mang = pair.Value;
 
                         TraceGC _gc = TraceGarbageCollector.GetCurrentGC(mang);
                         // If we are in the middle of a GC.
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -196,12 +196,14 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
             // See if the source knows about the CLR Private provider, if it does, then 
             var clrPrivate = new ClrPrivateTraceEventParser(source);
 
+            Dictionary<TraceProcess, TraceLoadedDotNetRuntime> processRuntimes = new Dictionary<TraceProcess, TraceLoadedDotNetRuntime>();
+
             // if any clr event is fired, this process is managed
             Action<TraceEvent> createManagedProc = delegate (TraceEvent data)
             {
                 var proc = data.Process(); // this will create an instance, if one does not exist
-                var mang = proc.LoadedDotNetRuntime();
-                if (mang == null)
+                TraceLoadedDotNetRuntime mang;
+                if (!processRuntimes.TryGetValue(proc, out mang))
                 {
                     // duplicate the TraceProcess and create an instance of TraceManagedProcess
                     mang = new TraceLoadedDotNetRuntime(proc);
@@ -214,6 +216,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
 
                     // fire callback and associate this DotNetRuntime with this process
                     proc.OnDotNetRuntimeLoaded(mang);
+                    processRuntimes.Add(proc, mang);
                 }
             };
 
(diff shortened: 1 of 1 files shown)


vancem:
+ @jeffschwMSFT @Maoni0  who care about GCStats

I am not opposed to this change.   However there is better way of doing this.
'Trace*' classes have a concept of '*Index', which are small integers that uniquely identify the thing.  Thus TraceProcess has a 'ProcessIndex' which is a small integer that is unique for that TraceProcess.

These indexes are designed specifically to allow higher level constructs (e.g. GCStats to 'attach' extra information TraceProcess in an efficient manner by having a side array.   Indeed the 'GrowableArray<T> has specific methods (Get and Set) that fetch or set things growing the array as necessary (thus you don't need to know the upper bound on indexes).    This allows you to replace the Dictionary lookup with a array fetch, which is faster.  

My only qualm is that I don't know if we were good about setting ProcessIndex well in all cases (it looks like we were not),   Ideally we would fix this at least in the case where this code is used.   

However if that is problematic, I am OK with the dictionary lookup.

jeffschwMSFT:
@sharwell this code relies on the fact that all processes are known by the first ThreadCSwitch - but that is not the case.  The processes list is built over time (as events are seen), similarly for ManagedRuntimes.  A strategy like this could work, but you would need to add to the cache over time.

sharwell:
@jeffschwMSFT If you use the **Review changes** drop down on the **Files changed** page, you can **Request changes** to transition the pull request into a state where it tells me that I have work to do on it. 👍

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -290,14 +290,27 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
...
+                        // lazy initialization
+                        processRuntimes = new Dictionary<TraceProcess, TraceLoadedDotNetRuntime>();
+                        foreach (var proc in source.Processes())
+                        {
+                            var runtime = proc.LoadedDotNetRuntime();
+                            if (runtime != null)
+                                processRuntimes[proc] = runtime;
+                        }
+                    }
+
```
jeffschwMSFT: The processRuntime cache needs to be built overtime.  The contents of both source.Processes and LoadedDotNetRuntime are built as events are processed.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -290,14 +290,27 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
                 var clrRundown = new ClrRundownTraceEventParser(source);
                 clrRundown.RuntimeStart += doAtRuntimeStart;
                 source.Clr.RuntimeStart += doAtRuntimeStart;
+                Dictionary<TraceProcess, TraceLoadedDotNetRuntime> processRuntimes = null;
 
                 CircularBuffer<ThreadWorkSpan> RecentThreadSwitches = new CircularBuffer<ThreadWorkSpan>(1000);
                 source.Kernel.ThreadCSwitch += delegate (CSwitchTraceData data)
                 {
```
jeffschwMSFT: I know that CSwitches are enormously verbose, but is there a reason this cache is not applied more broadly?  If you look above in the file you can see how the lookup is done.  Perhaps using the ProcessIndex could offer a faster lookup.

src/TraceEvent/Computers/TraceManagedProcess.cs
sharwell: I plan to trace through the code to understand this process, and apply a correction to account for it. If I run into problems taking a look on my own, I'll let you know. :smile:

src/TraceEvent/Computers/TraceManagedProcess.cs
sharwell: I tried to limit the scope of my implementation since I'm not very familiar with the code. The only reason I found it was through the use of PerfView to analyze opening the GCStats window for a large trace.

The use of `ProcessIndex` sounds promising, but based on @vancem's comments it sounds like it might not work without other changes. I plan to first make the limited corrections that we know are necessary to make this local cache accurate in its current form, and then hopefully we can file an issue to further improve it by using direct lookup instead of a dictionary.

sharwell: @jeffschwMSFT I updated the cache to apply to most uses of this data inside `SetupCallbacks`. The exception is the use of `MutableTraceEventStackSource` and `HasMutableTraceEventStackSource` which operate on a slightly different object and don't currently appear to be a problem.

src/TraceEvent/Computers/TraceManagedProcess.cs
sharwell: ➡️ This should be working now.

jeffschwMSFT:
@sharwell one of the more challenging parts of working within TraceEvent is lifetime.  The reason the current cache is a property of Source is that there are many scenarios where lifetime of the callbacks becomes overlapping.  I would recommend trying this code out with a live TraceEvent session and one that processes multiple etls within the same session.
How are the performance improvements looking?  The existing implementation differs really only by 1 indirection and ContainsKey v TryGet.  I would expect that any performance gain could be close to modifying the current assessor.

sharwell:
> one of the more challenging parts of working within TraceEvent is lifetime. The reason the current cache is a property of Source is that there are many scenarios where lifetime of the callbacks becomes overlapping.

It appears `SetupCallbacks` is called once per source. The cache is scoped to this context. The cache should be valid unless one of the following occurs:

1. The `UserData` is cleared somewhere, and then the callbacks are invoked again
2. `SetupCallbacks` is called multiple times for the same source, where the callbacks for one of these depends on data initialized by the other
3. The callbacks for a single source are invoked concurrently

> I would recommend trying this code out with a live TraceEvent session and one that processes multiple etls within the same session.

I'm not sure I know how to do this.

> The existing implementation differs really only by 1 indirection and ContainsKey v TryGet. I would expect that any performance gain could be close to modifying the current assessor.

So far I was unable to come close with a few other approaches I tried. For reference, here's how much difference this change makes on my current scenario:

* Without this change: **Opening GCStats (Elapsed Time: 180.646 sec)**
* With this change: **Opening GCStats (Elapsed Time: 6.511 sec)**

jeffschwMSFT:
Explore TraceEventSession("<Session Name>") for the live session.  Also do try various multiple processing configurations - as I mentioned this was a place of many bugs when I did this work last summer.
This portion of the code does not have good validation.  To accomodate that I did text diffs of GCStats' pane to compare.  How are those looking for this change?

sharwell:
> Explore TraceEventSession("") for the live session. Also do try various multiple processing configurations - as I mentioned this was a place of many bugs when I did this work last summer.

I'm pretty new to this, so I'm not sure what this means. I looked in PerfView but didn't see anything for "live session". I also don't know anything able "processing configurations" aside from just opening files in the PerfView UI.

> To accomodate that I did text diffs of GCStats' pane to compare. How are those looking for this change?

The source (Right Click -> View Source) was a binary match according to Beyond Compare. The input is a 5.28GiB trace.

sharwell:
> This portion of the code does not have good validation.

I'm hoping we can work to improve this after #195 is merged and ideally code coverage for the builds is enabled. :smile:

vancem:
I am persuaded by the argument that Sam made above that there should not be lifetime issues.   The logic for the correctness is pretty simple (every process can have at most one loaded runtime so if we computed it in the past it should still be valid).  

I think it is safe enough and has non-trivial performance value.   We should move forward, so I am merging it.

jeffschwMSFT:
I think that is fine.  There is a lifetime issue, though the worst case scenario is that the cache gets deleted and requires to be refilled, I believe.  I am still super surprised that the following code change is leading to this speedup:

From:
<method call>
Dictionary<ProcessIndex, DotNetRuntime> map = process.Source.UserData["Computers/LoadedDotNetRuntimes"] as Dictionary<ProcessIndex, DotNetRuntime>;
if (map.ContainsKey(process.ProcessIndex)) return map[process.ProcessIndex].Runtime;
else return null;

To:
processRuntimes.TryGetValue(data.Process(), out mang)

In essence it is going from method call + 1 string hash lookup + 2 int hash lookups TO 1 object hash lookup.  That still makes me feel there is something lurking in this change.  Though our auto etl processing engine (which runs multiple etl processing in parallel) would find any egregious issues.

PR #193, opened 2017-05-10
Improve the performance of user data lookups by changing from string to enum

sharwell:
This change updates the `UserData` dictionary to use enum keys instead of string keys, resulting in a marginal (observable to an astute user) performance improvement in opening the GCStats window. It may improve other areas as well, but I have not collected data to say for sure.

⚠️ I believe this is an API change in the TraceEvent library, and based on the name I must assume that users will notice. The `GetOrCreateKnownUserData` method aims to provide a new way to accomplish the same underlying operation. I'm not sure what the project stance on long-term compatibility is, or how users feel about API changes which are implemented to improve overall performance.


vancem:
This needs a more careful/thoughtful look.   If I don't get back to you by Monday, nudge me again.

sharwell:
I'm going to close this one for now. There are too many alternatives which seem better for the performance-critical paths where this data tends to be used.

PR #194, opened 2017-05-10
Add PerfView64, a wrapper around PerfView

### LINKED ISSUE #171: OOM Exception inside Perfview when trying to analyze a largish etl file
An unhanded exception occured.

At this point you can opt to continue, however it is possible that the aborted computation will cause additional failures.   Because PerfView generally only opens files for reading, there is no danger of corrupting  files, so it generally does not hurt to try.   However be on guard for unusual/incorrect behavior going forward.

You can of course exit and restart PerfView to be completely safe.

__The fact that this exception went unhanded is a programmer error.   It should be reported so it can be fixed.  Please set along the following stack trace information which will be useful in diagnosing the problem.__

StackTrace:
System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at System.Text.StringBuilder.ToString()
   at Controls.TextEditorWriter.<Flush>b__3_0()
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)

--------------------------

image

This occured when trying to open "Thread Time" on a 2.7GB etl file.

If you need it I can either send you the original etl, or a dump of perfview itself.

-------------------

image

(latest from official download center)

### LINKED ISSUE #175: 64-bit build of PerfView
### Steps to reproduce

1. Run PerfView with Thread Time enabled
2. Open Thread Time view
3. Set grouping to **[no grouping]**
4. (Sometimes required) Try to switch between views

### Expected results

The application works.

### Actual results

The application crashes after an OutOfMemoryException.

### Additional details

A local 64-bit build showed that use of PerfView regularly used 16+GB of memory for analyzing traces of even a few seconds (5-10 seconds of system time). The application is totally unusable for Thread Time when it fails to load as a 64-bit image.

sharwell:
This change introduces **PerfView64**, a 12KiB file wrapper executable which launches **PerfView** as a 64-bit application.

In this initial form, the intent is the **PerfView64** executable would be attached to the GitHub release alongside the **PerfView** executable. Users could optionally download the executable, and would be responsible for launching through it if they need to use more memory during analysis.

Fixes #175
Fixes #171

### DIFF
--- a/src/PerfView/Utilities/SupportFiles.cs
+++ b/src/PerfView/Utilities/SupportFiles.cs
@@ -177,26 +177,18 @@ namespace Utilities
             }
         }
         /// <summary>
-        /// The path to the executable for relaunching PerfView with administrator privileges.
+        /// The path to the entry executable.
         /// </summary>
-        public static string ExePathForRelaunch
+        public static string ExePath
         {
             get
             {
-                if (s_exePathForRelaunch == null)
+                if (s_exePath == null)
                 {
-                    // As a special case, handle the 64-bit entry point of PerfView64
-                    var entryAssembly = Assembly.GetEntryAssembly();
-                    if (string.Equals(entryAssembly.ManifestModule.Name, "PerfView64.exe", StringComparison.OrdinalIgnoreCase))
-                    {
-                        s_exePathForRelaunch = entryAssembly.ManifestModule.FullyQualifiedName;
-                    }
-                    else
-                    {
-                        s_exePathForRelaunch = ExePath;
-                    }
+                    s_exePath = Assembly.GetEntryAssembly().ManifestModule.FullyQualifiedName;
+                    Debug.Assert(s_exePath.EndsWith(".exe", StringComparison.OrdinalIgnoreCase));
                 }
-                return s_exePathForRelaunch;
+                return s_exePath;
             }
         }
         /// <summary>
@@ -346,7 +338,7 @@ namespace Utilities
         private static string s_supportFileDir;
         private static string s_supportFileDirBase;
         private static string s_mainAssemblyPath;
-        private static string s_exePathForRelaunch;
+        private static string s_exePath;
         private static List<string> s_managedDllSearchPaths;
 #endregion
     }
--- a/src/PerfView/Utilities/SupportFiles.cs
+++ b/src/PerfView/Utilities/SupportFiles.cs
@@ -124,7 +124,7 @@ namespace Utilities
             get
             {
                 {
-                    var exeLastWriteTime = File.GetLastWriteTime(ExePath);
+                    var exeLastWriteTime = File.GetLastWriteTime(MainAssemblyPath);
                     var version = exeLastWriteTime.ToString("VER.yyyy'-'MM'-'dd'.'HH'.'mm'.'ss.fff");
                     s_supportFileDir = Path.Combine(SupportFileDirBase, version);
                 }
@@ -144,14 +144,14 @@ namespace Utilities
             {
                 if (s_supportFileDirBase == null)
                 {
-                    string appName = Path.GetFileNameWithoutExtension(ExePath);
+                    string appName = Path.GetFileNameWithoutExtension(MainAssemblyPath);
 
                     string appData = Environment.GetEnvironmentVariable(appName + "_APPDATA");
                     if (appData == null)
                     {
                         appData = Environment.GetEnvironmentVariable("APPDATA");
                         if (appData == null)
-                            appData = Path.GetFileName(ExePath);
+                            appData = Path.GetFileName(MainAssemblyPath);
                     }
                     s_supportFileDirBase = Path.Combine(appData, appName);
                 }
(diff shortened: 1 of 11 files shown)


src/PerfView64/PerfView64.csproj
```diff
@@ -0,0 +1,79 @@
...
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <Prefer32Bit>false</Prefer32Bit>
+  </PropertyGroup>
+  <PropertyGroup>
+    <ApplicationIcon>..\PerfView\performance.ico</ApplicationIcon>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
```
vancem: Can you trim out the unnecessary references.   I think you can get rid of almost all of them.

src/PerfView64/App.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Diagnostics;
+
+namespace PerfView64
+{
+    public class App
```
vancem: Can you put a comment on the basic architecture here (it is a wrapper that is referring to the other EXE (and is meant to be deployed in the same directory as PerfView.exe.

sharwell:
@vancem I addressed your feedback 👍

src/PerfView/Utilities/SupportFiles.cs
```diff
@@ -178,6 +178,29 @@ public static string ExePath
             }
         }
         /// <summary>
+        /// The path to the executable for relaunching PerfView with administrator privileges.
```
vancem: I would like to get rid of this special case.     SupportFiles is supposed to be generic code that knows nothing about PerfView.  

The comments in the ExePath property tell the story.   Over time this has morphed and now needs fixing.   Here is my proposed fix.

We should rename ExePath to 'MainAssemblyPath'  (it is defined as the assembly that contains the SupportFiles class).  

Then we refine ExePath to simply use 'GetEntryAssembly()', and use that to relaunch.

vancem:
I left some line-level feedback.    The general direction is good.   Ultimately I think what we should do is 

1. Put the 64 bit exe in the support files.   
2. We can then have a File-> 'Launch as 64 bit'.   We can also make it so that if the EXE is named PerfView64 (or a particular environment variable is set or command line option is set), it will relaunch itself as 64 bit (we already have logic that does kind of thing (for elevation or if launched on a file share).  

This way you don't have to deploy the 64 bit version.  Every version has that capability and users can set that default in a number of easy ways if they so desire.   

I can give more details on exactly how to do this if needed/desired.

src/PerfView64/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,52 @@
...
+//      Major Version
+//      Minor Version
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.9.53")]
```
vancem: Don't make this version the same as the PerfView version, as I don't want to have to keep it in sync.   It is a wrapper and its version is 1.0.0, and it will probably never change again.

vancem:
I would like the SupportFiles.cs to be fixed to avoid special casing PerfView.

sharwell:
> We can then have a File-> 'Launch as 64 bit'.

How are we expecting a user to create a shortcut which always launches the application in 64-bit mode? I could, for example, create a command line argument which immediately relaunches if it started in 32-bit mode, but then we have other strange problems such as confusion regarding the ability to debug the application when running as a 64-bit process.

vancem:
> How are we expecting a user to create a shortcut which always launches the application in 64-bit mode? I could, for example, create a command line argument which immediately relaunches if it started in 32-bit mode, but then we have other strange problems such as confusion regarding the ability to debug the application when running as a 64-bit process.

Yes the idea is that either a command line arg or the name of the EXE can trigger a relaunch if necessary.   Yes, if you want to debug, it is harder or you have to know that you need to run the 64bit exe directly.   Note that we have this debugging problem in other things (we have to spawn a sub process to capture heap dumps to match the bitness of the target).     I don't view the debugger PerfView scenario as important compared to the 'convenience of non debugging scenarios'.

src/PerfView64/Properties/AssemblyInfo.cs
sharwell: ➡️ Updated this to 1.0.0.0

sharwell:
@vancem I'm concerned about the pending scope creep for this pull request for a couple of reasons.

1. The lack of this feature is a known blocker for multiple people trying to use PerfView, and it's demonstrated to cause problems for some users in even very simple use cases (i.e. it needs to be fixed soon)
2. The `ExePath` property and `SupportFiles` are used for much more than just relaunching the executable. This became immediately obvious to me when I tried to update `ExePath` to simply behave like the new `ExePathForRelaunch` (prior to splitting them into two methods): on first launch the EULA acceptance window reappeared.

At this point I would prefer to avoid turning a blocking bug fix into a full new feature design. I believe the following path forward would be a good way to unblock users:

1. Use this pull request in its current functionality (32-bit or 64-bit determined exclusively by which executable is clicked)
2. Attach the resulting PerfView64.exe executable to the GitHub release notes when a release is made, but don't complicated the official download page at this point
3. File a follow-up issue to merge or otherwise update PerfView.exe to address 64-bit systems without a separate executable. After this follow-up issue is resolved, 64-bit support can be announced and included via the official download page.

vancem:
First, I should have been more clear that the only change that is blocking this checkin is the change to SupportFiles.cs.   (the rest are suggestions for future work unrelated to this update).   

The change to SupportFiles.cs was not meant to be burdensome.     Basically it is
 
1. Renaming ExePath to MainAssemblyPath.   This is a trivial thing to do, and will not break anything.
2. Make up a new ExePath property that uses GetEntryAssembly() (and requires no special casing).  Use that when re-launching.  

This seems easy.  Am I missing something?

I am OK with creating a release which includes PerfView64.exe  after checkin.

sharwell:
@vancem Thanks for the update. I think I was confusing two separate comments from before. That should be an easy change which I can do now.

sharwell:
@vancem Updated with the renamed properties and removed the special case.

vancem:
LGTM,   Thanks for doing this.

PR #195, opened 2017-05-10
Added AppVeyor build configuration

sharwell:


README.md
```diff
@@ -4,6 +4,8 @@ If you are unfamiliar with PerfView, there are [PerfView video tutorials](http:/
 As well as [Vance Morrison's blog](http://blogs.msdn.com/b/vancem/archive/tags/perfview) which also gives overview and getting 
 started information. 
 
+[![Build status](https://ci.appveyor.com/api/projects/status/fxtu3xa874whk2w0?svg=true)](https://ci.appveyor.com/project/sharwell/perfview)
```
vancem: Put a small description before this:  e.g.  Perfview uses appveyor to produce automated builds (and eventually tests?).  We will also probably need a new .md file that talks quickly about the 'high level' view of this build / test system (where the yml file is and logically what it is doing, and security issues people need to know about, links to appVeyor, what the rules are (free for open source?). etc).  

Along those lines:  what happens with forks?   How does security work?

sharwell: > Put a small description before this

I can, but note that it's not a practice I generally see. Examples:

* https://github.com/Microsoft/vs-threading
* https://github.com/dotnet/roslyn
* https://github.com/NuGet/NuGetGallery

> where the yml file is and logically what it is doing [etc.]

Considering the prevalence of AppVeyor for builds using Microsoft tool chains, I'm not sure this is necessary. While it could be added, my two concerns are the limited value it would provide and the possibility that the information would become out of date over time.

> and eventually tests

I am expecting tests to run with this pull request. After the webhooks are in place I can confirm this before the PR is merged.

> what happens with forks?

By default, nothing. Users can create their own AppVeyor accounts and connect them to their forks if they wish to do so.

> How does security work?

The build does not use any non-public information for any purpose, nor has any been exposed to it by this process. Security works by developers not putting confidential information into the repository itself. :smile:

vancem: My experience is that with documentation you should err on the side of too much rather than too little (since the bad effect of too much is small and the bad effect of too little can be great).   I agree we should not put in information that quickly gets out of date.  I am talking about a sentence on this page, and a link to a new page with what can start as very small paragraph that tells you the value.     If you are averse to this, let me know and I will do it.  

On Security:  I mainly was interested in how trustworthy the output artifacts are (do we expect people to run the output from it, and if so, what risk are they taking?).

sharwell: > My experience is that with documentation you should err on the side of too much rather than too little 

I agree, and I don't mind adding this documentation. I will put something together with the intent of being the minimal amount of "stuff to read" (on the assumption that people have short attention spans) while still conveying the important information clearly.

> I mainly was interested in how trustworthy the output artifacts are (do we expect people to run the output from it, and if so, what risk are they taking?).

The build output is attached to each build. There is no guarantee whatsoever regarding the quality or security of these builds; any member of the public who is able to submit a pull request is able to get this system to create a build artifact.

sharwell: @vancem Does the following work for you?

---

### Developer information

This repository uses [AppVeyor](https://www.appveyor.com/) to automatically build and test pull requests, which allows
the community to easily view build results. The build status reflected here is the AppVeyor build status of the
**master** branch.

[Build status](https://ci.appveyor.com/project/sharwell/perfview)

> :warning: Builds produced by AppVeyor are not considered official builds of PerfView, and are not signed or otherwise validated for safety or security in any way. This build integration is provided as a convenience for community participants, but is not endorsed by Microsoft nor is it considered an official release channel in any way. For information about official builds, see [Getting PerfView]() below.

vancem: This is fine.    The one thing that would be good to change is to move the whole thing to right before the 'Building PerfView yourself'.    I realize that it makes finding the status a bit harder to find, but I also think I does make the most sense there (before this it was about using PerfView, after it is a building it yourself, this section is half way (getting daily builds).

sharwell: Thanks, that's part of the reason I asked before pushing an update to the PR. 😄

sharwell:
@vancem I updated the README file, but we'll need to set up the webhook prior to testing that everything works as expected.

sharwell:
Marked this **in progress** again since I do want to see a passing build prior to merge.

sharwell:
@vancem Do you have an update on the webhook for this (details were sent in email 5/10)? I'd like to get the build automated with tests and then add code coverage results using codecov.io so I can start looking for ways to add targeted tests for new things I'm working on. 😄

vancem:
I have created the webhook.

sharwell:
@vancem If you click the AppVeyor details button, you'll see one test failed for Visual Studio 2015 (Debug and Release), while all tests pass for 2017. Do you have any information about what might cause the test **TraceEventTests.GeneralParsing.ETW_GeneralParsing_Basic** to fail?

vancem:
The output gives you detail.  In this case 

 ERROR: File .\unzipped\net.4.6.2.x86.etl: event not equal to expected on line 850

      Expected: EVENT 4,548.216: MSNT_SystemTrace/SystemConfig/Defragmentation PID=24464; TID=13020; PName=PerfView; ProceNum=3; DataLen=228; AlignmentClusters=0; AvgFreeSpaceSize=0; ClustersPerSlab=0; FragmentedDirectoryExtents=0; FragmentedExtents=0; FreeSpaceCount=0; LargestFreeSpaceSize=0; LastRunActualPurgeClusters=0; LastRunClustersTrimmed=0; LastRunFullDefragTime=0; LastRunTime=131,078,897,202,240,...; MFTSize=517,734,400; TotalClusters=0; TotalUsedClusters=0; AvgFragmentsPerFile=0; BytesPerCluster=4,096; DirectoryCount=0; FragmentedDirectories=0; FragmentedFiles=0; FragmentedSpace=100; HardwareIssue=0; InUseMFTRecords=0; InUseSlabs=70; LastRunActualPurgeSlabs=0; LastRunInitialBackedSlabs=99; LastRunPercentFragmentation=100; LastRunPinnedSlabs=0; LastRunPotentialPurgeSlabs=0; LastRunSpaceInefficientSlabs=0; LastRunTrimmedSlabs=712; LastRunUnknownEvictFailSlabs=0; LastRunVolsnapPinnedSlabs=0; MFTFragmentCount=0; MovableFiles=0; TotalMFTRecords=0; TotalSlabs=99; UnmovableFiles=0; VolumeId=863c45e9-0000-0000-0...; VolumePathNames=C:\; 

      Actual  : EVENT 4,548.216: MSNT_SystemTrace/SystemConfig/Opcode(31) PID=24464; TID=13020; PName=PerfView; ProceNum=3; DataLen=228; 

-----------------

What is going on is that The MSNT_SystemTrace/SystemConfig/Defragmentation got parsed when the baseline was created, but was unknown when the test was run.  (It is Opcode(31).  

This is likely caused by the fact that the events are parsed by asking the OS, and the test were run on an older OS.   

I will fix the test to exclude events that we don't KNOW are supported on all OSes.

sharwell:
@vancem I'll rerun the test later after I see the update :)

vancem:
I have fixed the OS version problem above with PR #204 
The check above now seems to fail because a timeout fetching a Nuget package.

sharwell:
@vancem After this is merged, AppVeyor will cache the NuGet packages, so it will be less susceptible to these failures. I restarted the build in the meantime.

vancem:
I am going to merge this, as making it 'official' will help in getting information about how well it works.

Issue #196, opened 2017-05-12
ETW_GeneralParsing_Basic fails due to timezone differences


kayle:
I'm on EST, so my StartTime/EndTime are 3 hours after the expected baseline.

In EventTraceHeaderTraceData:
```
public DateTime EndTime { get { return ETWTraceEventSource.SafeFromFileTimeUtc(EndTime100ns).ToLocalTime(); } }
```

```
ERROR: File .\unzipped\sl.5.x86.etl: event not equal to expected on line 1
   Expected: EVENT 0.000: Windows Kernel/EventTrace PID=11636; TID=19564; PName=PerfView; ProceNum=0; DataLen=332; BufferSize=65,536; Version=83,951,626; ProviderVersion=10,586; NumberOfProcessors=4; EndTime=13:10:31.290843 (9,3...; TimerResolution=156,250; MaxFileSize=0; LogFileMode=65,537; BuffersWritten=631; StartBuffers=1; PointerSize=8; EventsLost=0; CPUSpeed=2,993; BootTime=5/14/2016 3:08:46 AM; PerfFreq=10,000,000; StartTime=13:10:21.975648 (0.0...; ReservedFlags=1; BuffersLost=0; SessionName=Relogger; LogFileName=[multiple files]; UtcOffsetMinutes=480; 
   Actual  : EVENT 0.000: Windows Kernel/EventTrace PID=11636; TID=19564; PName=PerfView; ProceNum=0; DataLen=332; BufferSize=65,536; Version=83,951,626; ProviderVersion=10,586; NumberOfProcessors=4; EndTime=16:10:31.290843 (9,3...; TimerResolution=156,250; MaxFileSize=0; LogFileMode=65,537; BuffersWritten=631; StartBuffers=1; PointerSize=8; EventsLost=0; CPUSpeed=2,993; BootTime=5/14/2016 6:08:46 AM; PerfFreq=10,000,000; StartTime=16:10:21.975648 (0.0...; ReservedFlags=1; BuffersLost=0; SessionName=Relogger; LogFileName=[multiple files]; UtcOffsetMinutes=480; 
```

vancem:
Fixed with PR #202

PR #202, opened 2017-05-15
Make tests work in all time zones

vancem:
The GenerParsing test had DateTimes that were specific to a time zone, change to UTC to work anywhere.


vancem:
This fixes issue #196

Issue #197, opened 2017-05-12
Fix HistoryDictionary to avoid slow lookup times when a handle is reused alot. 


vancem:
Submitted by Calvin Hsia

The following from the following directory

\\ddfiles\Team\XAML\User\erikd\Bugs\398113\Run4886725\PhotoSharingApp_NoUserCode-05-11-2017 022827

And it will be very slow.   Much time is spend in HistoryDictionary.Add.   

The issue seems to be that there are MANY reuses of a particular file handle and HistoryDictionary creates a long linked list in this case.

******************

Note that historyDictionary was designed for RARE reuse (like process IDs thread IDs etc).   For file handles we may want to use something else instead.

However I would like to see if a small tweek will fix things.    We are very likely to both add and fetch things in chronological order.   If we keep a one element cache to the last access, then it should speed things up dramatically for a very small effort.

vancem:
Fixed by PR #203

PR #203, opened 2017-05-15
Speed up HistoryDictionary

vancem:
HistoryDictionary keeps its values in a sorted linked list.  This can get long and sow if there are many entires.
We normally only access them sequentially, however so we can speed up access by remembering
the last entry we used and start from there if we can.


vancem:
This fixes issue #197

PR #209, opened 2017-05-16
Use codecov.io for code coverage

sharwell:


sharwell:
Something is hanging in the tests... will look into it and update this later.

Issue #198, opened 2017-05-13
CopyBinariesToOutDir in PerfView.SupportFiles does I/O at inappropriate times


rainersigwald:
The NuGet package `PerfView.SupportFiles` defines a target that copies some package-delivered binaries to the build's output folder:

```csproj
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <Target Name="CopyBinariesToOutDir" BeforeTargets="ResolveAssemblyReferences" Condition="('$(DesignTimeBuild)' != 'true')">
	<Copy SourceFiles="$(MSBuildThisFileDirectory)..\lib\native\x86\sd.exe" DestinationFolder="$(OutDir)x86"/>
	<Copy SourceFiles="$(MSBuildThisFileDirectory)..\lib\native\x86\DiagnosticsHub.Packaging.dll" DestinationFolder="$(OutDir)x86"/>
	<Copy SourceFiles="$(MSBuildThisFileDirectory)..\lib\native\x86\DiagnosticsHub.Packaging.Interop.dll" DestinationFolder="$(OutDir)x86"/>
	<Copy SourceFiles="$(MSBuildThisFileDirectory)..\lib\net40\Microsoft.Diagnostics.Tracing.EventSource.dll" DestinationFolder="$(OutDir)"/>
	<Copy SourceFiles="$(MSBuildThisFileDirectory)..\lib\net40\Microsoft.DiagnosticsHub.Packaging.InteropEx.dll" DestinationFolder="$(OutDir)"/>
    </Target>
</Project>
```

This goes against convention by doing disk operations before ResolveAssemblyReference. The `DesignTimeBuild` condition should keep it from firing during most VS operations (I actually came here to suggest adding just that condition), but it evidently doesn't keep it from happening in _enough_ situations.

This is the cause of the VS UI hang reported by @sharwell at https://github.com/Microsoft/msbuild/issues/2100.

rainersigwald:
I tried changing the condition to `'$(BuildingProject)' == 'true'`, which seems to work better. But while I'm here I'll propose a bit of an overhaul in the target.

* Do you need to explicitly copy the `lib/` assemblies? I would expect them to be copied from the package by virtue of being `<Reference>`d in the project.
* The copy operations could be done in bulk to reduce a tiny bit of overhead.
* Unless I'm misunderstanding, this target can run much later in the build, around the same time other copy-to-output-directory targets run.

I think something like this would do the trick:

```xml
[...]

PR #210, opened 2017-05-16
Simplify the logic associated binary Support Files

vancem:
There are two nuget packages that we used to carry binary files that are needed
by perfview but are not built by this repo.    There was some copy logic in those
that was causing grief for Visual Studio.   Simplify this to avoid the need for the
copy.

Upgraded these support nuget packages to version 1.0.2.   This is mostly a no-op (the binaries did not change, the the possible exception of small changes to OSExtentions.dll)
but I needed to update these nuget packages to simplify them.

Also unbroke the logic for generating these support files.

Finally removed some .SLN files that have not been used in a long time and are probably confusing.


vancem:
This fixes issue #198

src/NugetSupportFiles/Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles.Populate.bat
```diff
@@ -1,5 +1,5 @@
 REM copies from an existing build to a nuget package creation area (so that *.MakeNuget.bat works)
-xcopy /s ..\PerfView\packages\Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles.1.0.1\*.dll Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles
+xcopy /s ..\..\packages\Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles.1.0.1\*.dll Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles
```
sharwell: ❓ Should this be 1.0.**2**?

src/NugetSupportFiles/PerfView.SupportFiles.Populate.bat
```diff
@@ -1,6 +1,6 @@
 REM copies from an existing build to a nuget package creation area (so that *.MakeNuget.bat works)
-xcopy /s /y ..\PerfView\packages\PerfView.SupportFiles.1.0.1\*.dll PerfView.SupportFiles
-xcopy /s /y ..\PerfView\packages\PerfView.SupportFiles.1.0.1\*.exe PerfView.SupportFiles
+xcopy /s /y ..\..\packages\PerfView.SupportFiles.1.0.1\*.dll PerfView.SupportFiles
```
sharwell: :memo: Same question regarding 1.0.**2**.

src/NugetSupportFiles/Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles.Populate.bat
vancem: Well, at the time this was used, there did not exist a 1.0.2 to copy from (that was what was being created).    Still I can update it now, and I added some more comments / docs to make it less likely people mess this up.

vancem:
The following test fails  
  LinuxTracing.Tests.EventParseTests.OneStack [FAIL]

Because of a file-in-use error 

      System.IO.IOException : The process cannot access the file 'C:\projects\perfview\src\LinuxEvent.Tests\bin\Release\Sources\onegeneric.perf.data.dump' because it is being used by another process.
      Stack Trace:
           at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
           at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
           at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy)
           at System.IO.FileStream..ctor(String path, FileMode mode)
        src\PerfView\OtherSources\Linux\LinuxPerfScriptStackSource.cs(414,0): at Diagnostics.Tracing.StackSources.LinuxPerfScriptStackSource.GetPerfScriptStream(String path, ZipArchive& archive)
---------
Which seems to be unrelated to the checkin.   Should followup on this error eventually.

PR #217, opened 2017-05-17
Use System.IO.Compression instead of duplicating it

sharwell:


vancem:
Note that the whole CSVReader subdirectory is probably going to be abandoned.   This functionality was added very early on as a clone of an earlier tool called ETLStackBrowse, and was added because it allowed ETL data to be parsed with XPERF but read with PerfView for easy comparison.   This was useful early on because PerfView was had bugs or did not have all the functionality of ETLStackBrowse.    It has not been used in an non-trivial way in years (to my knowledge), but it was doing no harm, so we left it, but did not invest in it.

At some point it will be a pain and we will remove it.     

What testing did you do?    There is no automated testing of this code, but it did work the last time I checked.    You can create file that it operates on (a *.CSV.ZIP) with the \\clrmain\tools\xperfRun.exe command.   Assuming the most basic tests with such a file work (you can look at a stack view an the 'events' view), I am OK with this change.   

A readme.md conflict also would need to be resolved (trivial)

sharwell:
Tentatively marking this as blocked on #209. If the feature is removed altogether then it wouldn't make such a difference.

Issue #213, opened 2017-05-16
TextEditorWindow: resizing moves the text around too much


BruceForstall:
When choosing "Goto Source (Def)", by default a small TextEditorWindow is opened with the output. First, the default size is too small (IMO). Second, when resizing this window, the original line of focus (presumably the line with the highest hit count of events) becomes lost. Presumably this is because wrapped lines become not wrapped, and the "line to display" gets messed up. Sometimes it's hard to find where that line of focus disappeared to.

brianrob:
Fixed via #219.

PR #219, opened 2017-05-17
Make text editor windows bigger by default.

vancem:
This makes issue #213 less severe in a trivial way 

@BruceForstall


BruceForstall:
LGTM

PR #221, opened 2017-05-17
Change CommandToRun from a text box to a history text box

BruceForstall:
Thus, the most recently run commands will be available for
easy re-use.

### DIFF
--- a/src/PerfView/Dialogs/RunCommandDialog.xaml.cs
+++ b/src/PerfView/Dialogs/RunCommandDialog.xaml.cs
@@ -64,11 +64,18 @@ namespace PerfView
             m_mainWindow = mainWindow;
 
             CurrentDirTextBox.Text = Environment.CurrentDirectory;
+
+            // Initialize the CommandToRun history if available. 
+            var commandToRunHistory = App.ConfigData["CommandToRunHistory"];
+            if (commandToRunHistory != null)
+                CommandToRunTextBox.SetHistory(commandToRunHistory.Split(';'));
+
             if (args.CommandLine != null)
             {
                 CommandToRunTextBox.Text = args.CommandLine;
-                CommandToRunTextBox.SelectAll();
+                // CommandToRunTextBox.SelectAll();
             }
+
             DataFileNameTextBox.Text = args.DataFile;
             RundownTimeoutTextBox.Text = args.RundownTimeout.ToString();
             SampleIntervalTextBox.Text = args.CpuSampleMSec.ToString();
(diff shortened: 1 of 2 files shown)


src/PerfView/Dialogs/RunCommandDialog.xaml.cs
```diff
@@ -64,11 +64,18 @@ public RunCommandDialog(CommandLineArgs args, MainWindow mainWindow, bool isColl
...
+            // Initialize the CommandToRun history if available. 
+            var commandToRunHistory = App.ConfigData["CommandToRunHistory"];
+            if (commandToRunHistory != null)
+                CommandToRunTextBox.SetHistory(commandToRunHistory.Split(';'));
+
             if (args.CommandLine != null)
             {
                 CommandToRunTextBox.Text = args.CommandLine;
-                CommandToRunTextBox.SelectAll();
+                // CommandToRunTextBox.SelectAll();
```
BruceForstall: Apparently ComboBox doesn't have SelectAll but TextBox does. Not sure if it matters.

src/PerfView/Dialogs/RunCommandDialog.xaml.cs
```diff
@@ -64,11 +64,18 @@ public RunCommandDialog(CommandLineArgs args, MainWindow mainWindow, bool isColl
             m_mainWindow = mainWindow;
 
             CurrentDirTextBox.Text = Environment.CurrentDirectory;
+
+            // Initialize the CommandToRun history if available. 
+            var commandToRunHistory = App.ConfigData["CommandToRunHistory"];
+            if (commandToRunHistory != null)
+                CommandToRunTextBox.SetHistory(commandToRunHistory.Split(';'));
```
BruceForstall: I suppose I could change the name of `CommandToRunTextBox` to `CommandToRunHistoryComboBox` (which seems quite verbose), or just `CommandToRunControl`. Comments?

BruceForstall:
@vancem PTAL

src/PerfView/Dialogs/RunCommandDialog.xaml.cs
vancem: I prefer the current TextBox suffix (that is what we do elsewhere).

src/PerfView/Dialogs/RunCommandDialog.xaml.cs
vancem: The SelectAll just lets you delete the command quickly (since it is already selected).   
We can implement it trivially (it is implemented by PerfView), but I am OK with dropping it (remove the line).

src/PerfView/Dialogs/RunCommandDialog.xaml.cs
```diff
@@ -295,7 +302,25 @@ private void OKButtonClick(object sender, RoutedEventArgs e)
...
+                {
                     m_args.CommandLine = CommandToRunTextBox.Text;
+
+                    if (CommandToRunTextBox.AddToHistory(m_args.CommandLine))
+                    {
+                        StringBuilder sb = new StringBuilder();
+                        foreach (string item in CommandToRunTextBox.Items)
+                        {
+                            if (item != "")
+                            {
```
vancem: I worry if the command had a ; in it (although that is weird).  rather than implementing an escape mechanism, I would recommend simply putting !item.Contains(';') (thus we don't store in history things with ';').

BruceForstall:
@vancem Updated to respond to CR feedback.

PR #222, opened 2017-05-17
Faster testing

sharwell:
Improve testing speed and granularity

* Use `StringComparer.Ordinal` instead of the current culture
* Use `[Theory]` instead of `[Fact]` to separate a test with several sub-tests
* Fix test output not showing up when run from **Test Explorer** in Visual Studio (only covers one test where I was trying to review the output)


vancem:
LGTM

PR #223, opened 2017-05-17
Fix artifact path for debug builds

sharwell:
Currently build artifacts are only getting uploaded for the Release configuration jobs. This change updates the AppVeyor definition to include artifacts for all build jobs.


sharwell:
:memo: Verified the build output for the AppVeyor job before merging

Issue #201, opened 2017-05-14
Implement an object finalization tree view


sharwell:
This tree would allow the user to see how many objects of each type are getting finalized by each process. This would be a histogram on the process and type for the events described in #200.

I managed to extract this information manually using a pivot table in excel and looking up all the type names I could find in the **Microsoft-Windows-DotNETRuntime/GC/AllocationTick** event, but it was slow, painful, and missing a great number of type names.

stephentoub:
Just to clarify, you mean a tree view instead of or in addition to the table / Excel export it currently can generate?

sharwell:
@stephentoub If you are referring to the table on the GCStats page, yes. I'm not sure how that table works exactly, but it appeared to be very limited in its ability¹ and I question its accuracy². 

¹ There is no ability to restrict the results to a particular range of time, and the results are not available unless a special mode is enabled. Why are the results not available from just the **FinalizeObject** event?

² The results do not at all align with information observed in the Events table for the FinalizeObject event described in #200.

vancem:
Note that the current logic in the GCStats view works with the PRIVATE event (see my comment in #200 for more on ClrPrivate events).     

As suggested in #200, we could fix this by making the FinalizeObject event 'smart'  enough to use rundown information to look up the TypeName from the TypeId    When that is done we should also fix the GCStats view to use this new information.   

That would probably be sufficient to fix this issue.     I am hesitant to introduce new views that are truly new, so we should evaluate after that if the Events view and the GCStats view (and maybe some doecs) would be sufficient.   We may also wish to have a link on (either in the file view or in the GCStats view), that opens the 'Events' view with the proper filtering to make that discoverable.

sharwell:
@vancem The reason I was suggesting a new view is the following:

1. The GCStats view buckets events by process (good), but doesn't allow filtering the view to a specific start/stop point. The latter limitation would prevent me from using this view in a meaningful way during the investigation of a recent regression (it's linked above).
2. The Events view shows individual events, but it doesn't show a histogram. If the Events view were modified to allow a "group by" setting on one or more columns, it could certainly serve this purpose. Note that in the event too many event instances match the current filter, the group headings would need to show the number of visible events in the group, and the total number of events in the group.

For example, if we Group By: TypeID, you might have this:

* 0x00000001 (3 of 1005 events shown)
  * Event 1
  * Event 2
  * Event 3
* 0x0029993a (400 of 1000 events shown)
  * Event 5
  * Event 6
[...]

PR #224, opened 2017-05-18
Use the CLR GCFinalizeObject event to report finalized object counts

sharwell:
Improves the ability of the GCStats page to show accurate GC counts, and removes the requirement that the `/Finalizers` option be used. This addresses one of the minor items mentioned in the comments of #201.

### DIFF
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -1010,6 +1010,59 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                             finalizationCount + 1 :
                             1;
                 };
+
+                source.Clr.TypeBulkType += data =>
+                {
+                    var stats = currentManagedProcess(data);
+                    var typeMapping = stats.GC.m_stats.TypeMapping;
+
+                    var clonedData = data;
+                    clonedData = (GCBulkTypeTraceData)clonedData.Clone();
+
+                    var count = clonedData.Count;
+                    for (int i = 0; i < count; i++)
+                    {
+                        var bulkTypeValues = clonedData.Values(i);
+                        var typeId = bulkTypeValues.TypeID;
+                        if (typeMapping.ContainsKey(typeId))
+                            continue;
+
+                        typeMapping[typeId] = bulkTypeValues;
+                    }
+                };
+
+                source.Clr.GCFinalizeObject += data =>
+                {
+                    var stats = currentManagedProcess(data);
+                    long finalizationCount;
+                    stats.GC.m_stats.FinalizedObjectsSlim.TryGetValue(data.TypeID, out finalizationCount);
+                    stats.GC.m_stats.FinalizedObjectsSlim[data.TypeID] = finalizationCount + 1;
+                };
+
+                source.Completed += () =>
+                {
+                    foreach (var process in source.Processes())
+                    {
+                        var stats = process.LoadedDotNetRuntime();
+                        if (stats == null)
+                            continue;
+
+                        if (stats.GC.m_stats.FinalizedObjectsSlim.Count == 0)
+                            continue;
+
+                        var typeMapping = stats.GC.m_stats.TypeMapping;
+                        if (typeMapping.Count == 0)
+                            continue;
+
+                        stats.GC.m_stats.FinalizedObjects.Clear();
+                        var typeNameCache = new Dictionary<Address, string>();
+                        foreach (var pair in stats.GC.m_stats.FinalizedObjectsSlim)
+                        {
+                            var typeName = GetTypeName(pair.Key, typeNameCache, typeMapping);
+                            stats.GC.m_stats.FinalizedObjects[typeName] = pair.Value;
+                        }
+                    }
+                };
             }
 
             //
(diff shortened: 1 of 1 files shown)


src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -1010,6 +1010,59 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
...
+                        var bulkTypeValues = clonedData.Values(i);
+                        var typeId = bulkTypeValues.TypeID;
+                        if (typeMapping.ContainsKey(typeId))
+                            continue;
+
+                        typeMapping[typeId] = bulkTypeValues;
+                    }
+                };
+
+                source.Clr.GCFinalizeObject += data =>
```
sharwell: ❓ @vancem What is the relation between this added line and [this line](https://github.com/Microsoft/perfview/blob/06fb35cf868f32a730ed5b74c4e426d76f6475ee/src/TraceEvent/TraceLog.cs#L880)?

vancem: Roughly speaking there is no relationship between this code and the code you referred to.   The code you referenced is in the 'first pass' when ETL is converted t ETLX, and the 'doNothing' delegate is there only to insure that during this first pass we parse the event during that pass.   The code you have here is an independent pass over the events.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -1230,6 +1283,47 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
...
+            string typeName;
+            if (typeNameCache.TryGetValue(typeID, out typeName))
+                return typeName;
+
+            GCBulkTypeValues bulkTypeValues;
+            if (typeMapping.TryGetValue(typeID, out bulkTypeValues))
+            {
+                typeName = bulkTypeValues.TypeName;
+                var parameterCount = bulkTypeValues.TypeParameterCount;
+                if (parameterCount > 0)
```
sharwell: 📝 This is currently producing results like this:

```
System.WeakReference`1[System.Diagnostics.Tracing.EtwSession]<System.Diagnostics.Tracing.EtwSession>
```

❓ It seems `GCBulkTypeValues.TypeName` includes generic arguments already, but in a different form. Which of the following would you prefer the result be?

1. ``System.WeakReference`1[System.Diagnostics.Tracing.EtwSession]``
2. `System.WeakReference<System.Diagnostics.Tracing.EtwSession>`

If I go with the first approach, I can change `GCStats.TypeMapping` to just a `Dictionary<Address, string>` and avoid cloning the trace event. But the performance overhead didn't seem to necessitate one way or the other.

vancem:
I am conflicted about this change.    Basically you fixed things by making GCStats 'smart' and know about the TypeID->TypeName mapping, but you can see it is a pain, and anyone making use of GCFinalizeObject. events would have to clone similar logic.    This argues that the GCFinalizeObject should 'fix' the problem for all clients, rather than forcing it on the clients to do this.  

Of course the problem is that you extra information (from two passes) to do this.   However this problem has occurred before (we have the same issue with FileNames, Registry names, Disk I/O and stacks), and this is basically what the ETLX file is about (storing useful information discovered in a pass over the data, and indexing it so that you can get it when you need it.  

I think it would be better to fix is in this way (Basically GCFinalizeObject would have a 'TypeName' property which 'works' if the data is coming from and ETLX file.   

Rather than trying to explain how this works, I think it is faster to simply do it and show you a PR.    I will be out of the office until next Tuesday, but at that time either I will have a counter-proposal, or I will let this PR get merged.

sharwell:
@vancem Keep in mind it's not uncommon for me to run up against the file size limitations of ETLX files, so we should make sure the overhead this introduces is minimal. I'm hoping to find a way to remove the 20M object limit that currently gets applied in some cases.

src/TraceEvent/Computers/TraceManagedProcess.cs
vancem: You have basically uncovered a bug in the GCBulkType event.    However there is a larger issue of what the proper format for type names is.  format (1) is what the runtime uses internally and for reflection (it is independent of any particular language), and (2) is of course the C# syntax for it.    

Now the question is what should PerfView display?   I would argue that displaying the .NET reflection syntax will just confuse most people.    On the other hand PerfView mostly just displays strings that it was given and thus .NET Reflection syntax is basically the standard used today in PerfView (you will see it in overloads and other things).   Thus we have a policy question (do we care more about being pretty or being consistent?) 

My recommendation is the following: start with the easiest thing (that is fix the bug, and make it easy to do policy changes later).    Handle the policy question as a separate item.  This argues for not displaying C# conventions that is go with (1), at least in the short term. 

The way to do this matter.  We should fix it so that we fix the bug as early as possible, and in a way that allows us to fix the runtime events and not break.    

I believe that GCBulkTImeValue.TypeName should not have generic parameters but should otherwise conform to .NET runtime standards (it does not try to demangle the name). Thus it should return System.WeakReference`1.   We should fix this property so that if it notices a bug, it fixes it (if you have generic parameters it strips the [...] suffix.  

We should create a method 'FullTypeName' that generates (1).   This should be documented to return the .NET Runtime convention for types.   

Optionally we could create a CSharpFullTypeName that generates (2), but we only need this if we change policy.   For now we can skip it.    Logically this is a helper method (users could have defined it, but it is likely to be useful to many clients so we should provide it).     

We should log a bug in .NET Core to fix the GCBulkType event.  

Note that all of this is (logically) independent of where the ID->Name mapping is done.

vancem:
> @vancem Keep in mind it's not uncommon for me to run up against the file size limitations of ETLX files

The approach I am suggesting does work on raw ETL files and real time sessions, however it does have to live with the constraints that data later in the file is not available until you scan it (or you do two passes).  

In any case, I think the problems of ETLX size limitations and this issue are reasonably independent, and you are probably tied to ETLX in other ways already (if you are using PerfView in normal ways, you have a heavy dependency).   

I am just arguing that we should try to fix similar problems in similar ways (and if those similar ways have limitations, fix those once for all).

sharwell:
@vancem Thanks for the detailed feedback. I'm still trying to wrap my head around everything so it may take a few tries before I manage to piece together things that everyone likes 😁

sharwell:
Superseded by #237

PR #226, opened 2017-05-20
Fix heap dump when running PerfView as a 64-bit process

sharwell:
Currently the heap dump for a 32-bit process only works when running PerfView as a 32-bit process. This pull request updates the code launching HeapDump to work with both PerfView and PerfView64.


vancem:
LGTM

PR #227, opened 2017-05-20
Fix build warnings

sharwell:
💭 In the future we can consider alternatives to `[Obsolete]` for experimental features, but for now it seemed to make sense to clean up the build.

### DIFF
--- a/src/PerfView/Extensibility.cs
+++ b/src/PerfView/Extensibility.cs
@@ -2843,6 +2843,7 @@ namespace PerfViewExtensibility
         /// Mainly here for testing
         /// </summary>
         /// <param name="dllName"></param>
+        /// <param name="ILPdb"></param>
         public void LookupSymbolsFor(string dllName, string ILPdb=null)
         {
             var symbolReader = App.GetSymbolReader();
@@ -3096,9 +3097,9 @@ namespace PerfViewExtensibility
                 OpenHtmlReport(outputFileName, "JITStats report");
         }
 
-        /// <summary>
-        /// Given a PDB file, dump the source server information in the PDB file.
-        /// </summary>
+        ///// <summary>
+        ///// Given a PDB file, dump the source server information in the PDB file.
+        ///// </summary>
         //public void DumpSourceServerStream(string pdbFile)
         //{
         //    SymbolReader reader = new SymbolReader(LogFile);
--- a/src/PerfView/PerfView.csproj
+++ b/src/PerfView/PerfView.csproj
@@ -49,7 +49,7 @@
     <UseVSHostingProcess>false</UseVSHostingProcess>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <DocumentationFile>bin\Debug\PerfView.XML</DocumentationFile>
-    <NoWarn>1591,0436</NoWarn>
+    <NoWarn>1591,0436,0618</NoWarn>
     <Prefer32Bit>true</Prefer32Bit>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
@@ -63,7 +63,7 @@
     <UseVSHostingProcess>false</UseVSHostingProcess>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <DocumentationFile>bin\Release\PerfView.XML</DocumentationFile>
-    <NoWarn>1591,0436</NoWarn>
+    <NoWarn>1591,0436,0618</NoWarn>
     <GenerateSerializationAssemblies>Auto</GenerateSerializationAssemblies>
     <Prefer32Bit>true</Prefer32Bit>
   </PropertyGroup>
--- a/src/PerfView/PerfView.csproj
+++ b/src/PerfView/PerfView.csproj
@@ -123,9 +123,7 @@
     <Compile Include="OtherSources\Linux\LinuxPerfScriptEventParser.cs" />
     <Compile Include="OtherSources\Linux\LinuxPerfScriptStackSource.cs" />
     <Compile Include="OtherSources\XmlTreeStackSource.cs" />
-    <Compile Include="sigparser.cs">
-      <Link>Utilities\sigparser.cs</Link>
-    </Compile>
+    <Compile Include="sigparser.cs" />
     <Compile Include="..\heapDump\GCHeapDump.cs">
       <Link>memory\GCHeapDump.cs</Link>
     </Compile>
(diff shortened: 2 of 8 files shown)


src/PerfView/Extensibility.cs
```diff
@@ -3620,7 +3621,9 @@ public static MutableTraceEventStackSource ThreadTimeStacks(this TraceLog eventL
         var stackSource = new MutableTraceEventStackSource(eventLog);
         stackSource.ShowUnknownAddresses = App.CommandLineArgs.ShowUnknownAddresses;
 
+#pragma warning disable CS0618 // Type or member is obsolete
```
vancem: I don't really want lots of #pragma warning disable statements littered throughout the code.    This is a case where I think the cure is worse then the disease.      The warnings are meant to force us to fix things, and suppressing them is going in the wrong direction and making the code uglier.  

We can either
1. Fix the code not to use the Obsolete thing (may be hard)
2. If the obsolete was marking something experimental, determine if the Obsolete attribute can be removed. 
3. Make a wrapper with one #pragma warning disable in the wrapper
4. If there really is only one call site and (1) is not an option, it is OK to use #pragma warning disable. 
5. Simply disable CS0618 and log a issue to do options 1-4 and turn it back on.  
6. Just live with the warnings.  (they remind us of unfinished work).

src/PerfView/HeapView/GcInfoView.cs
```diff
@@ -129,6 +129,7 @@ public double PauseDuration
             }
         }
 
+        [Obsolete("This is experimental, you should not use it yet for non-experimental purposes.")]
```
vancem: Adding @jeffschwMSFT how owns TraceGC.  

Jeff: we made TraceGC experimental by marking most of its methods Obsolete with the special message.   
Do we feel comfortable making most of it at least non-experimental?   We should discuss if there is overlap/conflict in the current APIs.

src/PerfView/HeapView/HeapDiagram.cs
```diff
@@ -696,10 +696,12 @@ void SetYScale(Renderer render)
 
             foreach (TraceGC gc in m_data.events)
             {
+#pragma warning disable CS0618 // Type or member is obsolete
```
vancem: As mentioned above, one way or the other, I don't want most of these pragmas here.

src/PerfView/Extensibility.cs
sharwell: My preference would be to replace all such `[Obsolete]` items with one of the following:

1. Use [`<preliminary/>`](http://ewsoftware.github.io/XMLCommentsGuide/html/c16bece7-694e-48ca-802d-cf3ae9205c55.htm) in a documentation comment (requires maintenance and attention to assembly versioning)
1. Do nothing

If we keep using `[Obsolete]` for this manner, which it clearly wasn't designed for, my preference would be to go with the approach used in this pull request since technically that is adhering precisely to what was requested in the first place. The options with wrappers (3 and 4 in list) aren't really options because then we're letting the fact that something is new influence the design in ways that will carry over even when it's not new. That either increases design debt over time (if the wrappers stay) or increases our risk of breaking something during stabilization (if wrappers are removed).

src/PerfView/PerfView.csproj
```diff
@@ -80,9 +80,9 @@
       <Private>True</Private>
       <EmbedInteropTypes>True</EmbedInteropTypes>
     </Reference>
-    <Reference Include="Interop.Dia2Lib, Version=2.0.0.0, Culture=neutral, processorArchitecture=MSIL">
+    <Reference Include="Dia2Lib, Version=2.0.0.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\..\packages\Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles.1.0.2\lib\net40\Interop.Dia2Lib.dll</HintPath>
-      <EmbedInteropTypes>True</EmbedInteropTypes>
+      <EmbedInteropTypes>False</EmbedInteropTypes>
```
vancem: Can you explain the rationale for not embedding the interop types for this DLL?   We do it for others, and seems like generally a good idea.

sharwell: Embedding interop types is only really useful if you can avoid shipping the interop assembly. In this case, the compiler was producing a warning saying the reference could not be eliminated, so I set this back to false to reduce the assembly size.

src/PerfView/Triggers.cs
```diff
@@ -189,9 +189,11 @@ public override void Dispose()
         {
             m_monitoringDone = true;
 
+#if PERFVIEW
```
vancem: Outside of TraceEvent, we should not need the PERFVIEW define.   These defines date to a time when the code was shared with Visual Studio, but we don't do that anymore.    Thus I don't want to add more (I would like to get rid of it).     Thus I would like to know what forced you to use it here.   With luck we can simply not have it.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -1231,8 +1231,6 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
         }
 
         private Version runtimeVersion;
```
vancem: I assume these were unused.

src/PerfView/Triggers.cs
sharwell: `m_cmd` was only set when this is defined (another part of the same file).

src/TraceEvent/TraceEvent.csproj
```diff
@@ -41,7 +40,7 @@
     <WarningLevel>3</WarningLevel>
     <DocumentationFile>bin\Debug\Microsoft.Diagnostics.Tracing.TraceEvent.xml</DocumentationFile>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <NoWarn>0649 0618</NoWarn>
+    <NoWarn>0649,0618</NoWarn>
```
vancem: As mentioned if we can't get rid of them other ways, I would prefer the blanked NoWarn on the obsolete flag over many #pragma no warns.

src/TraceEvent/Computers/TraceManagedProcess.cs
sharwell: :arrow_right: Yes that's correct.

src/TraceEvent/TraceEvent.csproj
sharwell: Note that this is a bug fix. It appears the project already intended to disable CS0618, but it wasn't applied because the syntax for `<NoWarn>` is a comma-separated list.

sharwell:
@vancem I'm planning to move commit 5a66428 from this PR to a separate PR, and remove it from this one.

src/TraceEvent/TraceEvent.csproj
sharwell: @vancem The #pragma is no more in this PR. Any objection to adding a `<NoWarn>` for the projects which still report that warning and filing an issue to address it later?

src/PerfView/Triggers.cs
sharwell: ➡️ Filed #241 to address the definition's existence separately.

jeffschwMSFT:
@sharwell any chance you could merge the commits?  Vance added me to a change comment, but I am having a hard time finding it.  WRT using obsolete as a means to indicate experimental, I am OK ingeneral with removing them.  There were a few that are there really only to support PerfView and we should come up with another strategy to help identify those (I saw the comments where y'all were discussing options).

sharwell:
@jeffschwMSFT The comment appears above but it's now collapsed since I removed the commit that contained the original change:

image

> any chance you could merge the commits?

I can, but my preference is always to keep logically separate commits separate. For this PR, each commit is an independent unit that would successfully build and test.

src/TraceEvent/TraceEvent.csproj
vancem: I am not quite sure what you are proposing, but I am fine with the general principle.

sharwell: @vancem Added commit d1251f3.

vancem:
I am going to pull out the hasProcessEvents variable since it is no longer used, but I will do that after merge.

PR #228, opened 2017-05-20
Simplify handling of processor architecture

sharwell:
:memo: I'm going to make a few notes in the diff with questions on this one.

### DIFF
--- a/src/HeapDump/CrossGenerationLiveness/CrossGenerationLivenessCollector.cs
+++ b/src/HeapDump/CrossGenerationLiveness/CrossGenerationLivenessCollector.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Reflection;
 using System.Runtime.InteropServices;
 using Microsoft.Diagnostics.HeapDump;
 using Microsoft.Diagnostics.Runtime;
@@ -19,7 +20,7 @@ namespace Microsoft.Diagnostics.CrossGenerationLiveness
         private const int AttachTimeoutInMsec = 60000;
         private const int WaitForEventTimeoutInMsec = 5000;
 
-        private static string s_ProcessArch;
+        private static string s_ProcessArchDirectory;
 
         private int _PID;
         private int _GenerationToTrigger;
@@ -212,7 +213,7 @@ namespace Microsoft.Diagnostics.CrossGenerationLiveness
         /// </summary>
         public static void LoadNative(string relativePath)
         {
-            var fullPath = Path.Combine(RootDir, ProcessArch, relativePath);
+            var fullPath = Path.Combine(RootDir, ProcessArchitectureDirectory, relativePath);
             var ret = LoadLibrary(fullPath);
             if (ret == IntPtr.Zero)
             {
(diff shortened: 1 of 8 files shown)


src/HeapDump/CrossGenerationLiveness/CrossGenerationLivenessCollector.cs
```diff
@@ -223,19 +224,28 @@ public static void LoadNative(string relativePath)
         /// <summary>
         /// Get the name of the architecture of the current process.
         /// </summary>
-        private static string ProcessArch
+        private static ProcessorArchitecture ProcessArch
         {
             get
             {
-                if (s_ProcessArch == null)
+                return Environment.Is64BitProcess ? ProcessorArchitecture.Amd64 : ProcessorArchitecture.X86;
```
sharwell: ❓ Do we need to support ARM here? I figured no since HeapDump only works on AMD64 or X86.

src/HeapDump/GCHeapDumper.cs
```diff
@@ -1020,7 +1020,7 @@ private bool TryGetDotNetDump(int processID)
                     if (iCorDebug != null)
                     {
                         m_log.WriteLine("Attaching to silverlight process {0} from a {1} process.", processID,
-                            Environment.GetEnvironmentVariable("PROCESSOR_ARCHITECTURE"));
+                            Environment.Is64BitProcess ? ProcessorArchitecture.Amd64 : ProcessorArchitecture.X86);
```
sharwell: ❓ Does silverlight work on ARM? Even if it does, do we need to handle it since HeapDump doesn't work on ARM?

src/PerfView/App.cs
```diff
@@ -44,7 +44,7 @@ public static int Main(string[] args)
             try
             {
                 // Can't display on ARM because the SplashScreen is WPF
-                var noGui = SupportFiles.ProcessArch == "ARM" ||
+                var noGui = SupportFiles.ProcessArch == ProcessorArchitecture.Arm ||
```
sharwell: ❓ Do we need to keep PerfView working on ARM?

src/PerfView/Utilities/SupportFiles.cs
```diff
@@ -191,42 +191,59 @@ public static string ExePath
...
-                if (s_ProcessArch == null)
+                if (s_ProcessArch == ProcessorArchitecture.None)
                 {
-                    s_ProcessArch = Environment.GetEnvironmentVariable("PROCESSOR_ARCHITECTURE");
-                    // This should not be needed, but when I run PerfView under VS from an extension on an X64 machine
-                    // the environment variable is wrong.  
-                    if (s_ProcessArch == "AMD64" && System.Runtime.InteropServices.Marshal.SizeOf(typeof(IntPtr)) == 4)
-                        s_ProcessArch = "x86";
+                    var processorArchitecture = Environment.GetEnvironmentVariable("PROCESSOR_ARCHITECTURE");
+                    if (!Enum.TryParse(processorArchitecture, out s_ProcessArch))
```
sharwell: ❗️ This needs to be case-insensitive.

src/TraceEvent/NativeDlls.cs
```diff
@@ -25,22 +25,33 @@ public static void LoadNative(string simpleName)
         }
     }
 
-    public static string ProcessArch
+    public static ProcessorArchitecture ProcessArch
     {
         get
         {
-            if (s_ProcessArch == null)
+            return IntPtr.Size == 8 ? ProcessorArchitecture.Amd64 : ProcessorArchitecture.X86;
```
sharwell: ❓ Does TraceEvent need to support ARM?

src/HeapDump/CrossGenerationLiveness/CrossGenerationLivenessCollector.cs
vancem: We had some support for ARM ETW collection but even that rotted after the windows team requires special  signing to run things on ARM.   We never made HeapDump work.  It is fine to ignore ARM for the foreseeable future.

src/HeapDump/GCHeapDumper.cs
vancem: We can ignore ARM support.

src/PerfView/App.cs
vancem: At one point (and it probably is still true), PerfView could collect ETW data on ARM (Mostly by simply avoiding using GUI things since WPF was not supported on ARM).   It has not been used in a while because ARM is just not interesting anymore and even when it was there were signing requirements that got in the way.

However ARM may yet come back for the server.   I would not invest in it right now, but I also would not proactively rip it out (the arm specific things should be small).

PR #229, opened 2017-05-21
Build TraceEvent NuGet package

sharwell:
* Build the **Microsoft.Diagnostics.Tracing.TraceEvent** NuGet package with the TraceEvent project.
* Attach the NuGet package to the set of AppVeyor artifacts.


src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
```diff
@@ -0,0 +1,54 @@
...
+      The library includes
+      -- TraceEventSession which can enable ETW providers,
+      -- EtwTraceEventSource which lets you read the stream of ETW events, and
+      -- TraceLog which is is digested form of ETW events which include decoded stack traces associated with the events.
+      -- A Programmers Guide for using the classes above.
+
+      Also see the TraceEvent Samples NuGet package for example uses of this library.
+    </description>
+    <summary>TraceEvent is a .NET Framework library for capturing and analyzing ETW events.</summary>
+    <releaseNotes>https://github.com/Microsoft/perfview/releases/tag/V$version$</releaseNotes>
```
sharwell: :bulb: This link is much more maintainable than the previous approach.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
```diff
@@ -0,0 +1,54 @@
...
+    <copyright>Copyright © Microsoft Corporation</copyright>
+    <tags>TraceEvent EvenSource Microsoft ETW Event Tracing for Windows</tags>
+  </metadata>
+
+  <files>
+    <!-- User information -->
+    <!--<file src="License-Stable.rtf" target="License-Stable.rtf" />-->
+    <file src="ReadMe.txt" target="ReadMe.txt" />
+    <file src="ReleaseNotes.txt" target="ReleaseNotes.txt" />
```
sharwell: :bulb: I would recommend removing this from the repository now that the project is on GitHub. It's an unnecessary maintenance burden.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
```diff
@@ -0,0 +1,54 @@
...
+
+      Also see the TraceEvent Samples NuGet package for example uses of this library.
+    </description>
+    <summary>TraceEvent is a .NET Framework library for capturing and analyzing ETW events.</summary>
+    <releaseNotes>https://github.com/Microsoft/perfview/releases/tag/V$version$</releaseNotes>
+    <copyright>Copyright © Microsoft Corporation</copyright>
+    <tags>TraceEvent EvenSource Microsoft ETW Event Tracing for Windows</tags>
+  </metadata>
+
+  <files>
```
sharwell: :memo: I intentionally left out the **content** files which are part of the current NuGet package. This is not a standard convention for NuGet packages and despite the intention makes it harder to adopt the package in a project.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0"?>
+<package>
+  <metadata minClientVersion="2.5">
+    <id>Microsoft.Diagnostics.Tracing.TraceEvent</id>
+    <version>0.0.0</version>
```
sharwell: :memo: This along with `$version$` are automatically populated from the `[AssemblyInformationalVersion]` attribute value.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.targets
```diff
@@ -0,0 +1,37 @@
...
+
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (C) Microsoft Corporation. All rights reserved.
+***********************************************************************************************
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="TraceEventInjectReference" BeforeTargets="ResolveAssemblyReferences" Condition="('$(DesignTimeBuild)' != 'true')">
```
sharwell: ❓ Why is this a build target? Can't we just place the `<ItemGroup>` directly in the file?

/cc @rainersigwald

src/TraceEvent/Properties/AssemblyInfo.cs
```diff
@@ -17,3 +17,6 @@
 [assembly: NeutralResourcesLanguage("en")]
 
 // [assembly: InternalsVisibleTo("PerfView")]
+[assembly: AssemblyVersion("1.0.42.0")]
```
sharwell: :memo: There are multiple approaches available which either eliminate the need to manually update these, or dramatically reduce the frequency of changes. If you describe your desired versioning behavior either @AArnott or myself will implement that approach as described or will propose changes in order to make things even simpler. If you are unsure of what you want, we can make recommendations.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.targets
rainersigwald: Yes: since there's no glob expansion in the output directories of the project, these should be outside of a target.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.targets
```diff
@@ -0,0 +1,37 @@
+<!--
+***********************************************************************************************
+Microsoft.Bcl.Compression.targets
+
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
```
rainersigwald: I find this header overly chatty, plus it doesn't make sense in a NuGet package since they aren't installed machine state. Plus the filename is wrong.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
rainersigwald: I'm confused by this statement. `content` is a NuGet standard: https://docs.microsoft.com/en-us/nuget/schema/nuspec#including-content-files

It's a confusing standard that has changed behavior over time, which is probably enough to avoid using it though.

src/TraceEvent/packages.config
```diff
@@ -1,4 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles" version="1.0.2" targetFramework="net45" />
+  <package id="NuGet.CommandLine" version="2.8.3" targetFramework="net45" />
+  <package id="Tvl.NuGet.BuildTasks" version="1.0.0-alpha002" targetFramework="net45" developmentDependency="true" />
```
rainersigwald: Could also use NuGetizer-3000 for this for a "more supported" approach: https://github.com/NuGet/NuGet.Build.Packaging.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
sharwell: @rainersigwald What I mean is, the TraceEvent NuGet package currently adds its readme and getting started guide to the project whenever you install it into a project. This is allowed, but makes installing the package more burdensome when you have to go delete these files every time.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.targets
sharwell: ➡️ I copied this from the existing TraceEvent NuGet package without modification.

src/TraceEvent/packages.config
sharwell: I don't have particular experience with that one, but I've been quite happy with the performance of these build tasks for cases where one NuGet package is created for the assembly contained in one C# project.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
vancem: I agree we should not put docs in the package.    All of that predated GitHub and easy web links.  We have wanted to get rid of it for a while.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
vancem: Yes, you can remove these as well.   In general we want only one place for information to be updated.  Putting it in the releases page on GitHub is very reasonable.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.targets
vancem: Note that the previous Nuget package should be viewed as experimental.   It should probably not be used as a template if there are better templates available.    Ideally we start with a completely vanilla nuspec and tweek it just enough to get what we want.   The goals are

1. Simple and straightforward.   
2. A resulting nuget package that current users can pick up very easily.   (Ideally it is compatible, but even that is  negotiable if we can get significant improvements by a modest breaking change).  
3. We wire into existing infrastructure to get the package signed and or released.   

The thing that makes this package a bit harder than normal is that it has native components.    The solution in the past was to include all architectures for the native components in the one package, but we may wish to change to the mechanism used by other packages where you refer to one of several architecture-specific packages.

Our goal is to NOT be innovative, but simply follow a standard for doing what we need (access native components).  

We may wish to split the package (symbol resolution, collection and processing are the obvious groupings).   However I think we can defer that issue for now (for better or worse the are currently together and there is no reason to force this transition now).

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.targets
vancem: Again, I would recommend starting with a completely vanilla file and build up from there.   
I do not know the purpose of TraceEventInjectReference (I think it was there a bug fix for a scenario that we ultimately dropped).    I think it is OK to re-learn why things were done by starting with something simple and understandable, and work up from there.

src/TraceEvent/Properties/AssemblyInfo.cs
vancem: I would like to follow the simplest/most commonly used approach (to NOT be innovative).   Ideally humans don't deal with version numbers much at all (only on major transitions).

src/TraceEvent/TraceEvent.csproj
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\packages\Tvl.NuGet.BuildTasks.1.0.0-alpha002\build\Tvl.NuGet.BuildTasks.props" Condition="Exists('..\..\packages\Tvl.NuGet.BuildTasks.1.0.0-alpha002\build\Tvl.NuGet.BuildTasks.props')" />
```
vancem: What is this for?    Presumably it is a Nuget building thing?

src/TraceEvent/Properties/AssemblyInfo.cs
AArnott: Agreed. Which is exactly why a couple groups of people developed a NuGet package to establish a simple pattern you can follow and then basically forget about versioning except when you really want to bump the major (or minor) versions.
Check out [Nerdbank.GitVersioning](https://github.com/aarnott/nerdbank.gitversioning). Disclosure: I wrote it.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.targets
sharwell: > Note that the previous Nuget package should be viewed as experimental. It should probably not be used as a template if there are better templates available. Ideally we start with a completely vanilla nuspec and tweek it just enough to get what we want.

I'm fine with that. I started with what existed to be less likely to "rock the boat". I'll work with @AArnott and/or @rainersigwald to recommend specific changes for a great long-term solution (dependable yet low maintenance).

> The thing that makes this package a bit harder than normal is that it has native components. 

My work was the foundation for libgit2sharp's native binaries distribution (libgit2/libgit2sharp#821), which has since been modified for improved cross-platform support ([libgit2/libgit2sharp.nativebinaries](https://github.com/libgit2/libgit2sharp.nativebinaries)). I plan to build on that work to the extent it relates to this project, but simplify it where it makes sense in cases that don't apply to us.

> Our goal is to NOT be innovative, but simply follow a standard for doing what we need (access native components).

There's a reason I suggested the two people I suggested for helping. :smile:

image image

> We may wish to split the package (symbol resolution, collection and processing are the obvious groupings).

This tends to cause more headaches than it solves. We should approach that if/when it becomes a clear requirement.

src/TraceEvent/Properties/AssemblyInfo.cs
sharwell: @vancem From what you have said, my versioning approaches target a different audience and are not likely what you want to deal with. My recommendation is to let @AArnott file a pull request later after this one is completed and merged which incorporates his versioning tools. They are as low maintenance as you can get.

src/TraceEvent/TraceEvent.csproj
sharwell: It's added by a package that tells MSBuild how to convert the **.nuspec** into a **.nupkg** as part of the build. The source is here:

https://github.com/tunnelvisionlabs/NuGetBuildTasks

vancem:
I want to look at the versioning scheme suggested in the conversation and take a look at the nuget package building utility over the weekend.   We are setting precedent here.

sharwell:
> I want to look at the versioning scheme suggested in the conversation and take a look at the nuget package building utility over the weekend. We are setting precedent here.

No problem. Keep in mind that I made no changes to versioning as part of this PR, at least none that I was aware of. I was planning to address that separately.

The NuGet package building tool is the one we've been using for StyleCop Analyzers for a couple years and thousands of builds without hitting problems. It's lightweight and reliable.

vancem:
I have looked this over more careful.  I do see that this change has nothing to do with version numbering.  

My largest concern is that I want to be as 'boring' as possible, using the most standard way of creating a nuget package.    This update uses the Tvl.NuGet.BuildTasks to do this building, which I do see is a very thin wrapper but it is still non-standard.    I notice that there seems to be built in MSBuild support now with the <pack> target (see https://docs.microsoft.com/en-us/nuget/schema/msbuild-targets).  I don't know what limitations it has, but it seems like the most 'boring' way, so I want to at least understand what drove us elsewhere.   

If that built-in support is too simplistic, the https://github.com/NuGet/NuGet.Build.Packaging. also seems worthy of consideration because it comes from the Nuget team presumably is recommended by them (I would be interested in their official guidance on the matter).  

Since most of the delta here has nothing to do with the Tvl.Nuget.BuildTasks and switching to something else should be very straightforward, I am not averse to merging this request as is (since we are making forward progress), but my goal would be to replace Tvl.Nuget.BuildTasks rather aggressively once we determine the 'boring' guidance we should be following.

vancem:
The only thing blocking this is the merge conflict.

sharwell:
> The only thing blocking this is the merge conflict.

Fixed

> This update uses the Tvl.NuGet.BuildTasks to do this building, which I do see is a very thin wrapper but it is still non-standard.

On the plus side, I went with one I *know* works. The reason the package is still "alpha" is because in the two years it's been available, despite thousands of users using it for various projects I never had a single negative report come back after my initial publish.

PR #231, opened 2017-05-21
Remove reference to non-existent file

sharwell:
Now that we switched from MSTest to XUnit, the \*.testsettings files are no longer needed.


vancem:
LGTM

PR #236, opened 2017-05-22
Fix handling of start/stop times in other cultures

### LINKED ISSUE #235: SetTimeRange does not work when digit grouping symbol is space
Suppose my regional settings have space as a grouping character:

image

Then in the Stacks view times are formatted accordingly:

image

When Set Time Range is executed the result is bad (although predictable):

image

And Copy Time Range also formats full time range with space in between:

image

Does PerfView just use space to separate Start and End values?

Is it something that is only in use internally by PerfView? If this is true, then it should not be too difficult to fix (I can try to).

sharwell:
Fixes #235


sharwell:
I reached out to @stevepeak regarding the lack of coverage for this build. It should be showing a rather large coverage improvement since it's the first time this project has any tests.

vancem:
We are setting precedent on how we do testing.     There is a 'fork in the road'.   You can test the 'GUI' by running GUI stuff and scrapping answers out of GUI objects.   Or you can tests underlying APIs (which are not as fragile and more understandable and don't take a dependencies on the GUI.   This later point is especially important because I think we could do much better organizing the GUI layer and I would prefer not to wire in more dependencies on exactly how it works (which make changing it that much harder).  

My dream is that we have a reasonably clear separation between GUI and 'model'   The 'model' is described by the API in Extensibiliy.cs as well as all the APIs in TraceEvent.    Adding tests that take advantage of that is certainly goodness.      

So that is the general direction.   We also need to be pragmatic, some testing is better than none, and we always have the option of abandoning tests if we need to.   Still I would like to think about this over the weekend.     In the past we had some GUI tests and we ultimately turned them off because they broke easily and yet did not seem to find true regressions.    The cost outweighed the benefit.

vancem:
I note you marked this review only.   What prompted that?

src/PerfView/GuiUtilities/StatusBar/StatusBar.xaml.cs
```diff
@@ -203,54 +204,67 @@ public void StartWork(string message, Action work, Action finally_ = null)
...
-            ThreadPool.QueueUserWorkItem(delegate
+            ThreadPool.QueueUserWorkItem(delegate (object state)
             {
+                var oldCulture = Tuple.Create(CultureInfo.CurrentCulture, CultureInfo.CurrentUICulture);
+                var culture = (Tuple<CultureInfo, CultureInfo>)state;
                 try
                 {
+                    Thread.CurrentThread.CurrentCulture = culture.Item1;
+                    Thread.CurrentThread.CurrentUICulture = culture.Item2;
+
```
vancem: My philosophy on culture things is the less the code sees of it the better.    I am completely OK with fixing the ad-hoc parsing that happened to be culture specific.    However when I see explicit manipulation of a thread's culture, I get worried.

Can you describe the rationale here?   Was it for testing?    Ideally very little code 'knows' about the current culture.

sharwell:
> I note you marked this review only. What prompted that?

Mostly so you'd know not to look at it first. We can talk about test strategies later; wasn't quite ready to put this one forward but still wanted to see the code coverage impact so I created a PR for it.

sharwell:
> We are setting precedent on how we do testing. There is a 'fork in the road'. You can test the 'GUI' by running GUI stuff and scrapping answers out of GUI objects. Or you can tests underlying APIs (which are not as fragile and more understandable and don't take a dependencies on the GUI.

Yes and no. The only real precedent would be establishing the acceptability of testing a bug at the UI layer. Other tests can still occur at lower layers. I've done a fair bit of GUI application testing in the past, and found the following:

* You can't test the GUI without testing the GUI.
* High-speed and reliable testing at the GUI layer is achievable. And really not that difficult if you set out to enable that scenario. Of the multiple applications/platforms we worked at like this, WPF provided one of the more stable foundations with our test instability sitting at near 0.
* The test I've written provides the highest degree of confidence you can get from a high-speed automated test that the underlying bug has been (and remains) resolved, provided that it is not a flaky test.

For me, the most important tests are the application interface behavior tests:

1. Tests at the boundary of an exposed "module" or "unit of code" to ensure that the underlying implementation behaves in accordance with its documented interface.
2. Tests at the boundary of an application with a user, to ensure that the application behaves as expected in the presence of user behaviors.
3. Tests at the boundary of the application with the system, to ensure that unexpected occurrences in the host environment are handled by the application (e.g. missing and/or inaccessible files, OS version differences etc.).

This test sits in the second interface layer which cleanly aligns with the original bug report.

Typically, the real challenge with an approach like this is ensuring you've adequately covered edge cases since it's so difficult to reason about edge cases deep in an implementation when looking at tests in the GUI layer. The answer to this is incorporating code coverage tooling throughout the development process - codecov.io's handling of pull requests allows one to easily review the coverage in the implementation internals while writing tests that make sense from an end-user perspective (or from the perspective of a developer using one module from another). If you have an edge case where code isn't getting covered ([like this]()), it's easier and often better to answer the question by providing a new user input that shows the importance of that edge case than to write a unit test that executes directly against the edge case.

Everything is a balance of course.

> This later point is especially important because I think we could do much better organizing the GUI layer and I would prefer not to wire in more dependencies on exactly how it works (which make changing it that much harder).
>
> My dream is that we have a reasonably clear separation between GUI and 'model' The 'model' is described by the API in Extensibiliy.cs as well as all the APIs in TraceEvent. Adding tests that take advantage of that is certainly goodness.

I would advise caution before going too far down this road. There is a common misconception that clean separation of a view into a view and a view model improves the ability to test. In the end, what it really means is you have an easier ability to test the view's model, but you haven't really removed any of the requirements for testing of the view, so you just end up with more work. For as many cases as it improves things, there are cases where the separation increases complexity, exhibits certain differences in behavior between platforms/versions of WPF, adds runtime overhead, and decreases overall maintainability.

When I started working on the Coverity products, everything was separated and testing didn't rely on the GUI layer. Over time, working towards a goal to improve coverage and confidence in delivered software quality, we moved much of the testing up a layer to include the GUI components. We ended up revealing many bugs that users were hitting, increasing coverage, and to the surprise of some, decreasing overall test execution time.

Which leads to a very important question: How do I feel about the specific test that I've implemented here? Well, it's overly complex to say the very least. Much of it can be moved into helpers so there's much less effort involved in adding tests and maintaining the current ones, but for a proof of concept it does a reasonable job of showing the code sequences involved with such an approach. There are a few determinism issues as well that need to be resolved, but this part isn't so bad anymore now that I've done it before at a much larger scale and in in a much less controllable environment.

sharwell:
@vancem I would consider the updated test clean enough to merge. The key change which gives me more confidence in was the change from this in the original implementation:

```csharp
// Yield so command takes effect
await Task.Delay(20);
```

to this in the latest version:

```csharp
// Wait for any background processing to complete
await stackWindow.StatusBar.WaitForWorkCompleteAsync().ConfigureAwait(true);
```

vancem:
First, the fix for the StartTime parsing is fine.    

The original code for StatusBar.xml.cs predated Task support.  (perfVIew used to work on V3.5).  Moving to use tasks is good.   

The only other obvious change in that file was to propagate the Culture and UI culture to the worker thread (and put it back on exceptions).  This is also fine.     You also added your WaitForWorkCompleteAsync (for testing).   

I am more conflicted about the test.   I do understand your point about needing to test the GUI, and thus I am OK with having such tests.    Ideally the code would supply most of the functionality to open GUI items (effectively we thought through a nice model for controlling the GUI programatically)    There is a start for this in Extensibiliy.cs (e.g.  OpenStackView and OpenEventViewer).    Ideally this 'gui control' stuff was what the GUI actually uses (right now there is some cloned code due to some structural problems in the PerfViewItem architecture).    

You can see that currently the cost of creating this particular test was reasonably high, but most of this is because infrastructure (I am wondering if TimeRangeStackSource should be in its own file to promote reusability).    It all feels like a bit of thought is needed to determine the guidance for making tests.  Ideally any particular test would require no more than 1-2  dozen lines of code specific to that test (and we have guidance that makes it discoverable how to write those 1-2 dozen lines.  

Anyway, most of this borders on the philosophical.    Pragmatically, the test represents progress, so I am OK moving forward with it.  

Thanks

Vance

PR #237, opened 2017-05-23
Update the CLR FinalizableObject to include TypeName

vancem:
The event only contained a  TypeID without the name.   Add the logic
to keep track of the ID->Name mapping so that we can add a TypeName
property.   Note that this DOES require two passes to work (since the
ID->Name information is given later in the trace.

@sharwell - this is my counter-proposal for PR #224.   I am not sure I am completely happy with this PR.
We should think about removing the /Finalizers switch (but I want to think about the loss of discovery).


sharwell:
> We should think about removing the /Finalizers switch (but I want to think about the loss of discovery).

I'm not worried about this, since there is no loss of information from the user's perspective. We could always allow the `/Finalizers` switch to be recognized as an argument but ignored.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -1677,7 +1716,7 @@ protected internal override void EnumerateTemplates(Func<string, string, EventFi
                 templates[29] = new GCMarkTraceData(null, 26, 1, "GC", GCTaskGuid, 29, "MarkFinalizeQueueRoots", ProviderGuid, ProviderName);
                 templates[30] = new GCMarkTraceData(null, 27, 1, "GC", GCTaskGuid, 30, "MarkHandles", ProviderGuid, ProviderName);
                 templates[31] = new GCMarkTraceData(null, 28, 1, "GC", GCTaskGuid, 31, "MarkCards", ProviderGuid, ProviderName);
-                templates[32] = new FinalizeObjectTraceData(null, 29, 1, "GC", GCTaskGuid, 32, "FinalizeObject", ProviderGuid, ProviderName);
+                templates[32] = new FinalizeObjectTraceData(null, 29, 1, "GC", GCTaskGuid, 32, "FinalizeObject", ProviderGuid, ProviderName, null);
```
sharwell: :bulb: Use named parameter for literal (i.e. `state: null` instead of just `null`)

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -191,7 +193,44 @@ public enum Keywords : long
             /// </summary>
             GCHeapSnapshot = GC | GCHeapCollect | GCHeapDump | GCHeapAndTypeNames | Type,
         };
-        public ClrTraceEventParser(TraceEventSource source) : base(source) { }
+        public ClrTraceEventParser(TraceEventSource source) : base(source)
```
sharwell: ❓ How do changes like this work with the "TraceParserGen" tool mentioned at the top of this file?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -5347,6 +5413,8 @@ public override object PayloadValue(int index)
         }
 
         private event Action<FinalizeObjectTraceData> Action;
+        protected internal override void SetState(object newState) { state = (ClrTraceEventParserState)newState; }
```
sharwell: ❗️ Comment on the base method needs to be updated.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -5313,8 +5376,9 @@ protected internal override void Validate()
         public override StringBuilder ToXml(StringBuilder sb)
         {
             Prefix(sb);
-            XmlAttribHex(sb, "TypeID", TypeID);
+            XmlAttrib(sb, "TypeName", TypeName);
```
sharwell: ❓ Why change the order of `ObjectID` and `TypeID`?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -5335,10 +5399,12 @@ public override object PayloadValue(int index)
             switch (index)
             {
                 case 0:
-                    return TypeID;
+                    return TypeName;
```
sharwell: ❓ What happens when this code tries to load data produced by an earlier version of the tool?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -9882,4 +9950,85 @@ public enum GCGlobalMechanisms
...
+
+    #region private types
+    /// <summary>
+    /// ClrTraceEventParserState holds all information that is shared among all events that is
+    /// needed to decode Clr events.   This class is registered with the source so that it will be
+    /// persisted.  Things in here include
+    /// 
+    ///     * TypeID to TypeName mapping, 
+    /// </summary>
+    // [SecuritySafeCritical]
```
sharwell: ❓ Why is this commented out?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -9882,4 +9950,85 @@ public enum GCGlobalMechanisms
...
+    internal class ClrTraceEventParserState : IFastSerializable
+    {
+        internal void SetTypeIDToName(int processID, Address typeId, long timeQPC, string typeName)
```
sharwell: ❓ Why bother passing `timeQPC` here? Seems for this map it's always just set to 0. (Same question applies for `TypeIDToName`.)

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -9882,4 +9950,85 @@ public enum GCGlobalMechanisms
...
+        {
+            if (_typeIDToName == null)
+                _typeIDToName = new HistoryDictionary<string>(500);
+
+            _typeIDToName.Add(typeId + ((ulong)processID << 48), timeQPC, typeName);
+        }
+
+        internal string TypeIDToName(int processID, Address typeId, long timeQPC)
+        {
+            lazyTypeIDToName.FinishRead();      // We don't read fileIDToName from the disk unless we need to, check
```
sharwell: :bulb: Comment references wrong field name. Also consider placing the comment on the line before the code so it's easier to be consistent across code formatters.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -191,7 +193,44 @@ public enum Keywords : long
...
+
+            // Subscribe to the GCBulkType events and remember the TypeID -> TypeName mapping. 
+            ClrTraceEventParserState state = State;
+            AddCallbackForEvents<GCBulkTypeTraceData>(delegate (GCBulkTypeTraceData data)
+            {
+                for (int i = 0; i < data.Count; i++)
+                {
+                    GCBulkTypeValues value = data.Values(i);
+                    string typeName = value.TypeName;
+                    // TODO the GCBulkType events are logged after the event that needed it.  It really
```
sharwell: :bulb: I wouldn't bother with the TODO here. What's done is done. A simple comment that we're ignoring the event time since they are reported after the event that uses them in some cases will suffice. (If you remove the `timeQPC` parameter of `SetTypeIDToName`, this comment could be moved inside that method.)

src/TraceEvent/Parsers/ClrTraceEventParser.cs
vancem: At one time the ClrTraceEventParser was generated using TraceParserGen.   However TraceParserGen does not support a bunch of cases, including arrays of structs, and other things that were added later on by hand.   It also does not support parsers with state.   You could imagine doing this, but it is relatively rare and has never been done.   When we add new events to ClrTraceEventParser we run the tool and then bascally hand merge the new parts into the existing code.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
vancem: You want the most important/useful fields first (so the default list shows things in the best order).    It is a minor point

src/TraceEvent/Parsers/ClrTraceEventParser.cs
vancem: Done

src/TraceEvent/Parsers/ClrTraceEventParser.cs
vancem: The attribute should not be there.  Removing.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
vancem: The model is that any ID -> Name mapping has a lifetime (it begins and ends).   In this case we ended up not having a clean model, and I don't want to hide that fact.

sharwell: 💭 If you can't foresee a case where we would likely want to use a non-zero value going in, it seems like removing the parameter and commenting in the body of the method makes it less likely to misuse this. Otherwise the knowledge requirement is pushed onto the caller. But it's internal and probably rarely updated, so minor.

sharwell:
@vancem My only real concern is how the system handles cases where the ETLX data format changes. Is it going to quietly discard the old file and create a new one, or will it potentially crash or have corrupt state in some scenarios?

sharwell:
💭 @vancem I felt like this was close to ready. Did the serialization issue end up ruling it out?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
vancem: It should just see no ID->Name mappings and thus TypeName will be "" for everything.  
However just to be super-safe I have added a change to bump the version number on the ETLX file which will force it to be re-created in PerfView.

vancem:
Sorry, the branch was removed by mistake.   I put it back.   
As mentioned in other feedback the change should be version compatible, but I have bumped the version so that ETLX files will get re-generated.

sharwell:
@vancem Awesome. In addition to being "safe", this does have the additional advantage that compatible ETL files recorded and analyzed with the old version will automatically get the new feature.

PR #238, opened 2017-05-23
Fix it so that clean installs of 2017 works

vancem:
It turns out that the header files that are needed for PerfView's ETWClrProfiler code
are in a different place in newer versions of the Windows SDK.

If you had installed an older SDK, and upgraded, the header files ended up both places and so things 'just worked'
But if you installed the SDK clean (as you do when you install VS 2017 on a new machine) it is only in the new place.

Basically add the new place to the include search paths to fix.

### DIFF
--- a/src/EtwClrProfiler/ETWClrProfilerX86.vcxproj
+++ b/src/EtwClrProfiler/ETWClrProfilerX86.vcxproj
@@ -66,7 +66,7 @@
       <ExceptionHandling>false</ExceptionHandling>
       <BufferSecurityCheck>false</BufferSecurityCheck>
       <PreprocessToFile>false</PreprocessToFile>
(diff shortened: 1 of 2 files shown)


src/EtwClrProfiler/ETWClrProfilerX86.vcxproj
```diff
@@ -66,6 +66,7 @@
       <ExceptionHandling>false</ExceptionHandling>
       <BufferSecurityCheck>false</BufferSecurityCheck>
       <PreprocessToFile>false</PreprocessToFile>
+      <AdditionalIncludeDirectories>$(NetFxKitsDir)\include\um</AdditionalIncludeDirectories>
```
sharwell: ❔ Is there a reason why this one doesn't inherit previous values (i.e. `%(AdditionalIncludeDirectories)`)?

PR #239, opened 2017-05-24
Fix release build include paths to match Debug build

vancem:
In PR #238 I did not apply the fix to the Release variant.  This fixes that.


sharwell:
👍

PR #240, opened 2017-05-26
Reduce the size of coverage reports sent to codecov

sharwell:
Reduce the load on codecov.io in preparation for sending larger reports covering more of the project.

* Don't track coverage information in xunit assemblies, since they aren't relevant to this project
  * Adds the `-[Xunit*]*` filter
* Combine reports for the same assembly, e.g. **Microsoft.Diagnostics.Tracing.TraceEvent**, when it appears multiple times in different directories (the build directory for each of the test projects)
  * Adds the `-mergebyhash` argument

For OpenCover usage information see this page:
https://github.com/opencover/opencover/wiki/Usage


sharwell:
This dropped the report size from 86601273 bytes in build 1.0.71 to 48066432 bytes in build 1.0.73.

PR #244, opened 2017-05-28
Move OBSERVABLE_TESTS and DISPATCH_TESTS to proper tests

sharwell:
I found these in conditional code blocks and moved them to proper test cases that run during the builds.


vancem:
It is not clear to me that these tests are worth running routinely, but they are better than nothing, so as long as they don't take too long, it is forward progress.    We can change our mind later.

sharwell:
@vancem The big advantage I saw was one more step towards getting rid of conditional compilation elements. Also it seems to have revealed a possible bug (#249) but I haven't worked through it in detail. This is an area I'll likely end up coming back to after a larger portion of a proper test suite is implemented.

PR #247, opened 2017-05-28
Remove the /Finalizers switch

### LINKED ISSUE #242: Remove /Finalizers switch
Starting with #237, the **/Finalizers** switch is no longer required for users to get object finalization information. References to it can be removed for the most part.

:memo: It is recommended that **/Finalizers** still be accepted as a command line argument, but a message can print out when it is seen saying the value is no longer used or required.

sharwell:
The option itself is preserved for compatibility, but the behavior is disabled
and the checkbox for it is removed from the PerfView UI.

Fixes #242

### DIFF
--- a/src/PerfView/CommandLineArgs.cs
+++ b/src/PerfView/CommandLineArgs.cs
@@ -180,7 +180,6 @@ namespace PerfView
         public double SkipMSec;
         public bool ForceNgenRundown;
         public bool DumpHeap;
-        public bool Finalizers;
 
         // Collect options
         public bool NoGui;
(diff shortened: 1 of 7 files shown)


src/PerfView/CommandLineArgs.cs
```diff
@@ -421,8 +420,9 @@ private void SetupCommandLine(CommandLineParser parser)
                     KernelTraceEventParser.Keywords.Interrupt;
             }
 
+            bool Finalizers = false;
             parser.DefineOptionalQualifier("DumpHeap", ref DumpHeap, "Capture a heap snapshot on profile stop");
-            parser.DefineOptionalQualifier("Finalizers", ref Finalizers, "Track the number of objects of each type finalized.");
+            parser.DefineOptionalQualifier("Finalizers", ref Finalizers, "(Obsolete) This option is accepted for backward compatibility with tooling, but is no longer used.");
```
vancem: Please simply remove this option (don't bother with a no-op obsolete qualifier).    We don't guarantee compatibility of command line options.  The idea is that PerfView's xcopy deployment allows any scripts that have wired in particular command line options to also carry the particular version of PerfView as well.

sharwell: Done.

vancem:
LGTM

PR #252, opened 2017-06-02
Update LibraryProvider to avoid loading a v2 CLR

sharwell:
Prior to this change, the `LibraryProvider` always tried to load the V2 CLR libraries when converting a **.dmp** to a **.gcDump**. This resulted in a return code of **CORDBG_E_LIBRARY_PROVIDER_ERROR** which [results in an exception](https://github.com/Microsoft/perfview/blob/c1d0048aa563b0b902295edb46163c13a7549503/src/HeapDump/Debugger/Debugger.cs#L76).

### DIFF
--- a/src/HeapDump/Debugger/DumpDebugging/LibraryProvider.cs
+++ b/src/HeapDump/Debugger/DumpDebugging/LibraryProvider.cs
@@ -20,6 +20,10 @@ namespace Profiler
             CLRMetaHost mh = new CLRMetaHost();
             foreach (CLRRuntimeInfo rti in mh.EnumerateInstalledRuntimes())
             {
+                string versionString = rti.GetVersionString();
+                if (versionString.StartsWith("v2."))
+                    continue;
+
                 string libPath = Path.Combine(rti.GetRuntimeDirectory(), fileName);
                 if (DoesFileMatch(libPath, timestamp, sizeOfImage))
                 {


vancem:
I would like to understand the conditions.   I just tried it myself and opening a .dmp file seems to work.   (I could not repro the failure)  What are the conditions where it fails?   

My other concern is that while I have not tested it in a while, I *think* the existing code would allow dumping heap from processes using a V2.0 runtime.   The change would presumably cut off that capability.   While I agree we should not spend much effort keeping V2.0 scenarios alive, I would also not to preclude it without good reason.

sharwell:
As far as I can tell, the condition for failure is attempting to perform the conversion on a machine that has the v2 CLR installed. The `EnumerateInstalledRuntimes` *always* returns the v2 installation first on my machine.

Technically it is still able to create a dump after this failure, but it reports the following message:

```
Trying to get a ICorDebugProcess object.
Warning: Failed to get a V4.0 debugger Message: The ICLRDebuggingLibraryProvider callback returned an error or did not provide a valid handle. (Exception from HRESULT: 0x80131C43)
Continuing with less accurate GC root information.
```

Also I'm not so sure a v2 dump is meant to be supported by this code given [this exception](https://github.com/Microsoft/perfview/blob/c1d0048aa563b0b902295edb46163c13a7549503/src/HeapDump/Debugger/Debugger.cs#L87):

```csharp
throw new InvalidOperationException("No V4.0 .NET Runtime found in dump file.");
```

vancem:
OK I tested this on a V2 scenario (both 64 and 32 bit) and falls back to something useful (you get a useful heap dump).   I agree that it improves the V4 experience, so this is an improvement.

PR #253, opened 2017-06-02
Check syntax of XML comments during a build

sharwell:
This change ensures features like Add Using and Remove Unused Usings work with documentation comments. See dotnet/roslyn#2773.

:memo: This is limited to test projects as part of a specific change I'm working on.


vancem:
LGTM

PR #254, opened 2017-06-02
Consolidate xunit

sharwell:
Update test projects to reference the same versions of the .NET Framework and xunit.

* .NET Framework 4.6
* xunit 2.2.0


sharwell:
@vancem I'm going ahead with the merge on this to simplify some other work. I am not anticipating problems for the following reasons:

1. This change only affects test code
1. We already had one test project targeting .NET 4.6, and it wasn't a problem (i.e. no new framework dependency is being introduced)
1. We already had one test project targeting xunit 2.2.0, and it wasn't a problem (i.e. no new test library dependency is being introduced)

PR #255, opened 2017-06-02
Fix assertion handling during unit tests

sharwell:
⚠️ This pull request builds on the work in #274, and its diff will be simplified after that one is merged. 

This change alters the behavior of `Debug.Assert`, `Debug.Fail`, and the related methods in the `Trace` class to throw exceptions rather than show the assertion UI during testing. The assertion UI would hang the AppVeyor tests making it near impossible to understand the cause of a failure.

The approach has been adopted from the one used by the Roslyn project.

### DIFF
--- a/src/PerfView.TestUtilities/DebugAssertionTests.cs
+++ b/src/PerfView.TestUtilities/DebugAssertionTests.cs
@@ -1,0 +1,45 @@
+﻿namespace PerfView.TestUtilities
+{
+    using System;
+    using System.Diagnostics;
+    using Xunit;
+
+    /// <summary>
+    /// This class verifies the behavior of <see cref="Debug.Assert(bool)"/> and <see cref="Debug.Fail(string)"/> when
+    /// called during unit testing.
+    /// </summary>
+    /// <remarks>
+    /// <para>This file can be linked into any project which needs to validate that assertions are behaving correctly
+    /// for the purpose of unit testing.</para>
+    /// </remarks>
+    public class DebugAssertionTests
+    {
+#if DEBUG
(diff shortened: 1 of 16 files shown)


src/PerfView.TestUtilities/DebugAssertionTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        [Fact]
+        public void TestTraceAssertThrowsException()
+        {
+            Assert.ThrowsAny<Exception>(() => Trace.Assert(false));
+        }
+
+        [Fact]
+        public void TestTraceFailThrowsException()
+        {
+            Assert.ThrowsAny<Exception>(() => Trace.Assert(false));
```
pharring: `Trace.Fail` ?

src/PerfView.TestUtilities/DebugAssertionTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            Debug.Assert(true);
+
+            Assert.ThrowsAny<Exception>(() => Debug.Assert(false));
+        }
+
+        [Fact]
+        public void TestDebugFailThrowsException()
+        {
+            Assert.ThrowsAny<Exception>(() => Debug.Assert(false));
```
pharring: `Debug.Fail` ? (Otherwise this is the same as the method above)

src/PerfView.TestUtilities/DebugAssertionTests.cs
sharwell: Yes, that was the intent. Thanks!

PR #256, opened 2017-06-02
Remember history of additional providers

### LINKED ISSUE #88: The Additional Providers text box should have a history
I often type "*Microsoft-VisualStudio-Common" in this text box every time I start PerfView. It would be nice if this text box remembered it like the recent file/directory combobox or the Group Pats combobox.

ilabutin:
This is a naive attempt to fix #88 - I just implemented the same as for command history (with same history length == 10).
I was thinking to refactor code a little and extract concatenating logic into separate method, but decided not to, because actual logic varies a little (reg. ';' handling in items).
But if this is good idea, I can introduce that as well.


vancem:
LGTM

PR #257, opened 2017-06-04
Fix ObservableTests

### LINKED ISSUE #249: Assertion evaluating TraceEvent.ID in tests
Some of the `ObservableTests` added in #244 are skipped because they occasionally trigger an assertion in debug builds:

```
Microsoft.Diagnostics.Tracing.TraceEvent.dll!Microsoft.Diagnostics.Tracing.TraceEvent.ID.get() Line 568	C#
Microsoft.Diagnostics.Tracing.TraceEvent.dll!Microsoft.Diagnostics.Tracing.TraceEventParser.RemoveCallback<Microsoft.Diagnostics.Tracing.TraceEvent>(System.Action<Microsoft.Diagnostics.Tracing.TraceEvent> callback, object subscriptionId) Line 2344	C#
Microsoft.Diagnostics.Tracing.TraceEvent.dll!Microsoft.Diagnostics.Tracing.ObservableExtensions.Observe.AnonymousMethod__1(System.Action<Microsoft.Diagnostics.Tracing.TraceEvent> callback, object subscriptionId) Line 3906	C#
Microsoft.Diagnostics.Tracing.TraceEvent.dll!Microsoft.Diagnostics.Tracing.ObservableExtensions.TraceEventObservable<Microsoft.Diagnostics.Tracing.TraceEvent>.TraceEventSubscription.Dispose() Line 3978	C#
TraceEventTests.dll!TraceEventTests.ObservableTests.RunTests(string etlFileName) Line 70	C#
[External Code]	
```

sharwell:
Fixes #249

The root cause of the test failures is the following:

1. [This sanity check](https://github.com/Microsoft/perfview/blob/da21a66c08816ca95664adcd91461300b21d992c/src/TraceEvent/DynamicTraceEventParser.cs#L513) failed with an exception [here](https://github.com/Microsoft/perfview/blob/da21a66c08816ca95664adcd91461300b21d992c/src/TraceEvent/DynamicTraceEventParser.cs#L1009), seemingly due to an `addDomain` field which is possible related to [this TODO](https://github.com/Microsoft/perfview/blob/da21a66c08816ca95664adcd91461300b21d992c/src/TraceEvent/Computers/ActivityComputer.cs#L818) (e.g. if the appDomain field doesn't appear in the baseline files due to some omission in the TPL parser; the details here are totally outside my current understanding)
1. [This exception handler](https://github.com/Microsoft/perfview/blob/da21a66c08816ca95664adcd91461300b21d992c/src/TraceEvent/DynamicTraceEventParser.cs#L531) resulted in a string being assigned to a field
1. The string value [failed to convert to a long](https://github.com/Microsoft/perfview/blob/da21a66c08816ca95664adcd91461300b21d992c/src/TraceEvent/DynamicTraceEventParser.cs#L831)
1. The exception was [caught in `Dispatch`](https://github.com/Microsoft/perfview/blob/da21a66c08816ca95664adcd91461300b21d992c/src/TraceEvent/TraceEvent.cs#L2960-L2964), bypassing the code which assigns `eventRecord` back to null
1. When the event was eventually unsubscribed, [this assertion failed](https://github.com/Microsoft/perfview/blob/da21a66c08816ca95664adcd91461300b21d992c/src/TraceEvent/TraceEvent.cs#L568) causing the test to hang

This pull request modifies the event dispatch mechanism to ensure `eventRecord` is set back to null after dispatch in all cases.

### DIFF
--- a/src/TraceEvent/Ctf/CtfTraceEventSource.cs
+++ b/src/TraceEvent/Ctf/CtfTraceEventSource.cs
@@ -517,11 +517,18 @@ namespace Microsoft.Diagnostics.Tracing
                 var hdr = InitEventRecord(header, entry.Reader, etw);
                 TraceEvent traceEvent = Lookup(hdr);
                 traceEvent.eventRecord = hdr;
-                traceEvent.userData = entry.Reader.BufferPtr;
-                traceEvent.EventTypeUserData = evt;
+                try
+                {
+                    traceEvent.userData = entry.Reader.BufferPtr;
+                    traceEvent.EventTypeUserData = evt;
 
-                traceEvent.DebugValidate();
-                Dispatch(traceEvent);
+                    traceEvent.DebugValidate();
+                    Dispatch(traceEvent);
+                }
+                finally
+                {
+                    traceEvent.eventRecord = null;
+                }
             }
 
             sessionEndTimeQPC = (long)lastTimestamp;
--- a/src/TraceEvent/ETWReloggerTraceEventSource.cs
+++ b/src/TraceEvent/ETWReloggerTraceEventSource.cs
@@ -368,19 +368,25 @@ namespace Microsoft.Diagnostics.Tracing
 
                 // Lookup the event;
                 TraceEvent anEvent = source.Lookup(rawData);
+                try
+                {
+                    source.m_curITraceEvent = eventData;
+                    source.m_curTraceEventRecord = anEvent.eventRecord;
 
-                source.m_curITraceEvent = eventData;
-                source.m_curTraceEventRecord = anEvent.eventRecord;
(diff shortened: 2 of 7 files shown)


vancem:
I need to think about this more.    

I really want the dispatch code to be as lean (and simple) as possible.  Thus I really don't want us establishing an invariant that anEvent.eventRecord should be null on return from processing  Indeed in dispatch we say 
```
    anEvent.eventRecord = null;      // Technically not needed but detects user errors sooner. 
```
because it is a nice to have (you get AVs if it is not set) but it is not essential to the code.   My expectation is that eventRecord should always be set in dispatch and while it may have a stale value after processing, that should not matter.

sharwell:

brianrob:

PR #260, opened 2017-06-04
Fix race conditions due to static state in ComputerNameExists

sharwell:


vancem:
I don't really like the idea of a cache at all because the value of this could change over time.   

A review of the callers, shows that the caching is not necessary (they are all write-once locations).   

My recommendation is to rip the caching out completely.

sharwell:
Works for me; updated to remove the caching altogether.

vancem:
LGTM.  I want to do one or two scenarios two to confirm we don't get big delays because of this.

sharwell:
@vancem Thanks. I would recommend testing the worst case by locally forcing the timeout as follows:

```csharp
Thread.Sleep(timeout);
return false;
```

vancem:
LGTM

PR #261, opened 2017-06-04
Eliminate static state in extension features

sharwell:
* Eliminate static state in `TraceLoadedDotNetRuntimeExtensions`
* Eliminate static state in `TraceProcessesExtensions`

### DIFF
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -32,22 +32,37 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
     /// </summary>
     public static class TraceLoadedDotNetRuntimeExtensions
     {
+        internal static readonly string LoadedDotNetRuntimesKey = "Computers/LoadedDotNetRuntimes";
+
+        /// <summary>
+        /// Starts gathering .NET runtime information for a trace. If the trace was previously gathering this
+        /// information, the information collected so far is cleared.
+        /// </summary>
+        /// <remarks>
+        /// <para>This method also calls <see cref="TraceProcessesExtensions.NeedProcesses"/> to gather process
+        /// information from a trace.</para>
+        /// </remarks>
+        /// <param name="source">The source.</param>
         public static void NeedLoadedDotNetRuntimes(this TraceEventDispatcher source)
         {
-            // ensure there are base processes
+            // initialize and/or reset the processes information
             source.NeedProcesses();
 
-            if (m_currentSource != source) TraceLoadedDotNetRuntime.SetupCallbacks(source);
-            source.UserData["Computers/LoadedDotNetRuntimes"] = new Dictionary<ProcessIndex, DotNetRuntime>();
-
-            m_currentSource = source;
+            bool previouslyRegistered = source.UserData.ContainsKey(LoadedDotNetRuntimesKey);
+            source.UserData[LoadedDotNetRuntimesKey] = new Dictionary<ProcessIndex, DotNetRuntime>();
+            if (!previouslyRegistered)
+                TraceLoadedDotNetRuntime.SetupCallbacks(source);
         }
 
         public static TraceLoadedDotNetRuntime LoadedDotNetRuntime(this TraceProcess process)
         {
             Debug.Assert(process.Source != null);
-            Debug.Assert(m_currentSource == process.Source);
-            Dictionary<ProcessIndex, DotNetRuntime> map = process.Source.UserData["Computers/LoadedDotNetRuntimes"] as Dictionary<ProcessIndex, DotNetRuntime>;
+
+            object runtimes;
+            if (!process.Source.UserData.TryGetValue(LoadedDotNetRuntimesKey, out runtimes))
+                throw new InvalidOperationException("This source is not gathering .NET runtime information.");
+
+            Dictionary<ProcessIndex, DotNetRuntime> map = runtimes as Dictionary<ProcessIndex, DotNetRuntime>;
             if (map.ContainsKey(process.ProcessIndex)) return map[process.ProcessIndex].Runtime;
             else return null;
         }
@@ -55,8 +70,12 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
         public static void AddCallbackOnDotNetRuntimeLoad(this TraceProcess process, Action<TraceLoadedDotNetRuntime> OnDotNetRuntimeLoaded)
         {
             Debug.Assert(process.Source != null);
-            Debug.Assert(m_currentSource == process.Source);
-            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)process.Source.UserData["Computers/LoadedDotNetRuntimes"];
+
+            object runtimes;
+            if (!process.Source.UserData.TryGetValue(LoadedDotNetRuntimesKey, out runtimes))
+                throw new InvalidOperationException("This source is not gathering .NET runtime information.");
+
+            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)runtimes;
             if (!map.ContainsKey(process.ProcessIndex)) map.Add(process.ProcessIndex, new DotNetRuntime());
             map[process.ProcessIndex].OnLoaded += OnDotNetRuntimeLoaded;
         }
@@ -64,8 +83,12 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
         public static void SetMutableTraceEventStackSource(this TraceProcess process, MutableTraceEventStackSource stackSource)
         {
             Debug.Assert(process.Source != null);
-            Debug.Assert(m_currentSource == process.Source);
-            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)process.Source.UserData["Computers/LoadedDotNetRuntimes"];
+
+            object runtimes;
+            if (!process.Source.UserData.TryGetValue(LoadedDotNetRuntimesKey, out runtimes))
+                throw new InvalidOperationException("This source is not gathering .NET runtime information.");
+
+            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)runtimes;
             if (!map.ContainsKey(process.ProcessIndex)) map.Add(process.ProcessIndex, new DotNetRuntime());
             map[process.ProcessIndex].StackSource = stackSource;
         }
@@ -73,15 +96,23 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
         public static MutableTraceEventStackSource MutableTraceEventStackSource(this TraceProcess process)
         {
             Debug.Assert(process.Source != null);
-            Debug.Assert(m_currentSource == process.Source);
-            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)process.Source.UserData["Computers/LoadedDotNetRuntimes"];
+
+            object runtimes;
+            if (!process.Source.UserData.TryGetValue(LoadedDotNetRuntimesKey, out runtimes))
+                throw new InvalidOperationException("This source is not gathering .NET runtime information.");
+
+            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)runtimes;
             if (map.ContainsKey(process.ProcessIndex)) return map[process.ProcessIndex].StackSource;
             else return null;
         }
 
         public static bool HasMutableTraceEventStackSource(this TraceEventDispatcher source)
         {
-            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)source.UserData["Computers/LoadedDotNetRuntimes"];
+            object runtimes;
+            if (!source.UserData.TryGetValue(LoadedDotNetRuntimesKey, out runtimes))
+                throw new InvalidOperationException("This source is not gathering .NET runtime information.");
+
+            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)runtimes;
             return map.Any(kv => kv.Value.StackSource != null);
         }
 
@@ -96,7 +127,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
         internal static void OnDotNetRuntimeLoaded(this TraceProcess process, TraceLoadedDotNetRuntime runtime)
         {
             Debug.Assert(process.Source != null);
-            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)process.Source.UserData["Computers/LoadedDotNetRuntimes"];
+            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)process.Source.UserData[LoadedDotNetRuntimesKey];
             if (!map.ContainsKey(process.ProcessIndex)) map.Add(process.ProcessIndex, new DotNetRuntime());
             map[process.ProcessIndex].Runtime = runtime;
             if (map[process.ProcessIndex].OnLoaded != null) map[process.ProcessIndex].OnLoaded(runtime);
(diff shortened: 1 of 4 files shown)


vancem:
+ @jeffschwMSFT

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -32,56 +32,82 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
...
+
         public static void NeedLoadedDotNetRuntimes(this TraceEventDispatcher source)
         {
+            if (source.userData.ContainsKey(LoadedDotNetRuntimesKey))
+            {
+                // The callbacks are already registered for this source
+                return;
+            }
+
             // ensure there are base processes
```
jeffschwMSFT: This change is not resilient to calling source.Process twice.  The data in DotNetRuntime is accumulative and if not cleared, will be wrong.

src/TraceEvent/Computers/TraceProcess.cs
```diff
@@ -27,24 +27,27 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
...
     public static class TraceProcessesExtensions
     {
+        internal static readonly string ProcessesKey = "Computers/Processes";
+
         public static void NeedProcesses(this TraceEventDispatcher source)
         {
             TraceProcesses processes = source.Processes();
-            if (processes == null || m_currentSource != source)
+            if (processes == null)
             {
```
jeffschwMSFT: Help me understand why the check for the source was removed?  This was explicitly added to avoid adding callbacks more than once on the same source.

jeffschwMSFT:
@sharwell I added a few comments - they hit similar scenarios, reuse and multiple calls on the same source.  What sort of validation suite are you using for these?  Many of the harder issues with TraceEvent come up when you drill into source reuse, concurrency, and live events.  All these scenarios are not just possible they are hit in our live etl processing.

src/TraceEvent/Computers/TraceProcess.cs
sharwell: ➡️ In the pull request as it currently stands, it is no longer possible for the callbacks to be registered multiple times. In the past, it was possible under concurrent usage scenarios for the callbacks to be registered multiple times (frighteningly, we're talking about two independent sections of code processing two independent traces here, with no shared objects).

src/TraceEvent/Computers/TraceManagedProcess.cs
sharwell: ❓ Can you describe the method by which you expect this information to be cleared in order to restart processing using the same source but without the previously accumulated data?

sharwell:
@jeffschwMSFT I don't currently have tests for this. If you can describe a test case in a fair amount of specificity, I can get it coded as a unit test.

These changes are focused on eliminating potential negative interactions under concurrent use of independent objects. I was not looking at source reuse. In the latter case, this pull request only has one substantial impact, which is the elimination of clearing the accumulated data under a second call to `Need*` for the same source. If this is a scenario you need, I can easily add it back without losing the benefits this pull request offers for concurrent usage scenarios.

src/TraceEvent/Computers/TraceManagedProcess.cs
sharwell: 💭 From the comments as a whole, it sounds like what you are saying is `NeedLoadedDotNetRuntimes` should behave as follows:

* The first time it's called, sets up callbacks and initializes the dictionary for information
* If called again later, leaves the callbacks in place but otherwise performs the same initialization - clearing the existing dictionary

If this is how you would describe this method, I should be able to implement/document/test this behavior in a straightforward manner. But I'll wait to hear back before proceeding down this path.

sharwell: @jeffschwMSFT I went ahead and modified the new implementation to behave the same as the old implementation, except without the race condition for concurrent use. The one part that doesn't make sense to me right now is why you'd want to reset the collected information. Do you have examples of this case?

The scenario that still makes me uncomfortable (even though it's the same as the previous implementation) is the following:

1. Call `NeedProcesses` to track process information
2. Wait some amount of time
3. Call `NeedLoadedDotNetRuntimes` to start tracking .NET runtime information
  * ➡️ Information in `TraceProcesses` is reset

jeffschwMSFT: The existing semantics must be persisted, they exist for a reason.   Keep in mind that the state of the items attached to source only evolve during *.Process().  So there is no scenario like the case you mention above.  If NeedLoadedDotNetRuntimes is called multiple times before you call process, it resets, if you call NeedLoadedDotNetRuntimes after calling Process it resets.

src/TraceEvent/Computers/TraceProcess.cs
jeffschwMSFT: That is not true.  Please bear this out with your concurrent test suite?  The previous check is a loose guarantee, but it is required.

jeffschwMSFT:
@sharwell I am having a hard time understanding the motivation and underlying value of these changes.  Can you help me understand?  The net-net of the change is removal of one static to track source - which frankly was a loose guarantee / hack.  The other changes are just moving things around.

src/TraceEvent/Computers/TraceProcess.cs
sharwell: > That is not true.

Can you clarify which part this refers to? I made two claims. :smile:

sharwell:
> I am having a hard time understanding the motivation and underlying value of these changes. Can you help me understand?

My general motivation comes on two fronts:

1. Ensure that two independent well-behaved¹ users of the TraceEvent library (i.e. users operating on distinct objects and distinct input files) do not interfere with each other in a concurrent application
2. Where reasonable, protect against mutations to static state which is not supposed to change.

When applied, these practices improve the reliability of the library in distribution and also help our own development through details like preventing one unit test from affecting the behavior of a following unit test.

¹ Well-behaved means adhering to the requirements documented for various methods in the library with respect to exposed static state that is technically mutable but is expected to not actually be mutated by library users.

jeffschwMSFT:
@sharwell What case is failing today with the current implementation?  In my mind the well-behaved case is load an ETL, process it, then shutdown the process.  We do this regularly and much more.  A great example of the well behaved case is Perfview and I am not aware of a failure/issue there.

sharwell:
@jeffschwMSFT The failing case today is two independent threads concurrently processing two independent trace files. The last one to call `NeedProcesses` will result in the first, on an unrelated processing thread, having assertion failures.

jeffschwMSFT:
We do that all the time - this is a common case.  What is the result when this happens?  I am seeing it succeed.  Related, what portion of this change materially makes it better?

src/TraceEvent/Computers/TraceManagedProcess.cs
sharwell: > The existing semantics must be persisted, they exist for a reason.

Note that I'm not trying to claim otherwise. However, since these reasons weren't completely clear to me at the start of the work, I am interested in using the additional information from you to 1) help future readers understand these scenarios and 2) formulate some amount of automated testing to ensure the scenarios do not regress in the future.

sharwell:
> We do that all the time - this is a common case. What is the result when this happens? I am seeing it succeed. Related, what portion of this change materially makes it better?

The most obvious problematic cases only occur in debug builds of the TraceEvent library (i.e. you actually execute your code against a debug build of TraceEvent, as opposed to the more common action of executing a debug build of a tool which references the TraceEvent library). For release builds, the only observable impact in TraceProcesses would be extra-accumulation of the amount of time spent in a process [here](https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/Computers/TraceProcess.cs#L126) after using `NeedLoadedDotNetRuntimes` or `NeedProcesses` a second time to reset accumulated state.

For duplicate callback registration in `NeedLoadedDotNetRuntimes`, the resulting aggregation errors are a bit more difficult to be certain of without more work, but I couldn't find anything to indicate that you would experience anything more severe than (presumably subtle) inaccuracy of results if this race condition is hit.

jeffschwMSFT:
Debug TraceEvent has a number of issues - it is noble and much needed to clean those up.  Though the asserts I have seen are far deeper within the setup callbacks and not at this level.  
I agree the following scenario is problematic (a good bug find): using(source) { source.NeedProcesses(); source.Process(); source.NeedProcesses(); source.Process(); }.  Though the fix for that is different than the proposed.  I would recommend tackling that issue within the context of also keeping the right semantic for the following, which is far more common: using(source) { source.NeedProcesses(); source.NeedProcesses(); source.Process(); }  
We need to find a way to fix the issue you encountered and keep the existing compatibility.

sharwell:
@jeffschwMSFT It is my belief that the current implementation (after my last modification) does not change the behavior for processing a single source under any situation aside from the race condition involving two different sources. From the perspective of compatibility, it is intended to be a fully-compatible change. Is there a case where it doesn't appear to do so?

src/TraceEvent/Computers/TraceProcess.cs
```diff
@@ -27,24 +27,30 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
...
             }
-
-            m_currentSource = source;
         }
         public static TraceProcesses Processes(this TraceEventSource source)
         {
-            if (source.UserData.ContainsKey("Computers/Processes")) return source.UserData["Computers/Processes"] as TraceProcesses;
-            else return null;
+            object processes;
+            if (!source.UserData.TryGetValue(ProcessesKey, out processes))
```
sharwell: ❗️ This condition should not be negated.

sharwell:
I wrote some initial tests and should have a new commit ready in the morning with them. 👍

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -32,56 +32,87 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
...
+            // initialize and/or reset the processes information
             source.NeedProcesses();
 
-            if (m_currentSource != source) TraceLoadedDotNetRuntime.SetupCallbacks(source);
-            source.UserData["Computers/LoadedDotNetRuntimes"] = new Dictionary<ProcessIndex, DotNetRuntime>();
-
-            m_currentSource = source;
+            bool previouslyRegistered = source.UserData.ContainsKey(LoadedDotNetRuntimesKey);
+            source.UserData[LoadedDotNetRuntimesKey] = new Dictionary<ProcessIndex, DotNetRuntime>();
+            if (previouslyRegistered)
```
sharwell: ❗️ This condition should be negated.

sharwell: ➡️ Now fixed

src/TraceEvent/Computers/TraceProcess.cs
sharwell: ➡️ Now fixed

src/TraceEvent/TraceEvent.Tests/TraceEventDispatcherExtensionsTests.cs
```diff
@@ -0,0 +1,225 @@
...
+                    Assert.Equal(processDuplicatePinningReports[process.ProcessID], loadedDotNetRuntime.GC.GCs.Sum(gc => gc.duplicatedPinningReports));
+                    Assert.Equal(processGCTotalCpu[process.ProcessID], loadedDotNetRuntime.GC.Stats().TotalCpuMSec);
+                    Assert.Equal(processGCCpuTime[process.ProcessID], loadedDotNetRuntime.GC.GCs.Sum(gc => gc.GCCpuMSec));
+                }
+            }
+        }
+
+        [Fact]
+        [WorkItem(261, "https://github.com/Microsoft/perfview/pull/261")]
+        public void TestNeedProcessesTwoPipelines()
```
sharwell: :memo: This test was failing prior to the changes made in this pull request.

src/TraceEvent/TraceEvent.Tests/TraceEventDispatcherExtensionsTests.cs
```diff
@@ -0,0 +1,225 @@
...
+                foreach (var process in source2.Processes())
+                {
+                    Assert.Equal(processTimes2[process.ProcessID], process.CPUMSec);
+                }
+            }
+        }
+
+        [Fact]
+        [WorkItem(261, "https://github.com/Microsoft/perfview/pull/261")]
+        public void TestNeedLoadedDotNetRuntimesTwoPipelines()
```
sharwell: :memo: This test was failing prior to the changes made in this pull request.

jeffschwMSFT:
To start please merge the relevant commits, after that I will then take a deeper look.
The semantics for NeedProcesses and NeedLoadedDotNetRuntimes should be as follows: 
- On first call - setup callbacks and create state
- On subsequent calls - do nothing
- After processing and calling - clear state, but do not re-establish callbacks

As for scenarios to test I would do the following:
- PerfView case - using(sourceA) { sourceA.NeedsProcesses(); sourceA.NeedsLoadedDotNetRuntimes(); sourceA.Process(); }
- for(many different sources) { using(source) { source.NeedsProcesses(); source.NeedsLoadedDotNetRuntimes(); source.Process(); } }
- using(sourceA) { sourceA.NeedsProcesses(); sourceA.NeedsLoadedDotNetRuntimes(); sourceA.Process(); } /* the same for sourceB */ /* then sourceA again */
- etc. the cases in the tests seem to cover interesting scenarios - the ones I am most concerned with are thread concurrent and multiple different ETLs with occassional duplication

Some of the test cases are really non-scenarios.  What is the customer use case for doing sourceA.Process() multiple times in a row?  This feels like a debug scenario.

sharwell:
> Some of the test cases are really non-scenarios. What is the customer use case for doing `sourceA.Process()` multiple times in a row?

The test doesn't actually call `sourceA.Process()` two times in a row. The test is intended to deterministically reproduce a concurrency bug by forcing what would normally be two independent operations on two different sources to execute in a specific order. It is a specific linearization of two independent operations.

sharwell:
> To start please merge the relevant commits, after that I will then take a deeper look.

@vancem Can you take a look at merging #274 and then #255 to help out with this?

@jeffschwMSFT Prior to that, I would suggest you focus exclusively on reviewing the tests in **TraceEventDispatcherExtensionsTests.cs**. They are unique to this pull request and are intended to be valid test cases regardless what, if any, other code changes are made.

> The semantics for NeedProcesses and NeedLoadedDotNetRuntimes should be as follows:
>
> * On first call - setup callbacks and create state
> * On subsequent calls - do nothing
> * After processing and calling - clear state, but do not re-establish callbacks

Currently we are close:

* On first call: setup callbacks and create state ✔️ 
* On subsequent calls: do not setup callbacks ✔️  but reestablishes state, which *should* be idempotent ❓ 
* After processing and calling: do not setup callbacks ✔️ and reestablishes state ✔️ **but** I was not able to create a test that proved it

Note that the following test failed when I attempted it, which suggests *something* is not getting cleared properly:

```csharp
source.NeedProcesses();
source.Process();

source.NeedProcesses();
source.Process();

processTimes = source.Processes().ToDictionary(p => p.ProcessID, p => p.CPUMSec);
```

However, the equivalent test for `NeedDotNetLoadedRuntimes` was working correctly, so the state reset failure appears to be limited to `TraceProcesses`. I filed #276 for this issue.

> As for scenarios to test I would do the following:

Most of these were covered already, and I added additional coverage in 86b70e5df5b9706028c57f20ecbdff45b3967745.

src/TraceEvent/TraceEvent.Tests/TraceEventDispatcherExtensionsTests.cs
```diff
@@ -0,0 +1,489 @@
...
+                foreach (var process in source.Processes())
+                {
+                    Assert.Equal(processTimes[process.ProcessID], process.CPUMSec);
+                }
+            }
+        }
+
+        [Fact(Skip = "https://github.com/Microsoft/perfview/issues/276")]
+        [WorkItem(261, "https://github.com/Microsoft/perfview/pull/261")]
+        public void TestNeedProcessesResetsState()
```
sharwell: :memo: This test was failing prior to the changes made in this pull request, and currently continues to fail.

src/TraceEvent/TraceEvent.Tests/TraceEventDispatcherExtensionsTests.cs
```diff
@@ -0,0 +1,489 @@
...
+                    Assert.Equal(processDuplicatePinningReports[process.ProcessID], loadedDotNetRuntime.GC.GCs.Sum(gc => gc.duplicatedPinningReports));
+                    Assert.Equal(processGCTotalCpu[process.ProcessID], loadedDotNetRuntime.GC.Stats().TotalCpuMSec);
+                    Assert.Equal(processGCCpuTime[process.ProcessID], loadedDotNetRuntime.GC.GCs.Sum(gc => gc.GCCpuMSec));
+                }
+            }
+        }
+
+        [Fact]
+        [WorkItem(261, "https://github.com/Microsoft/perfview/pull/261")]
+        public void TestNeedLoadedDotNetRuntimesResetsState()
```
sharwell: :memo: This test was failing prior to the changes made in this pull request.

jeffschwMSFT:
@sharwell merge as in merge the commits, not merge with master.  Please collapse the changes so I can see the true deltas from the current source.  What realworld scenarios are represented in the test cases?

sharwell:
@jeffschwMSFT This has now been updated.

jeffschwMSFT:
@sharwell as discussed please merge relevant commits and then I will take a deeper look.

sharwell:
@jeffschwMSFT I rebased and merged to have three distinct units of work:

1. Add tests which demonstrate the observable problems and known use cases
2. Self-contained corrections to `TraceProcessesExtensions`
3. Self-contained corrections to `TraceLoadedDotNetRuntimeExtensions`

jeffschwMSFT:
This change looks good.  Thanks for adding tests.

sharwell:

brianrob:

PR #264, opened 2017-06-04
Remove unnecessary static state

sharwell:
* Make `DisableParallelism` an instance variable
* Remove static state in `StartStopActivityComputer` for an unused field


src/TraceEvent/Stacks/CallTree.cs
```diff
@@ -50,10 +50,10 @@ public StackSource StackSource
 
                 // TODO FIX NOW I turned of Parallelism because there were signs of races and
```
vancem: Please move the TODO comment to the DIableParallelism property (Where the default is set to true).

sharwell: Good idea, fixed.

PR #265, opened 2017-06-04
Make static state immutable where possible

sharwell:


vancem:
LGTM, but has conflicts

sharwell:
@vancem Thanks, fixed

PR #266, opened 2017-06-04
Namespace and preprocessor directives cleanup

sharwell:
* Move command classes from `global::Utilities` to `global::Microsoft.Diagnostics.Tracing.Utilities`
* Move PE file classes from `global::PEFile` to `global::Microsoft.Diagnostics.Tracing.PEFile`
* Remove the unnecessary preprocessor directives `COMMAND_PUBLIC` and `PEFILE_PUBLIC`


vancem:
This is tied up with the issue whether these classes are exposed in TraceEvent.

Currently they are NOT, as they are not really logically related to logging, and I think that is probably for the best.

sharwell:
@vancem Currently these files are exposed in the TraceEvent library and NuGet package, and PerfView consumes them from there. Since it appeared to be a true single point of definition in what is intended to be a reusable library, it seemed to make sense to place them in a namespace used by the rest of the same library.

sharwell:

brianrob:

PR #267, opened 2017-06-04
Enable Ctrl+C support

sharwell:
I noticed this block of disabled code, which should either be completed/included, or removed.

⚠️ I am not necessarily in favor of including this feature. The two acceptable outcomes appear to be including this code (similar to what I've done here, possibly with modifications), or simply removing the code altogether.

### DIFF
--- a/src/TraceEvent/Utilities/command.cs
+++ b/src/TraceEvent/Utilities/command.cs
@@ -498,11 +498,10 @@ namespace Utilities
                 outputStream.WriteLine("RUN CMD: " + commandLine);
             }
 #endif
-#if false
             // Make sure we kill this task when Ctrl C or appdomain unloads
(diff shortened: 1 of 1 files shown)


vancem:
It has been a while, but I believe the reason I pulled this code out is that it does not establish a useful invariant because sub-processes of the processes that are directly started might still live (and this is super-common (e.g. batch file that spawns MSBUILD which spawns other tools ...)  

This is best solved by OS features (process groups etc), but I have not investigated exactly how to make that work.  

Thus my recommendation is to remove it, since the 'right way' to do the feature works in a completely different way.

sharwell:
> Thus my recommendation is to remove it, since the 'right way' to do the feature works in a completely different way.

On further thought, I totally agree with this.

PR #268, opened 2017-06-04
Remove unnecessary preprocessor directives

sharwell:


vancem:
We should not remove these until we figure out some packaging issues.   We may wish to keep them forever. 

These directives are an artifact of the fact that we built both the TraceEvent Nuget package (currently built from a non-GIT branch of this code but we want to build it here), and in this package utility functionality should be internal.      

Previously (and is still the case), we build a special version of TraceEvent that is used by perfView and to avoid cloning, in this version we make these utility methods public (since we need them from outside the TraceEvent DLL).  

Ultimately we do want PerfView to NOT create its own version o TraceEvent but to use the Nuget version.  If we do that we either need to 

1. Use the source code in both projects (as internal?)
2. Make several small Nuget packages for the functionality.   
  
I would prefer that we don't rip out the #ifdefs until the ultimate packaging scheme implemented.

sharwell:
Another option is to keep the types exposed by the TraceEvent assembly for now, and if it ever becomes an issue where you want to reference FastSerialization, PEFile, or Command but don't want to reference TraceEvent, we can move one or more items to their own NuGet package and add a `[TypeForwardedTo]` to the TraceEvent library. I'm not currently convinced such a situation will arise, suggesting we can avoid some complexity by just shipping one package.

sharwell:

brianrob:

brianrob:

PR #273, opened 2017-06-05
Include Microsoft.Diagnostics.FastSerialization in the TraceEvent package

sharwell:
Currently the **Microsoft.Diagnostics.Tracing.TraceEvent** assembly depends on **Microsoft.Diagnostics.FastSerialization**, but no dependency was added to the NuGet package. Rather than deal with separate NuGet packages at this time, this change simply includes the dependency in the original NuGet package.

:memo: If we ever need to split these in the future, we can trivially do so and add a dependency from the current NuGet package to the new one without breaking upgrade scenarios. Since this makes distribution a little bit more complex (managing multiple NuGet releases from one repository) and didn't seem essential at this time, I chose the simplest route to packaging for my initial proposed approach.

### DIFF
--- a/src/FastSerialization/FastSerialization.csproj
+++ b/src/FastSerialization/FastSerialization.csproj
@@ -45,6 +45,17 @@
     <OutputPath>bin\release\</OutputPath>
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
+  <PropertyGroup>
+    <!--
+      Make sure any documentation comments which are included in code get checked for syntax during the build, but do
+      not report warnings for missing comments.
+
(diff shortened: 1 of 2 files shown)


vancem:
This is reasonable, but I am mulling over whether we event want a separate DLL for the FastSerialization code (what makes this different from the other utility code that happens to be in TraceEvent.dll (e.g. PEFile  ...).    This was originally spit out to allow reuse by Visual Studio, but in retrospect, that was not a clear win.   Hmm...

sharwell:
@vancem If you aren't sure, we could file that as a follow-up issue to take on separately.

vancem:
I have been mulling this too long.   It is now blocking someone, and so we will do it this way and we can change it later as a separate issue.

PR #274, opened 2017-06-06
Test utilities

sharwell:
* Create a utilities library **PerfView.TestUtilities**
* Move `WorkItemAttribute` and `UseCultureAttribute` to the new utilities library so it can be used in any test


vancem:
LGTM

PR #279, opened 2017-06-08
Several minor updates

vancem:
Added a link to more documenation about collecting on Linux machines.
Updated logic so that /OnlyProviders will collect addtion events if the user asks for stacks on those events (there was logic but it was geared to old syntax).
Fix a bug with creating large data that might exceed 4GB (happened with some large heap dumps).
Fixed bug where /CpuCounters argument was lost if PerfView had to elevate.

### DIFF
--- a/src/FastSerialization/StreamReaderWriter.cs
+++ b/src/FastSerialization/StreamReaderWriter.cs
@@ -360,6 +360,10 @@ namespace FastSerialization
         protected virtual void Dispose(bool disposing) { }
         #endregion 
         #region private
+
+        /// <summary>
+        /// Makespace makes at least sizeof(long) bytes available (or throws OutOfMemory)
+        /// </summary>
         internal /* protected */ virtual void MakeSpace()
         {
             const int maxLength = 0x7FFFFFC7; // Max array length for byte[]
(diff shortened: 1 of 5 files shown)


src/PerfView/CommandLineArgs.cs
```diff
@@ -352,7 +352,9 @@ private void SetupCommandLine(CommandLineParser parser)
                 bool hasTpl = false;
                 foreach (var provider in onlyProviders)
                 {
-                    if (0 <= provider.IndexOf(":stack", StringComparison.OrdinalIgnoreCase))
+                    if (0 <= provider.IndexOf("@StacksEnabled=true", StringComparison.OrdinalIgnoreCase))
```
sharwell: :bulb: I believe you can use `Contains` instead of `IndexOf` to make the condition a bit more readable

src/PerfView/SupportFiles/UsersGuide.htm
```diff
@@ -6766,7 +6766,8 @@ <h3 id="BatchXMLCreation">Step 1: Preprocessing ETL Data and Forming the Scenari
         Linux has a kernel level event logging system called <a href="https://en.wikipedia.org/wiki/Perf_%28Linux%29">Perf Events</a> which is
         not unlike ETW, and in particular knows how to capture CPU stacks at a periodic interval (e.g. 1msec)  PerfView knows how to read this data,
         so it is possible to collect data using the Perf Events tool on Linux copy the data over to a Windows machine and view it with PerfView's
-        stack viewer.
+        stack viewer.   Much of the rest of this section is a clone of the <a href="https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/linux-performance-tracing.md">linux-performance-tracing.md</a>
+        document.   You may wish to check there as well to see if there for the lastest version of these instructions.
```
sharwell: :bulb: la~~s~~test

src/FastSerialization/StreamReaderWriter.cs
```diff
@@ -368,11 +372,12 @@ public void Dispose()
             // Make sure we don't exceed max possible size
             if (bytes.Length < (maxLength / 3) * 2)
                 newLength = (bytes.Length / 2) * 3;
-            else if (bytes.Length < maxLength)
-                newLength = maxLength;
+            else if (bytes.Length < maxLength - sizeof(long))      // Write(long) expects Makespace to make at 8 bytes of space.
```
sharwell: :bulb: I always find the comments more readable when they are on the preceding line (or immediately inside the block when commenting on a condition) rather than at the end, but either way works

src/PerfView/CommandLineArgs.cs
vancem: Sadly, Contains does not have the overload that allows it to ignoring case (at least Intellisense does not give the option).   We should probably add that overload going forward.

PR #280, opened 2017-06-09
Remove unused code to attach a CLR debugger to a heap dump

sharwell:
It turns out the change in #252 was the result of a misleading message in the console. In reality, due to the change to use clrmd at some point, the attached process is not actually used during extraction of a GC heap from a process dump.

:memo: This change leads to a massive cascade of dead code within heapdump.exe. Since some of that code is technically publicly accessible right now (i.e. could be called by some other library even though we don't ourselves), I'm going to file that as a separate pull request.


vancem:
I see that we we have been relying on CLRMD for a while now and thus a bunch of code is now dead.   
This change could go further (in particular the logic for ICorDebugProcess debugProcess is no longer needed).   
I would prefer that all disruptive changes (like that) to eliminate dead / pointless code be done 'together in a single PR as much as possible (can be left as multiple commits).   
I want this so that if for whatever reason we want to bring this code back it is should make it easier.  

Don't worry about public APIs for this file.  The public APIs here are either not used outside perfView are we are willing to break them.    

So I would like the ICorDebugProcess debugProcess logic removed, and anything else you think is dead as part of this PR.

vancem:
The 'Debugger' directory is a clone from another code base (I believe it is mdbg).  which we have never had to modify.   I am completely OK with the idea of removing whole files that are not needed from this clone.   It is not as clear whether it is a good idea to modify these files (even if only to prune) since it really does 'fork' the code base (it becomes hard to know that we can simply 'update' this code by replacement).   

If you want a quick merge, we should just remove the full files that we can.     Otherwise some thought will be required...

sharwell:
@vancem I reverted all the changes from the last commit except the two small changes which enabled the removal of all the other whole files.

vancem:
Could you undo the changes to Debugger.cs and NativeImports.cs since these changes to the file in the Debugger directory that I would like to avoid.   

We should also get rid of the code in GCHeapDumper.cs where ICorDebugProcess is passed around (since it is no longer used and can only be confusing).   This is the change that I wanted in 'one go' so that if we wanted undo it we could also do it in 'one go'.

sharwell:
@vancem I can't undo the changes in Debugger.cs/NativeImports.cs *and* remove the files I removed. We can either take neither (my original PR) or both (current PR). Let me know your preference on this.

I'll look at the ICorDebugProcess change now.

sharwell:
@vancem The use of `ICorDebugProcess` seems to still be relevant for extracting a GC heap from a currently running process.

sharwell:

brianrob:

brianrob:

PR #282, opened 2017-06-12
Fix flaky tests due to file access errors

sharwell:
The tests all read from files located in source control. As long as the tests only take non-exclusive read locks, they will no longer fail due to exceptions seen during several recent AppVeyor builds.


vancem:
LGTM

PR #285, opened 2017-06-13
Support new SHA-256 source file hash type in DIA

### LINKED ISSUE #284: Assert: "Unknown hash type"
With newly synced/built PerfView, and https://github.com/dotnet/coreclr, I'm doing a PerfView collect->Run using:
```
c:\gh\coreclr\bin\tests\Windows_NT.x86.Release\Tests\Core_Root\superpmi.exe c:\spmi\x86\alltests_win_minopts.mch c:\gh\coreclr\bin\tests\Windows_NT.x86.Release\Tests\Core_Root\clrjit.dll
```

(This requires a `alltests_win_minopts.mch` that I have built)

In the CPU stacks view, I choose the top `Exc%` line and hit ALT-D (go to source), and then get this assert:

```
        unsafe internal SourceFile(SymbolModule module, IDiaSourceFile sourceFile)
        {
            m_symbolModule = module;
            BuildTimeFilePath = sourceFile.fileName;

            // 0 No checksum present.
            // 1 CALG_MD5 checksum generated with the MD5 hashing algorithm.
            // 2 CALG_SHA1 checksum generated with the SHA1 hashing algorithm.
            m_hashType = sourceFile.checksumType;
            if (m_hashType != 2 && m_hashType != 1 && m_hashType != 0)
            {
                Debug.Assert(false, "Unknown hash type");
                m_hashType = 0;
            }
```

in this case, m_hashType is 3.

Has something changed? Namely, is there a new hash type that needs support?

Is there a workaround for using PerfView without this, or forcing a different hash type?

BruceForstall:
Fixes #284

### DIFF
--- a/src/TraceEvent/Symbols/SymbolReader.cs
+++ b/src/TraceEvent/Symbols/SymbolReader.cs
@@ -2681,16 +2681,16 @@ sd.exe -p minkerneldepot.sys-ntgroup.ntdev.microsoft.com:2020 print -o "C:\Users
             // 1 CALG_MD5 checksum generated with the MD5 hashing algorithm.
             // 2 CALG_SHA1 checksum generated with the SHA1 hashing algorithm.
             m_hashType = sourceFile.checksumType;
-            if (m_hashType != 2 && m_hashType != 1 && m_hashType != 0)
-            {
-                Debug.Assert(false, "Unknown hash type");
-                m_hashType = 0;
-            }
             if (m_hashType != 0)
             {
+                uint hashSizeInBytes;
+                fixed (byte* bufferPtr = m_hash)
+                    sourceFile.get_checksum(0, out hashSizeInBytes, out *bufferPtr);
+
                 // MD5 is 16 bytes
                 // SHA1 is 20 bytes  
-                m_hash = m_hashType == 1 ? new byte[16] : new byte[20];
+                // SHA-256 is 32 bytes
+                m_hash = new byte[hashSizeInBytes];
 
                 uint bytesFetched;
                 fixed (byte* bufferPtr = m_hash)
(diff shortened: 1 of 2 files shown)


src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -2681,16 +2681,16 @@ unsafe internal SourceFile(SymbolModule module, IDiaSourceFile sourceFile)
...
-            if (m_hashType != 2 && m_hashType != 1 && m_hashType != 0)
-            {
-                Debug.Assert(false, "Unknown hash type");
-                m_hashType = 0;
-            }
             if (m_hashType != 0)
             {
+                uint hashSizeInBytes;
+                fixed (byte* bufferPtr = m_hash)
+                    sourceFile.get_checksum(0, out hashSizeInBytes, out *bufferPtr);
```
BruceForstall: Passing zero causes the API to tell you how big the hash needs to be. I'm not sure the syntax is optimal, though: in C++, you pass NULL for the final argument; I don't know what the equivalent is in managed code.

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -2699,9 +2699,28 @@ unsafe internal SourceFile(SymbolModule module, IDiaSourceFile sourceFile)
...
+                case 2:
+                    crypto = new System.Security.Cryptography.SHA1CryptoServiceProvider();
+                    break;
+
+                case 3: 
+                    crypto = new System.Security.Cryptography.SHA256CryptoServiceProvider();
+                    break;
+
+                default:
+                    Debug.Assert(false);
```
vancem: Don't just assert, throw an IllegalOperationException.   Otherwise we will nullRef below which is unhelpful.

sharwell: 😆 I thought you were copying me but then I realized I forgot to hit submit.

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -2681,16 +2681,16 @@ unsafe internal SourceFile(SymbolModule module, IDiaSourceFile sourceFile)
...
             {
+                uint hashSizeInBytes;
+                fixed (byte* bufferPtr = m_hash)
+                    sourceFile.get_checksum(0, out hashSizeInBytes, out *bufferPtr);
+
                 // MD5 is 16 bytes
                 // SHA1 is 20 bytes  
-                m_hash = m_hashType == 1 ? new byte[16] : new byte[20];
+                // SHA-256 is 32 bytes
+                m_hash = new byte[hashSizeInBytes];
```
vancem: The API has the concept of having a hash checksum or not.   (see HashCheckSum, and does nothing if it does not have a checksum).   We want this code to fall back to this 'don't do anything' if it sees a hash that it does not understand (in fact if you had not run debug, the existing code would have given you the source (just not given you reassurance that it was the right source).

Currently HasCheckSum is defined to be m_hash != null, which means by setting this unconditionally here, we are saying we always have a checksum, which is not true.

My recommendation:

Create a method 'GetCryptoProvider() that returns the Crypto Provider.
Make ComputeHash take that as a parameter.  
Make HasChecksum be GetCryptoProivder() != null.

BruceForstall:
@vancem @sharwell Updated with feedback.

Also fixed one RVA log output, and added more info to another RVA log output (VA and ImageBase).

PR #294, opened 2017-06-20
Support EventPipe

northtyphoon:
1. EventPipeFileEventSource. 
 - The new EventSource allows us to process the data from EventPipe file and convert to TraceLog.

2. SampleProfilerThreadTimeComputer
- The new thread time computer is a variant from existing ThreadTimeComputer. The key change is the ThreadState tracking is based on CoreClr ThreadSample events which provide On/Off CPU stack traces.

@vancem @brianrob @pharring


src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -74,8 +74,8 @@ public StartStopActivityComputer(TraceLogEventSource source, ActivityComputer ta
 
             Dictionary<long, int> mapConnectionToTargetProcess = new Dictionary<long, int>();
 #endif 
-            var dynamicParser = source.Dynamic;
-            dynamicParser.All += delegate (TraceEvent data)
+            //var dynamicParser = source.Dynamic;
```
northtyphoon: I haven't figured out how to integrated the dynamic templates generated from EventPipeTraceEventParser to DynamicParser.  So I need to subscribe AllEvents to get EventSource events. That's also the reason I add EventPipeTraceEventParser in TraceLog.RegisterStandardParsers. Need suggestion.

vancem: My recommendation is to use the source.AddCallbackForProviderEvents(CouldBeStartStop, callback) to create a filter (CouldBeStartStop) that includes events with 'Start' or 'Stop' suffixes, or are from a provider that is special cased in the code).

src/TraceEvent/TraceEvent.cs
```diff
@@ -397,6 +398,8 @@ string ITraceParserServices.ProviderNameForGuid(Guid taskOrProviderGuid)
         internal /*protected*/ DynamicTraceEventParser _Dynamic;
         internal /*protected*/ RegisteredTraceEventParser _Registered;
 #endif //  !NOT_WINDOWS
+
+        internal /*protected*/ string origin;
```
vancem: I don't want this string because it is relative inefficient to test for it (it is a string comparison).   I would prefer that you simply test with a type test (that is 'source is EventPIpeSource'), for ugly ad hoc things or put in a real virtual method if we think we new abstractions (see comments at use points).

src/TraceEvent/TraceEvent.cs
```diff
@@ -1869,6 +1872,11 @@ protected internal virtual void Validate()
         [Conditional("DEBUG")]
         protected internal void DebugValidate()
         {
+            // TODO FIX NOW For EventPipe data, the validation of EventDataLength for the following data always fail. 
+            // GCSuspendEETraceData, GCNoUserDataTraceData, MethodLoadUnloadVerboseTraceData, MethodJittingStartedTraceData
+            if (this.source.origin == EventPipeEventSource.EventPipe)
```
vancem: We need to understand why EventDataLength is wrong for those events.   It should not be... 

In general if we need to relax asserts we should do so at the particular event (and add a source is EventPipeEventSource test if it specific to EventPipes, todo to FIX It.

src/TraceEvent/TraceEventStacks.cs
```diff
@@ -239,8 +240,12 @@ public override StackSourceCallStackIndex GetCallerIndex(StackSourceCallStackInd
                     nextIndex += (int)threadIndex;
 
                     // Mark it as a broken stack, which come after all the indexes for normal threads and processes. 
-                    if (!ReasonableTopFrame(callStackIndex, threadIndex))
+                    // TODO Fix NOW Clr thread sample only has managed call stack. Do we have a better solution to differentiate it from normal call stack?
+                    if (!string.Equals(m_log.origin, EventPipeEventSource.EventPipe, StringComparison.OrdinalIgnoreCase)
```
vancem: We should make up a property onlyManagedCodeStacks for this purpose instead of origin.

src/TraceEvent/TraceEventStacks.cs
```diff
@@ -758,14 +761,18 @@ public StackSourceCallStackIndex GetCallStack(CallStackIndex callStackIndex, Sta
             {
                 callerIdx = top;
 
-                var frameName = GetFrameName(frameIdx, false);
-                var bangIdx = frameName.IndexOf('!');
-                if (0 < bangIdx)
+                // TODO Fix NOW Clr thread sample only has managed call stack. Do we have a better solution to differentiate it from normal call stack?
+                if (!string.Equals(m_log.origin, EventPipeEventSource.EventPipe, StringComparison.OrdinalIgnoreCase))
```
vancem: We should make up a property onlyManagedCodeStacks for this purpose instead of origin.

src/TraceEvent/TraceEvent.cs
```diff
@@ -2432,6 +2440,11 @@ void ConfirmAllEventsAreInEnumeration()
             if (GetType().Name == "JScriptTraceEventParser")
                 return;
 
+            // TODO FIX NOW ApplicationServerTraceEventParser contains some templates sharing the same event (task) name but different event ids
+            // The validation rule here doesn't quite work for them.
+            if (GetType().Name == "ApplicationServerTraceEventParser")
```
vancem: I would like to understand more about exactly what fails.   In general we should make the hacks to turn off asserts as narrow as possible.

src/TraceEvent/EventPipe/EventPipeTraceEventParser.cs
```diff
@@ -0,0 +1,214 @@
...
+namespace Microsoft.Diagnostics.Tracing.EventPipe
+{
+    public sealed class EventPipeTraceEventParser : ExternalTraceEventParser
+    {
+        public EventPipeTraceEventParser(TraceEventSource source)
+            : base(source)
+        {
+        }
+
+        public void AddTempalte(EventMetadata eventMetadata)
```
vancem: AddTempalte -> AddTemplate

vancem:
There is a failure  in the checkin test 

ERROR: File .\unzipped\netfx.4.6.1080.0.netfxrel3stage.x64.etl: event not equal to expected on line 339
   Expected: EVENT 2,876.635: System.Diagnostics.Eventing.FrameworkEventSource/ThreadPoolDequeueWork PID=1040; TID=6384; PName=Microsoft.VsHub.Server.HttpHost; ProceNum=3; DataLen=8; workID=242,867,940; 
   Actual  : EVENT 2,876.635: FrameworkEventSource/ThreadPoolDequeueWork PID=1040; TID=6384; PName=Microsoft.VsHub.Server.HttpHost; ProceNum=3; DataLen=8; workID=242,867,940; 

But I did not see anything you changed that might cause this  *it is only a question of the name of the name of the provider).  These will be rerun when you update the PR, so we will see if they persist.

src/TraceEvent/EventPipe/EventPipeEventSourceFactory.cs
```diff
@@ -0,0 +1,57 @@
+﻿using FastSerialization;
```
vancem: There is an expected pattern to using the serialization library that you are not following.   Basically things you want to deserialize you add the IFastSerializable interface.  THis gives you two methods to implement (ToStream and FromStream).  YOu put all the decoding logic into FromStream.  SInce we never create a EventPipeEventSource you can simply throw NotImplemented for ToStream.   

This is what EventPipeEventSource should do.  Then you don't need to have a factory, since you can say 

deserializer.GetEntryObject();
 
And it gets the 'first' object (which is represented by your whole file).   

Also you don't need an explicit V1 class for the EventPipeEventSource.   The idea is that the FromStream should remember the version number if needed (it does not in this case), and adapt what it does accordingly (that is users should not care about versions, by the time FromStream is done it is deserialized (and looks like the latest version).   



TraceLog.InitializeFromFile has an example, but you don't need the 'RegisterFactory' if the type has a default constructor that takes no arguments.

src/TraceEvent/EventPipe/EventMetadata.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Runtime.InteropServices;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Tracing.EventPipe
```
vancem: I recommend that this type  be a IFastSerializable (since it is something in the file), and its FromStream read in its fields.   But I also don't recommend that it have most of the fields below.   Instead it should have the TraceEvent type for that meta-data set up that has been filled out (including a EVENT_TRACE_HEADER blob as well as the blob size needed for the payload.   Basically anything you can do before you get the event data should be done here.  

The idea is that to dispatch an events you only need to deserialize the event number, look up its EventMetatData, copy the right number of bytes from the ETLX file to to the buffer associated with the EventMetatData, and return the TraceEvent (super fast).

src/TraceEvent/EventPipe/EventPipeEventSourceV1.cs
```diff
@@ -0,0 +1,373 @@
...
+            _header->EventHeader.Version = (byte)eventMetadata.Version;
+            _header->EventHeader.Level = (byte)eventMetadata.Level;
+            _header->EventHeader.Opcode = GetOpcodeFromEventName(eventMetadata.EventName);
+            _header->EventHeader.Id = (ushort)eventMetadata.EventId;
+            _header->EventHeader.Keyword = eventMetadata.Keywords;
+
+            _header->EventHeader.ActivityId = activityId;
+            _header->ExtendedDataCount = 1;
+            _header->ExtendedData = (TraceEventNativeMethods.EVENT_HEADER_EXTENDED_DATA_ITEM*)Marshal.AllocHGlobal(sizeof(TraceEventNativeMethods.EVENT_HEADER_EXTENDED_DATA_ITEM));
+            _header->ExtendedData->ExtType = TraceEventNativeMethods.EVENT_HEADER_EXT_TYPE_RELATED_ACTIVITYID;
```
vancem: You don't need to do this extended type thing.  TraceLog already had this problem of how to serialize extended data, and that is what the 'TraceEvent.myBuffer' field gets used for.   See TraceEvent.Clone() for more details, but basically things a TraceEvent understands how to be 'own' its own native memory (It will clean it up too),  You just need to allocate the extra space (for both its payload and if needed RelatedActivityID) and set it up load 'Clone()' does and TraceEvent will decode it properly.

vancem:
In thinking about this a bit more, my structural suggestions (avoid doing lots of work on a per-event basis, while good, could wait for a second phase (lets get something in that works).    Thus we can defer the issues I raised in EventPipeEvent*.cs and EventMetaData.cs.

src/TraceEvent/Computers/StartStopActivityComputer.cs
northtyphoon: I want to enable DynamicParser to handle EventSource events from EventPipe. Similar to RegisteredTraceEventParser, I add EventPipeTraceEventParser to DynamicTraceEventParser. It seems working. Can you take a look at the latest change?

src/TraceEvent/TraceEvent.cs
northtyphoon: I re-enable the validation. I can work with you to check why the EventDataLength is wrong.

src/TraceEvent/TraceEvent.cs
northtyphoon: I re-enable the validation in latest change to repro the issue.

northtyphoon: The following is an example in  ApplicationServerTraceEventParser that causes the assert failure. The two templates has the same eventName (taskName/opcodeName).

        public event Action<Multidata4TemplateHATraceData> ActivityStateRecord
        {
            add
            {
                source.RegisterEventTemplate(new Multidata4TemplateHATraceData(value, 103, 2609, "WorkflowTracking", WorkflowTrackingTaskGuid, 0, "Info", ProviderGuid, ProviderName));
            }
            remove
            {
                source.UnregisterEventTemplate(value, 103, ProviderGuid);
                source.UnregisterEventTemplate(value, 0, WorkflowTrackingTaskGuid);
            }
        }
        public event Action<Multidata3TemplateHATraceData> ActivityScheduledRecord
        {
            add
            {
                source.RegisterEventTemplate(new Multidata3TemplateHATraceData(value, 104, 2609, "WorkflowTracking", WorkflowTrackingTaskGuid, 0, "Info", ProviderGuid, ProviderName));
            }
            remove
            {
                source.UnregisterEventTemplate(value, 104, ProviderGuid);
                source.UnregisterEventTemplate(value, 0, WorkflowTrackingTaskGuid);
            }
        }

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -48,6 +49,8 @@ public DynamicTraceEventParser(TraceEventSource source)
             registeredParser = new RegisteredTraceEventParser(source, true);
             // But cause any of its new definitions to work on my subscriptions.  
             registeredParser.NewEventDefinition = OnNewEventDefintion;
+            // make an eventPipeTraceEventParser to resolve EventPipe events
+            eventPipeTraceEventParser = new EventPipeTraceEventParser(source, dontRegister: true);
```
vancem: I do NOT wish to have DynamicTraceEventParser (which is basically knows about inline manifest, tracelogging and classic ETW based events (but they are all ETW), to have any relationship with EventPipeTraceEventParser (which is its own format and technically has nothing to do with ETW).     

So ideally we do not couple them here.   However I am OK with fixing this after checkin.

vancem:
I have structural changes I would like to make but I don't want to block checkin of what we have so far (so that work can go on in parallel).    So I am merging this now.

PR #311, opened 2017-07-15
Fix LinuxPerfScriptEventParser when parsing timespamp event

gleocadie:
When parsing timestamp of an event ("0.0") with the current culture set to "fr-FR", a System.FormatException is raised.

### DIFF
--- a/src/LinuxEvent.Tests/EventParseTests.cs
+++ b/src/LinuxEvent.Tests/EventParseTests.cs
@@ -1,11 +1,9 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using Diagnostics.Tracing.StackSources;
 using Microsoft.Diagnostics.Tracing.Stacks;
+using PerfView.TestUtilities;
 using Xunit;
 
 namespace LinuxTracing.Tests
@@ -94,6 +92,16 @@ namespace LinuxTracing.Tests
 				new List<string>{ "module!symbol", "Thread (0)", "comm", null }
 			});
 		}
+
+        [Fact]
+        [UseCulture("fr-FR")]
+        public void OneStackButWithAnotherCulture()
+        {
+            string path = Constants.GetTestingPerfDumpPath("onegeneric");
+            this.DoStackTraceTest(path, doBlockedTime: false, callerStacks: new List<List<string>> {
+                new List<string>{ "module!symbol", "Thread (0)", "comm", null }
+            });
+        }
 
         [Fact(Skip = "FIX NOW")]
         public void ResolvingSymbols()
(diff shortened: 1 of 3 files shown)


sharwell:
Thanks for submitting this! After a quick look at codecov, it appears this line was getting covered by tests prior to your change. To add a test validating your change, a new test could be added next to one of the existing tests that covers the line, and apply the [`UseCulture`](https://github.com/Microsoft/perfview/blob/master/src/PerfView.Tests/StackViewer/StackWindowTests.cs#L22) attribute to force the new test to run with `fr-FR`.

gleocadie:
Thanks for your quick answer.
Prior to the PR, I tried to set the culture to "en-US" (using the UseCulture attribute) for one test, but it kept failing. When debugging step by step, I could see the Culture still set to "fr-FR". I dig deeper and noticed that this method was run in another thread (using  a Task) and that would explain why the test is failing. (https://msdn.microsoft.com/fr-fr/library/system.globalization.cultureinfo.currentculture(v=vs.110).aspx)

I'm not sure this test will bring any value, since on a machine with the Culture set to "en-US" with or without the fix, the test still will pass (it may confuse people playing with it).

But if you think it's worth it, I will add the test.

sharwell:
> this method was run in another thread

Do you know where exactly this transition occurs? If so, we can [explicitly preserve the culture at the boundary](https://github.com/Microsoft/perfview/blob/316ebcea0ac19b70c89e3b88cdc55654dbb98a08/src/PerfView/GuiUtilities/StatusBar/StatusBar.xaml.cs#L214-L215).

gleocadie:
> Do you know where exactly this transition occurs? If so, we can explicitly preserve the culture at the boundary.

Sorry, I removed the link when I cleaned up my message
https://github.com/Microsoft/perfview/blob/9dd9b1acee1f5f35ec28bb220777f81e64de22b8/src/PerfView/OtherSources/Linux/LinuxPerfScriptStackSource.cs#L51

Ok so, I will preserve the culture and add a test.

Thanks for your comment

src/PerfView/OtherSources/Linux/LinuxPerfScriptStackSource.cs
```diff
@@ -48,29 +47,45 @@ protected override void DoInterning()
                     threadBlockedTimeAnalyzers[i] = new List<BlockedTimeAnalyzer>();
                 }
 
+                var currentCulture = Thread.CurrentThread.CurrentCulture;
```
sharwell: :memo: (For reviewers) these changes are more easily reviewed with whitespace ignored:
https://github.com/Microsoft/perfview/pull/311/files?w=1

src/LinuxEvent.Tests/EventParseTests.cs
```diff
@@ -95,6 +93,16 @@ public void OneStack()
 			});
 		}
 
+        [Fact]
+        [UseCulture("fr-FR")]
```
sharwell: ❓ I'm assuming you ran this test on your machine without your code change, and the test failed due to the described localization problem. Can you confirm that? I will try it locally as well. :smile:

gleocadie: Yes, I ran the test without the change on double.Parse but with the preservation of the culture. The test failed. Setting the culture to **CultureInfo.InvariantCulture** in _double.Parse_ make the test green.

Issue #165, opened 2017-03-02
Port TraceEvent to .NET Core 


vancem:
This is transferring the following bug that was in the corefx repo https://github.com/dotnet/corefx/issues/4864

Basically we want the Microsoft.Diagnostics.Tracing.TraceEvent nuget package to work on .NET Core.

adamsitnik:
@vancem I just want to let you know that BenchmarkDotNet would love to have `Microsoft.Diagnostics.Tracing.TraceEvent` ported to .NET Core. We are currently using the existing version, and would like to give our .NET Core users great experience.

ViktorHofer:
@vancem Is there any progress on that matter or are you still investigating? We have some disabled tests in corefx for that.

vancem:
I have not made progress.  It should be a straightforward exercise (at least to make SOME progress), I just have not had time.    I will raise its priority a bit, but I still offer no promises.

vancem:
This has been fixed by PR #360.

adamsitnik:
@vancem When do you plan to release the new NuGet package?

vancem:
> When do you plan to release the new NuGet package

There is some work to do before I can release the package on Nuget.   In particular it needs to be signed, and some testing has to be added.   Also it has to be tested on the scenarios we have for that package (e.g.  there are samples).    This is not my top priority right now so it is hard to give a deadline, but it will be at least another week or so (the wildcard is how much time I can give it)

TyOverby:
Hi @vancem, it looks like this package has not made its way to nuget yet.  Is there anything I can do to help publish it?  I'm just down the hall from you if you want to chat!

benaadams:
Was trying to listen to some kernel events with .NET Core to detect blocking calls; but alas...

Would be great when the packages are shipped! 😉

PR #315, opened 2017-07-20
Trace event.net core2.0

vancem:
This is the beginning of a PR to address issue #165.   Basically we want a to have a TraceEvent Nuget package that builds against NetStandard and thus will work both on Desktop tools as well as .NET Core based tools.  

I attempted this some months ago but at the time the NetStandard V1.X API surface would have required non-trivial effort to make it work.   Because NetStandard V2.0 held the promise of making this port much easier, I simply waited.  

This pull request is my first attempt to build a TraceEvent Nuget package by using the V2.0 APIs.   The very good news is that getting it to build was surprisingly easy.   

At this point I have just gotten it to build (type build.cmd in src\TraceEvent\.NETCore).   It will make the Microsoft.Diagnostics.Tracing.TraceEvent.dll as well as a nuget package (currently only for X64).  

There were a number of fragile hacks that were done to make this work, and I KNOW that the we may have troubles with some interop and support DLLs.    I also believe I will probably reorganize where the build artifacts go (currently .NET Core builds are completely separate and not built by default, we ultimately will want to change that).  In short it is not done.  

For this reason I have put the Do Not Merge label on this as it is not really ready. 

However I wanted to put this PR out there in case anyone really needs this functionality, as gets you 80-90% of the way there, and anyone wishing to help with this effort will want to start here.


sharwell:
@vancem Let me know if/when you'd like me to look at this more closely. I've converted several projects to the new project system now so this would be just another one. :smile:

vancem:
This has been subsummed by PR #359

PR #317, opened 2017-07-20
Fix handling of commands in tree views

### LINKED ISSUE #316: Commands not working in several views
I noticed that commands like Include Item, Exclude Item, and Lookup Symbols appear to have no effect in certain views, such as the Call Tree view. As a workaround, users have to switch to the By Name view to apply the command, then switch back to where they were previously.

sharwell:
Fixes #316


src/PerfView.Tests/StackViewer/StackWindowTests.cs
```diff
@@ -18,6 +25,230 @@ namespace PerfViewTests.StackViewer
...
+                    await WaitForUIAsync(stackWindow.Dispatcher, cancellationToken);
+                }
+            }
+
+            return dataGrid;
+        }
+
+        // Since the implementation relies on hit testing and not just events coming from the proper sources, we must
+        // move the mouse prior to clicking.
+        private static void RaiseMouseInputReportEvent(Visual eventSource, int timestamp, int pointX, int pointY, int wheel)
```
sharwell: :link: this method taken from: https://stackoverflow.com/a/21276269/138304

vancem:
LGTM,  Just resolve the merge conflict

PR #319, opened 2017-07-21
[WIP] Move entire solution to new project system

sharwell:


sharwell:
:memo: This should now be ready for review

@vancem I used this branch as the basis for implementing .NET Standard builds of some of the projects, but I'll leave that for a follow-up review.

src/HeapDump/HeapDump.csproj
```diff
@@ -0,0 +1,83 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- net451 is actually just another net45 build build with the platform set to x64 -->
```
vancem: This is a bit strange to use a net451 to represent the X64 build.     Is there not more straightforward way of doing this?  Can we simply make up a new targetFramework (e.g. net45-X64)?

vancem:
Generally looks good (simpler).    There is a bunch of dead code removal, which I am generally OK with (most of it is reasonably old, and not likely to have going-forward value).

The one thing that I don't like is how we built both X64 and X86 version of HeapDump.exe.   If we can't find something nice with targetFramework, or other MSBUILD techniques, then I would recommend two .csproj files that pretty much just define one variable and include a .targets file that specifies the rest.  

You will need to resolve the conflict with appveyor.yml

src/HeapDump/HeapDump.csproj
sharwell: This limitation is due to a bug in one or more of dotnet/project-system, dotnet/roslyn, dotnet/sdk, and/or Microsoft/msbuild. It is my understanding that the teams do want to allow configurations on facets other than the predefined set of target frameworks. Ideally, a future update to this project file would have the following:

```xml
<TargetFramework>net45</TargetFramework>
<Platforms>AnyCPU;x64</Platforms>
```

Part of the ongoing effort to understand scenarios for the new project system is implementing things like we have in this file and making sure bugs are filed to make improvements based on real-world use cases. In this case, dotnet/sdk#350 suggests what we want might possible. dotnet/project-system#1938 (and the issues linked to it) track various issues related to multi-building on a facet other than just the target framework.

Given that the "trickery" is limited to a small number of lines, it seems like this approach should be fine even without a follow-up issue. But if it's concerning I could create an issue to make adjustments as soon as the project system supports it.

src/HeapDump/HeapDump.csproj
```diff
@@ -0,0 +1,83 @@
...
+    <TargetFrameworks>net45;net451</TargetFrameworks>
+    <OutputType>Exe</OutputType>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Prefer32Bit>true</Prefer32Bit>
```
sharwell: :memo: This line is unnecessary (set below for 32-bit build, and meaningless for the 64-bit build).

sharwell:
@vancem I wanted to make sure you are explicitly aware of the following ~~limitations~~ "changes" following this change (I'll wait for confirmation before fixing the conflict):

* The solution will no longer build prior to Visual Studio 2017. The *code* is still set to language version 6 and all the compiled outputs are the same, but the *projects* have changed.
* I removed the targets related to "razzle". I'm not sure what that is or whether or not it was still in use. If it's part of the signed build release process I was assuming we'd use something closer to what Roslyn is using.

src/HeapDump/HeapDump.csproj
vancem: My concern is that this looks like a feature that may never happen ( https://github.com/dotnet/sdk/issues/350 is closed (long ago), and 1938 does not address the issue squarely).    I would not bank on the fix, and even if it comes, it would tie us to a very new version of VS, which is not particularly desirable.   

While small hacks may be reasonable especially if they are time-bounded,, it also seems like the idea of two .csproj files that define one property and include a common .targets file is also very clean, supported now, and about the same total lines of code.   Seems like a win...

vancem:
Yes, I saw the two things you mentioned.   It is unfortunate to abandon 2015, but as I have said in the past, I am willing to drop compatibility there if there is obvious value in the 2017 features.    The desire to make TraceEvent build from this repo for .NET core, seems to be the tipping point for me. 

I am glad that we don't need the absolute latest version of 2017 (if that were so I would ask that you put a check in the build that gives a CLEAR error of what to do, because people WILL hit it).   (In fact if it is not already the case that it will clearly fail with an appropriate message on 2015, we should see if we can make that happen).

Removing the razzle stuff is goodness.   I was used a while ago when the code was shared with VS and that stopped but we never removed it.  It is just unnecessary complexity, and I am happy it is gone.

sharwell:
> The desire to make TraceEvent build from this repo for .NET core, seems to be the tipping point for me.

I have this change ready. It's a very small change on top of this pull request, but I was planning to submit it separately.

> I am glad that we don't need the absolute latest version of 2017

I was worried, but AppVeyor shows it works at least in 15.2.

> In fact if it is not already the case that it will clearly fail with an appropriate message on 2015, we should see if we can make that happen.

Good idea. Currently the error is terrible (says it's an incompatible VS 2003 solution), but I'm going to push a commit that marks the solution as requiring VS 2017. Opening in 2015 will have this message:

image

sharwell:
:memo: I have a couple additional things to check before we merge this

vancem:
I just tried syncing up to this and building as a final check before merging and I get the following errors trying to compile the ETWClrProfiler DLL. 

This is on VS 15.2.  Do you believe that I need to upgrade to VS15.3?    That seems odd since I am assuming the 8.1 SDK is old.   

1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VC\VCTargets\Platforms\Win32\PlatformToolsets\v141\Toolset.targets(34,5): error MSB8036: The Windows SDK version 8.1 was not found. Install the required version of Windows SDK or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution".
1>Done building project "ETWClrProfilerX86.vcxproj" -- FAILED.
2>C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VC\VCTargets\Platforms\x64\PlatformToolsets\v141\Toolset.targets(36,5): error MSB8036: The Windows SDK version 8.1 was not found. Install the required version of Windows SDK or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution".
2>Done building project "ETWClrProfilerX64.vcxproj" -- FAILED.

sharwell:
@vancem Did you install the C++ workload for Visual Studio 2017? I haven't made any changes to the C++ projects in this branch, so I would expect if they worked for you in Visual Studio 2017 on the master branch then they would still work for you in this branch.

vancem:
Hmm, it certainly worked before, but it was not a controlled experiment.   Let me try some things.

sharwell:
Here you go:
https://developer.microsoft.com/en-us/windows/downloads/windows-8-1-sdk

vancem:
The machine I happened to be working from I had to reinstall VS recently.  I just installed 15.3 and by default it does not seem to include the windows 8.1 SDK.   Thus there is potentially an issue where our instructions for new users are not correct.   

However this is independent of these changes, so I will not block this merge. 

Thanks.  This is a very nice improvement.    Sorry it took so long to get it reviewed

PR #323, opened 2017-07-23
Missing containsSelfDescribingMetadata in serialization

### LINKED ISSUE #322: Processing EventSourceRichpayload events in Windows Server 2012 R2 is not reliable.
I can share the large etl separately if needed. As shown in the comparison between WS2012R2 and Win10, the Microsoft-ApplicationInsights-Data/Operation/Start events are processed as Microsoft-ApplicationInsights-Data/Operation/Stop events.

WS2012R2
image

Win10
image

The difference is because the call to **TdhGetEventInformation** in [RegisteredTraceEventParser.cs](https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/RegisteredTraceEventParser.cs#L506) returns a non-empty EventGuid for self-describing TraceLogging events in Windows Server 2012 R2. The EventGuid is used as TaskGuid in event template. When the event template is registered, the non-empty TaskGuid causes the template [double-registered](https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/TraceEvent.cs#L3364). For the repro case, the Operation/Stop event template is even registered as a classic template for Operation/Start event.

A related issue is [DynamicTraceEventData](https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/DynamicTraceEventParser.cs#L1460) doesn't persist containsSelfDescribingMetadata in trace log (.etlx). The information can avoid the above template registration issue.

northtyphoon:
The change is to fix #322. 

@vancem @pharring

### DIFF
--- a/src/TraceEvent/DynamicTraceEventParser.cs
+++ b/src/TraceEvent/DynamicTraceEventParser.cs
@@ -1470,6 +1470,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
             serializer.Write(MessageFormat);
             serializer.Write(lookupAsClassic);
             serializer.Write(lookupAsWPP);
+            serializer.Write(containsSelfDescribingMetadata);
 
             serializer.Write(payloadNames.Length);
             foreach (var payloadName in payloadNames)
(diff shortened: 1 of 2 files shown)


vancem:
LGTM

PR #335, opened 2017-08-02
Ensure StatusBar.m_work is assigned before the work starts

sharwell:
Fixes an occasional assertion failure (and subsequent test hang) where `EndWork` was called before `m_work` was assigned.


sharwell:
@vancem I merged this since it was showing up in several recent build failures.

Issue #332, opened 2017-08-01
TraceEvent Stack Traces not working for x64 Processes


Alois-xx:
I have tried the TraceLogMonitor sample which is a great tool for real time exception tracing. But unfortunately it looks like TraceEvent is only capable to decode stack traces from x86 processes. Stack traces from x64 stacks show up as empty. 

Am I missing something obvious? 
The stacks for x64 look like 
`System.IO.FileNotFoundException: Could not load file or assembly 'MemoScope.XmlSerializers' or one of its dependencies. The system cannot find the file specified.

`
Whereas for x86
```
System.IO.FileNotFoundException: Could not load file or assembly 'MemoScope.XmlSerializers' or one of its dependencies. The system cannot find the file specified.
        EtwCallout
        CoTemplate_zzpqhh
        ETW::ExceptionLog::ExceptionThrown
        COMPlusThrowCallback
        Thread::StackWalkFramesEx
        Thread::StackWalkFrames
        CPFH_RealFirstPassHandler
        CPFH_FirstPassHandler
        _COMPlusFrameHandler
        _COMPlusNestedExceptionHandler
        ExecuteHandler2
        ExecuteHandler
        KiUserExceptionDispatcher
        RaiseTheExceptionInternalOnly
        UnwindAndContinueRethrowHelperAfterCatch
        AssemblyNative::Load
        System.Reflection.RuntimeAssembly.nLoad(System.Reflection.AssemblyName, System.String, System.Security.Policy.Evidence, System.Reflection.RuntimeAssembly, System.Threading.Sta
        System.Reflection.RuntimeAssembly.LoadWithPartialNameInternal(System.Reflection.AssemblyName, System.Security.Policy.Evidence, System.Threading.StackCrawlMark ByRef)
        System.Reflection.Assembly.LoadWithPartialName(System.String, System.Security.Policy.Evidence)
        System.Xml.Serialization.TempAssembly.LoadGeneratedAssembly(System.Type, System.String, System.Xml.Serialization.XmlSerializerImplementation ByRef)
        System.Xml.Serialization.XmlSerializer..ctor(System.Type, System.String)
        System.Xml.Serialization.XmlSerializer..ctor(System.Type)
        CallDescrWorkerInternal
        CallDescrWorkerWithHandler
        DispatchCallDebuggerWrapper
        DispatchCallSimple
        MethodTable::RunClassInitEx
        MethodTable::DoRunClassInitThrowing
        CEEInfo::initClass
[...]

PR #341, opened 2017-08-04
Fix for missing callstacks in real time Tracelog sessions (issue 332)

Alois-xx:
When using  Realtime tracing the call stacks of x64 processes did come up empty because the code addresses were treated as kernel addresses which did result in the assignment of 0 (IDLE) process instead of the real processes where the stack was generated.  See issue #332

### DIFF
--- a/src/TraceEvent/ETWReloggerTraceEventSource.cs
+++ b/src/TraceEvent/ETWReloggerTraceEventSource.cs
@@ -407,7 +407,7 @@ namespace Microsoft.Diagnostics.Tracing
                         // first event (real time case).   This is really a problem, as we need that information, but we will go ahead and
                         // try to initialize as best we can. 
 
-                        m_source.pointerSize = sizeof(IntPtr);
+                        m_source.pointerSize = Environment.Is64BitOperatingSystem ? 8 : 4;
                         m_source.numberOfProcessors = Environment.ProcessorCount;
                         m_source._QPCFreq = Stopwatch.Frequency;
                         m_source._syncTimeUTC = DateTime.UtcNow;
@@ -515,6 +515,6 @@ namespace Microsoft.Diagnostics.Tracing
         TraceEventNativeMethods.EVENT_RECORD* m_curTraceEventRecord;             // This is the TraceEvent eventRecord that corresponds to the ITraceEvent. 
         TraceLoggingEventId m_traceLoggingEventId;                               // Used to give TraceLogging events Event IDs. 
 
-#endregion
+        #endregion
     }
 }
--- a/src/TraceEvent/ETWTraceEventSource.cs
+++ b/src/TraceEvent/ETWTraceEventSource.cs
@@ -455,9 +455,9 @@ namespace Microsoft.Diagnostics.Tracing
                 sessionEndTimeQPC = this.UTCDateTimeToQPC(maxSessionEndTimeUTC);
             }
             Debug.Assert(_QPCFreq != 0);
-            if (pointerSize == 0)       // Real time does not set this (grrr). 
+            if (pointerSize == 0 || IsRealTime)  // We get on x64 OS 4 as pointer size which is wrong for realtime sessions. Fix it up. 
             {
-                pointerSize = sizeof(IntPtr);
+                pointerSize = Environment.Is64BitOperatingSystem ? 8 : 4;
                 Debug.Assert((logFiles[0].LogFileMode & TraceEventNativeMethods.EVENT_TRACE_REAL_TIME_MODE) != 0);
             }
             Debug.Assert(pointerSize == 4 || pointerSize == 8);
--- a/src/TraceEvent/NativeDlls.cs
+++ b/src/TraceEvent/NativeDlls.cs
@@ -15,7 +15,9 @@ class NativeDlls
     /// <param name="simpleName"></param>
     public static void LoadNative(string simpleName)
     {
-        var thisDllDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().ManifestModule.FullyQualifiedName);
+        // When TraceEvent is loaded as embedded assembly the executing assembly is mscorlib which points to the wrong path.
+        // We use therefore the executable as entry point
+        var thisDllDir = Path.GetDirectoryName(Assembly.GetEntryAssembly().ManifestModule.FullyQualifiedName);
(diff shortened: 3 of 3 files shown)


src/TraceEvent/ETWReloggerTraceEventSource.cs
```diff
@@ -407,7 +407,7 @@ private unsafe void Initialize(TraceEventNativeMethods.EVENT_RECORD* rawData)
                         // first event (real time case).   This is really a problem, as we need that information, but we will go ahead and
                         // try to initialize as best we can. 
 
-                        m_source.pointerSize = sizeof(IntPtr);
+                        m_source.pointerSize = Environment.Is64BitOperatingSystem ? 8 : 4;
```
jorive: Is it possible to use `IntPtr.Size` instead?

src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -455,9 +455,9 @@ private void Initialize(string fileOrSessionName, TraceEventSourceType type)
                 sessionEndTimeQPC = this.UTCDateTimeToQPC(maxSessionEndTimeUTC);
             }
             Debug.Assert(_QPCFreq != 0);
-            if (pointerSize == 0)       // Real time does not set this (grrr). 
+            if (pointerSize == 0 || IsRealTime)  // We get on x64 OS 4 as pointer size which is wrong for realtime sessions. Fix it up. 
             {
-                pointerSize = sizeof(IntPtr);
+                pointerSize = Environment.Is64BitOperatingSystem ? 8 : 4;
```
jorive: Is it possible to use IntPtr.Size instead?

src/TraceEvent/ETWReloggerTraceEventSource.cs
Alois-xx: I am not sure what is right here. When an ETL is recorded it can contain x64 and x86 stacks. So the structures in the ETL file should be all dependant on the bitness of the OS and not the process starting the trace. 
It could be that depending on the source.pointersize various assumptions are encoded which will lead to followup errors. 
I can strip the change down to the concrete fix to make realtime tracing work and leave the rest as is.

vancem: I agree that pointerSize should be the underlying OS pointer size (we should never be using IntPtr (that is the size for the local process) to determine this value.    This is a good change.

src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -455,9 +455,9 @@ private void Initialize(string fileOrSessionName, TraceEventSourceType type)
                 sessionEndTimeQPC = this.UTCDateTimeToQPC(maxSessionEndTimeUTC);
             }
             Debug.Assert(_QPCFreq != 0);
-            if (pointerSize == 0)       // Real time does not set this (grrr). 
+            if (pointerSize == 0 || IsRealTime)  // We get on x64 OS 4 as pointer size which is wrong for realtime sessions. Fix it up.
```
vancem: Ideally we would not have the || IsRealTime in this condition.   Basically the ETW system in theory knows the pointer size and we should just be using that.  We only 'take control' and set it if it is clear that the ETW system did not (it set it to 0).   

Is it really the case that the pointerSize comes back as 4 on a when you create a real-time ETW source?

src/TraceEvent/NativeDlls.cs
```diff
@@ -15,7 +15,9 @@ class NativeDlls
     /// <param name="simpleName"></param>
     public static void LoadNative(string simpleName)
     {
-        var thisDllDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().ManifestModule.FullyQualifiedName);
+        // When TraceEvent is loaded as embedded assembly the executing assembly is mscorlib which points to the wrong path.
```
vancem: I would like to understand this scenario more.   GetExeuctingAssembly gets the assembly of the CALLER of that API (otherwise it would be useless, as you point out it would always return mscorlib).   

Thus the existing code will look for the DLL 'next to' the assembly that compiled NativeDlls.cs (which would be Microsoft.Diagnostics.Tracing.TraceEvent.dll).  That is what I want.   There are scenarios where you don't control
the entry assembly (that is the EXE) so you don't really want to use that.   

Basically I am asking for more details about the issue that drove this change.   I think the code as it was should work.

src/TraceEvent/ETWTraceEventSource.cs
Alois-xx: Yes I have checked that on x64 I have got 4 as pointer size back. I guess that can happen if you start tracing from an x86 process but I have not looked deeper but the  set value was definitely wrong.

src/TraceEvent/NativeDlls.cs
Alois-xx: I have several tools which rely on TraceEvent in different versions. I would like to embed TraceEvent and related things and load them as byte arrays into the AppDomain so I do not need to deploy the referenced assemblies to prevent breaking any tool relying on TraceEvent. 

I just have found that OSExtensions.dll
`private static void LoadKernelTraceControl()
{
	if (!ETWKernelControl.s_KernelTraceControlLoaded)
	{
		try
		{
			string arg_27_0 = Path.GetDirectoryName(Assembly.GetExecutingAssembly().ManifestModule.FullyQualifiedName);
`

uses the same trick which breaks the approach. I know that these wrapper are searched relative to the dll but if it is loaded from a byte array the Manifest path will be <Unknown> which then breaks the Path.Combine call due to invalid characters. 
A non breaking workaround would be to check first the returned ManifestModule.FullyQualifiedName for <Unkown>. If this is bad then use the executable directory as fallback in TraceEvent and OsExtensions. That way the old deployments are not broken and new ones are enabled.

Alois-xx: By the way the publicly used KernelTraceControl.dll does not work on Windows 7 because it is linked with ApiSets agains Win8? It fails with  

`Application popup: LiveExceptionMonitor.exe - Entry Point Not Found : The procedure entry point ReleaseSRWLockExclusive could not be located in the dynamic link library api-ms-win-core-synch-l1-2-0.dll. 
`
That should be fixed as well. 
As feature request I would like to use "Circular Kernel Context Logger" as my second trace logger to get out of the way of procmon, procexp, ... and "real" profiling. That would also be a big help. This is a big help on Win7 where officially there is only one kernel logger allowed.

PR #344, opened 2017-08-09
Added TraceParserGen

rovarma:
I've added the source for TraceParserGen from [OneDrive](https://onedrive.live.com/?id=D068505F1C71AB61%21124&cid=D068505F1C71AB61)  (traceParserGen.src.zip). According to @vancem this is the most recent code for it. 

I've verified it works with my manifest, but haven't tested it extensively.

The code contains quite a lot of various #if preprocessor blocks; enabling some of them makes the code not compile or run again. One particular part that's disabled is the ability to generate a parser from a .mof file, which I'm interesting in having. I may spend some time getting that to work again, but that's for a different PR.


PerfView.sln
```diff
@@ -61,76 +61,146 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "PerfView.Tests", "src\PerfV
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "PerfView.TestUtilities", "src\PerfView.TestUtilities\PerfView.TestUtilities.csproj", "{FE5CC86D-E87E-4560-8004-8852F3DE6794}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TraceParserGen", "src\TraceParserGen\TraceParserGen.csproj", "{034CB9AF-7E9E-4D96-8F71-526999D284BA}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
sharwell: :memo: I wish I could force it to stop adding these, though I'm pretty sure that will mean renaming the current "Mixed Platforms" solution platform to "Any CPU".

sharwell:
@vancem I assume this is a straightforward inclusion of the original code. There are some minor tweaks to the project and solution file that will be needed for uniformity. I can push these changes to **rovarma/perfview** prior to merging this pull request, but need to know whether to do it before or after #319 is merged.

rovarma:
@sharwell Yes, the code is an unmodified version of the code on OneDrive. I'm not sure what tweaks are needed, but feel free to do so :)

vancem:
Looks good.     The most pressing things to update is a bit of documentation.   Still this is forward progress.

sharwell:
@vancem Cool, I'll merge this into #319 and do the necessary updates there.

PR #345, opened 2017-08-11
Update TraceEvent to include a .NET Standard 2.0 build

sharwell:
:memo: This updates the project to produce the new outputs, but the NuGet package specification still needs to be updated to include the new binaries.

### DIFF
--- a/src/TraceEvent/RegisteredTraceEventParser.cs
+++ b/src/TraceEvent/RegisteredTraceEventParser.cs
@@ -7,12 +7,14 @@ using System.Runtime.InteropServices;
 using System.Security;
 using System.Text.RegularExpressions;
 using FastSerialization;
-using System.Diagnostics.Eventing;
 using Microsoft.Diagnostics.Tracing.Session;
 using Microsoft.Diagnostics.Tracing.Extensions;
 using System.IO;
 using System.Threading;
 
+#if !NETSTANDARD2_0
+using System.Diagnostics.Eventing;
+#endif
 
 namespace Microsoft.Diagnostics.Tracing.Parsers
 {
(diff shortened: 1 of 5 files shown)


src/TraceEvent/RegisteredTraceEventParser.cs
```diff
@@ -1076,6 +1078,88 @@ internal struct TRACE_EVENT_INFO
             public EVENT_PROPERTY_INFO EventPropertyInfoArray;  // Actually an array, this is the first element.  
         }
 
+#if NETSTANDARD2_0
```
vancem: I prefer that the 'going forward' code base have as few #ifdefs as possible (that the #ifdefs are only there for hacks to make older things work).      System.Diagnostics.Eventing.EventDescriptor has not made it into NetStandard 2.0 so we should NEVER rely on it.   Looking around, I see that only use of it seems to be in the definition of TRACE_EVENT_INFO.  Thus the definition should be put there (it could also be simplified to be a simple struct with correct fields of the correct size, which is what we do with all the other interop structs.   This definition should be unconditional.   (it is just another interop struct that happens to be used in RegisteredTraceEventParser.cs

vancem: Otherwise looks good.

src/TraceEvent/TraceEvent.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net45</TargetFrameworks>
+    <TargetFrameworks>net45;netstandard2.0</TargetFrameworks>
```
vancem: One step at a time.  Lets get 2.0 working, and then we can see what delta we need to make 1.6 work.

Can you describe the value of supporting 1.6?

jorive: It would be nice to use the TraceEvent library down to .NET Core 1.0. The ability to harness, measure performance, and compare between 1.0 and 2.0 runtimes is important for the dotnet family (e.g. CoreClr, CoreFx), where we track the product performance over time on the different released branches and master.

src/TraceEvent/RegisteredTraceEventParser.cs
vancem: I have made the change I suggested above to remove the dependency to System.Diagnostics.Eventing.EventDescriptor  with PR #357

vancem:
I believe PR #359 which was just merged subsumes this change.    Basically I took this change, removed the part hat PR #357 addresses, and then added changes as I did testing.    The additional changes were mostly updates to the support package.

PR #347, opened 2017-08-14
Fix debug launch for PerfView

### LINKED ISSUE #346: By default launching the debugger (f5) from a clean PerfView build fails
The perfView readme.md instructions tells you all you need to do to debug PerfView is to set the PerfView project as the startup project and then start debugging (F5).  

If you do this currently with a clean repository you get a popup 
```
---------------------------
Microsoft Visual Studio
---------------------------
Unable to run your project. The "RunCommand" property is not defined.
---------------------------
```
You can work-around this getting to the Debug properties of the project (right clicking on the PerfView project in the solution explorer -> Properties), and changing the 'Launch' field from 'Project' to 'Executable' and setting the 'Executable' field to src\PerfView\bin\Debug\net45\PerfView.exe  (Debug->Release  if you are debugging a release build).  

However this work-around should not be necessary.   We need to update the project so that it just works (probably by defining the 'RunCommand' variable but I don't have a lot of experience with the new build infrastructure).  

@sharwell

sharwell:
Fixes #346


vancem:
LGTM

PR #349, opened 2017-08-17
support for GCs under new condition and bug fix

Maoni0:
+ added support for full GCs triggered during provisional mode
+ fixed a bug in LOH alloc calculation


Maoni0:
@swgillespie PTAL

PR #350, opened 2017-08-17
Adding support for parsing and Analyzing IIS ETW events in Perfview

puneetg1983:
In this pull request, I am adding support for parsing IIS events by adding the **IisTraceEventParser** which is generated with the help of TraceParserGen tool and after converting the classic IIS MOF to XML.
I also added a HTML view called "IIS Stats" which allows IIS administrators and ASP.NET developers to easily identify slow and failed requests. Using the same view you can drill down to see all the EVENTS fired for a particular request in the IIS stack and can drill down easily in the ThreadStacks view where we identify that the requests are slow in the ManagedPipeline handler.

**IIS Stats in PerfView**

The IIS Stats node under Perfview has 4 major sub-reports


vancem:
Hello Puneet, 
Sorry it has taken so long for me to review this.  

As mentioned, I think in the limit I would like to explore getting this IIS information into the ThreadTime with Start-Stop view.   However I am OK with the idea of other views as well.

What I like about this change is that it is almost entirely additive (you did not need to change existing behavior).     There is only one change I would like you to make. 

1. Please put the IISStats view in the 'AdvancedGroup' folder (not at the top level).   I really only want the views that I believe EVERYONE will want to know about.    This may become true for IISStats, but lets see how it goes first.  

I think I would like to move things like PerfVIewIISStats out of the PerfViewData.cs file, but that is how all the others work today, and it is best to follow suite as you have done until we decide to break it apart for everyone.  

So I guess it is just that one change, and I will merge things.

Note that I will be out on Monday watching the solar eclipse that is in the US that day, so my responses may be delayed until Tuesday.

puneetg1983:
Hi Vance,
I understand. I have made this change and pushed it to github.

[cid:image001.png@01D31A7C.189CEA40]

Waiting for you to merge 😊

Thanks,
Puneet Gupta

vancem:
Puneet, I have merged this. however I am having trouble opening your sample traces.   DO the *.diagsession files open for you?

puneetg1983:
Nope I opened an issue https://github.com/Microsoft/perfview/issues/356

Issue #354, opened 2017-08-17
TestIncludeItemOnCallerCalleeTabCallerAsync  fails locally


vancem:
WhenI run all tests in VS2017 I get one failure 

TestIncludeItemOnCallerCalleeTabCallerAsync  

It is not clear if this test is enabled (it does have a 'WorkItem' attribute which may mean it is disabled, but VS runs it.  

Anway it fails, because it is fetching the 'callers' of a caller-callee view and the focus node happens to be ROOT (which does not have a caller, so a Items[0] throws a invalid Index exception).  

It is not clear to me what the intent of the test is (it seems like the test should never have worked).  It should probably select some other node besides root for the focus node.  

@sharwell 

Also, the automated tests succeed.   It is not clear why this fails locally but not in the automated run.   That is concerning...

sharwell:
@vancem I can't reproduce this. What's supposed to happen is the following:

* If we're trying to test the callees grid and it's empty, double click the first item in the callers list, which changes the current focus node to a node we know has a callee.
* If we're trying to test the callers grid and it's empty, double click the first item in the callees list, which changes the current focus node to a node we know has a caller.

The handling is here:

https://github.com/Microsoft/perfview/blob/fbca7255dc5c3f78e2c5e5d33ce29001f8c99f50/src/PerfView.Tests/StackViewer/StackWindowTests.cs#L183-L210

This test is part of a comprehensive regression test for #316. It ensures the context menu behaves properly in all potential contexts. Each context is handled separately, and previously this handling had some bugs, so a simple test of a subset of the known grids is not sufficient to detect regressions.

mikkelbu:
@sharwell this test is failing on my machine every time I execute it. Setup:
* VisualStudio.15.Release/15.3.5+26730.16
* Standing on the commit 118f6bc, but with the addition of `<WindowsTargetPlatformVersion>10.0.15063.0</WindowsTargetPlatformVersion>` in both vcxproj-files (otherwise I get `Cannot open include file: 'assert.h': No such file or directory` when compiling `perfview\src\etwclrprofiler\stdafx.h`)
* The tests are executed following the description in CONTRIBUTING.md by selecting Test -> Run -> All Tests from within VS.

I've attached the full error message from the failing test (the stacktrace is not that usable, due to lambdas etc.). [Perfview_Test_ErrorMessage.txt](https://github.com/Microsoft/perfview/files/1484000/Perfview_Test_ErrorMessage.txt)

[...]

PR #358, opened 2017-08-22
Disabled test tracked by #354

vancem:


src/PerfView.Tests/StackViewer/StackWindowTests.cs
```diff
@@ -37,7 +37,7 @@ public Task TestIncludeItemOnByNameTabAsync()
             return TestIncludeItemAsync(KnownDataGrid.ByName);
         }
 
-        [WpfFact]
+        [WpfFact(Skip = "Failing with indexOutOfRange.  See issue https://github.com/Microsoft/perfview/issues/354")]
```
sharwell: :bulb: Overall this was handled great given I wasn't available for responding. For Roslyn we don't bother with a longer message since the issue will already contain it, so this would be considered sufficient:

```csharp
[WpfFact(Skip = "https://github.com/Microsoft/perfview/issues/354")]
```

PR #360, opened 2017-08-23
Fixes for TraceEvent .NET Core 2.0 support

vancem:
Additional changes to make TraceEvent actually work at runtime on .NET Core 2.0 

1. Updated OSExtensions to remove dependencies on APIs not in .NETStandard 2.0
2. Changed OSExtentions to use the new build system. (that supports .NET Core)


src/TraceEvent/TraceEvent.csproj
```diff
@@ -54,7 +54,7 @@
   </Choose>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles" Version="1.0.7" PrivateAssets="all"/>
+    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles" Version="1.0.7" />
```
sharwell: @vancem Can you describe the issue that led to this? I want to make sure I understand the intended distribution mechanism for these files.

vancem: I am not certain what your question is, but the goal was to simply get simple scenarios that need OSExtensions (Starting a kernel session, or a heap session).

PR #372, opened 2017-08-29
Support change from Guids to string names as EventProvider identifiers 

nategraf:
This PR supports the changes to the coreclr (https://github.com/dotnet/coreclr/pull/13370) which change the identifier for event providers from Guids to string names. It allows Perfview to read the new file format.


vancem:
Note that a checkin tests fails   You should be able to run it in VS with Test -> Run -> All Tests 
```
    TraceEventTests.EventPipeParsing.Basic(eventPipeFileName: "eventpipe-dotnetcore2.0-linux.netperf") [FAIL]
      The event statistics doesn't match C:\projects\perfview\src\TraceEvent\TraceEvent.Tests\bin\Debug\net46\inputs\eventpipe-dotnetcore2.0-linux.netperf.baseline.txt. It's saved in C:\projects\perfview\src\TraceEvent\TraceEvent.Tests\bin\Debug\net46\inputs\eventpipe-dotnetcore2.0-linux.netperf.actual.txt.
      Expected: True
      Actual:   False
      Stack Trace:
        src\TraceEvent\TraceEvent.Tests\EventPipeParsing.cs(107,0): at TraceEventTests.EventPipeParsing.ValidateEventStatistics(SortedDictionary`2 eventStatistics, String eventPipeFileName)
        src\TraceEvent\TraceEvent.Tests\EventPipeParsing.cs(87,0): at TraceEventTests.EventPipeParsing.Basic(String eventPipeFileName)
      Output:
        OutputDir: C:\projects\perfview\src\TraceEvent\TraceEvent.Tests\bin\Debug\net46\output\1b29b2a8
        Processing the file C:\projects\perfview\src\TraceEvent\TraceEvent.Tests\bin\Debug\net46\unzipped\eventpipe-dotnetcore2.0-linux.netperf, Making ETLX and scanning.
```

vancem:
Documenting that this change goes hand in hand with https://github.com/dotnet/coreclr/pull/13370 in CoreCLR which uses this new format.

nategraf:
@vancem I think the test was actually slightly off. Have a look at the changes [1] and let em know if you agree. Prior to this PR two of the events in the test case failed to set their EventNames. A tangential change I made [2] addresses this. 

Now I may be wrong, and that's the ways it was supposed to be. If so, let me know and I will change [2] back


[1] https://github.com/Microsoft/perfview/pull/372/files#diff-494690ca4ea31318bb46dd3f9d5b9e89L50

[2] https://github.com/Microsoft/perfview/pull/372/files#diff-dca02247a9f2ee8460bd8eea72b43dc3L46

nategraf:
There is an issue that just appeared. The asserts in the Validate functions within ClrTraceEventParser will fail because the payloads now contain strings instead of Guids (and are longer). What should be done about this validation code?

vancem:
> The asserts in the Validate functions within ClrTraceEventParser will fail because the payloads now contain strings instead of Guids (and are longer). What should be done about this validation code?

I don't actually understand the problem (can you give details?)   I think you are talking about

vancem:
By the way I like the update to fix the baseline.   Cleary event names are better than event numbers.

brianrob:
Agreed, I like the event names as well.  Much clearer.

nategraf:
My assumption based on the changes I made was that this has to do with the EventMetaData. On the PerfView [1] side and on the and CLR [2] side the identifier (previously Guid, now string) is factored into the payload size.

Would a change in the size of the meta data instances trigger the Validate function's asserts?

[1] https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/EventPipe/EventPipeEventSourceV1.cs#L149
[2] https://github.com/dotnet/coreclr/blob/master/src/vm/eventpipeconfiguration.cpp#L357

vancem:
@nategraf  what exactly fails?     Yes it is true that you have changed the size of the event (before it had a guid and now a string), but CLRTRaceEventParser has no knowledge of that.   It can see the PAYLOAD size, but that is set in ReadAndDispatchEvent and refers to the PAYLOAD (user defined fields) that you did not change. 

Now if you are getting an assert, that can't be argued with, but lets start from there and figure out exactly why it is happening because your change should not have caused it...

nategraf:
When I try to see all the events of the "Microsoft-Windows-DitNETRuntimeRundown" type it throws the asserts for each event. 

These are the exact asserts that are triggering:
1. https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/Parsers/ClrTraceEventParser.cs#L9729
2. https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/Parsers/ClrTraceEventParser.cs#L8153
3. https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/Parsers/ClrTraceEventParser.cs#L8258
4. https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/Parsers/ClrTraceEventParser.cs#L8339

This is the call stack which is the same each time, except for the top frame: https://pastebin.com/59hiv4kE

I don't know if that answers your question on what failed exactly. I'll keep looking for a more specific source

vancem:
You can see in these asserts that EventDataLength is supposed to be a number like 2 (small).   You can see that it should only be the size of the payload.   What value is it?  Please track down how the payload was initialized for that event and see how you could have changed that logic.   Right now things don't really make sense...

vancem:
I am going to merge this, and so I can easily repo any asserts you are hitting.

nategraf:
So far I've tracked it back to an issue in coreclr eventpipebuffer.cpp. It's actually an error I introduced with the other eventpipe PR I wrote. I just never noticed it because I was always using the Release version of PerfView. 

I should be able to handle it from here

vancem:
Great.   When you have a clean version 2 netstat.zip file please put the file in src\TraceEvent\TraceEvent.Tests\inputs and make a baseline.txt file for it so that the EventPipeParsing.Basic test will simply pick it up and test if any asserts happen.

northtyphoon:
EventPipeParsing.Basic test currently is using DebugListenerBlock to disable all assert failures (otherwise the testcase couldn't get in and validate the subsequent checkins). After you clean up all the assertion, please remember to remove the DebugListenerBlock.

vancem:
Oh my, we should not have that.    I have created issue #373 to track it.

vancem:
@nategraf  I turned the asserts back on and most of them seem to be that the payload that is passed is bigger than it needs to be (often it rounds up to 32).   Is this what you are fixing?  

We should fix this (presumably it is a bug on the CLR side).

nategraf:
Yes, that's what I am working on. I'm confirming what I think is the cause now

nategraf:
Actually, my theory was incorrect. These assert failures did not originate from my changes. I built from 0bd583ccecc46d106ccb6af9f2f1e75f515460f8 (before my first change to EventPipe) and it has the same asserts come up in the Validate functions. 

@vancem Do you want me to keep investigating this as a new task, or hand it off to someone else?

northtyphoon:
It's one of the two unsolved assertion issues in my previous change. I can help to chase it down. I need to better understand what the assertion is checking.  I was confused why the payloads were still readable even the length was off according to the assertion.

vancem:
The assertions I see firing are that the EventDataLength is LARGER than expected.   The property is the length of all the data files that are SPECIFIC TO THE EVENT (that is it does not include fields that are in every event like, eventID, providerID, Opcode, TimeStamp, ...)     This cooresponds to the 'payload' byte[] associated with the event.   The asserts say that for a particular event with a particular version, we know that the length is exactly X.   What happens is that it is Bigger.  Now that OK (the data needed is there, but implies that creator of that event was confused about how much payload was needed and sent too much, which is a bug in CoreCLR (your files are bigger than they should be for no good reason).  

I get that it may predate Nate's change, but it is still a bug (probably in CoreCLR), but we should track it down.

For example when the AppDomainResourceManagementThreadCreated event is created it has a payload of the following fields at the following displacements

        public long ManagedThreadID { get { return GetInt64At(0); } }
        public long AppDomainID { get { return GetInt64At(8); } }
        public int Flags { get { return GetInt32At(16); } }
        public int ManagedThreadIndex { get { return GetInt32At(20); } }
        public int OSThreadID { get { return GetInt32At(24); } }
        public int ClrInstanceID { get { return GetInt16At(28); } }

Note that the last field takes 2 bytes and thus 30 bytes total are needed.   However the payload length is 32 not 30.   We should see on the CLR side when this event is created what number it is and why it is not 30.

Another example is the GCSuspendEEStart event (Search for GCSuspendEE in the code base)

It has the following fields

        public GCSuspendEEReason Reason 
        public int Count { get { if (Version >= 1) return GetInt32At(4); return 0; } }
        public int ClrInstanceID { get { if (Version >= 1) return GetInt16At(8); return 0; } }

Thus for version 1 it should have a length of 10, however EventDataLength is 32 not 10.  

Makes me think that we are passing a buffer size rather than the data size of an array in the CLR code.

Nate can you look into this?    I can help if you get stuck.

nategraf:
yes, I can look into it. Here is the issue in CoreCLR https://github.com/dotnet/coreclr/issues/13706

PR #374, opened 2017-08-31
Cleaning up the IisTraceEventParser.cs to have the right opcode names

puneetg1983:
Regenerated the IisTraceEventParser.cs after fixing some issues in TraceParserGen. Also adding the w3core.mof and w3core.mof.man file generated as a result of TraceParserGen exe.

Updated the relevant code in PerfviewData to reflect the new function names.


vancem:
Looks good.

PR #376, opened 2017-08-31
Enable Puneet's IIS parser by default

vancem:
This allows PerfVIew users to decode the IIS events without having to install IIS ETW providers on the machine doing the analysis.

Also cleaned up an unrelated help file fix (pointing people at Github for logging feedback) in UsersGuide.htm

@puneet-gupta


vancem:
I have tested that you get good parsing with this on a machine that does not have IIS ETW events installed.   We should do a comparison between this and an older version running on a machine with IIS ETW just to see if there are interesting differences.   

I have pushed this to \\clrmain\tools\PerfView.exe 

Remind me next week to make a PerfView release if I have not done it by then.

PR #378, opened 2017-08-31
missed out initializing the process Id in one place

puneetg1983:
A very slight change in the logic which might cause it to break to launch the right viewer


vancem:
LGTM

PR #380, opened 2017-09-06
Add test data for eventpipeeventinstance v2 objects

nategraf:
Adds the test data needed to verify changes in the new file format dictated by #372 and it's related CoreCLR PR.

Also fixes name resolution for SampleProfiler

Follow-up: #372


nategraf:
@vancem This is the test data you requested

nategraf:
Added FramWorkEventSource as a special case name to guid conversion because it is explicitly set as a random GUID [here](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Diagnostics/Eventing/FrameworkEventSource.cs#L31)

PR #382, opened 2017-09-10
Remove TraceEventNativeMethods.ZeroMemory

sharwell:
I tried other approaches as well:

1. Using `initblk` via `ILGenerator` (fast but high overhead)
1. Using `UnmanagedMemoryStream` to wrap a `byte*` and calling `SetLength` (incurs allocation penalty or object sharing costs)

The approach I used in this pull request is simple, but in the end it works well for each case.

### DIFF
--- a/src/TraceEvent/Ctf/CtfTraceEventSource.cs
+++ b/src/TraceEvent/Ctf/CtfTraceEventSource.cs
@@ -65,9 +65,9 @@ namespace Microsoft.Diagnostics.Tracing
                 }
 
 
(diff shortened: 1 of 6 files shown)


vancem:
Do we know why this is failing checkin tests?

sharwell:
@vancem I'm not sure. Looking like a real regression though.

sharwell:
@vancem Fixed the regression (as a separate commit so the cause is revealed in the diff)

PR #386, opened 2017-09-18
Adding ASP.NET events to the IIS Parser to get a combined view

puneetg1983:
Combining ASP.NET events with the IIS events so that we can get a combined view of all the pipeline events that happen for ASP.NET. The advantage of doing this along with the existing ASP.NET view is that the end user gets a better picture of both the combined events in one shot and can drill down faster than by looking at IIS or ASP.NET events individually.

### DIFF
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -1534,215 +1534,41 @@ table {
             };
 
             // Lets look at the rest of Enter/Leave events in AspNetReq now.
+            
+            aspNet.AddCallbackForEvents(name => name.EndsWith("Enter"), null, (TraceEvent traceEvent) => {
 
-            aspNet.AspNetReqPageInitEnter += delegate (AspNetPageInitEnterTraceData traceEvent)
-            {
-                IisRequest iisRequest;
-                if (m_Requests.TryGetValue(traceEvent.ContextId, out iisRequest))
+                // We are using AspNetReqAppDomainEnter to compute for ClrThreadPool so exclude that for now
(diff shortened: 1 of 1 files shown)


src/PerfView/PerfViewData.cs
```diff
@@ -1511,6 +1511,238 @@ protected override void WriteHtmlBody(TraceLog dataFile, TextWriter writer, stri
                 AddGenericStopEventToRequest(traceEvent.ContextId, traceEvent);
             };
 
+            var aspNet = new AspNetTraceEventParser(dispatcher);
+
+            // The logic used here is that delays between "AspNetTrace/AspNetReq/Start" and "AspNetTrace/AspNetReq/AppDomainEnter"
+            // will be due to the delay introduced due to the CLR threadpool code based on how
+            // ASP.NET code emits these events.
+            aspNet.AspNetReqStart += delegate (AspNetStartTraceData traceEvent)
+            {
```
vancem: You can handle all the starts (enter) events with 

aspNet.AddCallbackForEvents(name => name.EndsWith("Enter"), null, delegate(TraceEvent data ) {

       object contextObj = data.PayloadByName("ContextId");
       if (contextObj != null && contextObj is typeof(Guid))
              Guid contextGuid = (Guid) contextObj;
   
               // Code below 
}

You can do something similar for Exit.   that reduces the number of cases dramatically.

puneetg1983: @vancem awesome !!! Will try this out

vancem:
LGTM

Issue #387, opened 2017-09-18
Enable all TraceEvent Asserts for EventPipe


vancem:
Currently there is something called DebugListenerBlock in the EventPipeParsing.cs tests for EventPipe.  

This class is designed to suppress assertions while the test is running.  This was because previously EventPipe would emit junk at the end of an event and cause many asserts to fire.   

Supressing asserts  is  hack.   With https://github.com/dotnet/coreclr/pull/14049 it should be the case that this extra space is no longer emitted.   Thus it should be possible to remove the assert supression, at least after the traces used in the tests have been updated.

Sadly it is not trivial to emit old formats that have been fixed.  Thus we will need to keep the supression around for V1 format testing, but other than that we can remove it.

nategraf:
There is another assert as well, here https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/TraceLog.cs#L8322

I can take on looking into that if that'd be fruitful.

vancem:
That assert is surprising to me.    It is simply saying that that there should be a entry in the list of all modules for every file that we have a file path for.   It should be true based on the code just above it (since we only add things to moduleFilesByName if we have already added it to moduleFiles.   

I would suggestion commenting it out temporarily just to see what is beyond it.

vancem:
Looking at the code, it is more subtle, because there is a 'SetModuleFileName' which also updates one of the tables.   You can look if you want, but I am happy to look at it if you check in a file that I can use to run tests.

vancem:
The evil DebugListenerBlock  has been removed.   I am now happy

PR #389, opened 2017-09-18
Add updated tests data and remove DebugListenerBlock

nategraf:
This PR removes the DebugListenerBlock and adds updated test data to eliminate asserts on donetcore 2.0 EventPipe generated data so that asserts will not fire on load

Related: #387


nategraf:
@vancem The test for v1 EventPipe objects currently fails because it uses data without the write size fix. Is the best solution to cobble together a v1 object codebase with the size fix cherry-picked in and submit that as the new test data?

vancem:
> The test for v1 EventPipe objects currently fails because it uses data without the write size fix. Is the best solution to cobble together a v1 object codebase with the size fix cherry-picked in and submit that as the new test data?

No.   For better or worse, the V1 format has effectively shipped in .NET Core, and we can't change it.   The current test file accurately represents what a 2.0 .NET Core runtime will produce it is correct the way it is.

What this means is that we can't eliminate the DebugListenerBlock completely.  But we should only have it when running V1 format tests.  We probably should create a separate test for the V1 format (Don't lump it in with the others).  and when we don't care about the V1 format anymore we can eliminate it (and DebugListenerBlock), as a test case.

nategraf:
The fix to the payload sizes doesn’t effect the format of the data (or the ability of PerfView to read the files); it prevents extra empty/garbage bits from being written which TraceEvent correctly ignored, except in the DebugValidate function. The v1 test data with the fix (which was easier to generate than I originally thought) still verifies the ability of PerfView to read data from Core CLR that does not contain the fix, but it will not trigger asserts. In this sense I think we should use the new data because it still follows the v1 format, and will not degrade the value of the test, but it will allow us to fully eliminate the DebugListenerBlock (instead of creating a special case)

vancem:
LGTM

PR #397, opened 2017-10-09
Added PerfViewCollect non-gui perfView collector to build

vancem:
There was a desire to have PerfView collection on Windows Nano (which only supports .NET Core, and does not support any GUI operations).   This change adds the  PerfVIewCollect project that is a version of PerfView that has no GUI (thus is only really useful for collection), but also will compile for .NET Core.   In theory this can be used on Windows Nano (have not tried this yet).  

This change adds a few !DONET_CORE ifdefs around things that are not supported on .NET Core (basically perf counters and eventlog).   However I was also able to remove a number o !DOTNET_CORE ifdefs that are no longer needed (they were added before .NET Core 2.0 (which increased compat quite a bit).  

It also added PERFVIEW_COLLECT ifdefs around GUI things in the App.cs and CommandProcessor.cs files.  This is basically calls to 'MessageBox' for user prompts are not that important.  

Finally I had to move UnzipIfNecessary to place it where it was in a file that was not eliminated in the PerfViewCollect build.  Some renames were needed.

I also moved FetchSymbolsForProcess from being a direct command to being a user defined command.   I wanted to do this anyway, and it was useful here because normal commands are included in PerfVIewCollect (But not user commands), and FetchSymbolsForProcess has dependencies on things that  are not easily brought in to PerfViewCollect.

### DIFF
--- a/src/PerfView/App.cs
+++ b/src/PerfView/App.cs
@@ -32,7 +32,7 @@ namespace PerfView
         /// Main is also responsible for doing the 'install On First launch' logic that unpacks the EXE if needed.  
         /// </summary>  
         [System.STAThreadAttribute()]
-        [System.Diagnostics.DebuggerNonUserCodeAttribute()]
+        // [System.Diagnostics.DebuggerNonUserCodeAttribute()]
         public static int Main(string[] args)
         {
             CommandProcessor = new CommandProcessor();
@@ -157,7 +157,6 @@ namespace PerfView
                 CommandLineArgs.DataFile.IndexOf('.') < 0 && CommandLineArgs.DataFile.IndexOf('\\') < 0)
                 throw new ApplicationException("Error " + CommandLineArgs.DataFile + " not a perfView command.");
 
-#if !DOTNET_CORE
             // Check for error where you have a TraceEvent dll in the wrong place.
             var traceEventDllPath = typeof(TraceEvent).Assembly.ManifestModule.FullyQualifiedName;
             if (!traceEventDllPath.StartsWith(SupportFiles.SupportFileDir, StringComparison.OrdinalIgnoreCase))
@@ -171,7 +170,6 @@ namespace PerfView
                             "   You cannot place a version of Microsoft.Diagnostics.Tracing.TraceEvent.dll next to PerfView.exe.");
                 }
             }
-#endif
 
             // For reasons I have not dug into SetFileName does not work if you attach to a session.  Warn the user.  
             if (CommandLineArgs.InMemoryCircularBuffer && CommandLineArgs.DoCommand == CommandProcessor.Start)
@@ -813,8 +811,9 @@ namespace PerfView
             }
 #endif
         }
+#if !PERFVIEW_COLLECT
         private static WeakReference s_splashScreen;
-
+#endif
         private const string EulaVersion = "1";
         private static ConfigData s_ConfigData;
         private static string s_ConfigDataName;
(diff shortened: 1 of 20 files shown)


src/PerfView/App.cs
```diff
@@ -113,9 +112,11 @@ private static int DoMain(string[] args, ref bool newConsoleCreated, ref StreamW
 
             // Figure out where output goes and set CommandProcessor.LogFile
 
+#if !PERFVIEW_COLLECT
             // On ARM we don't have a GUI
             if (SupportFiles.ProcessArch == ProcessorArchitecture.Arm)
-                CommandLineArgs.NoGui = true;
+#endif
+            CommandLineArgs.NoGui = true;
```
sharwell: :bulb: This formatting is strange. Consider using this approach instead:

```csharp
if !PERFVIEW_COLLECT
    // On ARM we don't have a GUI
    if (SupportFiles.ProcessArch == ProcessorArchitecture.Arm)
        CommandLineArgs.NoGui = true;
#else
    // We never support GUI in PerfViewCollect
    CommandLineArgs.NoGui = true;
#endif
```

src/PerfView/App.cs
```diff
@@ -701,16 +708,18 @@ public static SymbolReader GetSymbolReader(string etlFilePath = null, SymbolRead
...
 
+#if !PERFVIEW_COLLECT
             if (!AppLog.InternalUser && !App.CommandLineArgs.TrustPdbs)
             {
                 ret.SecurityCheck = delegate (string pdbFile)
                 {
-                    var result = MessageBox.Show("Found " + pdbFile + " in a location that may not be trustworthy, do you trust this file?",
-                        "Security Check", MessageBoxButton.YesNo);
-                    return result == MessageBoxResult.Yes;
+                    var result = System.Windows.MessageBox.Show("Found " + pdbFile + " in a location that may not be trustworthy, do you trust this file?",
```
sharwell: :bulb: It would be preferable to place the `using` directives in conditional block instead of fully-qualifying everything.

sharwell:
This feels like a "dirty" approach to addressing this. Is it not possible to extract a clean command line application for collecting traces (multi-target to netcoreapp2.0 and net45), and then use that command line application within PerfView?

src/PerfView/App.cs
vancem: This kind of things is a bit of a tossup, but I don't mind if code that really is taking a bad dependency (e.g. taking a GUI dependency when I would rather App.cs be close to gui-free) be more explicit.    Basically I would like to move the code the direction of not having the ifdefs at all, but in the mean time making it slightly more painful to add more GUI dependencies is not a bad thing.

sharwell: Ironically, the reason I prefer the `using` directives is the same as the reason you prefer to not have them. 😆 I often "get a feel" for the amount a conditional dependency is used by deleting its using directive, but that doesn't work when things are fully-qualified.

vancem: And yet, you can achieve your effect of removing the using with a simple find string on the prefix.   In the end it seems like something that is not worth mandating one way or the other.

vancem:
We do need to enable the build of PerfViewCollect by default.    The faster we figure out what appveyor's nuget does not like TraceEvent the better.    

Making the GUI use the command line tool is interesting, but has to earn its keep from a cost-benefit perspective.     I certainly would love to get rid of the ifdefs.   Making a separate process for the collector that the gui spawns seems to be painful (you have to serialize things back and forth on those message boxes), and does not seem to have interesting value.  Calling the collector code cleanly in proc, that is probably better.   I am open to that, but I am not willing to pay too much for it (at least of my own time).

PR #402, opened 2017-10-12
[WIP] Enable GCStats and JITStats on Linux in PerfViewCollect

brianrob:
Enables two user commands for PerfViewCollect:

 - LinuxGCStats
 - LinuxJITStats

These user commands generate HTML reports for GC and JIT stats respectively.


brianrob:
Will re-open this against master.

PR #403, opened 2017-10-14
Enable UserCommands on Linux in PerfViewCollect

brianrob:
Enables two user commands for PerfViewCollect:

LinuxGCStats
LinuxJITStats

These user commands generate HTML reports for GC and JIT stats respectively.

### DIFF
--- a/src/PerfView/App.cs
+++ b/src/PerfView/App.cs
@@ -35,6 +35,8 @@ namespace PerfView
         // [System.Diagnostics.DebuggerNonUserCodeAttribute()]
         public static int Main(string[] args)
         {
+            Console.WriteLine("Attach debugger and hit any key.");
+            Console.ReadKey();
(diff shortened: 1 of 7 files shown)


brianrob:
I still need to fix the broken HtmlUsersGuide.  It looks like the file is not part of the assembly - I need to figure out why.

brianrob:
@vancem, this should have all of the fixes that we discussed:

 - SupportFiles are placed in ~/.config/PerfView[Collect] on Linux and in AppData on Windows by calling Environment.GetFolder(ApplicationData).
 - The HtmlUsersGuide is now embedded in PerfViewCollect.dll and so the ifdef for that has been removed.

vancem:
I would eventually like to factor things so there are not so many ifdefs for PerfViewCollect, but Iike that PerfViewCollect can have user commands (since they are useful for the command line version).   

LGTM

PR #411, opened 2017-10-19
Improve support for deeply-nested stacks

### LINKED ISSUE #161: Stack overflow viewing thread times in a specific trace
PerfView crashes with a stack overflow while opening a specific trace stored internally on the fileshare where perfview comes from at `\\<server>\public\writable\users\kevinpi\ReproTrace.etl.zip`

### LINKED ISSUE #348: PerfView crash on opening the Thread time stacks view
From Mohit Chakraborty mail on 8/10/17 

File is 

\\\\clrmain\public\writable\users\vancem\PerfViewIssues\PerfViewDataDev15.3.Picasso_2.etl.zip

Open Thread Time (with start stop tasks) view.   It will crash.

Seems to be a stack overflow.

sharwell:
Fixes #348 

Likely fixes #161, but I did not test this

### DIFF
--- a/src/TraceEvent/Stacks/CallTree.cs
+++ b/src/TraceEvent/Stacks/CallTree.cs
@@ -7,6 +7,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;                        // For TextWriter.  
+using System.Threading.Tasks;
 
 namespace Microsoft.Diagnostics.Tracing.Stacks
 {
@@ -148,7 +149,7 @@ namespace Microsoft.Diagnostics.Tracing.Stacks
         /// </summary>
         public void Sort(Comparison<CallTreeNode> comparer)
         {
-            m_root.SortAll(comparer);
+            m_root.SortAll(comparer, 0);
         }
         /// <summary>
         /// Sorting by InclusiveMetric Decending is so common, provide a shortcut.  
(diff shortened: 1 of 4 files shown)


src/TraceEvent/Stacks/CallTree.cs
```diff
@@ -271,8 +272,17 @@ struct TreeCacheEntry
             public CallTreeNode Tree;
         }
 
-        private CallTreeNode FindTreeNode(StackSourceCallStackIndex stack)
+        private CallTreeNode FindTreeNode(StackSourceCallStackIndex stack, int depth = 0)
```
sharwell: @vancem I'm not particularly happy with the parameter name `depth`. Let me know if there's some other name you would prefer.

💭 For me the redeeming factor may be twofold:

1. The methods are private
2. The recursion guard immediately inside the method is super obvious

vancem: I think 'callStackDepth' is better than 'depth'.

src/TraceEvent/Stacks/CallTree.cs
```diff
@@ -271,8 +272,17 @@ struct TreeCacheEntry
...
         {
+            if (depth > 400)
```
vancem: Can you put a comment about why this is here and what the magic number 400 is (that you are restarting the recursive function on a new thread so you don't get stack overflow.  Give 400 a symbolicName like 'MaxCallStacksBeforeNeedingANewThread' (or something like that).

sharwell:
@vancem We've used this successfully in a heap that had a `LinkedList<T>` with ~21 million nodes.

PR #414, opened 2017-10-23
make Set Time Range work for single cell selection

adamsitnik:
Currently, the user needs to select "First" and "Last" columns cells to make "Set Time Range" menu item work.

This is rather non-intuitive, you most probably use keyboard shortcuts so you don't hit this issue.

The PR is trivial. Single cell selected? If it's CallTreeNode, use the right values to set the time filter.

image

### DIFF
--- a/src/PerfView/StackViewer/StackWindow.xaml.cs
+++ b/src/PerfView/StackViewer/StackWindow.xaml.cs
@@ -1586,22 +1586,13 @@ namespace PerfView
             var cells = SelectedCells();
             if (cells != null)
             {
-                if (cells.Count == 2) // First and Last column cells
-                {
-                    StartTextBox.Text = GetCellStringValue(cells[0]);
-                    EndTextBox.Text = GetCellStringValue(cells[1]);
-                    Update();
-                }
-                else if (cells.Count == 1) // single cell from a row
-                {
-                    var callTreeNode = cells[0].Item as CallTreeNodeBase;
+                var callTreeNodes = cells.Select(cell => cell.Item).OfType<CallTreeNodeBase>();
 
-                    if (callTreeNode != null)
-                    {
-                        StartTextBox.Text = callTreeNode.FirstTimeRelativeMSec.ToString("n3");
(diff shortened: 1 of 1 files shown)


src/PerfView/StackViewer/StackWindow.xaml.cs
```diff
@@ -1586,12 +1586,23 @@ private void DoSetTimeRange(object sender, ExecutedRoutedEventArgs e)
...
             if (cells != null)
             {
-                if (cells.Count == 2)
+                if (cells.Count == 2) // First and Last column cells
                 {
                     StartTextBox.Text = GetCellStringValue(cells[0]);
                     EndTextBox.Text = GetCellStringValue(cells[1]);
                     Update();
                 }
+                else if (cells.Count == 1) // single cell from a row
```
vancem: I would prefer that this look just like the 2 element case as much as possible.  

Indeed by preferred code is  to fold it into the case above 

```
if (cells.Count == 2 || cells.Count == 1) {
    StartTextBox.Text = GetCellStringValue(cells[0]);
    if (cells.Count == 1)
        EndTextBox.Text = GetCellStringValue(cells[0]);
    else 
        EndTextBox.Text =  StartTextBox.Text ;
    Update();
}
```

adamsitnik: in case cells.Count == 1 we have single cell selected. We don't know if it's a Start/Last column. It can be something else (like name), in that case your proposal would fail.

adamsitnik:
@vancem I just submitted more general approach.

I set Start as Min(FirstTimeRelativeMSec) and End as Max(LastTimeRelativeMSec). In this case the order and number of cells does not matter.

From Perf perspective the code is not perfect, but the method is not on a hot path so I think that using LINQ should not be a problem.

vancem:
I actually prefer the simpler approach since it is easy to explain what it does (it copies the two cells to the start and end boxes) but I can live with this approach as well.  

I will merge it.

PR #415, opened 2017-10-23
Support dynamic merge for DiagSession ETL files

AaronRobinsonMSFT:
When opening a DiagSession, there could be multiple ETL files. If there are multiple files, attempt to merge them into a single file for analysis using the NONE option. This approach is valid since the auxillary events normally injected during a merge by KTC are instead insert into a aux ETL file in DiagSession scenarios (i.e. Visual Studio).

### DIFF
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -7670,7 +7670,7 @@ table {
                 });
 
                 // Get all ETL files
-                AddResourcesAsChildren(worker, dhPackage, "DiagnosticsHub.Resource.EtlFile", ".etl", (localFilePath) =>
+                AddEtlResourcesAsChildren(worker, dhPackage, (localFilePath) =>
                 {
                     return ETLPerfViewData.Get(localFilePath);
                 });
@@ -7711,13 +7711,8 @@ table {
             ResourceInfo[] resources;
             dhPackage.GetResourceInformationByType(resourceIdentity, out resources);
 
-            IEnumerable<ResourceInfo> orderedResources = resources
-                    .OrderBy(r => DhPackagingExtensions.ToDateTime(r.TimeAddedUTC));
-
             foreach (var resource in resources)
             {
-                Guid resourceId = resource.ResourceId;
-
                 string localFilePath = GetLocalFilePath(FilePath, dhPackage, resource, fileExtension);
 
                 worker.Log("Found '" + resource.ResourceId + "' resource '" + resource.Name + "'. Loading ...");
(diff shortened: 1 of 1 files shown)


vancem:
Hello Aaron,

There are two ways of dealing with multiple ETL files.   The first is to merge them into a new file (which is what you have done).   The other is to simply open all the files simultaneously (basically it merges on the fly).    TraceEvent/PerfView does this today when want to look at 'raw' files.   Indeed when you open 'X.etl'  by default it probes for  X.kernel*.etl  X.clr*.etl X.user*.etl.    

My general bias is NOT try to create persistent state (files) if I can avoid it, because it introduces failure modes that are easy to get wrong (locked files, corrupted files, cleanup ...).  I also expect it to be simpler in this case as well.   

What do you think about the idea of avoiding creating a merged file?   

What are the actual file names generated in the DiagSession?

AaronRobinsonMSFT:
I see. I didn't realize PerfView can do this on the fly. The sole reason for this was because we are going to be changing the contents to have multiple ETL files instead of a single merged file and we don't want to break some of our users workflow. Several of our customers open up DiagSession files in PerfView when they want more details. With the change to how we collect, a DiagSession can potentially contain up to 6 ETL files:

`System.*.etl`
`Heap.*.etl`
`UserPagedMemory.*.etl`
`UserNonPagedMemory.*.etl`
`ManualInjection.*.etl`
`ktc_aux.etl`

The `ktc_aux.etl` file will contain all the events that KTC would have injected into the merged ETL file. If PerfView can process the files as a group without an explicit merge, then I am absolutely fine not making this change. How can I test this scenario?

arr

vancem:
Hello Aaron

As mentioned TraceEvent/PerfView already has support where it looks for X.user*.etl X.kernel*.etl when it is given a file X.etl.     It basically opens them all in one OpenTrace command.   

Ideally we just rename the files so that they look like *.User_System.etl  *.User_Heap.etl *.User_PagedMemory.etl and *.User_NonPagedMemory.etl and *.Kernel.etl.

Note that technically you would not need to do this in the archive itself, we could fix it so that PerfView unpacks them to these names.   Then when you open *.etl file you will get all these merged in by default.     I think it is a bit nicer to fix the names in the archive, because I do like the convention of putting 'slices' of the ETL data as a suffix (rather that a prefix).   Note we require .kernel* or .user* to be a prefix because it is pretty common for users to want to have their own suffixes in their ETL file naming (e.g. what if you had X.Run1.etl and X.System.etl.   Are these part of the same logical ETL file or are they separate data?  By requiring 'slices' to have 'User' or 'Kernel' prefixes, we can avoid this ambiguity (I don't have a problem with having one or two more 'allowed' prefixes for slices, but the idea is that the likelihood of collision with 'normal' user suffixes should be low.

What do you think of this?   Is it easy to change the names in the archive, or should we have PerfView do the renaming as it unpacks them?    

Note that either way small changes to PerfView will be needed (if only to hide the *.user*.etl and *.kernel*.etl  files in the GUI).   

If you have an example of a new diagsession file (after the renaming if you are OK doing that). I would be interested in getting access to it.

AaronRobinsonMSFT:
I am absolutely fine with whatever naming convention is desired for PerfView. User are not permitted to name the ETL files for the scenario so there is little concern for collision on our end, but there is nothing stopping people from creating a new DiagSession and inserting whatever kind of ETL files they desire - so having a convention probably prudent in this case.

Your proposed suggestion of `*.user*.etl` and *.kernel*.etl will work for us and I can update the DiagSession names for this. We don't rely on the names at all. My plan would be to having the following possible names:

`*.Kernel_<iter>.etl`
`*.User_Heap_<iter>.etl`
`*.User_PagedMemory_<iter>.etl`
`*.User_NonPagedMemory_<iter>.etl`
`*.User_ManualInjection_<iter>.etl`
`*.User_aux.etl`

I think I know where this work needs to go and what the next steps are. I will update this commit and send you a link to the location of the sample DiagSession with the new format.

arr

vancem:
Great.  Sounds like a good plan.   Thanks Aaron.

vancem:
LGTM

PR #416, opened 2017-10-27
Update the DiagnosticHub DLLs

vancem:
Updated to the version in VS 15.4.   These DLLs are responsible for opening diagsession files.
They made a format change, so getting these udpated allows it to read the new format.

Note that I also took this as an opportunity to change the CSPROJ files so that the version
numbers of the suport Nuget packages are now centralized into the dir.props at the top level
of the repo.   This will make doing an update like this in the future simpler.

@AaronRobinsonMSFT


AaronRobinsonMSFT:
LGTM

PR #418, opened 2017-10-29
Create new thread objects when necessary to avoid assigning events to the wrong thread

sharwell:
@heejaechang This bug is the reason why we weren't seeing correct allocation data this past week.

### DIFF
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -5275,6 +5275,10 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
         internal TraceThread GetOrCreateThread(int threadID, long timeQPC, TraceProcess process, bool isThreadCreateEvent = false)
         {
             TraceThread retThread = GetThread(threadID, timeQPC);
+
+            if (process != null && retThread != null && retThread.process.ProcessID != -1 && retThread.process.ProcessID != process.ProcessID)
+                retThread = null;
+
             if (retThread == null || isThreadCreateEvent)
             {
                 InitThread();
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -5276,14 +5276,14 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
         {
             TraceThread retThread = GetThread(threadID, timeQPC);
 
-            // ThreadIDs are machine wide, however, they are also reused.   Thus GetThread CAN give you an OLD thread IF 
-            // we are missing thread Death and creation events (thus silently it gets reused on another process).   This 
-            // can happen easily becasue Kernel events (which log thread creation and deaths), have a circular buffer that
-            // might get exhausted before user mode events (e.g. CLR events).   Thus try to keep as much sanity as possible
-            // by confirming that if the thread we get back had a process, that that process is the same as the process for 
-            // the currnet event.  If not, we assume that there were silent thread deaths and creations and simply create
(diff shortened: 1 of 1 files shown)


sharwell:
@vancem I'm not sure why the assertion is failing in the debug build. It's this assertion:

https://github.com/Microsoft/perfview/blob/3c4d2b4526fe7ff9393d9a2dc2510120f05590a6/src/TraceEvent/TraceLog.cs#L2286

vancem:
Sam can you point me at an example where this problem occurs.  The fix suggest that thread IDs are not machine wide but only process wide, and that surprises me.   I would like to understand it better. 

As to the assert: A stack can have up to two parts (a user and kernel part).   This is basically saying we should not have a stack with two user parts (the algorithm will clobber the first one).    It probably means that we are not correct about assigning parts to the event.    Details on the ETL file to repro it is certainly interesting information.

sharwell:
@vancem A reproducer is attached to internal bug DevDiv 513576. Observe that the GC Heap Alloc Ignore Free view does not show the same information (at all) as the GC Stats view.

pharring:
Thread IDs might be machine-wide but I assume, like PIDs, they can be recycled. i.e. if a thread with ID x exits, then 'x' is available for use and may be assumed by a new thread being created later (either in the same process or in a completely different process).

vancem:
@sharwell  I have the trace, can you point me at a specific thread that has the issue?   It would save me the trouble of teasing it out.

@pharring  The current logic looks up thread by (ID- Timestamp) pairs specifically to guard against ID reuse (you ask for the ID at a particular time, and there should be only one).

src/TraceEvent/TraceLog.cs
```diff
@@ -5275,6 +5275,10 @@ private void InitThread()
         internal TraceThread GetOrCreateThread(int threadID, long timeQPC, TraceProcess process, bool isThreadCreateEvent = false)
         {
             TraceThread retThread = GetThread(threadID, timeQPC);
+
+            if (process != null && retThread != null && retThread.process.ProcessID != -1 && retThread.process.ProcessID != process.ProcessID)
+                retThread = null;
+
```
vancem: OK I investigated the scenario .  Please use the following code for the fix (notice it has another condition as well as the explanation of why it is needed.  
```
            // ThreadIDs are machine wide, however, they are also reused.   Thus GetThread CAN give you an OLD thread IF 
            // we are missing thread Death and creation events (thus silently it gets reused on another process).   This 
            // can happen easily becasue Kernel events (which log thread creation and deaths), have a circular buffer that
            // might get exhausted before user mode events (e.g. CLR events).   Thus try to keep as much sanity as possible
            // by confirming that if the thread we get back had a process, that that process is the same as the process for 
            // the currnet event.  If not, we assume that there were silent thread deaths and creations and simply create
            // a new TraceThread.   Note that this problem mostly goes away when we have just a single circular buffer since
            // we won't lose the Thread death and creation events.   
            if (process != null && process.ProcessID != -1 && retThread != null && retThread.process.ProcessID != -1  && process.ProcessID != retThread.process.ProcessID)
                retThread = null;
```
Thanks

Vance

sharwell:
@vancem I'm testing out one more edge case observed locally with the old approach to make sure it no longer is a problem.

vancem:
LGTM assuming the assert no longer fires (give me info about how to repro and I will look into it if needed).  

Merge when you are happy with it.

sharwell:
@vancem The anomaly I observed was also corrected by your suggested change. Things look great from automated and manual testing sides now. 👍

PR #419, opened 2017-10-31
Remove unnecessary non-standard extension point dir.props

sharwell:


sharwell:
@vancem This seems like an easy/obvious simplification to me, but I wanted to make sure you're aware of the relocation of the `PerfViewVersion` definition.

PR #423, opened 2017-11-02
Default heap dump processing to all objects instead of a sampling

sharwell:
Based on my experiences with processing heap dumps to improve reliability in Visual Studio over the past several months, I found that processing all objects in a heap provided such an overwhelming advantage for actionable results that it was always worth the cost (i.e. the larger the dump, the more valuable the unsampled result was). This change preserves the ability to sample the heap, but makes two notable changes to help users make the right decision on their first pass through the UI:

1. Removes the warning about heap dump extraction being too large
2. Defaults the UI and command line to not sample unless explicitly specified

### DIFF
--- a/src/HeapDump/GCHeapDumper.cs
+++ b/src/HeapDump/GCHeapDumper.cs
@@ -48,8 +48,6 @@ public class GCHeapDumper
         m_origLog = log;
         m_copyOfLog = new StringWriter();
         m_log = new TeeTextWriter(m_copyOfLog, m_origLog);
-
-        MaxDumpCountK = 250;
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


vancem:
This is obviously something that we need to weigh the costs and benefits carefully.     

Can you describe why the extra details helps (or how the sampling is misleading)?   (that is: what is the benefit).   

We also need information on the cost (presumably a 50GB heap will now be a file that is many GB in size).      

Keep in mind that the VS experience is not likely to be the most common.  The sweet spot is likely to be a naive user that has a 'easy' problem and does not need detailed knowledge (Because indeed his is not willing to invest much in learning the tool), but wants it fast (and would be very dissatisfied if the GUI was slow on his very large heap).     In general it is easier to teach a few motivated / sophisticated users than it is to teach many unmotivated transient users.   

I am putting in guidance to help the more sophisticated users to things that will be helpful for their scenario, but the defaults probably should be tuned for the casual user.

sharwell:

brianrob:
Closing as this PR is almost 2 years old.  Per @vancem's feedback, we likely do not want to process all objects by default.

PR #424, opened 2017-11-02
Increase the default buffer size to 256MiB

sharwell:
Reduces the chance of missed events.

I was missing some events (often thousands) measuring Visual Studio even though I have NVMe drives and 28 logical processors. With this change, the events are no longer dropped.

### DIFF
--- a/src/PerfView/CommandLineArgs.cs
+++ b/src/PerfView/CommandLineArgs.cs
@@ -104,7 +104,7 @@ namespace PerfView
 
         // Start options.
         public bool StackCompression;           // Used compresses stacks when collecting traces. 
-        public int BufferSizeMB = 64;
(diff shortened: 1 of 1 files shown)


felipepessoto:
Have we changed the message to suggest more than 256MB buffer size?

image

I can't find it in source code

vancem:
We did not update that.  (that text lives in PerfViewData.cs.   I have updated my local copy and will push it along with other changes when next push changes.   Thanks for pointing this out.

felipepessoto:
Do you already have the next version ETA ?

vancem:
PerfView is released as needed.   Basically when enough new things/fixes accumulate then we release again.   I have promised others that we would do this by the first week in Feb.    If something important comes up before that, it will be sooner.

PR #429, opened 2017-11-03
Improve the performance of CallTreeNode.AllCallees

sharwell:


splusq:
When do you plan on publishing the netstandard2.0 version of the TraceEvent library?

vancem:
An unsigned copy of it is buildable right now from the repo (it supports NetStandard 2.0).   Getting the TraceEvent package signed and up on nugget is trickier and I am targeting the end of the calendar year.

splusq:
can you publish the unsigned version with an alpha tag? I just need to consume from an official source - since this is at the leaf of our dependency, I am unable to move any of our projects to .netstandard because of it.

vancem:
I will check, but I was assuming that even pre-release packages need to be signed.   Microsoft would get blamed if any malicious modifications were made.

PR #433, opened 2017-11-03
Move GrowableArray tests to a proper unit test

sharwell:
Blocked on #290. 

📝 Please make sure to not rebase or squash this change during the merge process.


vancem:
Most of this is the change to eliminate float (making it double) in most if not all uses in PerfView. 

I know that Sam has noticed that various statistics do not 'add up' properly in some scenarios which is what prompted the change.    Generally speaking I don't have a problem with using double instead of float in any place where size is not an issue (which is many places).

The place that gives me pause, are the values in CallTreeNodeBase and HIstogram as TreeNodes are very common and this change is likely to make these nodes increase  in size by a substantial factor (I am guessing 60-80%).  We KNOW that we allocate GB of this kind of stuff (since we get out of Memory exceptions on 32 bit), so size really SHOULD matter, and increasing it by substantial amount is a worrisome prospect.

We could try to measure this impact, but it is not clear to me when we decide the perf loss is too much and we live with the inaccuracy.   Clearly we want both.

So I have an alternate suggestion.    Change this checkin so that you use double for all the CONTRACTs and frankly anything that is not perf-critical (which I am saying is anything generated in call trees including CallTreeNodeBase, and Histogram).   Leave those as floats for now (you can put in an #ifdef to flip it, but leave it as float for now).   

That can be checked in now, and gets rid of most of the possible merge conflicts (which are undoubtedly with the signatures not with the fields).   

This also lets us do side-by-side comparisons to determine if size really does matter.  

Assuming that we do want to optimize size, what I would LIKE is to see what we can do to get the precision AND the smaller size .   Maybe just leaving the Histogram as float, or encoding some of the fields as deltas from other fields can work.    I could imagine all sorts of things that probably would work (e.g. having two versions of CallTree that have more precision).   I don't really know what the right answer is.

However we don't need the right answer to move forward.  I am OK making interfaces use double, and having an #ifdef to do the flip.   We can then explore to either find a way to get both small size and good precision, or gather enough data to make use comfortable making the policy that we swallow the larger size for the sake of accuracy.

This first step, however should be a reasonably small change from this PR.

src/TraceEvent/Stacks/Histogram.cs
```diff
@@ -181,11 +181,11 @@ private IEnumerable<float> GetEnumerable()
         /// </summary>
         private readonly HistogramController m_controller;
 
-        private float[] m_buckets;               // If null means is its single value or no values
```
vancem: As mentioned in the general comments, For histograms as well as CallTreeNodeBase  (anything that is a substantial part of a CallTreeNode), I would like this to be under #ifdef control (currently defaulted to float), so we can experiment / measure and see if we can't get BOTH size and accuracy.

sharwell:
@vancem As mentioned in the summary, I wasn't expecting this to be reviewed prior to dealing with #290 (which is the change from float->double).

sharwell:

brianrob:
@sharwell is it possible to split this into two PRs?  I see two distinct pieces of this PR:

1. FastSerializationTests
2. Changes from float to double for the stack viewer statistics.

It seems that 1 is fairly straightforward and non-controversial, but 2 is not.  I'd love to merge 1 and then consider 2 separately.

sharwell:
@brianrob This pull request is blocked on #290

brianrob:

brianrob:

PR #436, opened 2017-11-10
Fix Linux NGEN/R2R symbols

brianrob:
As of .NET Core 2.0, crossgen no longer includes the .ni.dll suffix for native images.  PerfView still expects the .ni.dll suffix, and so it is not able to resolve any NGEN or R2R image symbols.

This PR fixes the issue in a backwards compatible way, allowing resolution of the image with or without the ni suffix.

### DIFF
--- a/src/PerfView/OtherSources/Linux/LinuxPerfScriptEventParser.cs
+++ b/src/PerfView/OtherSources/Linux/LinuxPerfScriptEventParser.cs
@@ -250,8 +250,9 @@ namespace Diagnostics.Tracing.StackSources
             return path.Substring(index + 1);
         }
         static char[] pathSeparators = new char[] { '/', '\\' };
+        const string NISuffix = ".ni.";
 
-        internal static string GetFileNameWithoutExtension(string path)
+        internal static string GetFileNameWithoutExtension(string path, bool stripNiSuffix = false)
         {
             var start = path.LastIndexOfAny(pathSeparators);
             if (start < 0)
@@ -262,7 +263,22 @@ namespace Diagnostics.Tracing.StackSources
             var end = path.LastIndexOf('.');
             if (end < start)
                 end = path.Length;
-            return path.Substring(start, end - start);
+            var name = path.Substring(start, end - start);
+
+            if (stripNiSuffix)
+            {
+                var suffixStart = name.IndexOf(NISuffix);
+                var suffixEnd = suffixStart + NISuffix.Length - 1; // Leave the trailing '.' from ".ni."
+                start = 0;
+                end = name.Length;
+
+                var first = name.Substring(0, suffixStart);
+                var second = name.Substring(suffixEnd, name.Length - suffixEnd);
+
+                name = first + second;
+            }
+
+            return name;
         }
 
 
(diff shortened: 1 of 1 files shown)


vancem:
Note that there is a break in the tests that looks like it is caused by this change.

src/PerfView/OtherSources/Linux/LinuxPerfScriptEventParser.cs
```diff
@@ -262,7 +263,22 @@ internal static string GetFileNameWithoutExtension(string path)
...
                 end = path.Length;
-            return path.Substring(start, end - start);
+            var name = path.Substring(start, end - start);
+
+            if (stripNiSuffix)
+            {
+                var suffixStart = name.IndexOf(NISuffix);
+                var suffixEnd = suffixStart + NISuffix.Length - 1; // Leave the trailing '.' from ".ni."
+                start = 0;
+                end = name.Length;
```
vancem: The assignments of start and end are dead at this point.  You should remove them.

src/PerfView/OtherSources/Linux/LinuxPerfScriptEventParser.cs
```diff
@@ -262,7 +263,22 @@ internal static string GetFileNameWithoutExtension(string path)
             var end = path.LastIndexOf('.');
             if (end < start)
                 end = path.Length;
-            return path.Substring(start, end - start);
+            var name = path.Substring(start, end - start);
+
+            if (stripNiSuffix)
+            {
+                var suffixStart = name.IndexOf(NISuffix);
```
vancem: You should not assume that this IndexOf will succeed.  Whenever you deal with paths you should use the 'ignore case' option.   The NISuffix variable I don't think should end in a '.'. because name as this point does not end in a '.'.

brianrob:
Thanks for the review @vancem.  These should be fixed now.

src/PerfView/OtherSources/Linux/LinuxPerfScriptEventParser.cs
```diff
@@ -262,7 +263,21 @@ internal static string GetFileNameWithoutExtension(string path)
...
```
vancem: I am surprised this works as you intend because NISuffix =".ni." (that is it has a trailing dot).  

I would have expected that typically you would have a path = Foo.ni.dll and name would be Foo.ni (that is no trailing dot), and thus name.IndexOf(NISuffix) would FAIL, and the new code will never do anything.  

Are we sure that actually works (it puts both Foo AND Foo.ni in the table)?

brianrob: The input path is for the map file and not for the NI.  An example path looks like "traces/perfdemo.trace/System.Console.ni.{d2de1c76-d19d-4c52-8ff5-70e88349ea91}.map"  Originally, this function reduced this to "System.Console.ni.{d2de1c76-d19d-4c52-8ff5-70e88349ea91}".  If you specify true for stripNiSuffix then you get "System.Console.{d2de1c76-d19d-4c52-8ff5-70e88349ea91}".  For .NET Core 2.0, the perfinfo-$pid.map file no longer contains the ".ni" and that's where things were breaking.

I have confirmed that it does work, and that we do put both Foo and Foo.ni in the table.  You can see where we put both into the table here: https://github.com/Microsoft/perfview/pull/436/files#diff-151b5622877b35646b2ea5e3c9466af6R701

PR #438, opened 2017-11-14
Toggle column visiblity in PerfDataGrid

WardenGnaw:
Ability to select columns in a view. This is so that you can have more space for columns you are actively interested in. You can save the column view preferences. 

image


vancem:
I am wondering if this is the best user experience.     I would have expected it to be the case that I would want to choose columns for EVERY view at once (not to have to change them one at a time for each view.    The list would be the union of the column names over all views.  

Do you really think there is value in having different settings for each view?

WardenGnaw:
That seems like a better user experience to have consistent column view settings across all the PerfGridViews. I'll change it to do that.

WardenGnaw:
@vancem When you have the time, could you look at my changes? 

Thanks!

vancem:
Looks Good To Me.     Thanks for your contribution.

PR #440, opened 2017-11-16
Flame Graph

adamsitnik:
PerfView is very powerful tool, but sometimes for some inexperienced users like me it is hard to get a fast overview of what takes most of the CPU if there are plenty of methods on the stack.

This PR provides a simple implementation of the "Flame Graph".

It is just different visualization of the "call tree" view.

Sample result (for the "tutorial.exe")

image

Sample result for more complicated app (dotnet cli):

image

1. I added new tab (before the "notes")
2. I did not use SVG (most common way of implementing flame graph) because I would have to add external dependency to the project ([link](https://social.msdn.microsoft.com/Forums/vstudio/en-US/22978b1c-be4d-475d-8a57-371f324b4697/does-wpf-support-svg?forum=wpf))
3. I decided to draw the rectangles/textblocks on a simple WPF Canvas. For small boxes, I use rectangles (too many textblocks on a single canvas = bad perf)
4. I have used some fancy color pallet to differentiate the boxes 
5. I added possibility to export the flame graph to png file (right click => save)
6. I added the docs
7. The canvas is redrawn on resize/when the tab gets the focus
8. Both tool tip and bottom panel contains method name, samples count and % of total time
9. Since the flame graph is using data from call tree view, any filter applied to call tree view works for flame graph as well


goldshtn:
I love this! I know it will help a lot of users figure out at a glance what’s happening in their call tree. 

Would the folding and grouping from the tree view be applied to the flame graph as well? 

Finally in terms of the palette, it might make sense to use a default palette for .NET processes where the .NET frames, CLR frames, Windows library frames, and kernel frames are colored using different palettes. (Like the flamegraph.pl Java palette, for example.)

+100000

adamsitnik:
> Would the folding and grouping from the tree view be applied to the flame graph as well?

@goldshtn yes, I am reusing the data generated for the call tree, so I get all of it's advantages out of the box

tutorial.exe with default grouping:

image

tutorial.exe with no grouping

image

adamsitnik:
> Finally in terms of the palette, it might make sense to use a default palette for .NET processes where the .NET frames, CLR frames, Windows library frames, and kernel frames are colored using different palettes. (Like the flamegraph.pl Java palette, for example.)

@goldshtn I really like this idea, but I have not found a way to implement it yet. My problem is that based on the name of the method I currently can not say to which group it belongs (it is just a string). I need to dig into the code of PerfView and do some PoC. Maybe I could use the existing grouping regexes to do it? It looks like perfect candidate

ilabutin:
@adamsitnik Wait, if you're reusing same call tree, then all grouping is already done for you. So maybe all CLR calls are already grouped into "CLR" nodes. So, I would suggest to start with highlighting different folded groups with different palettes - it will already give nice visual understanding of how these groups are distributed.
And then, in addition, you can add more advanced highlighting based on module/method names which were not grouped.

adamsitnik:
due to some problems with GIT I need to recreate my fork. Flame graphs will be back ;)

adamsitnik:
Recreated #502

PR #441, opened 2017-11-16
Set filename from "save as" dialog

### LINKED ISSUE #212: TextEditorWindow "File->Save As" doesn't work
When choosing "Goto Source (Def)" and viewing the results in the default (?) TextEditorWindow, there is a "File" menu with a "Save As" command. That doesn't appear to do anything.

mikkelbu:
Fixes #212


vancem:
LGTM

PR #444, opened 2017-11-20
Make invalid data file error go to status bar, not pop exception dialog

danmosemsft:


src/PerfView/Dialogs/RunCommandDialog.xaml.cs
```diff
@@ -544,8 +544,16 @@ private void OKButtonClick(object sender, RoutedEventArgs e)
...
-                // This also throws an exception if there are illegal chars.  
-                var fullPath = System.IO.Path.GetFullPath(App.CommandLineArgs.DataFile);
+                string fullPath;
+                try
+                {
+                    fullPath = System.IO.Path.GetFullPath(App.CommandLineArgs.DataFile);
+                }
+                catch (Exception ex)
+                {
+                    m_mainWindow.StatusBar.LogError("Invalid datafile '" + App.CommandLineArgs.DataFile + "'. " + ex.Message);
```
sharwell: ❓ Can we still print the full exception text to the log?

danmosemsft: It's that the path is invalid (blank, invalid characters). It's not a bug. Is it useful to write out the .NET stack for that?

sharwell: If that's the only case where it's hit, then I'm fine with it either way.

danmosemsft:
@sharwell can I hit merge, or does @vancem need to sign off?

sharwell:
@danmosemsft I typically let Vance sign off on anything except test/infrastructure changes.

vancem:
LGTM

PR #445, opened 2017-11-20
Make error text bold briefly in the status bar

danmosemsft:
A few times I did not notice that an error message had been written to the status bar. This change causes error message text (but not other messages) to briefly bold twice so the error is not overlooked then return to normal weight. I did some experimentation and this did the job without being irritating.


vancem:
LGTM

Issue #448, opened 2017-11-26
Assertion in `SessionEndTime` fails on computer with Danish culture (UTC+1)


mikkelbu:
The assertion `Debug.Assert(SessionStartTime <= ret)` fails on a computer with Danish culture and UTC+1. The problem is that `SessionStartTime` has already been converted to local time, but `ret` has not, hence `SessionStartTime` will be approx 1 hour larger in the comparison.

vancem:
Fixed by PR #451

PR #451, opened 2017-11-27
Three Minor fixes

vancem:
Fix assert, fix spelling, add more diagnostics.

### DIFF
--- a/src/TraceEvent/Symbols/SymbolReader.cs
+++ b/src/TraceEvent/Symbols/SymbolReader.cs
@@ -2465,7 +2465,7 @@ namespace Microsoft.Diagnostics.Symbols
                     }
                     else
                     {
-                        log.WriteLine("Source Server command is not recognized as safe (sd.exe or tf.exe), failing.");
+                        log.WriteLine("Source Server command {0} is not recognized as safe (sd.exe or tf.exe), failing.", fetchCmdStr);
                         return null;
                     }
                     Directory.CreateDirectory(Path.GetDirectoryName(target));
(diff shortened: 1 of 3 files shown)


vancem:
Fixes issue #448

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -2465,7 +2465,7 @@ string GetSourceFromSrcServer()
                     }
                     else
                     {
-                        log.WriteLine("Source Server command is not recognized as safe (sd.exe or tf.exe), failing.");
+                        log.WriteLine("Source Server command {0} is not recognized as safe (sd.exe or tf.exe), failing.", fetchCmdStr);
```
sharwell: :bulb: Consider using new syntax for readability:

```
log.WriteLine($"Source Server command {fetchCmdStr} is not recognized as safe (sd.exe or tf.exe), failing.");
```

src/TraceEvent/TraceEvent.cs
```diff
@@ -242,9 +242,9 @@ public DateTime SessionEndTime
         {
             get
             {
-                var ret = QPCTimeToDateTimeUTC(sessionEndTimeQPC);
+                var ret = QPCTimeToDateTimeUTC(sessionEndTimeQPC).ToLocalTime();
```
sharwell: :memo: I generally prefer to use `DateTimeOffset` instead of `DateTime` to avoid these kinds of problems, but that would be an API change instead of just a bug fix.

Issue #452, opened 2017-11-28
TraceLog is not properly disposed in call to CreateFromEventTraceLogFile


nategraf:
The TraceLog object is not disposed at the end of ETLX generation, as is noted in the comment linked here: https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/TraceLog.cs#L823

This seems to be a known issue that slipped through the cracks

I ran into this issue when processing a large number of ETL files and becoming unable to delete processed ETL files because the file handle had leaked

PR #453, opened 2017-11-28
Add using statement to mange TraceLog disposal

nategraf:
Closes: #452


nategraf:
@vancem this change **seems** pretty straight forward. Are there any concerns with using the `using` statement to dispose the tracelog?

PR #455, opened 2017-11-29
Target netstandard1.6

nategraf:
As it is needed to run some of new tests I am writing for CoreCLR, I have managed to retarget TraceEvent to netstandard 1.6

I've opened this PR to discuss the idea of making this mainstream. The alternative is that we will need to publish a non-standard version on MyGet which CoreCLR tests builds will target

@vancem what do you think?


nategraf:
It won't build in the CI because I reference a modified version of OSExtensions. That version can be build from the encrypted source included

src/TraceEvent/Computers/ActivityComputer.cs
```diff
@@ -1145,7 +1145,7 @@ private void ResolveWellKnownSymbols()
                 }
             }
             if (!foundThreadingAPIs)
-                throw new ApplicationException("Could not resolve symbols for Task library (mscorlib), task stacks will not work.");
+                throw new Exception("Could not resolve symbols for Task library (mscorlib), task stacks will not work.");
```
vancem: In PerfView I use ApplicationException for things that you want a user to see.   Things that are not ApplicationExceptions are assumed to be programmer errors and are not surfaced to the end user, but simply say that PerfView had an internal error.    By changing this you lose this distinction.  

I am assuming that ApplicationException is not in V1.6?   

I am OK having an #ifdef that makes it throw Exception if you are on V1.6 but ApplicationException on desktop.

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -1996,7 +1996,7 @@ internal void ParseProviderEvents(Func<DynamicTraceEventData, EventFilterRespons
 
             THROW:
             if (!noThrowOnError && error != null)
-                throw new ApplicationException("Error parsing the manifest for the provider " + (this.name ?? "UNKNOWN"), error);
```
vancem: Same issue with ApplicationException

src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -592,7 +592,7 @@ private bool ProcessOneFile()
         {
             int dwErr = TraceEventNativeMethods.ProcessTrace(handles, (uint)handles.Length, (IntPtr)0, (IntPtr)0);
             if (dwErr == 6)
-                throw new ApplicationException("Error opening ETL file.  Most likely caused by opening a Win8 Trace on a Pre Win8 OS.");
+                throw new Exception("Error opening ETL file.  Most likely caused by opening a Win8 Trace on a Pre Win8 OS.");
```
vancem: Same issue with Application Exception.  We should probably make a helper and fix it one place.

src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -611,7 +611,7 @@ private bool ProcessOneFile()
         // method (see http://msdn2.microsoft.com/en-us/library/aa364089.aspx) We set it up so that we call
         // back to ETWTraceEventSource.Dispatch which is the heart of the event callback logic.
         // [SecuritySafeCritical]
-        [AllowReversePInvokeCalls]
+        //[AllowReversePInvokeCalls]
```
vancem: Go ahead and remove these instead of commenting them out.

src/TraceEvent/Computers/ActivityComputer.cs
nategraf: That's correct. ApplicationException is not available in 1.6. What if we created a new exception type and threw that? 

That way the developer using TraceEvent can catch that Exception and know it was one of these errors. I notice we catch ApplicationException in a few places within PerfView

src/TraceEvent/NativeDlls.cs
```diff
@@ -24,7 +24,7 @@ public static void LoadNative(string simpleName)
         // When TraceEvent is loaded as embedded assembly the manifest path is <Unknown>
         // We use as fallback in that case the process executable location to enable scenarios where TraceEvent and related dlls
         // are loaded from byte arrays into the AppDomain to create self contained executables with no other dependent libraries. 
-        string assemblyLocation = Assembly.GetExecutingAssembly().ManifestModule.FullyQualifiedName;
+        string assemblyLocation = typeof(NativeDlls).AssemblyQualifiedName;
```
vancem: This is not correct.  The AssemblyQualifedName is aa name like 'mscorlib, Version=4.0.0.0' and we want a file path name.   However you can fix this simply by typeof(NativeDlls).Assembly.ManifestModule.FullyQualifiedName

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -880,8 +880,10 @@ internal bool GetPhysicalFileFromServer(string serverPath, string pdbIndexPath,
                             m_log.WriteLine("FindSymbolFilePath: In task, sending HTTP request {0}", fullUri);
 
                             var req = (System.Net.HttpWebRequest)System.Net.HttpWebRequest.Create(fullUri);
-                            req.UserAgent = "Microsoft-Symbol-Server/6.13.0009.1140";
```
vancem: Was this intentional to remove the 'UserAgent' property set?   If so we need to do a bit more testing.   At one point at least symbol servers did not reply properly if this user agent field was not set to something it recognized (hopefully they just dropped that).

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -2437,7 +2439,7 @@ string GetSourceFromSrcServer()
                 if (!File.Exists(target) && fetchCmdStr != null)
                 {
                     log.WriteLine("Trying to generate the file {0}.", target);
-                    var toolsDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().ManifestModule.FullyQualifiedName);
+                    var toolsDir = Path.GetDirectoryName(typeof(SourceFile).AssemblyQualifiedName);
```
vancem: Same issue.  AssemblyQualifiedName -> Assembly.ManifestModule.FullyQualifiedName

src/TraceEvent/Utilities/command.cs
```diff
@@ -422,16 +422,14 @@ public Command(string commandLine, CommandOptions options)
             if (options.elevate)
             {
                 options.useShellExecute = true;
-                startInfo.Verb = "runas";
+                startInfo.Arguments = "runas " + startInfo.Arguments;
```
vancem: Have we tested this?

src/TraceEvent/Symbols/SymbolReader.cs
nategraf: Yes. That isn't supported in 1.6. I'll go ahead and test that

vancem:
In general it does not look to bad to support V1.6.    We should try to make the official version work.   

I can make a new SupportFiles DLL when we are happy with any changes (send me the OSExtensions.cs file you modified.

src/TraceEvent/Computers/ActivityComputer.cs
vancem: Make up a new type called ApplicationException in the Microsoft.Diagnostics.Tracing.  Have this exception inherit from the System.Application when we are on desktop (plumb through a new define, let me know if you need help doing this).     

This way on Desktop, code will simply just work.

src/TraceEvent/Utilities/command.cs
nategraf: I hadn't. I thought I knew what it was doing after reading the docs, but I should have, because this doesn't work. I am working on a alternate solution.

src/TraceEvent/Computers/ActivityComputer.cs
nategraf: Should we maybe call it "TraceEventApplicationException" or something like that to avoid someone ripping out their hair in the future trying to catch our ApplicationException from a netstandard compiled package when they are on desktop?

src/TraceEvent/Utilities/command.cs
nategraf: I've been looking for a solution to this and so far nothing great. Furthermore, it seems that `UseShellExecute` is actually [not supported on core 1.1](https://github.com/dotnet/corefx/blob/release/1.1.0/src/System.Diagnostics.Process/src/System/Diagnostics/ProcessStartInfo.Windows.cs#L42-L49)

Seeing as this code is only concerned with Windows anyway it would be possible to solve this by adopting a portion of the source from corefx 2.0, but I haven't determined just how big that chunk would need to be yet

I'll keep looking for another solution in the meantime

nategraf: The prospect of supporting this in netstandard1.6 seems to be surprisingly difficult. My current thought is to throw `PlatformNotSupportedException` on netstandard1.6 through the use of an #if condition. That way we can still support perfview working just as before, but leave this unimplemented in netstandard1.6. If this is the approach then it seems reasonable to additionally target netstandard2.0 so users in core 2.0 on Windows can use this function

nategraf:
@vancem one thing I noticed you didn't comment on was the bump to net461 for desktop consumers. I did this to avoid some apis that are different between  net45 and netstandard1.6. Will that be an not be an issue, or should I go back and modify the code with some precompiler defines to revert that change?

src/TraceEvent/Computers/ActivityComputer.cs
vancem: My goal is to avoid making the code 'worse' for the sake of old things (V1.6 support).   I don't mind if old things have poorer functionality.    My personal opinion is that you should not have many different exception types (who uses them?) but you doe need a few 'groups' and this idea that knowing just the exception type you can usefully predict if it is likely some human should see.  

I really don't want to make users do 'wierd' things for V2.0 or beyond (I don't care about 1.6), and frankly ANY changes are suspect (as this one is).  

One 'easy' way of doing this to have a V2.0 and a V1.6 builds in the nuget package   Then simply have ifdefs for the V1.6 version.   This will also be useful for the issue you ran into with elevation, and I suspect we might run into it in the future as well.   

What do you think about having a helper that throws the applicationException and having #ifdef code that throws Exception on V1.6 in the body?

src/TraceEvent/Symbols/SymbolReader.cs
vancem: Note an easy way of doing this is to build PerfView with this line removed, remove the %TEMP%\symbols directory and then resolve some native OS Dlls from a trace.   If that works I am happy.

src/TraceEvent/Utilities/command.cs
vancem: Yes, as mentioned above, making a new target for 1.6 and #ifdefs seems and throwing when needed is very reasonable.   Do that.

vancem:
The move to net461 is an interesting one.    I personally don't have too much of a care factor for V4.5 but there are users of TraceEvent that at least care at the moment (although they do have their private copy right now).   

My take is that we should be minimizing effort for old systems until shown to have value.   Thus unless the effort is small to support 4.5 I am OK deferring it.   However I do want to understand how hard it would be to do that work if we decide it is important.   Can you have some idea what would be needed to keep the V4.5 support (can we do an experiment where we that this delta and simply put back all the net45 targets and see what explodes).  If it is easy, we should just do it.  If it is larger, We can document it in the csproj file how much effort it is likely to be.

src/OSExtensions/OSExtensions.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net45;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net461;netstandard1.6</TargetFrameworks>
```
vancem: Because OSExtensions is its own DLL (and a pain to update) and probably does not use APIs that are different between net45 and netStandard1.6 we should leave this as net45.   This allows us to move to net45 without having to update OSExtensions in the future.

sharwell:
@nategraf :bulb: You can update your GitHub profile to add your gmail account and it will link the remaining commits to you as the author.

nategraf:
I actually don't own that email (and I am mad about it, but that's a different story) so I suppose some mysterious other Nate Graf will be getting the credit for those commits.... or I could rewrite history, but I've read enough sci-fi to know that is rarely as successful as the hero hopes

src/TraceEvent/Symbols/SymbolReader.cs
nategraf: It works without setting user agent. I captured the [network log](https://pastebin.com/4uzUqdUg) just to confirm

src/OSExtensions/OSExtensions.csproj
```diff
@@ -12,8 +12,12 @@
     <Copyright>Copyright © Microsoft 2010</Copyright>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="Microsoft.Win32.Registry" Version="4.0.0" />
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.6'">
```
vancem: I was able to build OSExtensions with net45 rather than net461.   I also removed the dependency on RuntimeInformation (it was causing me failures during testing).  

In short,  you can simply remove your changes for OSExtensions.csproj.  The one that is there now is the one that I built with, and it should work.

nategraf: Alright, that's what I just did

src/TraceEvent/Utilities/command.cs
```diff
@@ -419,19 +420,27 @@ public Command(string commandLine, CommandOptions options)
             process = new Process();
             process.StartInfo = startInfo;
             output = new StringBuilder();
+
+// The functions needed to implement elevation are not supported in netstandard1.6
+// The command may just work anyway if TraceEvent's process is already elevated
+#if ! NETSTANDARD1_6
```
vancem: The library contains an 'isElevated' API so in the common case if you ask for elevation you need it.  

Thus this API is 'mostly broken'.  I would prefer in those cases that on NETSTANDARD1_6 we throw a clean NotImplementedException (indicating NetStandard1.6) to be unambiguous.

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -880,8 +880,13 @@ internal bool GetPhysicalFileFromServer(string serverPath, string pdbIndexPath,
                             m_log.WriteLine("FindSymbolFilePath: In task, sending HTTP request {0}", fullUri);
 
                             var req = (System.Net.HttpWebRequest)System.Net.HttpWebRequest.Create(fullUri);
+#if ! NETSTANDARD1_6
                             req.UserAgent = "Microsoft-Symbol-Server/6.13.0009.1140";
```
vancem: Don't bother making this 1.6 specific.  I am OK with it being unconditional.  I only put that in because I seemed to need it to make it work, but that does not seem to be the case anymore.

vancem:
We are very close now.   What do we know about how hard it is to work on V4.5?   (If you remember what forced you to 4.6.1 that is good enough, I just want a sense: is it a man-day, a man-week or longer?

nategraf:
It would take under a day to get it to work. I can't remember exact functions, but it was only a small subset of the changes and could be fixed, if by nothing else, with precompiler targeting

vancem:
There are errors in the automated build above 
```
     Creating library C:\projects\perfview\src\EtwClrProfiler\Debug\amd64\ETWClrProfiler.lib and object C:\projects\perfview\src\EtwClrProfiler\Debug\amd64\ETWClrProfiler.exp
  ETWClrProfilerX64.vcxproj -> C:\projects\perfview\src\EtwClrProfiler\Debug\amd64\ETWClrProfiler.dll
CSC : error CS1566: Error reading resource '.\OSExtensions.dll' -- 'Could not find a part of the path 'C:\Users\appveyor\.nuget\packages\microsoft.diagnostics.tracing.traceevent.supportfiles\1.0.9\lib\net461\OSExtensions.dll'.' [C:\projects\perfview\src\PerfView\PerfView.csproj]
  PerfView.TestUtilities -> C:\projects\perfview\src\PerfView.TestUtilities\bin\Debug\net461\PerfView.TestUtilities.dll
  CtfTracing.Tests -> C:\projects\perfview\src\TraceEvent\Ctf\CtfTracing.Tests\bin\Debug\net461\Tests.dll
  TraceEvent.Tests -> C:\projects\perfview\src\TraceEvent\TraceEvent.Tests\bin\Debug\net461\TraceEventTests.dll
  TraceParserGen -> C:\projects\perfview\src\TraceParserGen\bin\Debug\net40\TraceParserGen.exe
C:\projects\perfview\src\PerfView\App.cs(985,23): warning CS0169: The field 'App.s_threadToInterrupt' is never used [C:\projects\perfview\src\PerfViewCollect\PerfViewCollect.csproj]
C:\projects\perfview\src\PerfView\App.cs(986,20): warning CS0414: The field 'App.s_controlCPressed' is assigned but its value is never used [C:\projects\perfview\src\PerfViewCollect\PerfViewCollect.csproj]
```
Note that they turn on warning as errors, but also when you use the official SupportDLLs package I was able to leave it using the net45 libraries, which in turn affect the paths of OSExtentions.dll that need to be fixed. up.  

Let me know if you need more guidance on fixing these things.  

Finally, I do note that like 1/3 of the change is moving from 4.5 to 4.6.1.   If it really only takes < 5 code sit fixes to make it work we really should seriously consider just doing it.    I really do have a team that is likely to want to use this package on 4.5 it really is just a question that they probably can live with the existing one until maybe they don't care about 4.5.    My point is that while we MAY be able to avoid 4.5 support, it would very likely be at least convenient (and maybe REALLY convenient).   This probably justifies a couple hours of work if that is all that it would take.     Think about it, like i say I am willing to defer it if it is too much work.

nategraf:
It should take too long to reenable 4.5 support. I'll go ahead and get that done today and tomorrow.

nategraf:
@vancem I got it working on net45. If you think it looks good, this should be ready to go

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
```diff
@@ -32,15 +32,31 @@
...
+    <file src="$OutDir$netstandard1.6\Microsoft.Diagnostics.FastSerialization.dll" target="lib\netstandard1.6" />
+    <file src="$OutDir$netstandard1.6\Microsoft.Diagnostics.FastSerialization.xml" target="lib\netstandard1.6" />
+    <file src="$OutDir$netstandard1.6\Microsoft.Diagnostics.FastSerialization.pdb" target="lib\netstandard1.6" />
+
+    <!-- Support Dlls -->
+    <file src="$OutDir$net45\Dia2Lib.dll" target="lib\netstandard1.6" />
+    <file src="$OutDir$net45\TraceReloggerLib.dll" target="lib\netstandard1.6" />
+    <file src="$OutDir$net45\OSExtensions.dll" target="lib\netstandard1.6" />
+
+    <!-- NET 4.6.1 Framework -->
```
vancem: Change this back to 4.5

PR #460, opened 2017-12-08
Ensure RegisteredEventParser does call into tdh.dll on non-Windows platforms

nategraf:
Event processing which involved registered events currently fails on Linux when this parser tries to load tdh.dll (which is a Windows binary)

This PR fixes this by skipping steps in the lookup process which involve these calls.

### DIFF
--- a/src/TraceEvent/RegisteredTraceEventParser.cs
+++ b/src/TraceEvent/RegisteredTraceEventParser.cs
@@ -235,6 +235,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
                                     if (enumBuffer == null)
                                         enumBuffer = (byte*)System.Runtime.InteropServices.Marshal.AllocHGlobal(buffSize);
 
+#if !NOT_WINDOWS
                                     if (!enumIntern.ContainsKey(mapName))
                                     {
                                         EVENT_MAP_INFO* enumInfo = (EVENT_MAP_INFO*)enumBuffer;
@@ -269,6 +270,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
                                             }
                                         }
                                     }
+#endif //!NOT_WINDOWS
                                 }
 
                                 // Remove anything that does not look like an ID (.e.g space)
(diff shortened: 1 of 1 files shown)


src/TraceEvent/RegisteredTraceEventParser.cs
```diff
@@ -235,37 +235,40 @@ public static string GetManifestForRegisteredProvider(Guid providerGuid)
                                     if (enumBuffer == null)
                                         enumBuffer = (byte*)System.Runtime.InteropServices.Marshal.AllocHGlobal(buffSize);
 
-                                    if (!enumIntern.ContainsKey(mapName))
+                                    if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
vancem: I am surprised you care about GetManifestForRegisteredProvider.    It should not be called.   

Please hoist this check to the start of GetManifestForRegisteredProvider and throw an NotSupportedException.

nategraf: This seems comment valid for the other function I changed in this PR as well. The original reason I made this PR is to avoid crashing out on Linux, and adding NotSupportedException is slightly more clear than throwing when the dll is not found, but it doesn't really solve the issue, which I realize now must be whatever is causing this function to be called on Linux where we know it shouldn't work.

Should I maybe ditch this PR and go hunting for the logic error that is calling the RegisteredEventParser to get called in the first place? As I understand it, this parser is essentially entirely specific to Windows.

vancem: > This seems comment valid for the other function I changed in this PR as well. 

Actually the other if-check you inserted is OK, and expected.   Basically the issue is that the code that decodes EventSources called DynamicTraceEventParser) IS needed on Linux but is entangled with the code that does the OS lookup (RegisteredTraceEventParser) which is not.   While you could imagine disentangling it, it is certainly not worth it in the short term.   It is also pretty easy to neuter it because pretty much the main entry point to it is the 'TryLookupWorker' and if you simply return 'null' (lookup turned up nothing), it is fine.    The 'TryLookupWorker will be called any time you encounter an event that all other Parsers could not determine (which I WOULD expect to happen).

What is surprising about GetManifestForRegisteredProvider  is that you should only enter this method if some client calls it, and the scenarios where you need it are VERY rare (and only make sense on Windows where there are registered manifests).    Thus I would not have expected you to 'trip over' this one.

I think the right thing to do is to remove the if condition in GetManifestForRegisteredProvider  and leave the one in TryLookupWorker then see if you hit any errors on Linux.

nategraf: Alright that's what I did with e6302f8, and it works on Linux in my tests

vancem:
LGTM

PR #462, opened 2017-12-09
Add support .net native interop CCW events

GeorgePlotnikov:
Adding support .NET native interop CCW events into PerfViewData in callback manner.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -200,4 +200,3 @@ PerfView.exe
 src/TraceEvent/EventCounterHandler.cs
 OSExtensions.cs
 password.txt
-/src


src/TraceEvent/TraceEvent.cs
```diff
@@ -171,6 +171,20 @@ public ClrTraceEventParser Clr
             }
         }
 
+        /// <summary>
+        /// For convenience, we provide a property returns a ClrPrivateTraceEventParser that knows 
+        /// how to parse all the privte Common Language Runtime (CLR .NET) events into callbacks.
+        /// </summary>
+        public ClrPrivateTraceEventParser ClrPrivate
```
vancem: I have resisted making a shortcut for the ClrPrivate provider.   The idea is that (as its name suggests) is not public and in fact we want it to go away (move everything into the CLR provider).   

Also we only make these helpers for the TRULY important providers (kernel, clr, dynamic)., and only to help discoverability.    Arguably ClrPrivateTraceEventParser does not fall into that bucket.    

Every other provider you just need to do a 'New' on that one.  This is what we should do here.   

Remove this helper and simply new it up as needed.

src/TraceEvent/TraceEvent.cs
```diff
@@ -186,6 +200,20 @@ public KernelTraceEventParser Kernel
             }
         }
 
+        /// <summary>
+        /// For convenience, we provide a property returns a InteropTraceEventParser that knows 
+        /// how to parse the Interop events into callbacks.
+        /// </summary>
+        public InteropTraceEventParser Interop
```
vancem: Same as above.   I don't want TraceEvent (which is low in the dependency hierarchy) to take a dependency on basically 'random' parsers (which are high in the dependency hierarchy).  

Just new it up as you need it.   And remove it from here.

src/TraceEvent/Parsers/InteropTraceEventParser.cs
```diff
@@ -0,0 +1,1263 @@
+﻿using System;
```
vancem: Don't put this file in TraceEvent (it is not generally useful enough).   Put in in PerfView directory (you will see a PerfViewTRaceEventParser.cs there as well).

src/TraceEvent/TraceLog.cs
```diff
@@ -845,9 +845,9 @@ internal void RegisterStandardParsers()
             new HeapTraceProviderTraceEventParser(this);
             new MicrosoftWindowsKernelFileTraceEventParser(this);
             new IisTraceEventParser(this);
-
             new SampleProfilerTraceEventParser(this);
-#if false 
+            new InteropTraceEventParser(this);
```
vancem: Remove this as well.  You don't need it.   This means that your change does not affect TraceEvent at all (just PerfView).  This is how I want it.

src/PerfView/PerfViewData.cs
```diff
@@ -4120,9 +4121,54 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
+                    var ccwRefCount = "CCW NewRefCnt " + args.refCount;
+                    var ccwRefCountIndex = stackSource.Interner.FrameIntern(ccwRefCount.ToString());
+                    stackSource.Interner.CallStackIntern(ccwRefCountIndex, stackIndex);
+
+                    sample.StackIndex = stackIndex;
+                    stackSource.AddSample(sample);
+                });
+                eventSource.Interop.AddCallbackForEvents<TaskCCWRefCountDecArgs>(args =>
+                {
+                    sample.Metric = 1;
```
vancem: I believe we want the metric to be -1 so that decs 'cancel' the incs.   

It really would be useful to have the class name associated with the CCW as well.  We really should update the event to include this, but that is a separate issue.

src/PerfView/PerfViewData.cs
```diff
@@ -4120,9 +4121,54 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                 pinnedObjectAnalyzer.Execute(GCPinnedObjectViewType.PinnedObjectAllocations);
             }
             else if (streamName == "CCW Ref Count")
-            {
-                // TODO use the callback model.  We seem to have an issue getting the names however. 
-                foreach (var data in events.ByEventType<CCWRefCountChangeTraceData>())
+            {                
+                eventSource.Interop.AddCallbackForEvents<TaskCCWQueryRuntimeClassNameArgs>(args =>
+                {
```
vancem: Can you tell me more about when this fires?   Is it useful?

src/TraceEvent/TraceEvent.cs
GeorgePlotnikov: Got you

GeorgePlotnikov:
@vancem thanks for the review. Will apply it.

src/PerfView/PerfViewData.cs
GeorgePlotnikov: @vancem I assume this event might be useful because it contains information about the namespace and class name of CCW type. does it make sense?

vancem: I don't know enough about CCWQueryRuntimeClassName to understand if it is useful.   As you point out it might be because it has a name, but I suspect it won't be that useful because it fires only when the code actually calls a certain COM API (to get the class name), and that only happens rarely.

However even if the information is useful, we don't want it in this view because is throws off the ref count (we want increfCount to have a metric of 1 and decRefCount to have a metric of -1 so that we only see non-zero values for com objects that are still alive.   If you add in any other event with a non-zero metric you ruin that property.  

The fact that the Inc and Dec events have no way of getting a COM class name is unfortunate, and diminishes the value of the view but fixing that will require a change in CoreRT (which is not that big of a deal, we probably want to pursue it).  But that is a later step, for now we want to get the information we HAVE presented in a useful way (which means just using inc and dec).

vancem:
Hello George,

I see that you did a merge in your last commit.   You don't want to do that (you can now see that there are 250+ files changed in the branch.     Instead you want to rebase.    The good news is that it can be fixed.   Assuming your current branch is this one you can do
```
git reset --hard a1b24db
git push origin +support-dotnetnative-interop-events-for-ccw-ref-counting-views
```
The first command forceably set the current branch to a particular (the one right before the merge), and the second pushes it to your github fork.   The + sign is SUPER IMPORTANT here (and why you can't do this easily from Visual Studio (as far as I know).   Without this + sign the push will MERGE not REPLACE what is on github, which will suck in the bad merge, which is what we DON'T want.    We really need REPLACE semantics, and that is what the + does (it is called a forced push).  

You can actually rebase you change on top of the current master branch, if you have never done (it also requires a forced push), then just do the fix above and I will do the rest.

GeorgePlotnikov:
Hi @vancem, sincerely appreciate you for this info. Yesterday I struggled with this situation,  actually don't know how exactly it occurred. Today will try to apply your suggestion. Thank you very much, once again. Weird git glitch...

GeorgePlotnikov:
Hello @vancem, unfortunately, commands above said that did something, but actually wound gave any results. So I assumed the best way is to re-create branch and pull it. So [here](https://github.com/Microsoft/perfview/pull/477) it is. Actually, I still haven't realised what went wrong with the first branch.

PR #464, opened 2017-12-11
Open on a Diagsession file to call Open on the contained Etl resource

danlev1:
This change is to allow the RPS PerfView extension to work with .Diagsession file types. Extensions like RPS that extend the OnFileOpen for .diagsession files need to access the contained Etl resource. The PerfView.DiagSessionPerfViewFile object that is passed in (as pathToFile string) provides access to the Etl file, as an item in its Children collection. However, the PerfView.EtlPerfViewData that is returned does not have any useful data because the Etl file has not had its Open method exercised yet. I did not see a way to force Open on the Etl file from w/n the extension, and I'm not sure that would be the right approach. Instead, the change adds logic to the Open sequence of the .diagsession to also Open the contained Etl resource. The Open sequence calls AddEtlResourcesAsChildren, which is where I now call Open on the first discovered Etl file. I am doing the work on the existing worker for the .diagsession Open, so I created an OpenOnExistingWorker wrapper for the Etl file.


src/PerfView/PerfViewData.cs
```diff
@@ -5999,6 +6026,8 @@ protected override Action<Action> OpenImpl(Window parentWindow, StatusBar worker
         }
         // public override string DefaultStackSourceName { get { return "CPU"; } }
 
+
```
AaronRobinsonMSFT: Remove these extra lines.

vancem:
> @danlev1 

The PerfViewTreeItem object model definitely has some warts, but I don't wish to make it worse.    In particular making some parts of the object (in this case a DiagSession file) more expanded (but not completely expanded), feels bad, and that seems to be what is going on here.

It sounds like you are just trying to get your extension working, that has a 'OnFileOpen' hook.   Thus you have the PerfVIewFile, and you just need the ability to open any desired child.   

I want to understand why the solution is not just to use the 'Open' method on the child you need.    If it is simply that you don't know how to get a 'parentWindow' and 'StatusBar', it is the case that in other places we use GuiApp.MainWindow and GuiApp.MainWindow.StatusBar to do that.    I don't think that is so bad (indeed, if I were to do it over, I probably would probably create such a default).

There may be other issues that I don't understand.  Please bring them up.   I am interested in understanding your problem enough to give you good advice about the best way to accomplish it.

danlev1:
Thanks Vance! I was not aware of the GuiApp reference and will be exploring that route. I am not seeing GuiApp.MainWindow.StatusBar. Is it an extension?

vancem:
> I am not seeing GuiApp.MainWindow.StatusBar. Is it an extension?
StatusBar is a control on the Main window (See MainWindow.xaml line 135).   If you search for 'MainWindow.StatusBar' you will find it being used a few dozen places where there is not an obviously better place to log things.

danlev1:
Thanks for the feedback. While I see GuiApp.MainWindow.StatusBar calls in the PerfView codebase, I do not see a way to access the StatusBar via the extension project. The project has a reference to PerfView.exe and PresentationCore/PresentationFramework, but GuiApp.MainWindow.StatusBar does not resolve and the project will not build. Am I missing a required reference?

vancem:
The problem is that WPF items are internal by default, so the field is not accessible outside the PerfView executable.  

As a trial work-around, we could add this
```
       public StatusBar GetStatusBar() { return StatusBar; }
```
But I think a different change is better, where Open has a overload that takes no arguments and implicitly does its work in the 'default' window and StatusBar (that is GuiApp.MainWindow and GuiApp.MainWIndow.StatusBar).  

Vance

danlev1:
Added Open and OpenOnExistingWorker wrappers that default to use the internal GuiApp.MainWindow and GuiApp.MainWindow.StatusBar in order to simplify the call from an extension.

src/PerfView/PerfViewData.cs
```diff
@@ -7837,11 +7876,13 @@ private void AddResourcesAsChildren(StatusBar worker, DhPackage dhPackage, strin
         /// <summary>
         /// Adds child files from ETL resources in the DhPackage
         /// </summary>
-        private void AddEtlResourcesAsChildren(StatusBar worker, DhPackage dhPackage)
+        private void AddEtlResourcesAsChildren(Window window, StatusBar worker, DhPackage dhPackage)
         {
             ResourceInfo[] resources;
             dhPackage.GetResourceInformationByType("DiagnosticsHub.Resource.EtlFile", out resources);
 
+            var firstEtlHasBeenProcessed = false;
```
AaronRobinsonMSFT: Doesn't seem like this is needed anymore.

src/PerfView/PerfViewData.cs
```diff
@@ -7837,11 +7876,13 @@ private void AddResourcesAsChildren(StatusBar worker, DhPackage dhPackage, strin
         /// <summary>
         /// Adds child files from ETL resources in the DhPackage
         /// </summary>
-        private void AddEtlResourcesAsChildren(StatusBar worker, DhPackage dhPackage)
+        private void AddEtlResourcesAsChildren(Window window, StatusBar worker, DhPackage dhPackage)
```
vancem: It also seems that you don't need the new window parameter (it is now unused).   I would actually like to move in the direction of not having the Window parameter (since we should be able to derive it from the StatusBar)    Thus it is better if we remove this.

src/PerfView/PerfViewData.cs
```diff
@@ -493,6 +498,38 @@ public override void Open(Window parentWindow, StatusBar worker, Action doAfter
                     m_singletonStackSource.Viewer.Focus();
             }
         }
+
+        public void OpenOnExistingWorker(Action doAfter = null)
```
vancem: Can you describe the difference between OpenOnExistingWorker() and Open() and the scenarios where it is used and Open() does not work?    At the least we need to document it, but I would rather try to fold it into Open() somehow.

vancem:
Closing I rebased the files in PR #544

PR #471, opened 2017-12-20
Updating IIS Stats View to show Activity and Thread time view both

puneetg1983:
In this pull request, these are the issues fixed

- Parsed the Request CorrelationId which is used to accurately launch the Activity Start Stop Time View for slow requests.
- Added logic for the IIS's GENERAL_CHILD_REQUEST_START events which was causing us to incorrectly identify the slowest modules in the IIS request processing pipeline
- Add logic for more IIS related ETW events
- Parsing the ASP.NET Module for Pipeline Module events

Here is a screenshot of how the Stack Traces view looks like now. The option to launch Activity View is provided if the request has switched threads and the option to launch Threads View is provided to launch thread time view for synchronous requests.

image

### DIFF
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -1324,11 +1324,53 @@ table {
 
             var dispatcher = dataFile.Events.GetSource();
 
+            dispatcher.Dynamic.AddCallbackForProviderEvent("Microsoft-Windows-ASPNET", "Request/Send", delegate (TraceEvent data)
+            {
+                IisRequest request;
+                if (m_Requests.TryGetValue(data.ActivityID, out request))
+                {
+                    if (data.RelatedActivityID != Guid.Empty)
+                    {
+                        request.RelatedActivityId = data.RelatedActivityID;
+                    }
+                }
+            });
+
+            Dictionary<Guid, int> childRequests = new Dictionary<Guid, int>();
+
             var iis = new IisTraceEventParser(dispatcher);
 
             int startcount = 0;
             int endcount = 0;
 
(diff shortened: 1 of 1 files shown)


vancem:
I am OK with merging this, however I do think that the code has outgrown being 'inlined' into the PerfViewData.cs file.    On the one extreme, we could simply move PerfViewIisStats to its own file.  However I PerfView* classes are really logic associated with the GUI (how clicks work) and ideally what we want is decouple that logic from the GUI.    There are two ways of doing this

1.  Create a 'computer' for IIS / ASP.NET.     A computer is concept in the TraceEvent package (see the TraceEvent\Computers subdirectory).    The 'TraceProcess' or 'ActivityComputer' or GCReferenceComputer' are good examples.   Basically a computer is something that is given a TraceEventDispatcher (or TraceLogEventSoruce, both of which represent a 'stream' of events).   Given this it provides APIs that compute (hence the name) of other things derived from the events.   Thus its job is to 'hookup' to the event stream so that it can provide APIs for some other Information (e.g. a table of processes, or activities, or GCs or whatever).    

Computers are interesting, but tend to be 'low level' providing basic things.  (Basically what you want is the IIS requests (flesh out your IISRequest to be a public API).  

You can go further and may your computer also actually create the HTML reports (this is what the GCStats and JItStats do).  Not that this logic is GUI independent (it depends on HTML, but NOT The PerfViewData* data structures).   

I want to plant these ideas.   None of this is strictly needed, but you can see that the IIS logic is growing and the value of having 'reusable' code is likely growing (and even if it is not reused, it is valuable (because it is easier to read), to factor this a bit.    It is something to think about.

puneetg1983:
I totally agree. I will follow your recommendations and try to refactor this code in the next pull request.

PR #477, opened 2017-12-22
Add support dotnetnative interop events for ccw ref counting views

GeorgePlotnikov:
Instead of [this](https://github.com/Microsoft/perfview/pull/462) PR which branch was fatal corrupted. [Here](https://goo.gl/mKUgmo) is an example etl file which might be used to check how this PL works. 

At the collect data dialogue, there is a new checkbox "CCWRefCount" which allows collecting CCWRef events.


vancem:
I believe this PR is a rebasing of  PR #462.

GeorgePlotnikov:
Hello @vancem, true, it's the same PR but with applied code review comments. Git juggling didn't help eliminate "commits ahead" problem so I decided to re-create fork and provide new PR.

GeorgePlotnikov:
Hello @vancem, shall I apply some changes to this RP? Would you be so kind to review it?

vancem:
Sorry for the delay.  Indeed I wanted to look at this for over a month.   I am looked it over, you actually removed support for the old CCW refcount logic and that is actually still useful (it can be generated from the .NET Desktop runtime if you enable ClrPrivate events).    I also have to do some experimentation just to see that it actually is useful for its intended purpose.  I will be doing that tomorrow.

GeorgePlotnikov:
Thank you, @vancem

src/PerfView/PerfViewData.cs
```diff
@@ -4122,36 +4123,54 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
             }
             else if (streamName == "CCW Ref Count")
             {
-                // TODO use the callback model.  We seem to have an issue getting the names however. 
-                foreach (var data in events.ByEventType<CCWRefCountChangeTraceData>())
+                var interopTraceEventParser = new InteropTraceEventParser(eventSource);
+                interopTraceEventParser.AddCallbackForEvents<TaskCCWQueryRuntimeClassNameArgs>(args =>
```
vancem: It turns out that this QueryClassName event is probably not going to be useful to get a symbolic name, and it certainly should not be representing a 'increment' (which is what Metric=1) will do.   Just remove it.

src/PerfView/PerfViewData.cs
```diff
@@ -4122,36 +4123,54 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
+                    sample.Metric = 1;
+                    sample.TimeRelativeMSec = args.TimeStampRelativeMSec;
+                    var stackIndex = stackSource.GetCallStack(args.CallStackIndex(), args);
 
-                    var ccwRefKindName = "CCW " + operation;
-                    var ccwRefKindIndex = stackSource.Interner.FrameIntern(ccwRefKindName);
-                    stackIndex = stackSource.Interner.CallStackIntern(ccwRefKindIndex, stackIndex);
+                    var ccwRefKindName = "CCW RefCountInc";
+                    var ccwRefKindNameIndex = stackSource.Interner.FrameIntern(ccwRefKindName);
+                    stackSource.Interner.CallStackIntern(ccwRefKindNameIndex, stackIndex);
```
vancem: These frames don't actually show up on the view, because you did not do stackIndex = ....Intern(... stackIndex).  
This error was made several places.

src/PerfView/PerfViewData.cs
```diff
@@ -4122,36 +4123,54 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
             }
             else if (streamName == "CCW Ref Count")
             {
-                // TODO use the callback model.  We seem to have an issue getting the names however. 
-                foreach (var data in events.ByEventType<CCWRefCountChangeTraceData>())
+                var interopTraceEventParser = new InteropTraceEventParser(eventSource);
```
vancem: The old code is still useful.     Rather than try to make a view that is 'both' we should instead simply make a new view called '.NET Native CCW Ref Count'.

src/PerfView/Dialogs/RunCommandDialog.xaml
```diff
@@ -301,8 +301,12 @@
                     <Hyperlink Command="Help" CommandParameter="JITInliningCheckBox">JIT Inlining:</Hyperlink>
                     </TextBlock>
                     <CheckBox Grid.Row="0" Grid.Column="9" Name="JITInliningCheckBox" VerticalAlignment="Center" />
-                    
-                    <CheckBox Grid.Row="0" Grid.Column="11" Name="Unused1" Visibility="Hidden"  VerticalAlignment="Center"/>
+
+                    <TextBlock Grid.Row="0" Grid.Column="10" VerticalAlignment="Center" HorizontalAlignment="Right" Margin="8,2,5,2" 
+                                ToolTip="Log information about CCW reference counting.">
```
vancem: Instead of just CCW say 

NET Native Com Callable Wrapper (CCW)

src/PerfView/CommandLineArgs.cs
```diff
@@ -442,7 +443,7 @@ private void SetupCommandLine(CommandLineParser parser)
             parser.DefineOptionalQualifier("DotNetCalls", ref DotNetCalls, "Turns on per-call .NET profiling.");
             parser.DefineOptionalQualifier("DotNetCallsSampled", ref DotNetCallsSampled, "Turns on per-call .NET profiling, sampling types in a smart way to keep overhead low.");
             parser.DefineOptionalQualifier("JITInlining", ref JITInlining, "Turns on logging of successful and failed JIT inlining attempts.");
-
+            parser.DefineOptionalQualifier("CCWRefCount", ref CCWRefCount, "Turns on logging of information about CCW reference counting.");
```
vancem: Say .NET Native CCW

src/PerfView/SupportFiles/UsersGuide.htm
```diff
@@ -4028,6 +4028,11 @@ <h3>Advanced Options</h3>
             [MethodImpl(MethodImplOptions.AggressiveInlining)], etc.). This feature can also be
             activated with the /JITInlining command line option.
         </li>
+        <li>
+            The <strong><a id="CCWRefCountCheckBox">CCW References Count Checkbox</a></strong> - Causes
```
vancem: Say .NET Native CCW. and explain that it is only for .NET Native applications.

src/PerfView/PerfViewData.cs
```diff
@@ -5804,7 +5823,7 @@ protected override Action<Action> OpenImpl(Window parentWindow, StatusBar worker
                         hasMemAllocStacks = true;
                     if (name.StartsWith("GC/SampledObjectAllocation"))
                         hasMemAllocStacks = true;
-                    if (name.StartsWith("GC/CCWRefCountChange"))
+                    if (name.StartsWith("GC/CCWRefCountChange") || name.StartsWith("TaskCCWRef"))
```
vancem: I want this view to be independent of the existing CCW refcount view (because it is only for .NET Native but the other is useful for .NET Desktop).

src/PerfView/PerfViewData.cs
```diff
@@ -4122,36 +4123,54 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
-                    var ccwTypeName = "CCW Type " + data.NameSpace + "." + data.ClassName;
-                    var ccwTypeIndex = stackSource.Interner.FrameIntern(ccwTypeName);
-                    stackIndex = stackSource.Interner.CallStackIntern(ccwTypeIndex, stackIndex);
+                    var ccwRefKindName = "CCW RefCountDec";
+                    var ccwRefKindNameIndex = stackSource.Interner.FrameIntern(ccwRefKindName);
+                    stackSource.Interner.CallStackIntern(ccwRefKindNameIndex, stackIndex);
+
+                    var ccwRefCount = "CCW NewRefCnt " + args.refCount;
+                    var ccwRefCountIndex = stackSource.Interner.FrameIntern(ccwRefCount.ToString());
+                    stackSource.Interner.CallStackIntern(ccwRefCountIndex, stackIndex);
```
vancem: Please put the instance identifier (it is passed as objectID), as one of the pseudo-fields (as a hex number).  

For extra credit, add another callback that listens to the TaskCCWCreation event, and remembers a table that maps object ID to Type ID (the TaskCCWCreation has the Type ID).   Then use your ObjectID here to look up the TYpe ID and ALSO include a pseudo-field for that.  (it is useful to group things by type, and  as a next step, we can also turn this type ID into a actual name by looking it up in the PDB (but we can save that for later)).

GeorgePlotnikov:
@vancem, thanks for the review. NP, I'll do it

src/PerfView/PerfViewData.cs
GeorgePlotnikov: Thanks, done.

src/PerfView/PerfViewData.cs
GeorgePlotnikov: done

src/PerfView/Dialogs/RunCommandDialog.xaml
GeorgePlotnikov: done

src/PerfView/CommandLineArgs.cs
GeorgePlotnikov: done

src/PerfView/SupportFiles/UsersGuide.htm
GeorgePlotnikov: done

src/PerfView/PerfViewData.cs
GeorgePlotnikov: done

vancem:
Thanks for the response.  I don't see a new commit however.   Just let me know when you push the changes.

GeorgePlotnikov:
@vancem yes, I just mark with "done" for the convenience not to miss comment :-)

src/PerfView/PerfViewData.cs
GeorgePlotnikov: done

src/PerfView/PerfViewData.cs
GeorgePlotnikov: done

GeorgePlotnikov:
@vancem, please review changes

src/PerfView/PerfViewData.cs
```diff
@@ -4153,6 +4154,72 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
+                    var ccwRefCount = "CCW NewRefCnt " + args.refCount;
+                    var ccwRefCountIndex = stackSource.Interner.FrameIntern(ccwRefCount.ToString());
+                    stackIndex = stackSource.Interner.CallStackIntern(ccwRefCountIndex, stackIndex);
+
+                    sample.StackIndex = stackIndex;
+                    stackSource.AddSample(sample);
+                });
+                interopTraceEventParser.AddCallbackForEvents<TaskCCWRefCountDecArgs>(args =>
+                {
+                    sample.Metric = -1;
```
vancem: This body should be identical to the TaskCCWRefCountIncArgs case except that the metric is -1.     The method above is now complex enough that you should make a helper that both of these call.   (The helper will take the timestamp, object ID and the refcount and metric as parameters).

src/PerfView/PerfViewData.cs
```diff
@@ -4153,6 +4154,72 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
+            else if (streamName == ".NET Native CCW Ref Count")
+            {
+                var objectTypeMap = new Dictionary<long, long>(1000);
+                var interopTraceEventParser = new InteropTraceEventParser(eventSource);
+                interopTraceEventParser.AddCallbackForEvent<TaskCCWCreationArgs>(null, args =>
+                {
+                    if(!objectTypeMap.ContainsKey(args.objectID))
+                        objectTypeMap.Add(args.objectID, args.targetObjectIDType);
+                });
+                interopTraceEventParser.AddCallbackForEvents<TaskCCWQueryRuntimeClassNameArgs>(args =>
```
vancem: Do you have a reason for including TaskCCWQueryRuntimeClassName in the stacks?  Is it just that it has a string name?    At the very least you should insure that sample.Metric is set to 0 (otherwise it is whatever got left over from the last event).  

But I am thinking that mostly it is not clear that it useful to include it at all.   (If we do, it should be included in the information remembered by your objectTypeMap (but now it is an objectInfoMap).  I would drop it completely for now.

GeorgePlotnikov: okay, if we would get this info from pdb

src/PerfView/PerfViewData.cs
GeorgePlotnikov: done

vancem:
This all looks good.   I want to build it and play with it a bit one final time to see what the experience is like.  (for those who wish to do the same collecting a trace using the built in windows 10 'Photos' app (which is a .NET Native application) will collect interesting data.

GeorgePlotnikov:
thank you @vancem! I can implement receiving info about type from PDB. should I create an issue regarding it?

vancem:
Sure make an issue for looking up the type names.   Most of the support in theory is already there.  See TypeNameSymbolResolver (it is used in the GC Heap Alloc View).   In theory doing something just like it in the CCW view should 'just work'.

PR #479, opened 2018-01-01
Consistently place the exclusive value columns before the inclusive value columns

sharwell:
This change takes a bit of time to get used to, but in the end it reduces the number of cases where a developer incorrectly reports an **Inc Ct** count value as an **Exc Ct** value due to misreading the columns of the current view.

This error occurred many times when I started using PerfView.

### DIFF
--- a/src/PerfView/StackViewer/PerfDataGrid.xaml
+++ b/src/PerfView/StackViewer/PerfDataGrid.xaml
@@ -58,6 +58,58 @@
                     </TextBlock>
                 </WPFToolKit:DataGridTemplateColumn.Header>
             </WPFToolKit:DataGridTemplateColumn>
+
+            <WPFToolKit:DataGridTextColumn 
+                IsReadOnly="True"
+                Binding="{Binding ExclusiveMetricPercent, StringFormat={}{0:n1}}" 
+                HeaderStyle="{StaticResource CenterAlign}"
+                ElementStyle="{StaticResource RightAlign}">
+                <WPFToolKit:DataGridTextColumn.Header>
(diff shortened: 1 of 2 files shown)


src/PerfView/StackViewer/PerfDataGrid.xaml
```diff
@@ -58,120 +58,121 @@
                     </TextBlock>
                 </WPFToolKit:DataGridTemplateColumn.Header>
             </WPFToolKit:DataGridTemplateColumn>
+
             <WPFToolKit:DataGridTextColumn
```
sharwell: :memo: The diff here is bad, but this was just a drag/drop reordering of the columns.

vancem:
The current behavior of PerfView is that inclusive  things are first, except for the ByName view where exclusive things are first.    The rationale for this is the first column is the number you probably care about so put it first (it also helps when you cut and paste, because you don't have columns you don't care about in there).   It is also handy when screen space is at a premium.   

I see that there is a certain consistency if you always put exclusive counts first, but the value of that seems small, and you are losing different value (the first number is the most important, and what you should care about).

And of course, because it is a 'very noticeable change, lots of people will hate it, if only because it is a change.      

But I don't want to discourage something if users really want it.   But I need some demonstration that there will be more people THANKFUL for the change that ANNOYED by the change.   

Please put the question to the 'PerfViewUsers@microsoft.com' alias (explaining the feature and its value), and ask for opinions (both positive and negative).  If there is a clear majority that would prefer it then I will not stand in the way.

brianrob:
Closing for now.  Please re-open once feedback has been solicited per @vancem's request.  Thanks.

PR #480, opened 2018-01-01
Properly account for the size of memory mapped files

sharwell:


src/HeapDump/GCHeapDumper.cs
```diff
@@ -1765,6 +1765,22 @@ private void DumpAllSegments()
                     memoryGraphTypeIdx = GetTypeIndexForName(typeName, null, objSizeAsInt);
                 }
 
+                if (type.Name == "Microsoft.Win32.SafeHandles.SafeMemoryMappedViewHandle")
```
vancem: In general I like the idea of displaying memory that objects 'hold onto' (for example in this case native memory), but I don't like it to be just added to the size of the object (like you did there)

Instead I believe it is better to make up a PSEUDO-node that (we do so for RCWs above) and attach it there and point at it (thus you say 'Memory Mapped File'.  That way users can decide to exclude if if they don't care, but also it makes is clear this memory is not in the GC heap and is managed differently.

brianrob:
@sharwell, agreed with @vancem's comment on this.  Ideally we have a pseudo node that represents to "held" memory.  Closing this for now, but would be happy to merge this with the suggested feedback implemented.

PR #481, opened 2018-01-02
Added a Dispatch Hook to enable locking process thread and event filtering

jeffschwMSFT:
Added a hook right before event dispatch so that 1) programs can gain frequent access to the processing thread and 2) events can be filtered before dispatch.  The primary scenario is to allow a multi-threaded application inspect data accumulated on the source by using a synchronization mechanism in the dispatch hook.

The primary scenario for the DispatchHook is when a secondary thread wants to inspect the contents of accumulators (like TraceManagedDotNetRuntime) during a live ETW session.  The pattern would be to add a hook which sets a manual reset event.  This synchronization will block further processing as the secondary thread inspects the state, and is largely an no-op when no inspection is necessary.

### DIFF
--- a/src/TraceEvent/TraceEvent.cs
+++ b/src/TraceEvent/TraceEvent.cs
@@ -2816,6 +2816,28 @@ namespace Microsoft.Diagnostics.Tracing
         /// Subscribers of Completed will be called after processing is complete (right before TraceEventDispatcher.Process returns.    
         /// </summary>
         public event Action Completed;
+
+        /// <summary>
+        /// Add a hook right before dispatch that determines if an event should be dispatched
+        /// Multiple subscribers may hook this event, in that case they are called in lifo order,
+        /// once a subscriber indicates that an event should not be processed the chain of dispatchers is no longer searched
+        /// <input>TraceEvent - the event that is ready for dispatch</input>
+        /// <return>True - dispatch this event</return>
+        /// <return>False - not dispatch the event</return>
+        /// </summary>
+        public void AddDispatchHook(Func<TraceEvent, bool> onDispatch)
+        {
+            var hook = new DispatchHookEntry()
+            {
+                OnDispatch = onDispatch,
+                Next = null
+            };
+
+            // last in first out order
+            if (DispatchHook != null) hook.Next = DispatchHook;
+            DispatchHook = hook;
+        }
+
         #region protected
         /// <summary>
         /// Called when processing is complete.  You can call this more than once if your not sure if it has already been called.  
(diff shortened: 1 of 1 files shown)


src/TraceEvent/TraceEvent.cs
```diff
@@ -2943,6 +2965,19 @@ internal protected void Dispatch(TraceEvent anEvent)
             try
             {
 #endif
+                // check if there is a registered dispatch hook
+                if (DispatchHook != null)
+                {
+                    // check if we should dispatch this event, the first to indicate
+                    // false we will return without dispatching
+                    var hook = DispatchHook;
+                    while (hook != null && hook.OnDispatch != null)
```
sharwell: :bulb: Ensure that `hook.OnDispatch` can never be null and remove the unnecessary null check

src/TraceEvent/TraceEvent.cs
```diff
@@ -2943,6 +2965,19 @@ internal protected void Dispatch(TraceEvent anEvent)
...
+                    {
+                        if (!hook.OnDispatch(anEvent)) return;
+                        hook = hook.Next;
```
sharwell: 💡 Use a `for` loop instead of a `while` loop since you have all three parts of a typical `for` loop

src/TraceEvent/TraceEvent.cs
```diff
@@ -2816,6 +2816,28 @@ public virtual void StopProcessing()
...
+        public void AddDispatchHook(Func<TraceEvent, bool> onDispatch)
+        {
+            var hook = new DispatchHookEntry()
+            {
+                OnDispatch = onDispatch,
+                Next = null
+            };
+
+            // last in first out order
+            if (DispatchHook != null) hook.Next = DispatchHook;
```
sharwell: :bulb: This null check is unnecessary. You could just initialize `Next` to `DispatchHook` from the start.

src/TraceEvent/TraceEvent.cs
```diff
@@ -2943,6 +2965,19 @@ internal protected void Dispatch(TraceEvent anEvent)
             try
             {
 #endif
+                // check if there is a registered dispatch hook
+                if (DispatchHook != null)
```
sharwell: :bulb: This null check is unnecessary (duplicate of the one that follows)

jeffschwMSFT:
Thanks @sharwell, accommodated your suggestions and ensured to null our nextRecord.

src/TraceEvent/TraceEvent.cs
```diff
@@ -2816,6 +2816,30 @@ public virtual void StopProcessing()
...
+        /// Add a hook right before dispatch that determines if an event should be dispatched
+        /// Multiple subscribers may hook this event, in that case they are called in lifo order,
+        /// once a subscriber indicates that an event should not be processed the chain of dispatchers is no longer searched
+        /// <input>TraceEvent - the event that is ready for dispatch</input>
+        /// <return>True - dispatch this event</return>
+        /// <return>False - not dispatch the event</return>
+        /// </summary>
+        public void AddDispatchHook(Func<TraceEvent, bool> onDispatch)
+        {
+            if (onDispatch == null) throw new ArgumentException("Must provide a non-null callback", "onDispatch", null);
```
sharwell: :bulb: `nameof(onDispatch)`

jeffschwMSFT:
Thanks @sharwell.  I debated removing the null check prior to the for - I agree it is redundant, but I felt it left the code a bit more readable (?) stating that this was an exceptional path, not the norm.

sharwell:
@jeffschwMSFT Keep in mind that this is an extremely hot path (millions of calls per second). Might be good to measure and see if it makes a difference.

vancem:
I think I pulled the 'merge' button too soon, as I am having second thoughts about the design of this.    I will follow up with Jeff offline.

PR #482, opened 2018-01-02
Add source.AddProcessLifetimeLimits as a mechanism to reduce how much information that computers accumulate

jeffschwMSFT:
Added a Lifetime tracking mechanism for TraceEvent Computers.  The contract is that for primary caches they should consult the source if there is a lifetime model being applied, if so trim the cache based on the minimum relative timestamp (remove everything earlier).  This is a best effort mechanism to ensure that information is not accumulated indefinitely.

TraceGC.PinnedPlugs and PinnedObjects are now lazy initialized and eagerly cleaned up to further help keep accumulated memory consumption as low as possible.

The primary scenario is for live ETW sessions that want to run for long periods of time (or forever).  The mechanism will provide a best effort to restrict future memory growth by trimming core caches based on the lifetime provided to the source.  

Parameters:
Lifetime - The oldest item in a cache
Interval - Trimming may be expensive, so the interval between subsequent trims is configurable.  A smaller value indicates that the cache should be trimmed more frequently.

### DIFF
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -102,6 +102,17 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
             if (map[process.ProcessIndex].OnLoaded != null) map[process.ProcessIndex].OnLoaded(runtime);
         }
 
+        internal static void OnDotNetRuntimeUnloaded(this TraceProcess process)
+        {
+            Debug.Assert(process.Source != null);
+            Dictionary<ProcessIndex, DotNetRuntime> map = (Dictionary<ProcessIndex, DotNetRuntime>)process.Source.UserData["Computers/LoadedDotNetRuntimes"];
+            if (map.ContainsKey(process.ProcessIndex))
+            {
+                // remove this runtime, since the process has terminated
+                map.Remove(process.ProcessIndex);
+            }
+        }
+
         private static TraceEventDispatcher m_currentSource; // used to ensure non-concurrent usage
         #endregion
     }
@@ -219,6 +230,32 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                     processRuntimes.Add(proc, mang);
                 }
             };
+
+            // if applying lifetime, trim loaded managed runtimes when processes are terminated and are out of lifetime
+            source.AddCallbackOnProcessStop((p) =>
+            {
+                // check if we are applying a lifetime model
+                if (source.DataLifetimeEnabled())
+                {
+                    // iterate through all processes and unload the managed runtime from processes
+                    //  that have exited and are out of lifetime
+                    // immediately removing the runtime for stopped processes is possible, but that
+                    //  breaks the contract of how long data is kept with the lifetime
+                    foreach (var process in source.Processes())
+                    {
+                        // continue if the process has not exited yet
+                        if (!process.ExitStatus.HasValue) continue;
+
+                        if (process.EndTimeRelativeMsec < (p.EndTimeRelativeMsec - source.DataLifetimeMsec))
+                        {
+                            // remove this managed runtime instance
+                            process.OnDotNetRuntimeUnloaded();
+                            // remove from the local cache
+                            if (processRuntimes.ContainsKey(process)) processRuntimes.Remove(process);
+                        }
+                    }
+                }
+            });
 
             source.Clr.All += createManagedProc;
             clrPrivate.All += createManagedProc;
@@ -639,6 +676,20 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
 
                         // fire event
                         if (stats.GCStart != null) stats.GCStart(process, _gc);
+
+                        // check if we should apply a lifetime limit to the GC cache
+                        if (source.DataLifetimeEnabled() && data.TimeStampRelativeMSec >= stats.GC.NextRelativeTimeStampMsec)
+                        {
+                            // note the next time that lifetime should be applied, to avoid cleaningup too frequently
+                            stats.GC.NextRelativeTimeStampMsec = data.TimeStampRelativeMSec + (source.DataLifetimeMsec/2.0);
+                            // trim the GCs to only include those either incomplete or completed after lifetime
+                            stats.GC.m_gcs = stats.GC.m_gcs.Where(gc => !gc.IsComplete || gc.StartRelativeMSec >= (data.TimeStampRelativeMSec - source.DataLifetimeMsec)).ToList();
+                            // rewrite the index for fast lookup
+                            for (int i = 0; i < stats.GC.m_gcs.Count; i++)
+                            {
+                                stats.GC.m_gcs[i].Index = i;
+                            }
+                        }
                     }
                 };
 
@@ -649,6 +700,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                     TraceGC _gc = TraceGarbageCollector.GetCurrentGC(stats);
                     if (_gc != null)
                     {
+                        if (_gc.PinnedObjects == null) _gc.PinnedObjects = new Dictionary<Address, long>();
                         if (!_gc.PinnedObjects.ContainsKey(data.ObjectID))
                         {
                             _gc.PinnedObjects.Add(data.ObjectID, data.ObjectSize);
@@ -671,6 +723,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                     {
                         // ObjectID is supposed to be an IntPtr. But "Address" is defined as UInt64 in 
                         // TraceEvent.
+                        if (_event.PinnedPlugs == null) _event.PinnedPlugs = new List<TraceGC.PinnedPlug>();
                         _event.PinnedPlugs.Add(new TraceGC.PinnedPlug(data.PlugStart, data.PlugEnd));
                     }
                 };
@@ -729,6 +782,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                     {
                         // ObjectID is supposed to be an IntPtr. But "Address" is defined as UInt64 in 
                         // TraceEvent.
+                        if (_gc.PinnedPlugs == null) _gc.PinnedPlugs = new List<TraceGC.PinnedPlug>();
                         _gc.PinnedPlugs.Add(new TraceGC.PinnedPlug(data.PlugStart, data.PlugEnd));
                     }
                 };
@@ -1043,6 +1097,15 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
 
                     // fire event
                     if (stats.JITMethodStart != null) stats.JITMethodStart(process, _method);
+
+                    // check if we should apply a lifetime limit to the method cache
+                    if (source.DataLifetimeEnabled() && data.TimeStampRelativeMSec >= stats.JIT.NextRelativeTimeStampMsec)
+                    {
+                        // note the next time that lifetime should be applied, to avoid cleaningup too frequently
+                        stats.JIT.NextRelativeTimeStampMsec = data.TimeStampRelativeMSec + (source.DataLifetimeMsec/2.0);
+                        // trim the methods to only include those that were JITT'd after the lifetime timestamp
+                        stats.JIT.m_methods = stats.JIT.m_methods.Where(meth => meth.StartTimeMSec >= (data.TimeStampRelativeMSec - source.DataLifetimeMsec)).ToList();
+                    }
                 };
                 ClrRundownTraceEventParser parser = new ClrRundownTraceEventParser(source);
                 Action<ModuleLoadUnloadTraceData> moduleLoadAction = delegate (ModuleLoadUnloadTraceData data)
@@ -1289,11 +1352,12 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
             return null;
         }
 
-        private List<TraceGC> m_gcs = new List<TraceGC>();
+        internal List<TraceGC> m_gcs = new List<TraceGC>();
         private GCStats[] m_generations = new GCStats[3];
         internal GCStats m_stats = new GCStats();
         private int m_prvcount = 0;
         private int m_prvCompleted = 0;
+        internal double NextRelativeTimeStampMsec;
 
         private void Calculate()
         {
@@ -1430,7 +1494,8 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
 
         #region private
         internal JITStats m_stats = new JITStats();
-        private List<TraceJittedMethod> m_methods = new List<TraceJittedMethod>();
+        internal List<TraceJittedMethod> m_methods = new List<TraceJittedMethod>();
+        internal double NextRelativeTimeStampMsec;
         #endregion
     }
 
@@ -1630,9 +1695,12 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
             if (pinnedObjectSizes == -1)
             {
                 pinnedObjectSizes = 0;
-                foreach (KeyValuePair<ulong, long> item in PinnedObjects)
+                if (PinnedObjects != null)
                 {
-                    pinnedObjectSizes += item.Value;
+                    foreach (KeyValuePair<ulong, long> item in PinnedObjects)
+                    {
+                        pinnedObjectSizes += item.Value;
+                    }
                 }
             }
             return pinnedObjectSizes;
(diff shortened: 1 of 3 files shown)


vancem:
Hello Jeff,

I am trying to get in my brain the public contract for the various players.   Here is my summary

TraceEventSources now have a AddProcessLifetimeLimts.  method.   This is what users of TraceEventSource will use to convey that they don't want ANY computer to store things older than given time interval, and that ideally the computers would only check every so often (which is the other number) 

I realize you may not like the number of changes I am suggesting, but my goal is to define a public contract and make it as simple as possible while achieving the goal of trimming data growth (the most common thing we do in Microsoft is to make things to complex that we don't get a lot of value out of).   I am willing to help if that matters.  

The first new public API is: 
```
public static void AddProcessLifetimeLimits(this TraceEventSource source, TimeSpan lifetime, TimeSpan interval = default(TimeSpan))
```
I am assuming that 'Process' means the processing of events (not OS processes).   I would like to simplify it to 

First;   I am thinking of simply changing it to the following 
```
public double DataLifetimeMsec get; set
```
Basically there is a property on TraceEventSource that remembers this  0 means infinite, and TraceEventSource is simply just a place to hang this property.  We drop the need for a 'interval' because we just don't need that level of control (we can give guidance that there is an expectation that you might keep at LEAST this amount, but generally not more than TWICE this amount.   I don''t think we will ever care to 'tune' it more than that.   

Thus I don't think you need the ProcessLifeTimeToken class at all.    Notice I made a public getter, which allows every computer to get at the value.     We have also defined the contract (basically a TraceSource just holds this standard value that computers can choose to look at or not.   

Each computer that support lifetime limiting can hold the state it needs.   We may have helper methods if we find that computers all want to do the same operations (it can be a little helper class, if we can do it cleanly and it has non-trivial functionality in it).   

I would like to believe that this would cut the number of lines of code by maybe half (we will see).   Ideally for each new computer you only have to intialize the helper class, and call the  poll method on it  frequently (passing lambda that cleans up state for that particular computer).    The goal is just a couple lines of new code in each computer to add this support (you are not that far away).  

But the main feedback is
* Just have a property to declare the desire to trim.  That is the only new public area (but it is complete)
* A small helper class that keep state with a poll method that is passed a lambda that will do an expensive trim for a particular computer.
* Use this for your GC and JIT computer.   Hopefully each only has 2 new lines (construct + poll) + lambda.  

What do you think?   We can talk 1-1 if you like.

jeffschwMSFT:
@vancem trimming interval seems like a fine simplification.  The concept of interval would still exist, but like you said it will not likely need much tuning.
There are really 4 concepts that are included in this pull:
1. Overall lifetime support (feedback above about simplifying to a property)
2. OnTerminate for TraceProcess to enable proper downstream cleanup (eg. ManagedDotnetRuntime)
3. OnDotnetRuntimeRemoved to remove the associated ManagedDotnetRuntime
4. Small number of lazy init

#2-4 still are interesting pieces to the puzzle (and do not expand the public surface area)

Making this a property it would then be a first class citizen as opposed to an extension (more isolated within the Computers space).  I assume you are OK with that, even though the scope of this feature is really within the computer scope.

In general the code is laid out in _about_ the way you are recommending.  The key difference is that each poller retains the last time it checked and thus retains a per poller interval.  This enables all pollers to not be influenced by other pollers.  Said another way, if JIT polls infrequently but GC polls all the time, I strive to ensure that JIT and GC cleanup happens regardless of what other computer may be polling.  The reason ProcessLifetimeToken exists is to persist the time it polled last and make it unique for each poller.  Do you not see a need for that, or is that part of the per computer contract?

vancem:
I am OK with the OnTerminate for TraceProcess.

I don't really mind OnDotnetRuntimeRemoved but that given that we don't really support unloading runtimes without a process terminate, I am wondering if we can simply not have that (effectively we 'leak' runtimes until the process ends, but I can't imagine that is a problem).  

> Making this a property it would then be a first class citizen as opposed to an extension (more isolated within the Computers space). I assume you are OK with that, even though the scope of this feature is really within the computer scope.

Yes, basically I am OK adding direct support to the underlying class (e.g. TraceEventDispatcher) because I view its job to be support computers.    Thus anything that is used in MOST computers is reasonable to add to TraceEventDispatcher.   This lets us do things in the 'obvious' way, which appeals to me.

> The reason ProcessLifetimeToken exists is to persist the time it polled last and make it unique for each poller. Do you not see a need for that, or is that part of the per computer contract?

I agree that each computer should have a variable (or more than one) that keeps track of when it did it last so it can purge efficiently.    Indeed I see that a class that does this (which is what your Token class does today), is OK.   What I want, however is it NOT be part of the contract between TraceEventSource and the Computers.     

Thus TraceEventDispatcher should NOT know about your tokens (TraceEventDispatcher job is to simply publish the limit, not decide how to implement the cleanup).   It should not be mentioned in any communication (APIs) between  TraceEventDispatcher  and its computers.  

However It IS OK to have a utility class (e.g. we can call it CleanupScheduler) which basically has the logic that is in your Token class today.   This is totally optional for computers (they can choose to use this helper not (and we may or may not make it public, based on demand).  

As it is now, while ProcessLifetimeToken is not public it would HAVE to be if we wanted a computer outside of TraceEvent.dll to partipate in lifetime logic.   However in the scheme I outline above computers outside of TraceEvent.dll can participate, even if we don't make CleanupScheduler public.   I also think that explaining the contract for people implementing a computer is easier (There is one new value (DataLifetimeMSec), and that is all they need to know in order to implement the contract.

Is that reasonable?     In practical terms it means

1.   adding DataLifetimeMSec and removing the other lifetime logic from TraceEvent.cs.  Thus the only changes to TraceEvent.cs are the property and the changes for OnTerminate.  
2. Renaming ProcessLifetimeToken to something like CleanupScheduler, and putting in in either its own file (ComputerSupport.cs, CleanupScheduler.cs or TraceProcess.cs (since it will be the first to use it)).   

It is the kind of change that takes about an hour, and is likely to work the first time after you get it compiling (since it does not do much really).

jeffschwMSFT:
Makes sense.  I will make the suggested changes and push a change this week.  Thanks for reviewing.

jeffschwMSFT:
Thanks for the feedback @vancem updated to accommodate.

src/TraceEvent/Computers/TraceProcess.cs
```diff
@@ -138,9 +138,17 @@ public static void SetupCallbacks(TraceEventDispatcher source)
         }
 
         private static TraceEventDispatcher m_currentSource; // used to verify non-concurrent usage
+        
+        internal static void AddCallbackOnProcessCleanup(this TraceEventDispatcher source, Action<TraceProcess> OnProcessRemoved)
```
vancem: We already have a public AddCallbackOnProcessStart, and AddCallbackOnProcessCleanup is really just the flip side.  We should rename this method to AddCallbackOnProcessStart can make it public (and move it next to the other method).

src/TraceEvent/Computers/TraceProcess.cs
```diff
@@ -99,8 +99,8 @@ public static void SetupCallbacks(TraceEventDispatcher source)
 
             kernelParser.ProcessEndGroup += delegate (ProcessTraceData data)
             {
-                data.Process().ProcessEnd(data);
-                // Don't filter them out (not that many, useful for finding command line)
+                data.source.Processes().TerminateProcess(data);
```
vancem: I would prefer a slight reorganization/renaming.    I think 'ProcessEnd() should just do everything (you don't need to make up a new concept at this level of abstraction)    However you should remove ProcessEnd() from 'TerminateProcess() and call the method 'CleanupOldProcess() and call it from EndProcess().    From what I can tell this is equivalent  and is clearer (we are not introducing a new concept, we are simply having the existing EndProcess do some more work (cleaning up old processes).

src/TraceEvent/Computers/TraceProcess.cs
```diff
@@ -384,6 +450,7 @@ internal TraceProcess FindProcessAndIndex(int processID, long timeQPC, out int i
         private TraceLog log;
         private TraceEventDispatcher source;
         internal event Action<TraceProcess> OnInitialized;
+        internal event Action<TraceProcess> OnRemoved;
```
vancem: Lets call this OnProcessStop not on Removed, and go ahead and rename OnInitialized to OnProcessStart.   I realize that this Initialized was probably originally chosen because it is called slightly more places than a process start, but we actually have only one concept (we don't give two kinds of callbacks), and that while slightly inaccurate, calling these 'OnProcessStart' and 'OnProcessStop' will not cause important confusion (the only differences are in weird cases where you don't have a start, and even then, the right thing to do is to make believe you do (which is what happens).  

In short I just want naming to be consistent, and for us to not make fine distinctions unless we really believe it is important for end users to understand the fine distinction.

src/TraceEvent/Computers/TraceProcess.cs
```diff
@@ -342,21 +350,79 @@ internal TraceProcess GetOrCreateProcess(int processID, long timeQPC, bool isPro
...
+                }
+            }
+        }
 
         internal void FireOnInitialized(TraceProcess proc)
         {
             if (OnInitialized != null) OnInitialized(proc);
         }
 
+        internal void FireOnRemoved(TraceProcess proc)
```
vancem: nit, I would not bother with this method  and just inline the body (frankly I would get rid of FireOnInitlialized()) for the same reason.

src/TraceEvent/Computers/TraceProcess.cs
```diff
@@ -322,7 +330,7 @@ internal TraceProcesses(TraceLog log, TraceEventDispatcher source)
         }
         internal TraceProcess GetOrCreateProcess(int processID, long timeQPC, bool isProcessStartEvent = false)
         {
-            Debug.Assert(processes.Count == processesByPID.Count);
+            Debug.Assert(source.DataLifetimeMsec > 0 /* lifetime tracking enabled */ || processes.Count == processesByPID.Count);
```
vancem: Lets make a method DataLifetimeEnabled which is defined to be DataLifetimeMsec > 0 and use it here.   It seems useful in general to make code clearer.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -102,6 +102,17 @@ internal static void OnDotNetRuntimeLoaded(this TraceProcess process, TraceLoade
             if (map[process.ProcessIndex].OnLoaded != null) map[process.ProcessIndex].OnLoaded(runtime);
         }
 
+        internal static void OnDotNetRuntimeRemoved(this TraceProcess process)
```
vancem: Lets stick with the 'Start-Stop' terminology for things.  Below I suggests OnProcessRemoved be renamed OnProcessStop, in the same way I think we should name this OnDotNetRuntimeStop.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -639,6 +663,17 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
...
+
+                        // check if we should apply a lifetime limit to the GC cache
+                        if (source.DataLifetimeMsec > 0 && data.TimeStampRelativeMSec >= stats.GC.NextRelativeTimeStampMsec)
+                        {
+                            // note the next time that lifetime should be applied, to avoid cleaningup too frequently
+                            stats.GC.NextRelativeTimeStampMsec = data.TimeStampRelativeMSec + (source.DataLifetimeMsec/2.0);
+                            // trim the GCs to only include those either incomplete or completed after lifetime
+                            stats.GC.m_gcs = stats.GC.m_gcs.Where(gc => !gc.IsComplete || gc.StartRelativeMSec >= (data.TimeStampRelativeMSec - source.DataLifetimeMsec)).ToList();
+                            // rewrite the index for fast lookup
+                            for (int i = 0; i < stats.GC.m_gcs.Count; i++) stats.GC.m_gcs[i].Index = i;
```
vancem: nit, put the body on its own line.    I do this one line thing for some very short if conditions but I don't think it is a good idea for for loops because a for is longer.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -1043,6 +1081,15 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
 
                     // fire event
                     if (stats.JITMethodStart != null) stats.JITMethodStart(process, _method);
+
+                    // check if we should apply a lifetime limit to the method cache
+                    if (source.DataLifetimeMsec > 0 && data.TimeStampRelativeMSec >= stats.JIT.NextRelativeTimeStampMsec)
+                    {
+                        // note the next time that lifetime should be applied, to avoid cleaningup too frequently
+                        stats.JIT.NextRelativeTimeStampMsec = data.TimeStampRelativeMSec + (source.DataLifetimeMsec/2.0);
```
vancem: Nit.   The interval between updates is of course something we have to decide, and I have argued that it is not that important (because mostly people just want to avoid being unbounded).    Thus I would argue for making less frequent, so I would suggest the 'natural' default should be 1 (that is we keep events between DataLifeTIme and 2*DataLifetime.  It is easy to describe and probably fine.    We should probably add guidance about this in the doc for the DataLifeTimeMsec that this is the expectation.  

But if you really like the 2 factor, I really don't care that much.

jeffschwMSFT: As an interesting anecdote, the team that has this running did a factor of 1/2 as their starting point.  Without offering a knob, this may impact their memory sla.

src/TraceEvent/Computers/TraceProcess.cs
jeffschwMSFT: Done.  The one thing I am considering now is that the semantic of Stop to me says that the callback should be fired once we receive the EndGroup event.  Right now the code is structured to only fire this event if it is purged due to lifetime - that feels like it breaks the Stop contract.  I will think a bit more about this.

src/TraceEvent/Computers/TraceProcess.cs
jeffschwMSFT: Done.

src/TraceEvent/Computers/TraceProcess.cs
jeffschwMSFT: Done.

src/TraceEvent/Computers/TraceProcess.cs
jeffschwMSFT: Done.  I also removed FireOnInitialized to be consistent.

src/TraceEvent/Computers/TraceManagedProcess.cs
vancem: I do worry about scenarios that are overly sensitive to this factor.    In my current conception the contract does not state anything here.   

Anyway as mentioned I don't mind keeping it.    Ideally however, we make it the 'natural' (and easy to explain) value and they can hit their SLA by simply reducing the DataLifeTime value some.

src/TraceEvent/Computers/TraceProcess.cs
jeffschwMSFT: Done.

src/TraceEvent/Computers/TraceManagedProcess.cs
jeffschwMSFT: Done.  The interesting thing there is that there is a public AddCallbackOnDotNetRuntimeLoad, and the Start/Stop methods are internal.  If this were exposed externally perhaps we would use Unload?  Though I feel it is not necessary since ProcessStop has the same semantic as RuntimeStop, where as ProcessStart and RuntimeLoad happen on different events.

src/TraceEvent/Computers/TraceManagedProcess.cs
jeffschwMSFT: Done.

src/TraceEvent/Computers/TraceManagedProcess.cs
jeffschwMSFT: Understood.  For the concrete scenario I know there is a balance of overall memory consumption - which they want low - and the desire to retain a long enough time window to get good GC analysis - needs to be longer.  Given these are separate concepts just lowering the lifetime may not satisfy the scenario.  One place we could land is to remove the latency concept.  This tradesoff cpu time with speed.  I did measure this and in general Computers add CPU cost, which is very dependent on the load on the system.  For a system that has tremendous volume of GCs the overhead is much larger, etc.  Net-net, it sounds like you are OK (though not thrilled).  If this becomes an issue and people want to tune, perhaps we lead with reduce the lifetime.  If that is not satisfactory we can explore removing the concept of latency, as you said it is now opaque to the desire and internally tunable.

src/TraceEvent/Computers/TraceProcess.cs
vancem: This is an good point, and brings up other issues.    Processes can be very long lived, and I don't think we should be purging unterminated processes by default.     I think by default we should only remove things we DO have stops for.  

Now this means that we can still have leaks, but only if our process monitoring is got bugs where we 'miss' the termination of processes.    It is not clear this is important, and if it is I think we should have some sort of more explicit mechanism.

src/TraceEvent/Computers/TraceManagedProcess.cs
vancem: If we have a real need for introducing the latency parameter I am not opposed to doing it.  I just want to be driven there by verified constraints (we have a real world example, and just adjusting the lifetime does not cut it).

src/TraceEvent/Computers/TraceManagedProcess.cs
vancem: Unload has been used in some places (dlls are loaded and unload not started and stopped).   I see that we use 'Loaded' for the runtime start event, so I am OK using Unload for the stop.

src/TraceEvent/Computers/TraceProcess.cs
jeffschwMSFT: I changed the semantic to fire the event at Stop and only purge if a process was stopped.  I think this is fine, because this Lifetime is paired with the Dispatch hook only processes that matter will be analyzed.  If Dispatch hook is enabled then all live processes will be retained.  The big scenario here was for xmlconfiguration.exe which runs and stops all the time.  The current code addresses that case.

src/TraceEvent/Computers/TraceManagedProcess.cs
jeffschwMSFT: Done.

src/TraceEvent/Computers/TraceProcess.cs
vancem: I like that.

jeffschwMSFT:
Thanks @vancem, I have made the changes and am now validating.

jeffschwMSFT:
@vancem feedback accommodated.

Issue #399, opened 2017-10-10
Realtime ETW Session Causes Unhandled Exception


Alois-xx:
Currently any error in the realtime kernel ETW provider causes unhandled exceptions which is not good if I want to use TraceEvent in a GUI application:

https://github.com/Microsoft/perfview/blob/ca0f84651dc5919c21e4071d076a8bd50d9aa404/src/TraceEvent/TraceLog.cs#L3615-L3621

I would propose to add an event to it to catch it when someone has registered to it.

```
        /// <summary>
        /// Register to this event to prevent unhandled thread exceptions when a Realtime ETW kernel trace session could not be started.
        /// This is pretty common on Windows 7 machines where only one Kernel ETW session is allowed.
        /// </summary>
        public event Action<Exception> OnKernelETWSessionThreadException;

                        try
                        {
                            TraceLog.rawKernelEventSource.Process();
                            TraceLog.rawEventSourceToConvert.StopProcessing();
                        }
                        catch(Exception ex)
                        {
                            // prevent unhandled exceptions bubbling up 
                            if(OnKernelETWSessionThreadException != null)
                            {
                                OnKernelETWSessionThreadException(ex);
                            }
                            else
                            {
                                throw new InvalidOperationException("Realtime ETW Kernel thread processor did throw. Please register to TraceLogEventSource.OnKernelETWSessionThreadException to prevent this exception to become unhandled.", ex);
                            }
                        }
```

That should be an easy fix. I can provide a PR.

Alois-xx:
Any update on this? Seems like an easy fix.

vancem:
I want to solve this using PR #513.  

However there is another new feature that may be useful to you.    Basically #513 allows you to catch exceptions outside the call to TraceEventDispatcherProcess(), but does mean that any exception will destroy the Dispatcher.   This is unfortunate if you wish to simply continue (a likely thing for real time sessions).  

There is a way of doing this.  

TraceEventDispatchers support a 'AddDispatchHook'  API.  Here is its docs:
```
        /// Wrap (or filter) the dispatch of every event from the TraceEventDispatcher stream.   
        /// Instead of calling the normal code it calls 'hook' with both the event to be dispatched
        /// and the method the would normally do the processing.    Thus the routine has 
        /// the option to call normal processing, surround it with things like a lock
        /// or skip it entirely.  This can be called more than once, in which case the last
        /// hook method gets called first (which may end up calling the second ...)

[...]

PR #487, opened 2018-01-03
Issue 399 Unhandled exception real time session

Alois-xx:
Fix for issue #399


vancem:
I don't want expose an API like OnKernelETWSessionThreadException which exposes internal details (like the fact that kernel events are processed differently (on a separate thread).

I would like to solve the original #399 issue using the code in PR #513.     Basically what this will do is if there are exceptions during processing, the will be moved over to the thread that the USER called 'Process()' on (thus you can catch exceptions there if you like).   

Finally Please see the comment in #399 on using the 'AddDispatchHook' that may also be of interest.

PR #493, opened 2018-01-05
Enable row color changes

WardenGnaw:
This change allows row colors of the perf data grid to be set to either red, blue, or brown.

A nice feature to have when staring at two different call stack trees.

### DIFF
--- a/src/PerfView/StackViewer/CallTreeView.cs
+++ b/src/PerfView/StackViewer/CallTreeView.cs
@@ -301,6 +301,14 @@ namespace PerfView
         #endregion
     }
 
+    public enum BackgroundColor
+    {
+        Default = 0,
+        Blue = 1,
+        Brown = 2,
+        Red = 3
+    }
+
     /// <summary>
     /// This is basically a CallTreeNode with extra state (state of expand boxes) associated needed for the viewer 
     /// </summary>
@@ -313,6 +321,13 @@ namespace PerfView
             m_treeView.Validate();
 #endif
         }
+
+        public void SetBackgroundColor(BackgroundColor color)
+        {
+            State = ((int)color).ToString();
+        }
+
+        public string State { get; set; }
 
         /// <summary>
         /// Is the node expanded or not.  
--- a/src/PerfView/StackViewer/CallTreeView.cs
+++ b/src/PerfView/StackViewer/CallTreeView.cs
@@ -301,14 +301,6 @@ namespace PerfView
         #endregion
     }
 
-    public enum BackgroundColor
-    {
-        Default = 0,
-        Blue = 1,
-        Brown = 2,
-        Red = 3
-    }
-
     /// <summary>
     /// This is basically a CallTreeNode with extra state (state of expand boxes) associated needed for the viewer 
     /// </summary>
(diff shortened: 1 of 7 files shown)


src/TraceEvent/Stacks/CallTree.cs
```diff
@@ -603,6 +603,11 @@ public string DisplayName
         /// </summary>
         public CallTree CallTree { get { return m_callTree; } }
 
+        /// <summary>
+        /// This sets the background color of the CallTreeNode. For CallTreeNodeBase it should be the default color.
+        /// </summary>
+        public string State { get { return "0"; } }
```
vancem: You should not need to modify CallTree at all.    The CallTree is the 'model' and really should not have GUI state associated with it.  

I note that you have a 'State' property on CallTreeView, which the appropriate place for state like this.   I think you can simply not have this.  If you believe you do need it, we should talk more.

src/PerfView/StackViewer/CallTreeView.cs
```diff
@@ -314,6 +322,13 @@ public void ValidateTree()
 #endif
         }
 
+        public void SetBackgroundColor(BackgroundColor color)
+        {
+            State = ((int)color).ToString();
+        }
+
+        public string State { get; set; }
```
vancem: Why is this not a field called BackgroundColor?   I suspect it was when you though this was the model, but this a view object so it is very reasonable for it to have display things in it.  The type should also be color.

src/PerfView/StackViewer/CallTreeView.cs
```diff
@@ -301,6 +301,14 @@ private int IndexInParent(CallTreeNode node)
         #endregion
     }
 
+    public enum BackgroundColor
+    {
```
vancem: Why do we need a new color enumeration?  Can't we simply use the string name for the color)?  
I think that might allow you to have only one DataTrigger items below.

vancem:
Sorry it took so long for me to get back to review this again.     The changes look good.

I am going to merge this, but I do recommend a followup items:

1.  In the  'Stack Viewer Tips' section of the documentation I would recommend a line or two at highlight the feature.   Use that to link to a small section on the feature.   You might also say it does not work in the byname view (it really is a implementation issue that we did not do it there, and it is probably confusing to people).      You can tell them that the coloring only lasts until the next refresh (including any textbox changes).  

2. I would add keyboard shortcuts for these by modifying RoutedUICommand to add the key strokes I would recommend changing the color Brown to Green and then have the keyboard shortcuts Alt <B> (for background), and then either RGB for the color.

WardenGnaw:
@vancem The follow-up items sound good. However, do you have any good ideas of preserving the coloring between refreshes? @caslan would like the colors preserved when folding.

vancem:
The main problem with preserving the coloring between refreshes is what do you mean by 'the same entry' in the refreshed data.    Technically all data is thrown out and recreated, and there is not event a guarantee that the names are the same.   

We had this problem with the state of the 'opened' state of each node in the tree.   It would have been frustrating if after making a refresh you had to open everything again.  There we did a walk by name.  Starting at the root, as long as the old name was opened and matched the new name, we opened it in the new version (but it WAS explicit work to do this) See the CopyExpandedStateForNode methods).   

It would just be a line or two of code to make this propagate the background color as well as the expanded state.   

Note however that if @caslan change in folding affects any name in the path to root, then of course, you lose the coloring.  

Once you make the color more persistent, you need a way of getting rid of it.    This could be a 'clear' and 'clear all' commands, or it may be more convenient to simply have collapsing the code clear the state of any children (this is effectively what we do with 'isExpanded', when you collapse a node, it 'forgets' the expanded state of all children, so if you reopen it only reopens one level.   It has proven simple and convenient).

PR #507, opened 2018-01-09
Fix sporadic failures in UI testing

### LINKED ISSUE #354: TestIncludeItemOnCallerCalleeTabCallerAsync  fails locally

sharwell:
@vancem I was finally able to reproduce this by setting the display scaling to 200% before running tests.

Closes #354

### DIFF
--- a/src/PerfView.Tests/StackViewer/StackWindowTests.cs
+++ b/src/PerfView.Tests/StackViewer/StackWindowTests.cs
@@ -37,7 +37,7 @@ namespace PerfViewTests.StackViewer
             return TestIncludeItemAsync(KnownDataGrid.ByName);
         }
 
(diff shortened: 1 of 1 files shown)


sharwell:
Closing since it doesn't appear to fix the issue

PR #508, opened 2018-01-09
Change AddDispatch Hook to allow wrapping hooks

vancem:
Jeff checked in AddDispatch but after the review I wanted the API to easily support 'wrapping' (e.g. lock) scenarios.
(That is the hook is called which can set somethign up, call the original dispatch, and then tear it down).
I talked with Jeff, and I he is OK with this change.

### DIFF
--- a/src/TraceEvent/TraceEvent.cs
+++ b/src/TraceEvent/TraceEvent.cs
@@ -2818,26 +2818,25 @@ namespace Microsoft.Diagnostics.Tracing
         public event Action Completed;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


vancem:
@jeffschwMSFT  please let me know if you like this or not.

jeffschwMSFT:
LGTM.  Thanks @vancem

sharwell:
@vancem It looks like the test failure in this pull request is a true failure.
https://ci.appveyor.com/project/sharwell/perfview/build/1.0.664/job/phemuumu2ngch59n/tests

vancem:
Fix in #510

PR #509, opened 2018-01-09
Make Path Parsing logic work with slashes and backslashes.

vancem:
There is some logic that parses paths in TraceEvent.  They were wired to use Windows \
Change it so that / also works.

Note that this makes things consistant with the System.Path (which accepts either as a directory separator.


vancem:
Test failure caused by #508 fixed in #510.

PR #510, opened 2018-01-10
Fix for intermittent failure in ObservableTests in debug builds

vancem:


vancem:
The EventRecord field must be aggressively set to null for some asserts (most notably on in TraceEvent.ID) to work.

vancem:
The debug testing seems to be hung here
```
xUnit.net Console Runner (32-bit .NET 4.0.30319.42000)
  Discovering: PerfViewTests (app domain = on [no shadow copy], method display = ClassAndMethod)
  Discovered:  PerfViewTests (running 13 test cases)
  Starting:    PerfViewTests (parallel test collections = on, max threads = 2)
    PerfViewTests.StackViewer.StackWindowTests.TestIncludeItemOnByNameTabAsync [PASS]
      Output:
        PerfView.TestUtilities.ThrowingTraceListener+DebugAssertFailureException: User did not call EndWork before returning from work body.
        
           at PerfView.TestUtilities.ThrowingTraceListener.Fail(String message, String detailMessage) in C:\projects\perfview\src\PerfView.TestUtilities\ThrowingTraceListener.cs:line 24
``` 
Note that this test was just added back in, and passes on my machine locally.   I am going to merge this and see if this continues to fail with other PRs.

PR #512, opened 2018-01-10
Immediate ExpandAll

adamsitnik:
Whenever I clicked ExpandAll the app was freezing and it took a lot of time to do the job.

I used PerfView to profile itself and it turned out, that by default the expanded node was being selected after expanded. This envolved scroll of the grid, which took a lot of time.

expandall flamegraph1

The solution: don't select nodes when Expanding All.

Result:

expandallafter

### DIFF
--- a/src/PerfView/StackViewer/CallTreeView.cs
+++ b/src/PerfView/StackViewer/CallTreeView.cs
@@ -324,49 +324,63 @@ namespace PerfView
             {
                 if (m_isExpanded == value)
                     return;
+
                 if (value == true)
-                {
-                    ValidateTree();
-
-                    // We are trying to expand the treeNode, add the children after me. 
-                    var children = MakeChildren();
-                    m_treeView.m_flattenedTree.InsertRange(MyIndex + 1, children);
-                    m_isExpanded = true;
-
-                    ValidateTree();
-                    // Auto expand nodes that have only one real child.  (Don't do this for graph nodes as it may not terminate.  
-                    if (children.Count == 1 && !Data.IsGraphNode)
-                    {
-                        var onlyChild = children[0];
-                        if (onlyChild.HasChildren)
-                            onlyChild.IsExpanded = true;
-                    }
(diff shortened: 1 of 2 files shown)


src/PerfView/StackViewer/CallTreeView.cs
```diff
@@ -324,49 +324,63 @@ public bool IsExpanded
...
-                    // set the selected node to my caller (if available) or myself if there is none.  
-                    m_treeView.Select(Data.Caller ?? Data);  
-                }
-                ValidateTree();
+            ValidateTree();
+        }
+
+        private void CollapseNode()
+        {
+            ValidateTree();
```
vancem: Can you put a check (!m_isExpanded) return  at the start to be symmetric with your ExpandNode routine.    Seems safer now that others might call this from locations where the node happened to already be collapsed.

src/PerfView/StackViewer/CallTreeView.cs
```diff
@@ -481,16 +495,17 @@ public string InclusiveMetricByScenarioString
...
                 return;
             if (!expandGraphNodes && Data.IsGraphNode)
                 return;
 
-            IsExpanded = true;
+            ExpandNode(selectExpandedNode);
+
             foreach (var child in VisibleChildren)
-                child.ExpandToDepth(maxDepth - 1, expandGraphNodes);
+                child.ExpandToDepth(maxDepth - 1, expandGraphNodes, selectExpandedNode);
```
vancem: I note that previously the last node was selected, but now no node is selected.    I guess I am OK with no nodes being selected on expand-all so this is OK (nothing to do).

vancem:
One minor point.  Otherwise looks good to me.

adamsitnik:
@vancem done! thanks for a fast review!

vancem:
The fast review was by luck.   I happened to be trying to work off the backlog.  

Thanks for the contribution.

PR #513, opened 2018-01-10
Make TraceLog real time sessions handle exceptions well.

vancem:
Today if there is an exception when processing some events in a TraceLog real time session. it wil tear down the process.

This is because we need another thread to process kernel events and if an exception
happens there it reaches the base of that thread.    Instead of using Threads, use
Tasks which automatically capture exceptions thrown in them and re-raise them
when the Task is waited upon.  This is what we want here, so the fix was very easy.

This fixes issue #399

### DIFF
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -32,6 +32,8 @@ using Microsoft.Diagnostics.Tracing.Session;
 using System.Threading;
 using System.Runtime.InteropServices;
 using Microsoft.Diagnostics.Tracing.EventPipe;
+using System.Threading.Tasks;
+
 
 namespace Microsoft.Diagnostics.Tracing.Etlx
 {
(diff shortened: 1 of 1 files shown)


vancem:
@Alois-xx  Please see if this is a sufficient fix for your issue #399.

vancem:
Test failure
```
  Starting:    PerfViewTests (parallel test collections = on, max threads = 2)
    PerfViewTests.StackViewer.StackWindowTests.TestIncludeItemOnCallerCalleeTabCallerAsync [PASS]
      Output:
        PerfView.TestUtilities.ThrowingTraceListener+DebugAssertFailureException: User did not call EndWork before returning from work body.
        
           at PerfView.TestUtilities.ThrowingTraceListener.Fail(String message, String detailMessage) in C:\projects\perfview\src\PerfView.TestUtilities\ThrowingTraceListener.cs:line 24
        PerfView.TestUtilities.ThrowingTraceListener+DebugAssertFailureException: Called EndWork before work was started
        
           at PerfView.TestUtilities.ThrowingTraceListener.Fail(String message, String detailMessage) in C:\projects\perfview\src\PerfView.TestUtilities\ThrowingTraceListener.cs:line 24
```
Is not due to this change.   Seems to be non-deterministic.

PR #514, opened 2018-01-10
Improved real time Tracelog cleanup code a bit.

vancem:
We have some logic for keeping  some data structures from growing in an unbounded way when using TraceLog with real time sessions (that live forever). 

This logic was early, I have moved it late when the number of events 'in flight' (Not dispatched to the user), is at its smallest, and then only remember a number appropriate for that number of events.   This makes cleanup less costly, but perhaps more importantly insures that do not trim these tables too soon (which could happen before).  

@Alois-xx


vancem:
Closing because history is messed up #519 is the active pull request now.

PR #515, opened 2018-01-12
Fix issue where some traces had CPU samples attributed to the wrong thread

vancem:
This only happened when you had a large number of ETW CPU samples that had no stack associated with them (this is pretty rare, but can happen if you set CPU samples to simply not log stacks).

TraceEvent does not discard CPU samples with no stacks but rather creates a one element broken stack (somethign is better than nothing).   Unfortunately it has a
limit on how many of these it can make (it has to decide a maximum before knowing
how many it needs, so it guesses).

WHat it should have done is give up when this maxium was reachced, but instead it returned the first stack it ever encountered (thus many samples could be attributed
to this one stack.    Fixed this.


vancem:
Note that this code has a work-around in it that is tracked by #517.

PR #516, opened 2018-01-12
Support SourceLink in Microsoft PDBs

vancem:
This allows the 'Goto Source' functionality to work on PDBs (Microsoft or Portable) that uses SourceLink to describe how to fetch the source.  This is the mechanism used in .NET Core and thus Goto Source 'Just works' on .NET Core V2.0.3 or later.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -21,7 +21,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <PerfViewVersion>2.0.1.0</PerfViewVersion>
+    <PerfViewVersion>2.0.2.0</PerfViewVersion>
   </PropertyGroup>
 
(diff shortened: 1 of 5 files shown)


vancem:
Note that this code has a work-around in it that is tracked by #517.

MagicAndre1981:
A bit offtopic, but is SourceLink enabled for Eventsource and TraceEvent NuGet packages?

vancem:
@MagicAndre1981 

In a word.  Yes.   TraceEvent and PerfView have had SourceLink support for a while now.  

The .NET Core libraries are built with SourceLink that points at the GitHub repos for .NET Core and ASP.NET Core.   Thus 'goto source' just works with framework and asp.net.  (Well, that is the way it supposed to work, and does for 2.0, 2.03, and 2.05 version of the framework.

PR #523, opened 2018-01-15
[WIP] read pointer size from pipe file to get x86 working

adamsitnik:
**It's WIP. This requires more testing and approval of the overall idea.**

**Problem:** The pointer size for V1 of EvenPipeEventSource is [hardcoded to 8](https://github.com/Microsoft/perfview/blob/a70c6137ddc2e060204f0ca69f07b03eadb811fc/src/TraceEvent/EventPipe/EventPipeEventSourceV1.cs#L26) hence it does not work for x86.

**Some background:**

CoreCLR writes some important informations to the header of event pipe file ([link](https://github.com/dotnet/coreclr/blob/c1bbdae7964b19b7063074d36e6af960f0cdc3a0/src/vm/eventpipefile.cpp#L49-L58):

```cpp
// Write the header information into the file.

// Write the current date and time.
m_pSerializer->WriteBuffer((BYTE*)&m_fileOpenSystemTime, sizeof(m_fileOpenSystemTime));

// Write FileOpenTimeStamp
m_pSerializer->WriteBuffer((BYTE*)&m_fileOpenTimeStamp, sizeof(m_fileOpenTimeStamp));

// Write ClockFrequency
m_pSerializer->WriteBuffer((BYTE*)&m_timeStampFrequency, sizeof(m_timeStampFrequency));
```

This data is parsed in the ctor of EventPipeEventSourceV1 ([link](https://github.com/Microsoft/perfview/blob/a70c6137ddc2e060204f0ca69f07b03eadb811fc/src/TraceEvent/EventPipe/EventPipeEventSourceV1.cs#L69-#L98)):

```cs
private void ReadEventPipeTimeInfo()
{
	// Read and check the type name
	_deserializer.ReadStringAndVerify("Microsoft.DotNet.Runtime.EventPipeFile");

	// Read tag and check
	_deserializer.ReadByteAndVerify(0x6 /*EndObject tag*/);

	// Read end of event stream marker
	ForwardReference reference = _deserializer.ReadForwardReference();
	_endOfEventStream = _deserializer.ResolveForwardReference(reference, preserveCurrent: true);

	// Read the date and time of trace start.
	var year = _deserializer.ReadInt16();
	var month = _deserializer.ReadInt16();
	var dayOfWeek = _deserializer.ReadInt16();
	var day = _deserializer.ReadInt16();
	var hour = _deserializer.ReadInt16();
	var minute = _deserializer.ReadInt16();
	var second = _deserializer.ReadInt16();
	var milliseconds = _deserializer.ReadInt16();

	_syncTimeUTC = new DateTime(year, month, day, hour, minute, second, milliseconds, DateTimeKind.Utc);

	// Read the start timestamp.
	sessionStartTimeQPC = (long)_deserializer.ReadInt64();
	_syncTimeQPC = sessionStartTimeQPC;

	// Read the clock frequency.
	_QPCFreq = _deserializer.ReadInt64();
}
``` 

**My proposal** (implemented in this PR and in https://github.com/dotnet/coreclr/pull/15871 ): I believe that the header data should be extended with pointer size. One more field, which is parsed by new V2 version of the EventPipeEventSource.

What I don't like about this idea is that I am introducing new version. 

But is it a problem? @vancem @nategraf 

I think that I could also save some more info to avoid similar issues in the future (currently OS version, number of processors and cpu speed are missing ([link](https://github.com/Microsoft/perfview/blob/a70c6137ddc2e060204f0ca69f07b03eadb811fc/src/TraceEvent/EventPipe/EventPipeEventSourceV1.cs#L24-#L27)).

Vance you have mentioned that there are few things that can be improved, so maybe introducing V2 is not such a bad idea (assuming I could also fix some other issues?)


nategraf:
I talked with Vance about this when I added version 2 to the switch case. The guidance was that the V1 suffix was a result of the original development, but not really something we want to continue. Based on that you'd want to put the login into the `EventPipeEventSourceV1` class.

adamsitnik:
@nategraf thanks!

vancem:
Yes as part of the cleanup we should rename this to simply EventPipeEventSource.  The idea is that it is better to simply make the one class 'smart' and understand all versions (because hopefully all new versions should be a extension or modest tweek of the previous one (and thus a well placed if condition should be sufficient).

vancem:
I will take a look at this ;later today and make a concrete suggestion (probably just a pull request).

adamsitnik:
@vancem @nategraf 

I implemented new, more future proof approach and kept backward compat.

So far the events were written to the file right after the initial data (version, time, pointer size etc). The header contained only the info about events stream end position. The reader was reading initial data and events right after it until it was reaching the end.

The problematic scenario: new runtime has one more field in the header, old reader does not expect it, reads the fields it knows about and tries to read new field because it believes it's the begining of the events stream.

What I changed: I added forward reference to the begining of the event stream.

The reader now: reads the fields it's aware of and jumps to the begining of the events stream. If there are any new fields, they are ommited. 

I have bumped the version of the reader (not object) to keep the backward compat.

I also did some refactor and added tests. The key goal for the refactor was to make ctor have no side effects. Factory reads the initial data and creates the event source. Event source has the data ready with deserializer position set to the begining of the events stream.

adamsitnik:
I just discussed the PR with @brianrob 

I know what to change, once I am done I am going to ask Brian for final review.

vancem:
Adam,  

The original EventPipe* code in the TraceEvent library was written by someone who did not understand how to use the FastSerialization code properly.   It does things in a very convoluted way, that you are now extending.   

I realize that it means throwing away some of your work above, I would like to NOT build on this bad foundation but rather FIRST fix up the EventPipe code in the TraceEvent library to follow good patterns and then do the tweeks to make it work for X86 (and as you say, make it more robust to future additions).   

I don't know exactly where you are right now, but you should probably just stop until I can prepare this fixed code.

adamsitnik:
Closing, the code got merged in different PR (https://github.com/vancem/perfview/pull/1)

Issue #470, opened 2017-12-19
EventPipeEventSource needs to support x86 traces


nategraf:
With https://github.com/dotnet/coreclr/pull/14772 EventPipe will work on x86 Windows and therefore TraceEvent should recognise and read x86 traces generated by EventPipe

https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/EventPipe/EventPipeEventSourceV1.cs#L26

vancem:
Nate, can you provide a *.netperf file?

adamsitnik:
@vancem could you add me to the perfview collaborators and assign me to this issue?

vancem:
Done

nategraf:
Here is a trace collected on an x86 Windows build

https://microsoft-my.sharepoint.com/:u:/p/vigraf/EZI2YORpuadIk8Ge4R3NTXEBXbInuMRyY2yszI4C3veYng?e=WgvMri

brianrob:
This was implemented a while back.

PR #528, opened 2018-01-21
Reorganize event pipe

vancem:
This pull request is not meant for merging (yet), but rather to publish nicely so @adamsitnik can get at it.

@adamsitnik : This version builds and I have run some ad-hoc testing on a real *.netperf file.    It is at the point where I think if you built on top of it we would not conflict much (I will probably still tweek here and there, but nothing that wont merge easily.


vancem:
Some things I notices that need fixing when GENERATING the *.netperf file

Things to fix in the file format (for the Version 3 of the format)
1. There is an error in the EventPipeFile object.    It begins with a FastSerializerTags::BeginObject tag but does NOT end with a  FastSerializerTags::Endbject tag and it should.  we need to fix this (I have an ugly hack). 
2. As already discussed, we should have a forward reference tot the START of the event data.   That allows us to add more fields to the end of the File object in a way that will not break older readers.  
3. We should 4 byte align each Event data item.    This allows for efficient parsing.  
4. The GC/BulkSurvivingObjectRanges that the runtime generates seem to create events that exceed ETW's  64K limit.  This is very bad because ETW SILENTLY truncates the event.   Ack!

adamsitnik:
@vancem great, big thanks!

So what I am going to do for V3 now is:

1. Add pointer size and other missing information (see https://github.com/Microsoft/perfview/issues/470 and https://github.com/dotnet/coreclr/issues/11725)
2. Add forward reference to the beginning of the event stream
3. Add close EndObject tag for every event
4. 4-byte align the data serialized to event pipe file

I think that I can also add OpCode, it should be easy (see https://github.com/dotnet/coreclr/issues/11934)

I am going to call you later, I need to know more about the 64k ETW limit

Issue #529, opened 2018-01-22
Failure in the EventPipeParsing/Basic Tests (on the V3 format) on the Release build 


vancem:
@adamsitnik you already know about this.  This is for tacking as I want to merge

I have disabled the entire test for now (which is bad).

PR #530, opened 2018-01-23
remove old module name from the cache after update to new name

adamsitnik:
Fix for #529

Background: test passed for Debug, failed for Release.

It turned out, that due to some `#if DEBUG` the debug build was catching some Assert exceptions and ignoring few events.

[Failing Debug.Assert](https://github.com/Microsoft/perfview/blob/de2c10cd79ac2aa384b5694750719f71563990af/src/TraceEvent/TraceLog.cs#L8225) : `moduleFiles.Count >= moduleFilesByName.Count`

Assert is checking whether the total number of modules is bigger or equal the modules by name. It ofc should never be less.

Why it was failing here: some events were coming with `fileName = null` first. They were added to the dictionary with key = `ManagedModule`. Later on, some more events came, it was possible to get the real file name. The were name was being [updated](https://github.com/Microsoft/perfview/blob/de2c10cd79ac2aa384b5694750719f71563990af/src/TraceEvent/TraceLog.cs#L8200) and added to dictionary. The old name was staying.

I decided to remove the old name from the dictionary in that case. The other, less risky solution would be to remove mentioned `Debug.Assert`.

@vancem what do you think?


vancem:
Hello Adam,

I have learned through hard-won experience that the best solutions to tricky bugs, make the number of state transitions as small as possible.   That is, ideally something is set exactly once, but if we can't have that it should be monotonic (it keeps getting 'better' as information arrives).   We I want to avoid is having state 'change over time in a non-monotonic way (it is first one thing than another).  I am trying to apply this principle here.  Thus I would like to avoid the table ever having 'bad' information (a key of 'ManagedModule' if indeed the module did not have a file name (its filePath is null).  
```
It was failing here because some events were coming with fileName = null first. They were added to the dictionary with key = ManagedModule
```
I looked over the code and I don't really see how that happens, and you walk me through it?    

Really the solution I want is that we never put this first 'bad' entry into the table and thus everything is fine.

adamsitnik:
@vancem very good point! I am going to provide a fix with avoiding the problem

To reproduce:

Set conditional breakpoint [here](https://github.com/Microsoft/perfview/blob/de2c10cd79ac2aa384b5694750719f71563990af/src/TraceEvent/TraceLog.cs#L8246) with condition `moduleFile.fileName == null`

image

and debug this test:

image

src/TraceEvent/TraceLog.cs
```diff
@@ -8385,9 +8385,20 @@ IEnumerator<TraceModuleFile> IEnumerable<TraceModuleFile>.GetEnumerator()
...
+                return;
+            
+            var nameBefore = moduleFile.FilePath; // if fileName == null this property returns "ManagedModule"
             moduleFile.fileName = fileName;
             if (moduleFilesByName != null)
+            {
                 moduleFilesByName[fileName] = moduleFile;
+
+                // we have changed the name, so we also need to remove the old name from the dictionary
+                if (moduleFilesByName.ContainsKey(nameBefore))
```
sharwell: :bulb: There is no need for `ContainsKey` here. `Remove` is a NOP if the key isn't present.

adamsitnik: @sharwell thanks! I did not know that

adamsitnik:
@vancem done

vancem:
Thanks for doing this @adamsitnik.    Feel free to merge when tests pass (they are pending now).

adamsitnik:
@vancem tests have passed, but I don't have the rights to merge ;)

PR #531, opened 2018-01-23
add test to prove that old bug is gone, fixes #461

### LINKED ISSUE #461: EventPipeEventSource does not properly handle EventSource events
When using the TraceEventDispatcher with handlers registered on Dynamic.All and UnhandledEvents, EventSource events appeared in UnhandledEvents and the provider name did not resolve.

adamsitnik:
After recent @vancem rewrite of the consumer part of Event Pipe the #461 bug is gone

I added the test to make sure that it never comes back.

Vance what is the difference of using `EventPipeEventSource` directly (like it was before) vs `var traceLog = new TraceLog(TraceLog.CreateFromEventPipeDataFile(eventPipeFilePath))` ? Whis is this the recomended usage?


sharwell:
👏 Tests

PR #532, opened 2018-01-24
Changes for the next version of the EventPipe file format

vancem:
This change contains what I think should be the format for Version 3 (the next verision) of the EventPipe format.

@adamsitnik - This is what we talked about.  I need you to make coreclr emit this and then test this code.  This PR is meant simply to give an idea of what changes I wanted.   In particular
 
* Does not use stream location (but rather a field in the MetaData data) as the MetaData ID.  
* Removes the need for forward references (introduces a new EventPipeEventBlock class)
* Adds support for alignment (Does assume that all event payloads sizes will be rounded up).

Note that I have written a doc on the file format (see PR)

Finally Please see the TODOs 
* https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/EventPipe/EventPipeEventSource.cs#L280 
* https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/EventPipe/EventPipeEventSource.cs#L294

As mentioned there seems to be redundancy.  Indeed I don't actually think we need the metaData length field (we already have the length of all the fields.      In notices you added a Keywords field here  https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/EventPipe/EventPipeEventSource.cs#L272 that seems redudant with the field here https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/EventPipe/EventPipeEventSource.cs#L292

It would be good to clean this up for V3.   As always let me know if you have questions or think any of this is a bad idea, or have other suggestions.


adamsitnik:
@vancem  I am reading your PR. I am going to contact you in case something is not clear or I get stuck somehow

Speaking of `keywords` it is redundant when [there is](https://github.com/Microsoft/perfview/blob/10d1f92b242c98073b3817ac5ee6d98cd595d39b/src/TraceEvent/EventPipe/EventPipeEventSource.cs#L278) a metadata provided for given event type. If there is no metadata, it's needed.

Btw I tried to find the place where metadata is created. It seems like in the auto-generated `dotnetruntime.cpp`

```cpp
void InitDotNETRuntime()
{
    EventPipeProviderDotNETRuntime = EventPipe::CreateProvider(SL(DotNETRuntimeName));
    EventPipeEventGCStart = EventPipeProviderDotNETRuntime->AddEvent(1,1,0,EventPipeEventLevel::Informational);
    EventPipeEventGCStart_V1 = EventPipeProviderDotNETRuntime->AddEvent(1,1,1,EventPipeEventLevel::Informational);
    EventPipeEventGCStart_V2 = EventPipeProviderDotNETRuntime->AddEvent(1,1,2,EventPipeEventLevel::Informational);
}
```

The problem is that mine (Windows x86) has no metadata for any of the events. How can I find the place where the metadata is created? I would like to know what put things into this byte array.

vancem:
Hello Adam,

Adding @brianrob  who can tell you if I am leading you astray.  

I took a look at the details of the .NET Runtime's EventPipe stuff to answer your questions.   

First there are two main categories of event definitions in the code base today.  The first is for C++ code that you notice are generated at compile time.  These don't have information detailed meta-data information (like the event name or the parameter information),   This is sub-optimal but is not really necessary because readers like PerfView have built in knowledge of these events and thus don't need this information.   The other place we define events is in MANAGED code (see https://github.com/dotnet/coreclr/blob/master/src/mscorlib/shared/System/Diagnostics/Tracing/EventSource.cs#L690).   This is where we have non-null 'metaData' blobs.  

From an EventPipe perspective, the 'Provider' of event should also provide a Meta-data blob, and that blob completely describes the event.   However EventPipe needs some of the information (eventID version level and keywords), and so forces the client to provide these directly even though this information is ALSO in the meta data blob.   This is the source of the redundancy.  

This is actually pretty easy to fix.   The problem here entirely in EventPipeConfiguration::BuildEventMetadataEvent which is responsible for creating the payload for meta-data events.     The problem is that this method slavishly emits its state as it is represented in the class (thus its id, version keywords etc AND the meta-data blob (if present).   However as we have noted, this is redundant, as the meta-data ALREADY has everything in it.   So mostly we just need to dump the meta-data blob.

But what if we have no meta-data blob?   Conceptually we ALWAYS have a meta data blob, it is just that we have allowed some events to give us the 'bare minimum' information and not give us a blob, but we can ALWAYS form a blob from the information we have.

Thus this routine should the data in the blob (you don't need to actually serialize the blob length, just use it as the payload length).   If the blob is not present, you have to make one by using the data you do have to make something that looks just like a meta-data blob.  (You have all the information you need except EventName (you can make it the empty string), and the parameters (you will just emit a count of 0).  

This cleans up the file format.  Notice an Metadata blob in the file format now is EXACTLY the meta-data blob that is passed to DefineEvent (that is what Provider provide).  This is as it should be.  

Let me know if this is not clear.  I could give you actual code if you need it, but hopefully the explanation above is enough.  

Thanks

Vance

vancem:
I misspoke above.  The meta-data blob in the file format is not exactly the meta-data blob for an event.  
That is because there is one piece of information in the file format blob that is not in the event blob, namely the ProviderName.    Arguably EventPipe may want to just fix that so that the Meta-data blob also contains the ProviderName, but as  smaller first step BuildEventMetadataEvent  can simply concatenate them.

adamsitnik:
@vancem big thanks for the answer! and the docs are just great.

vancem:
I have pushed a commit that fixes the reader to read what I think is a sane (simple non-redundant) format for V3.

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -383,6 +434,77 @@ public EventPipeEventMetaData(PinnedStreamReader reader, int length, int fileFor
...
+        private void ReadEventMetaData(PinnedStreamReader reader, int fileFormatVersionNumber)
+        {
+            int eventId = (ushort)reader.ReadInt32();
+            _eventRecord->EventHeader.Id = (ushort)eventId;
+            Debug.Assert(_eventRecord->EventHeader.Id == eventId);  // No truncation
+
+            EventName = reader.ReadNullTerminatedUnicodeString();
+
+            // Deduce the opcode from the name.   
+            if (EventName.EndsWith("Start", StringComparison.OrdinalIgnoreCase))
```
brianrob: There has been a request to have EventPipe emit the opcode.  Do we want to just do this in CoreCLR instead of deducing it here?

vancem:
Adding the opcode to the meta-data is reasonable.

brianrob:
Overall, I like the changes - they definitely improve the file format.

One thing with respect to metadata blobs - it is a good idea to reduce the duplication that exists today especially for EventSource where metadata blobs exist, but I should call out that we do need to implement metadata blobs for the C++ events as well because they're going to be needed in order to take those events and dispatch them to EventListeners.  We're working towards this as part of the cross-platform performance monitoring work.  Thus, rather than having the runtime build a default metadata blob, we might like to just require a metadata blob and build the metadata blob for all events now.  This will likely just be some helper code to build the blobs and then modification to the python scripts that build the event firing functions and data structures.  This is tracked by https://github.com/dotnet/coreclr/issues/15126.

brianrob:
Ok, then let's just add opcode - tracked by https://github.com/dotnet/coreclr/issues/11934.

vancem:
It is good to hear that we have a scenario that forces us to 'do the right thing' and have meta-data for all eventPipe events.  

I note that this file format change is starting to get 'big' and I think there is value in doing things in stages (thus keeping this delta under control).     

Note that synthesizing a meta-data blob out of the fields of an EventPipeEventInstance is roughly what BuildEventMetadataEvent does today.   Thus it is very cheap (basically rearranging lines in BuildEventMetadataEvent).  We should get that working and checked in, and then tackle the other issue as a separate PR.

vancem:
One more thing.   Again, to decouple the work, we can add the opcode to the file format with an unconditional value of 0 right now and then fill it in for real as a separate step.  Again that allows us to finalize the file format (which has already got lots of delta to it), without making the CoreCLR delta very large.

vancem:
Actually on second, thought, I think we should NOT add the opcode.  I will give more detail in #15126.

brianrob:
I'm on-board with doing this in stages.  You're right, it's probably simpler (and not much throw-away work) to just do the default metadata payload first.

I'm also OK with not including opcode and closing https://github.com/dotnet/coreclr/issues/11934.

adamsitnik:
I am currently implementing the CoreCLR part and wondering about the blocks.

The question is: should I build the block first and then write it to the disk? Or start block and write events immediately to the disk?

Build & write to disk as a whole:

Pros:
+ when I finish the build I know few extra things that can be helpful (time of the first event, time of the last event, exact size)
+ less disk writes
+ my pedantic feelings are ok: I don't manually write Start and End tags, I just build entire object and call `FastSerialize()` when the block is full or the destructor got called

Cons:
+ delay - events are not immediately written to the disk
+ slightly higher memory usage (but never bigger than the size of a single block!)

@vancem Vance what do you think?

vancem:
The way EventPipe works is that when you actually write an event it does into a in-memory circular buffer (split by thread, so that it can be lock free, thus events are 'scattered' on many per-thread buffers.

Then, at some other time (often shutdown), you dump the circular buffer. This is what EventPipeBufferManager::WriteAllBuffersToFile does, and does this by repeatedly calling EventPipeFile::WriteEvent on different instances.  

So this tells us some things
1. The circular buffer is much bigger (many meg) than a file format block (probably 100K or less), so buffer is no big deal.  
2. That delay is not an issue (we are not doing real time here).  

Moreover you need to know the size of an EventBlock BEFORE the data is written, you basically need to buffer that block (that is the price for making the format streamable).    This forces buffering.

Now you could imagine avoiding a copy by making WriteAllBuffersToFile  accumulate say several hundred Events can calling a new EventPipeFile::WriteManyEvents with an array of EventInstancePointers.     This is interesting but probably not worth the complexity.

The easy thing to do is to modify EventPipeFile::WriteEvent to simply have a reasonably large output buffer that it accumulates events into (and another buffer that it accumulates event Metadata into).  When they get big enough (or on file close or a flush command), you make a EventBlock out of what you have buffered. 

In short this is your option 1 (in fact you HAVE to defer the write until you know the exact size, the only question was were you going to be 'smart' about it and not have another buffer that you copy into and then write from.  

I am saying you should have this buffer (it only lives while you are flushing it is fine), and extra copy (this is not that performance critical and we CAN fix it if we care). 

Simplicity first!

Vance

vancem:
Closing this as Adam has taken this commit and finished it off.

Issue #533, opened 2018-01-24
Realtime Tracing Crashes In Marshal.FreeHGlobal


Alois-xx:
This code sporadically fails in 

```
        private unsafe void DispatchClonedEvent(TraceEvent toSend)
        {
            TraceEvent eventInRealTimeSource = this.realTimeSource.Lookup(toSend.eventRecord);
            eventInRealTimeSource.userData = toSend.userData;
            eventInRealTimeSource.eventIndex = toSend.eventIndex;           // Lookup assigns the EventIndex, but we want to keep the original. 
            this.realTimeSource.Dispatch(eventInRealTimeSource);
            GC.KeepAlive(toSend);       // Keep the event alive because eventInRealTimeSource has pointers into its data.  

            // Optimization, remove 'toSend' from the finalization queue.  
            Debug.Assert(toSend.myBuffer != IntPtr.Zero);
            System.Runtime.InteropServices.Marshal.FreeHGlobal(toSend.myBuffer);   // Application thread is still here
            GC.SuppressFinalize(toSend);
        }
```
The crash occurs in the finalizer thread:

```
ntdll!KiUserExceptionDispatcher+0x26
ntdll!RtlReportCriticalFailure+0x88
ntdll!RtlpReportHeapFailure+0x2f
ntdll!RtlpLogHeapFailure+0xa7
ntdll!RtlFreeHeap+0xd8
KERNELBASE!LocalFree+0x53
DomainNeutralILStubClass.IL_STUB_PInvoke(IntPtr)
System.Runtime.InteropServices.Marshal.FreeHGlobal(IntPtr)(+0xe IL,+0x19 Native)
Microsoft.Diagnostics.Tracing.TraceEvent.Finalize()(+0x18 IL,+0x20 Native)
clr!FastCallFinalize+0x6d
clr!MethodTable::CallFinalizer+0x139
clr!CallFinalizer+0xa6
clr!FinalizerThread::DoOneFinalization+0x84
clr!FinalizerThread::FinalizeAllObjects+0xa6
clr!FinalizerThread::FinalizeObjectsOnShutdown+0x17
clr!ManagedThreadBase_DispatchInner+0x71
clr!ManagedThreadBase_DispatchMiddle+0x7e
clr!ManagedThreadBase_DispatchOuter+0x5b
clr!ManagedThreadBase::FinalizerBase+0x33
clr!FinalizerThread::FinalizerThreadStart+0x19c
[...]

PR #535, opened 2018-01-26
Fix finalization race

vancem:
This is a fix for https://github.com/Microsoft/perfview/issues/533


vancem:
I am removing changes I accidentally included.

PR #540, opened 2018-01-27
Enable ETWClrProfiler signing

vancem:


vancem:
@sharwell  I could use your MSBUILD expertise here. 

It would be nice if all the executable pieces of PerfView.exe would be microsoft signed.   In order to do this I need to add the following line to the project file 

<PackageReference Include="MicroBuild.Core" Version="0.2.0" />

We get the error that is shown in the CI run
```
C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\Microsoft\NuGet\15.0\Microsoft.NuGet.targets(186,5): error : Your project is not referencing the ".NETFramework,Version=v4.0" framework. Add a reference to ".NETFramework,Version=v4.0" in the "frameworks" section of your project.json, and then re-run NuGet restore. [C:\projects\perfview\src\EtwClrProfiler\ETWClrProfilerX64.vcxproj]
```
Looking around in other projects I did find the following property that may be of use as it was used in the init-tools.cmd and seemed to be associated with microbuild. 

    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>

I tried adding that but it did not help (also a word of warning, the error is 'sticky' after doing a msbuild /t:restore, you have to clean out the build after each experiment.

One possibility is to use 'old style' nuget references, but I would prefer not to do that.   If you have no advice I will try to reverse engineer what we do for other packages that need to sign native components, but hopefully you can short circuit that work.

Thank.

sharwell:
@vancem Let me check with some members of the infrastructure team next week. I believe they are moving towards a unified approach across open source projects.

sharwell:
@jaredpar @tannergooding There's a request here to enable signing on this repository. Can you take a look?

@vancem I was sent a link to this page, but I don't think it's enough to get started all by itself:

> https://github.com/dotnet/roslyn-tools/tree/master/src/SignTool

vancem:
@sharwell  to be clear I am already using Microbuild and getting things signed in the PerfView repository that way.   It works.  

However, in order to use Microbuild I need to include the MicroBuild.Core nuget package.   This works fine for C# projects, but I now want to compile a C++ project that is used by PerfView.     When I do the same thing to the *.vcxprog file it give the error above (Your project is not referencing the ".NETFramework,Version=v4.0" framework).   

Now I know that Microbuild can be used to sign native components, so there is a way around this particular error, and that is what I am looking for.   

Now I always have the option of making a dummy C# project whose only job is to sign the DLL the VC project created, but I was hoping for something more straightfoward than that.

vancem:
I have confirmed that the build of V2.0.5 of PerfView has a signed EtwClrProfiler dlls now too.

PR #544, opened 2018-01-29
Danlev1 dev/danlev/extension support for diagsessions

vancem:
This is a rebasing of pull request #464 by @danlev1

Dan, I had to do some history rewriting that has effectively broken all the commits before 12/20/2018.   I have rebased your commits on top of the new branch, but you should be sure to hard reset your local branch of master to the current one going forward (or simply delete and recreate your fork, which will also do it).  

I see that you did take my advice I gave back in Dec.   This is great.    

I want to make sure that this is still a useful delta.  Please let me know.   Also confirm that it is complete (I did not lose anything)> 

Finally please respond to the comments below (it is probably easier at this point that I make the edits.

### DIFF
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -493,6 +493,33 @@ namespace PerfView
                     m_singletonStackSource.Viewer.Focus();
             }
         }
+
+        public void OpenOnExistingWorker(Window parentWindow, StatusBar worker, Action doAfter = null)
+        {
+            if (!m_opened)
+            {
+                Action<Action> continuation = OpenImpl(parentWindow, worker);
+
+                m_opened = true;
+                FirePropertyChanged("Children");
+
+                IsExpanded = true;
+                var defaultSource = GetStackSource();
+                if (defaultSource != null)
+                    defaultSource.IsSelected = true;
+
+                if (continuation != null)
+                    continuation(doAfter);
+                else
+                    doAfter?.Invoke();
+            }
+            else
+            {
+                if (m_singletonStackSource != null && m_singletonStackSource.Viewer != null)
+                    m_singletonStackSource.Viewer.Focus();
+            }
+        }
+
         public override void Close()
         {
             if (m_opened)
@@ -6275,6 +6302,8 @@ table {
         }
         // public override string DefaultStackSourceName { get { return "CPU"; } }
 
+
+
         public TraceLog GetTraceLog(TextWriter log, Action<bool, int, int> onLostEvents = null)
         {
             if (m_traceLog != null)
@@ -8057,7 +8086,7 @@ table {
                 });
 
                 // Get all ETL files
-                AddEtlResourcesAsChildren(worker, dhPackage);
+                AddEtlResourcesAsChildren(parentWindow, worker, dhPackage);
             }
 
             return null;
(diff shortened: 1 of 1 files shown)


src/PerfView/PerfViewData.cs
```diff
@@ -493,6 +498,38 @@ public override void Open(Window parentWindow, StatusBar worker, Action doAfter
                     m_singletonStackSource.Viewer.Focus();
             }
         }
+
+        public void OpenOnExistingWorker(Action doAfter = null)
+        {
+            this.OpenOnExistingWorker(GuiApp.MainWindow, GuiApp.MainWindow.StatusBar, doAfter);
+        }
+
+        public void OpenOnExistingWorker(Window parentWindow, StatusBar worker, Action doAfter = null)
```
vancem: This feels like it should not be public (since it uses StatusBar, which is hard to get).   I would like it to be private if possible.

src/PerfView/PerfViewData.cs
```diff
@@ -8114,11 +8153,13 @@ private void AddResourcesAsChildren(StatusBar worker, DhPackage dhPackage, strin
         /// <summary>
         /// Adds child files from ETL resources in the DhPackage
         /// </summary>
-        private void AddEtlResourcesAsChildren(StatusBar worker, DhPackage dhPackage)
+        private void AddEtlResourcesAsChildren(Window window, StatusBar worker, DhPackage dhPackage)
         {
             ResourceInfo[] resources;
             dhPackage.GetResourceInformationByType("DiagnosticsHub.Resource.EtlFile", out resources);
 
+            var firstEtlHasBeenProcessed = false;
```
vancem: Dead code?  Can I remove?

src/PerfView/PerfViewData.cs
```diff
@@ -8114,11 +8153,13 @@ private void AddResourcesAsChildren(StatusBar worker, DhPackage dhPackage, strin
         /// <summary>
         /// Adds child files from ETL resources in the DhPackage
         /// </summary>
-        private void AddEtlResourcesAsChildren(StatusBar worker, DhPackage dhPackage)
+        private void AddEtlResourcesAsChildren(Window window, StatusBar worker, DhPackage dhPackage)
```
vancem: We don't use the window parameter, so this edit seems pointless.  Can I remove?

src/PerfView/PerfViewData.cs
```diff
@@ -493,6 +498,38 @@ public override void Open(Window parentWindow, StatusBar worker, Action doAfter
                     m_singletonStackSource.Viewer.Focus();
             }
         }
+
+        public void OpenOnExistingWorker(Action doAfter = null)
```
vancem: It is not clear what OpenOnExistingWorker does and what the scenario is.  Can you describe why you need it?   I would like to avoid a new API if I can.  Also, generally speaking public opens methods (which are called from the GUI thread typically should not run 'OpenImpl' on that thread (that is what the status bar is for, you run it on the 'background' and the status bar lets the users kill if if he wants).   

My knee-jerk reaction is that 'OpenOnExisting' should not exist because 'Open' will just 'do the right thing' if it is already open (if that is not true, we should fix it so it is true).  

Comments?

sharwell:
@vancem We have permissions in place to update the original PR if you want (rather than create a new replacement PR). Let me know if you want me to update the original PR with the changes you made here.

vancem:
I believe this issue was fixed with PR #687.   If not we can reopen.

Issue #552, opened 2018-02-02
Renaming unmerged *.etl file does not rename *.clrRundown.etl and *.kernel.etl


congyiwu:
1. Collect a PerfView trace w/ Zip and Merge options disabled.  This creates three files:
   * PerfViewData.etl
   * PerfViewData.clrRundown.etl
   * PerfViewData.kernel.etl
2. Right click on PerfViewData.etl in the file tree view, and click Rename
3. Rename to Renamed.etl.

I expected PerfViewData.clrRundown.etl and PerfViewData.kernel.etl to be renamed as well but they were not.

vancem:
Fixed by #559 (you will have to wait for release 2.0.5 or later)

PR #559, opened 2018-02-09
Make rename rename all unmerged files.

vancem:
Fixes issue #552

### DIFF
--- a/src/PerfView/MainWindow.xaml.cs
+++ b/src/PerfView/MainWindow.xaml.cs
@@ -1698,10 +1698,9 @@ namespace PerfView
             // rename all the other variations of the unmerged file
             if (selectedFileIsEtl && targetPath.EndsWith(".etl", StringComparison.OrdinalIgnoreCase))
             {
-                string selectedFileBase = selectedFilePath.Substring(0, selectedFilePath.Length - 4);
-                foreach (string relatedFile in System.IO.Directory.GetFiles(Path.GetDirectoryName(selectedFilePath), Path.GetFileName(selectedFileBase) + ".*"))
+                foreach (string relatedFile in System.IO.Directory.GetFiles(Path.GetDirectoryName(selectedFilePath), Path.GetFileNameWithoutExtension(selectedFilePath) + ".*"))
                 {
-                    Match m = Regex.Match(relatedFile, @"\.((clr.*)|(user.*)|(kernel.*)\.etl)$");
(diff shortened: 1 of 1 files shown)


src/PerfView/MainWindow.xaml.cs
```diff
@@ -1678,14 +1678,34 @@ private void DoRename(object sender, ExecutedRoutedEventArgs e)
...
+                targetPath = Path.ChangeExtension(targetPath, ".etl");
+
+            // Do the move.  
+            FileUtilities.ForceMove(selectedFilePath, targetPath);
+
+            // rename all the other variations of the unmerged file
+            if (selectedFileIsEtl && targetPath.EndsWith(".etl", StringComparison.OrdinalIgnoreCase))
+            {
+                string selectedFileBase = selectedFilePath.Substring(0, selectedFilePath.Length - 4);
+                foreach (string relatedFile in System.IO.Directory.GetFiles(Path.GetDirectoryName(selectedFilePath), Path.GetFileName(selectedFileBase) + ".*"))
```
sharwell: :bulb: `Path.GetFileNameWithoutExtension` here, and then `selectedFileBase` intermediate variable is unnecessary

src/PerfView/MainWindow.xaml.cs
```diff
@@ -1678,14 +1678,34 @@ private void DoRename(object sender, ExecutedRoutedEventArgs e)
...
+                {
+                    Match m = Regex.Match(relatedFile, @"\.((clr.*)|(user.*)|(kernel.*)\.etl)");
```
sharwell: :bulb: Would be good to use `$` at the end of this expression to force it to be at the end.

src/PerfView/MainWindow.xaml.cs
```diff
@@ -1678,14 +1678,34 @@ private void DoRename(object sender, ExecutedRoutedEventArgs e)
...
+            // Do the move.  
+            FileUtilities.ForceMove(selectedFilePath, targetPath);
+
+            // rename all the other variations of the unmerged file
+            if (selectedFileIsEtl && targetPath.EndsWith(".etl", StringComparison.OrdinalIgnoreCase))
+            {
+                string selectedFileBase = selectedFilePath.Substring(0, selectedFilePath.Length - 4);
+                foreach (string relatedFile in System.IO.Directory.GetFiles(Path.GetDirectoryName(selectedFilePath), Path.GetFileName(selectedFileBase) + ".*"))
+                {
+                    Match m = Regex.Match(relatedFile, @"\.((clr.*)|(user.*)|(kernel.*)\.etl)$");
```
sharwell: 💭 Do we want to make this a case-insensitive match?

PR #560, opened 2018-02-10
support Guids in Event Pipe metadata

adamsitnik:
Fix for https://github.com/dotnet/coreclr/issues/16105


vancem:
Looks good to me.

PR #563, opened 2018-02-15
User interface improvements for the CCW view for .NET Native

vancem:
Added the type name for the CCW (just had the ID before),
The refcount value that was expressed as a long is really two ints, (two ref counts).   Fix the code to recognise this.

Added a warning, because the CCW events doe NOT seem to be complete/correct.   The events have the currnet ref-count
and it is not consitant with the number of inc and dec events.   Needs to be investigated, but that is a runtime issue.


vancem:
This is a followon to PR #477 

+ @GeorgePlotnikov

vancem:
You should use the 'ignore whitespace (use the url https://github.com/Microsoft/perfview/pull/563/files?w=1) to look at the file diff.   Otherwise its much larger than it needs to be.

PR #568, opened 2018-02-20
Consume TraceEvent.SupportFiles 1.0.13

brianrob:
This updates TraceEvent to use a netstandard1.6 OSExtensions.dll for the netstandard1.6 version of TraceEvent.  Prior to this change TraceEvent is inadvertently using a net45 version of OSExtensions.dll which does not work on versions of .NET Core prior to 2.0.

Note - the addition of the dotnet-core MyGet feed is because the TraceEvent.SupportFiles package is posted on MyGet.

### DIFF
--- a/Nuget.config
+++ b/Nuget.config
@@ -2,5 +2,6 @@
 <configuration>
   <packageSources>
     <add key="Local" value=".\src\NugetSupportFiles" />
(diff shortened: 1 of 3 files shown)


vancem:
LGTM

Issue #478, opened 2017-12-25
x64 Realtime Stacks not working after switch to RuntimeInformation.OSArchitecture


Alois-xx:
Commit f3f49e9f contains the switch to System.Runtime.InteropServices.RuntimeInformation, Version=4.0.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a. That version breaks x64 Realtime stackwalks because RuntimeInformation.OSArchitecture returns x86 on x64 which is clearly an error.

The version is supplied with .NET 4.7.1 which forwards the type to mscorlib where it is defined as 
```
public static Architecture OSArchitecture
{
	[SecuritySafeCritical]
	get
	{
		object obj = RuntimeInformation.s_osLock;
		lock (obj)
		{
			if (!RuntimeInformation.s_osArch.HasValue)
			{
				Win32Native.SYSTEM_INFO sYSTEM_INFO;
				Win32Native.GetNativeSystemInfo(out sYSTEM_INFO);
				RuntimeInformation.s_osArch = new Architecture?(RuntimeInformation.GetArchitecture(sYSTEM_INFO.dwOemId));
			}
		}
		return RuntimeInformation.s_osArch.Value;
	}
}
```
That has several problems. 

1. dwOemId is at least on Win10 the wrong property of the struct to determine the OS architecture 
   The code in RuntimeInformation.GetArchitecture wrongly shifts it by 16 bits which results always in 0.
  
    ushort num = (ushort)(sysinfo.dwOemId >> 16);
   That must be fixed inside mscorlib inside .NET 4.7.1.

2. The supplied Nuget package for System.Runtime.InteropServices.RuntimeInformation 4.3 supplies an assembly for users who do not have .NET 4.7.1 installed but that version relies on API sets 

```
[DllImport("api-ms-win-core-sysinfo-l1-2-0.dll")]
internal static extern void GetNativeSystemInfo(out Interop.mincore.SYSTEM_INFO lpSystemInfo);
```
 which breaks Windows 7 users. API Sets are a feature introduced with Windows 8.

Point 2 seems to be already addressed with 4.3.0-preview1-24530-04 but that will still fail on all machines where .NET 4.7.1 is installed. 

[...]

PR #574, opened 2018-02-22
Ptr size fix

vancem:
Fixes issue #478

Basically we are using a RuntimeInformation.GetArchitecture property which seems to be busted on Desktop runtime.

Can avoid calling this API and thus avoids the problem.


sharwell:
Does this mean we can remove this?

https://github.com/Microsoft/perfview/blob/0ed309f46e7051736c21eb7d21a4222fd0109b95/src/TraceEvent/TraceEvent.csproj#L66

src/TraceEvent/ETWReloggerTraceEventSource.cs
```diff
@@ -407,7 +407,7 @@ private unsafe void Initialize(TraceEventNativeMethods.EVENT_RECORD* rawData)
                         // first event (real time case).   This is really a problem, as we need that information, but we will go ahead and
                         // try to initialize as best we can. 
 
-                        m_source.pointerSize = (RuntimeInformation.OSArchitecture == Architecture.X64 || RuntimeInformation.OSArchitecture == Architecture.Arm64) ? 8 : 4;
+                        m_source.pointerSize = IntPtr.Size;
```
sharwell: ❔ What is the result you want when PerfView is run as a 32-bit process on a 64-bit operating system? The approach taken here will produce `4` in this case.

sharwell:
:bulb: It appears you can rebase this pull request onto commit 0ed309f46e7051736c21eb7d21a4222fd0109b95 and drop the two extra merge commits. It looks like your fork is tracking ahead of the reference repository unnecessarily.

src/TraceEvent/ETWReloggerTraceEventSource.cs
sharwell: :bulb: My recommendation is to use this:

```csharp
#if !NETSTANDARD1_6
m_source.pointerSize = Environment.Is64BitOperatingSystem;
#else
m_source.pointerSize = (RuntimeInformation.OSArchitecture == Architecture.X64 || RuntimeInformation.OSArchitecture == Architecture.Arm64) ? 8 : 4;
#endif
```

vancem: Yes, I came to a similar conclusion.

sharwell: It sounds like it will make sense to bring back a **netstandard2.0** build, since that one could use the working API and most users will be able to leverage it. The maintenance cost would be almost nothing since the API surface area is already covered by other targets we are building.

vancem:
This technique does not work, will try a variation of @sharwell  suggestion

PR #575, opened 2018-02-23
Sql additions

vancem:
Some enhancements to improve diagnosis of SQL scenarios

1. Translate SQL BeginExecute EndExecute into start stop (so durations are calculated).   We do this for other events as well.
2. Add the SQL command text (first 50 chars) if it is available in the StartStop Activity view.


.gitignore
```diff
@@ -203,3 +203,4 @@ PerfView.exe
 src/TraceEvent/EventCounterHandler.cs
 OSExtensions.cs
 password.txt
+documentation/internal
```
sharwell: :bulb: If `internal` is a directory, this line should end with a `/`:

```
documentation/internal/
```

src/PerfView/EtwEventSource.cs
```diff
@@ -210,6 +210,22 @@ public override void ForEach(Func<EventRecord, bool> callback)
                             opcode = TraceEventOpcode.Stop;
                     }
 
+                    if (data.ProviderGuid == systemDataProviderGuid)
+                    {
+                        corelationOptions = CorelationOptions.UseActivityID;
+                        if ((int)data.ID == 1)          // BeginExecute
+                        {
+                            task = (TraceEventTask)0xFFFE;      // unique task but used for both BeginExecute and EndExecute.
```
sharwell: :bulb: It looks like this constant is a common pattern; seems like a candidate for extraction to a named constant.

src/PerfView/EtwEventSource.cs
```diff
@@ -309,6 +325,7 @@ enum CorelationOptions
             UseActivityID = 2,
         }
         static Guid httpServiceProviderGuid = new Guid("dd5ef90a-6398-47a4-ad34-4dcecdef795f");
+        static Guid systemDataProviderGuid = new Guid("6a4dfe53-eb50-5332-8473-7b7e10a94fd1");
```
sharwell: :bulb: `readonly`

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -920,8 +920,23 @@ void FixAndProcessAdoNetEvents(TraceEvent data)
                         return;
                     StartStopActivity creator = GetCurrentStartStopActivity(thread, data);
 
-                    Debug.Assert(data.payloadNames[1] == "dataSource");
-                    var extraStartInfo = data.PayloadValue(1) as string;
+                    string extraStartInfo = null;
+                    if (3 < data.payloadNames.Length)
```
sharwell: :bulb: Placing constants on the right side of expressions is typically more readable in C#

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -920,8 +920,23 @@ void FixAndProcessAdoNetEvents(TraceEvent data)
...
+                    {
+                        Debug.Assert(data.payloadNames[1] == "dataSource");
+                        Debug.Assert(data.payloadNames[3] == "commandText");
+                        string dataSource = data.PayloadValue(1).ToString();
+                        string commandText = data.PayloadValue(3).ToString();
+                        extraStartInfo = "DS=" + dataSource;
+                        if (!string.IsNullOrEmpty(commandText))
+                        {
+                            // Don't make it too long
```
sharwell: 💡 To me, the truncation seems like a normal coding pattern for "don't make it too long". More interesting to me is the value of `extraStartInfo`. I would guess that this allows people to identify specific queries getting executed from within stack trace views, but this may or may not be correct. If there is a concise explanation for the net value of this block it might be a good replacement for this comment.

src/PerfView/EtwEventSource.cs
```diff
@@ -210,6 +210,22 @@ public override void ForEach(Func<EventRecord, bool> callback)
                             opcode = TraceEventOpcode.Stop;
                     }
 
+                    if (data.ProviderGuid == systemDataProviderGuid)
+                    {
+                        corelationOptions = CorelationOptions.UseActivityID;
+                        if ((int)data.ID == 1)          // BeginExecute
```
sharwell: 💭 May be good to create internal static classes for constants like this. Future cleanup item since so many existing places already have the same code pattern.

src/TraceEvent/Computers/StartStopActivityComputer.cs
vancem: I use a convention that all comparisons go from small to big.

sharwell: `std::less` :trollface:

PR #576, opened 2018-02-23
Pointer size fix

vancem:
This fixes issue #478 

Basically it implements GetOSPointerSize() in a way that works around bugs see $478 for more

It also removes some unused ifdefs.

### DIFF
--- a/src/TraceEvent/ETWTraceEventSource.cs
+++ b/src/TraceEvent/ETWTraceEventSource.cs
@@ -458,7 +458,8 @@ namespace Microsoft.Diagnostics.Tracing
             Debug.Assert(_QPCFreq != 0);
             if (pointerSize == 0 || IsRealTime)  // We get on x64 OS 4 as pointer size which is wrong for realtime sessions. Fix it up. 
             {
-                pointerSize = (RuntimeInformation.OSArchitecture == Architecture.X64 || RuntimeInformation.OSArchitecture == Architecture.Arm64) ? 8 : 4;
+                pointerSize = GetOSPointerSize();
+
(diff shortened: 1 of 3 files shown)


vancem:
@Alois-xx  before submitting this, I if it is easy for you to test that this fixes your issue, I would appreciate you doing that test (if that is hard, at least you should confirm that it will work in theory)

src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -533,7 +551,7 @@ public static long GetUTCTimeAsQPC(DateTime utcTime)
                 return qpcTime._GetUTCTimeAsQPC(utcTime);
             }
 
-            #region private
+#region private
```
sharwell: :bulb: Probably unintentional formatting change

Alois-xx:
@vancem: I will take a look tomorrow.

Alois-xx:
@vancem: The pointer size issue is fixed with that one. But I have noticed that the exception handling does not work out with an older fix where I have proposed a event to get notified when ETW tracing could no be started. 

Tracelog.cs
`            if (TraceLog.rawKernelEventSource != null)
                {
                    kernelTask = Task.Factory.StartNew(delegate
                    {
                        TraceLog.rawKernelEventSource.Process();
                        TraceLog.rawEventSourceToConvert.StopProcessing();
                    });
                    kernelTask.Start();
                }`

a) If on Win7 two instances are started I never get an exception back because .Process will never be called at all. 
b) The current Task based code does not work because you call Task.Factory.StartNew and you then call start again where the task will then throw a Task Already started exception. Looks like this was never tested.

Issue #571, opened 2018-02-21
Go To Source Is Not Working


Alois-xx:
Symbol lookup for our IIS hosted Symbol Server is not working. 

\perfview\src\TraceEvent\Symbols\SymbolReader.cs

A fix is to include a valid User Agent string
```
                           m_log.WriteLine("FindSymbolFilePath: In task, sending HTTP request {0}", fullUri);

                            var httpClient = new HttpClient();

                            // Add a user-agent header
                            var headers = httpClient.DefaultRequestHeaders;
                            // Use Chrome User Agent strings
                            headers.UserAgent.Add(new System.Net.Http.Headers.ProductInfoHeaderValue("Mozilla", "5.0"));
                            headers.UserAgent.Add(new System.Net.Http.Headers.ProductInfoHeaderValue("AppleWebKit", "537.36"));
                            headers.UserAgent.Add(new System.Net.Http.Headers.ProductInfoHeaderValue("Chrome", "41.0.2228.0"));
                            headers.UserAgent.Add(new System.Net.Http.Headers.ProductInfoHeaderValue("Safari", "537.36"));
                            var responseTask = httpClient.GetAsync(fullUri);
                            responseTask.Wait();
                            var response = responseTask.Result;

                            alive = true;
                            if (!canceled)
                            {
                                if (contentTypeFilter != null && !contentTypeFilter(response.Content.Headers.ContentType.ToString()))
                                    throw new InvalidOperationException("Bad File Content type " + response.Content.Headers.ContentType.ToString() + " for " + fullDestPath);

                                using (var fromStream = response.Content.ReadAsStreamAsync().Result)
                                {
                                    if (CopyStreamToFile(fromStream, fullUri, fullDestPath, ref canceled) == 0)
                                    {
                                        File.Delete(fullDestPath);
                                        throw new InvalidOperationException("Illegal Zero sized file " + fullDestPath);
                                    }
                                }
                                successful = true;
                            }
```
With that change it works like a charm for source server indexed pdbs from our symbol server. I needed to switch over to HttpClient because the HttpRequest user agent property does not exist on .NET Core.

vancem:
Interesting.   When the source server stuff was first put in we had a user-agent header 
```
                        var fullUri = serverPath + "/" + pdbIndexPath.Replace('\\', '/');
                        try
                        {
                            var req = (System.Net.HttpWebRequest)System.Net.HttpWebRequest.Create(fullUri);
[...]

PR #577, opened 2018-02-23
Fix problem fetching from some symbol servers.

vancem:
This is bringing back old code that was removed because it was not available on NetStandard 1.6
and seemed to not matter.

Fixes issue #571


vancem:
@Alois-xx  please comment if you think this is enough to fix #571.

Alois-xx:
@vancem 👍 LGTM.

PR #585, opened 2018-03-01
Improved performance of drawing Flame Graph

adamsitnik:
As reported internally drawing Flame Graphs for some very big data sets was too slow (could even take few minutes). 

This PR solves the problem by replacing `FrameworkElement` with `DrawingVisual`.

`DrawingVisual` is simpler and faster. However, it does not support ToolTips OOTB so I had to implement this. I don't feel good about it, but I did not find any better idea.

For provided `.etl` file the time for drawing flame graph got from 5 minutes to 2 seconds.

I have also improved the way text is scaled and wrapped.

### DIFF
--- a/src/PerfView/StackViewer/FlameGraph.cs
+++ b/src/PerfView/StackViewer/FlameGraph.cs
@@ -5,7 +5,6 @@ using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Media;
 using System.Windows.Media.Imaging;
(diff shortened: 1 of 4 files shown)


vancem:
It looks like you just draw things instead of using WPF objects.  That is reasonable. 

I have not looked at the code closely, but it is isolated to flamegraph, which I consider your domain.   

I am OK with it.

PR #587, opened 2018-03-02
Trace parser manifest mods

andd64:
Manfiest files with a task opcode of "win:DCStart" and "win:DCStop" aren't correctly parsed.  Add these two names.


vancem:
I would like to fix this another way.   If look here

https://msdn.microsoft.com/en-us/library/windows/desktop/aa383956(v=vs.85).aspx

You will see that the official name for a data collection start and stop is DC_Start and DC_Stop.

You ran into problems because PerfView DumpRegisteredManifest use DCStop and DCStart instead.

In commit 09424e6b (3/2/2018) I have fixed PerfView so that it uses DC_Start and DC_Stop, which fixes the issue.

Thus I think we can do nothing in TraceParserGen.

Issue #446, opened 2017-11-20
Make GroupPats dropdown to remember user input between sessions


ilabutin:
Right now GroupPats dropdown remembers all user input, but does not keep it between sessions. It does remember last used value only.

I realize that when switching between different scenarios/apps this might not be valuable, but I primarily use PerfView for our applications where I would like to have my own groupings remembered.

So, suggestion is to let GroupPats to remember user input between sessions (but of course keep current logic about automatically adjusting "Just My App" option).

If this suggestion is OK, then I can do implementation.

sharwell:
My main concern here is how to handle a case where a user wants to get back to default values. It almost makes me wish I could have multiple saved presets, where each preset sets the grouping/folding/exclusions values.

vancem:
As @sharwell indicates, persistence of values can be negative (many of the entries are not worth saving and probably add to clutter), but that can undoubtedly be fixed (but please first suggest exactly what the experience is so we can agree we understand its positives and negatives)

Also note that there is already a feature in PerfView that should help you in the mean time.  

If you right click in the you will find 'Filter Params' -> 'Copy Filter Params' which basically copies all the filter parameters to the clipboard as XML.  You can also an operation to basically copy them back.   This allows you to remember 'special purpose' sets of parameters.   I realize this is not quite the experience you want, but may relieve some of the pressure.

ilabutin:
Idea of presets also came to my mind, but I thought it's too much. But let me describe how I see presets may be implemented.

_View preset_ is a combination of grouping/folding/exclusion fields. PerfView UI is modified to add extra dropdown of preset selection to the left of existing controls:
![perfview_2017-11-23_11-37-48](https://user-images.githubusercontent.com/1869103/33164578-822b8510-d044-11e7-99f5-4e0b6d96ccba.png)

Logic around this dropdown and old fields is following (I tried to put it into the sequence that users will experience, from basic usage to more advanced).

1. By default (if there we no saved presets) PerfView offers standard presets (those that are now in GroupPats dropdown).

1. Existing "Set as Default Grouping/Folding" menu item works as it is now - it just uses current view preset as default in all new views.

1. If one of predefined view presets is selected and user starts editing grouping/folding pattern manually, then preset selection dropdown is changed to "\*<Custom>" and save button becomes active. "\*" symbol in front of preset name means that current state of editboxes will not be preserved when window is closed.

[...]

PR #590, opened 2018-03-05
Added support for presets of grouping/folding

ilabutin:
Initial implementation of the grouping/folding presets (#446).
Presets only include Grouping/folding/include/exclude and do not include Memory Stack specific fields.
"Builtin" GroupPats are kept as is (only updated in the text boxes), so PerfView restart will restore them even if they were overwritten by dropdown history.

This implementation works for my needs, but I'm opened to suggestion about how to improve it for wider acceptance.

### DIFF
--- a/src/PerfView/StackViewer/StackWindow.xaml.cs
+++ b/src/PerfView/StackViewer/StackWindow.xaml.cs
@@ -767,7 +767,7 @@ namespace PerfView
             });
             symPathDialog.Show();
         }
-        private void DoSetDefaultGroupingFolding(object sender, RoutedEventArgs e)
+        private void DoSetStartupPreset(object sender, RoutedEventArgs e)
         {
             App.ConfigData["DefaultFoldPercent"] = FoldPercentTextBox.Text;
             App.ConfigData["DefaultFoldPat"] = FoldRegExTextBox.Text;
@@ -2926,10 +2926,8 @@ namespace PerfView
             // Make up a trivial call tree (so that the rest of the code works).  
             m_callTree = new CallTree(ScalingPolicy);
 
-            // Finalize the Preset menu (add default commands)
-            var presets = App.ConfigData["Presets"];
-            m_presets = Preset.ParseCollection(presets);
-            FinishPresetMenu();
+            // Configure the Preset menu (add standard commands and known presets)
+            ConfigurePresetMenu();
 
             StackWindows.Add(this);
 
@@ -3436,8 +3434,11 @@ namespace PerfView
             diffMenuItem.Items.Add(helpMenuItem);
         }
 
-        private void FinishPresetMenu()
+        private void ConfigurePresetMenu()
         {
+            var presets = App.ConfigData["Presets"];
+            m_presets = Preset.ParseCollection(presets);
+
             foreach (var preset in m_presets)
             {
                 var presetMenuItem = new MenuItem();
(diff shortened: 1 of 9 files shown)


ilabutin:
@vancem Do you have any plans to merge it or do I need to update it somehow? I wish to have this in perfview master, so I can use latest and greatest version also with my improvements.

vancem:
Sorry for letting this languish.   This is the kind of thing I have to pull down, build and play with a bit as well as look over the code.   I will try to do this late today.

vancem:
OK I finally got to this (sorry for taking so long).  

First, I like the new capability.  Thanks for taking the time to add it (and your patience in me getting to it).

1.  The help link called 'Present Name' in the 'New Preset' dialog does not exist in the help (so it just opens the help.
2. Same issue with the link called 'Presets' on the 'Manage Presets dialog. 
3. In the help you have Preset ->> I think you just mean Preset -> 
4. move the File -> Set As Default Folding into the presets. preset menu as 'Set Startup Preset'.   
5. If you put a _ before the letter in a Menu item name you make it a shortcut.   You should do this for your Menu items.   That way you can do Alt-P Alt-M to manage your presets.  

This is extra, but consider it:

Users will want some sort of export-import capability.   I would recommend a 'export' and 'import' button on the Manage Presets, which write the XML you wrote to UserConfig, to a file.   You can either have a file dialog or a textbox (But you should pick a nice name like PerfViewPresets.xml so that in the common case (so you can have your presets on another machine).

src/PerfView/StackViewer/StackWindow.xaml.cs
```diff
@@ -2926,6 +2926,11 @@ private void FinishInit()
             // Make up a trivial call tree (so that the rest of the code works).  
             m_callTree = new CallTree(ScalingPolicy);
 
+            // Finalize the Preset menu (add default commands)
+            var presets = App.ConfigData["Presets"];
+            m_presets = Preset.ParseCollection(presets);
+            FinishPresetMenu();
```
vancem: Nit.   Please rename FinishPresetMenu to 'ConfigurePresetMenu' and move the two lines  above this call into that method.

ilabutin:
Thanks for comments. I'll apply them. One question, though.

> which write the XML you wrote to UserConfig, to a file.

Are you saying you want me to write XML in the following form (same as UserConfig, but only presets)?
```
<ConfigData>
  <Presets>....</Presets>
</ConfigData>
```

where `...` is giant string where all presets are encoded separated by fancy separators (`####` and `$$$$`) that I hope will never appear in the real preset (see `Preset.ParseCollection` method).

Or you want full UserConfig to be exported?
In any case what do you think should happen during the import if file contains presets with the same name as already created? Replace or ignore or do something else?

@vancem

vancem:
I think the most interesting scenario is just the ability to share your presents (typically with yourself (on another machine), but it is also useful for users to share with each other. 

The format of the exported data, is flexible.   I would not wrap it in a <ConfigData> element because this is by design just presets.      I was not thinking of full user config, (I am assuming that is too user-specific).   

> In any case what do you think should happen during the import if file contains presets with the same name as already created? Replace or ignore or do something else?

I think if they are identical, it should just ignore the input.  If it is different you have several options, and I leave that to you (we can always change it).   To keep it simple, I would make it ignore new things (and warn in the log) since users can delete existing ones if they want them overridden.

ilabutin:
Applied all comments: help links, users guide, menu items and export/import.

I have one question though: I found how to write to log (`App.CommandProcessor.LogFile.WriteLine()`), but I wasn't able to find out how to update status bar (e.g. after import or export was completed). Is it done just by passing StatusBar instance into Manage Presets dialog or there exist a different way doing that?

@vancem

vancem:
I will take a look later today or tomorrow.  

> I have one question though: I found how to write to log (App.CommandProcessor.LogFile.WriteLine()), but I wasn't able to find out how to update status bar (e.g. after import or export was completed). Is it done just by passing StatusBar instance into Manage Presets dialog or there exist a different way doing that?

The Status bar has a feature that if you log a line with some text enclosed in [] then that also updates the status bar.   (This logic is in StatusTextWriter.UpdateStatus).

ilabutin:
Thanks for tip. Pushed these changes.
Do I need to squash all commits?

vancem:
Don't bother squashing the commits.

vancem:
Thanks again for your contribution.

Vance

PR #594, opened 2018-03-07
Migrate ServiceProfiler changes - baseline

xiaomi7732:
Migrate some easy fixes from Service Profiler to PerfView for TraceEvent library.

### DIFF
--- a/src/TraceEvent/Computers/StartStopActivityComputer.cs
+++ b/src/TraceEvent/Computers/StartStopActivityComputer.cs
@@ -167,9 +167,11 @@ namespace Microsoft.Diagnostics.Tracing
                     return;
                 }
 
-                // OK so now we only have EventSources with start and stop opcodes. There are a few that don't follow
-                // conventions completely, and then we handle the 'normal' case 
-                if (data.ProviderGuid == FrameworkEventSourceTraceEventParser.ProviderGuid)
+                // OK so now we only have EventSources with start and stop opcodes.
+                // There are a few that don't follow conventions completely, and then we handle the 'normal' case 
+                if (data.ProviderGuid == MicrosoftApplicationInsightsDataProvider)
+                    FixAndProcessAppInsightsEvents(data);
+                else if (data.ProviderGuid == FrameworkEventSourceTraceEventParser.ProviderGuid)
                     FixAndProcessFrameworkEvents(data);
                 else if (data.ProviderGuid == MicrosoftWindowsASPNetProvider)
                     FixAndProcessWindowsASP(data, threadToLastAspNetGuids);
@@ -656,7 +658,7 @@ namespace Microsoft.Diagnostics.Tracing
             {
                 uint nibble = (uint)(*bytePtr >> 4);
                 bool secondNibble = false;              // are we reading the second nibble (low order bits) of the byte.
-            NextNibble:
+                NextNibble:
                 if (nibble == (uint)NumberListCodes.End)
                     break;
                 if (nibble <= (uint)NumberListCodes.LastImmediateValue)
@@ -733,6 +735,10 @@ namespace Microsoft.Diagnostics.Tracing
         private static readonly Guid AdoNetProvider = new Guid("6a4dfe53-eb50-5332-8473-7b7e10a94fd1");
         private static readonly Guid MicrosoftWindowsHttpService = new Guid("dd5ef90a-6398-47a4-ad34-4dcecdef795f");
         private static readonly Guid MicrosoftDiagnosticsDiagnosticSourceProvider = new Guid("ADB401E1-5296-51F8-C125-5FDA75826144");
+
+        // EventSourceName: Microsoft-ApplicationInsights-Data
+        // Reference for definition: https://raw.githubusercontent.com/Microsoft/ApplicationInsights-dotnet/e8f047f6e48abae0e88a9c77bf65df858c442940/src/Microsoft.ApplicationInsights/Extensibility/Implementation/RichPayloadEventSource.cs
+        private static readonly Guid MicrosoftApplicationInsightsDataProvider = new Guid("a62adddb-6b4b-519d-7ba1-f983d81623e0");
 
         // The main start and stop logic.  
         unsafe private StartStopActivity OnStart(TraceEvent data, string extraStartInfo = null, Guid* activityId = null, TraceThread thread = null, StartStopActivity creator = null, string taskName = null, bool useCurrentActivityForCreatorAsFallback = true)
@@ -932,7 +938,7 @@ namespace Microsoft.Diagnostics.Tracing
                         if (!string.IsNullOrEmpty(commandText))
                         {
                             if (50 < commandText.Length)
-                                commandText = commandText.Substring(0, 50-3) + "...";
+                                commandText = commandText.Substring(0, 50 - 3) + "...";
                             extraStartInfo = extraStartInfo + ",CMD=" + commandText;
                         }
                     }
@@ -962,6 +968,7 @@ namespace Microsoft.Diagnostics.Tracing
                 string taskName = data.PayloadByName("EventName") as string;
                 if (taskName == null)
                     return false;
+
                 string extraInfo = null;
 
                 // In the converter from 'DiagnosticSource' to ETW, we have some default configurations.
@@ -970,11 +977,9 @@ namespace Microsoft.Diagnostics.Tracing
                 // The following code is used to make the call stack looks better (converting 'Activity2Start' to 'SQLCommand')
                 // and showing the SQL command.
                 // Details:
-                //https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSourceEventSource.cs
+                // https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSourceEventSource.cs
 
-                
-                var args = data.PayloadByName("Arguments") as DynamicTraceEventData.StructValue[];
-                if (args != null)
+                if (data.PayloadByName("Arguments") is DynamicTraceEventData.StructValue[] args)
                 {
                     var sb = Utilities.StringBuilderCache.Acquire(64);
                     bool first = true;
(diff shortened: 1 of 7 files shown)


src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -733,6 +735,9 @@ internal static unsafe string CreateActivityPathString(Guid guid)
         private static readonly Guid AdoNetProvider = new Guid("6a4dfe53-eb50-5332-8473-7b7e10a94fd1");
         private static readonly Guid MicrosoftWindowsHttpService = new Guid("dd5ef90a-6398-47a4-ad34-4dcecdef795f");
         private static readonly Guid MicrosoftDiagnosticsDiagnosticSourceProvider = new Guid("ADB401E1-5296-51F8-C125-5FDA75826144");
+        private static readonly Guid MicrosoftApplicationInsighsDataProvider = new Guid("a62adddb-6b4b-519d-7ba1-f983d81623e0");
+
+        private static readonly Guid GUID_ONE = new Guid("00000001-0000-0000-0000-000000000000");
```
vancem: Do we know when this guid gets generated?   I would like to know more about it (we should probably put it in a comment here.

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -957,53 +962,117 @@ void FixAndProcessAdoNetEvents(TraceEvent data)
...
-                            if (!string.IsNullOrEmpty(valueStr))
+                        case "Microsoft.EntityFrameworkCore.BeforeExecuteCommand":
+                            // In the converter from 'DiagnosticSource' to ETW, we have some default configurations.
+                            // By default, SQL (EntityFramework) DiagnosticSource event will be converted to 'Activity2Start' and 'Activity2Stop' events
+                            // of provider 'Microsoft-Diagnostics-DiagnosticSource'. Meanwhile, the sql command will be in the payload.
+                            // The following code is used to make the call stack looks better (converting 'Activity2Start' to 'SQLCommand')
+                            // and showing the SQL command.
+                            // Details:
+                            //https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSourceEventSource.cs
+
```
vancem: Is this comment line really not indented?   If so go ahead an indent it.  It is OK if the line is long.

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -957,53 +962,117 @@ void FixAndProcessAdoNetEvents(TraceEvent data)
...
+                return;
+            }
+            else
+            {
+                Debug.Assert(false, $"{data.TaskName} is not recognized");
+                return;
+            }
+        }
+
+        bool IsTrivialActivityId(Guid g)
```
vancem: Go ahead and make this static and move the declaration of GUID_ONE here.   If the language supported it we would make the static a local to this method, and placing them next to each other is our best approximation of that.

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -957,53 +962,117 @@ void FixAndProcessAdoNetEvents(TraceEvent data)
...
+                            // Details:
+                            //https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSourceEventSource.cs
+
+                            if (0 < data.PayloadNames.Length)
                             {
-                                if (!first)
-                                    sb.Append(' ');
-                                sb.Append(key).Append("=").Append('"').Append(valueStr).Append('"');
+                                var args = data.PayloadByName("Arguments") as DynamicTraceEventData.StructValue[];
+                                if (args != null)
```
vancem: Make a method DiagnosticSourceArgsToString(StructValue[] which encapulates this.  body here.  

I notice that the old code did this transformation for all DiagnosticSources, and you do it for only one particular one (BeforeExecuteCommand) Is that intentional?

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -957,53 +962,117 @@ void FixAndProcessAdoNetEvents(TraceEvent data)
...
             catch (Exception ex)
             {
                 Trace.WriteLine("Threw exception while processing DiagnosticSource start events: " + ex.Message);
             }
-            return false;
+
+            return result;
+        }
+
+        void FixAndProcessAppInsightsEvents(TraceEvent data)
```
vancem: I am OK with adding these for the ones we know about, but this is clearly a place where we should have an extensiblity point since client will care about ones we have not done, or want something different.   We can think about this.

src/TraceEvent/ZippedETL.cs
```diff
@@ -358,13 +357,13 @@ internal static List<string> GetNGenPdbs(string etlFile, SymbolReader symbolRead
...
 #endif
 
-            /// <summary>
-            /// ZippedETLReader is a helper class that unpacks the ZIP files generated
-            /// by the ZippedETLWriter class.    It can be smart about placing the 
-            /// symbolic information in these files on the SymbolReader's path so that
-            /// symbolic lookup 'just works'.  
-            /// </summary>
-        public class ZippedETLReader
+    /// <summary>
```
vancem: I would like to fix this differently.    We are going to make a new class ZippedDataReader and we are going to mark ZippedETLReader as obsolete.    We will implement ZippedETLReader by inheriting from ZippedDataReader.  

ZippedDataReader should have a property DataFileName, and should not have wired in dependencies.  It should simply look for the first file at the top level as the data file (and we can have another property called 'Extension' which allows us to look only for a file of a particular suffix).     

Now the Netperf case can use the ZippedDataReader without wired in changes.

src/TraceEvent/Computers/StartStopActivityComputer.cs
xiaomi7732: Bad view from github review page. The comment is indented well.

src/TraceEvent/Computers/StartStopActivityComputer.cs
xiaomi7732: Seems a regression for .NET Core trace introduced by commit 5a7b2a666998b54618a09b6b10707dc8eac1d26b in PerfView. I'll fix it.

src/TraceEvent/Computers/StartStopActivityComputer.cs
xiaomi7732: Are you saying this is a good point to allow user providing sort of pre-processors for the TraceEvents? That I agree.

src/TraceEvent/ZippedETL.cs
xiaomi7732: Synced offline. I'll leave this out from this effort and handle it from the callers' end. After all, I agree that it is a misuse for unzipping the netperf file using the ZippedETLReader.

src/TraceEvent/Computers/StartStopActivityComputer.cs
xiaomi7732: Fixed in i2.

src/TraceEvent/Symbols/SymbolPath.cs
```diff
@@ -254,6 +254,8 @@ public void ToXml(TextWriter writer, string indent)
         /// </summary>
         public static bool ComputerNameExists(string computerName, int timeoutMSec = 700)
         {
+            if (computerName == s_lastComputerNameLookupFailure)
```
sharwell: ❗️ This is an incorrect revert of #260

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -734,6 +736,10 @@ internal static unsafe string CreateActivityPathString(Guid guid)
         private static readonly Guid MicrosoftWindowsHttpService = new Guid("dd5ef90a-6398-47a4-ad34-4dcecdef795f");
         private static readonly Guid MicrosoftDiagnosticsDiagnosticSourceProvider = new Guid("ADB401E1-5296-51F8-C125-5FDA75826144");
 
+        // EventSourceName: Microsoft-ApplicationInsights-Data
+        // Reference for definition: https://raw.githubusercontent.com/Microsoft/ApplicationInsights-dotnet/e8f047f6e48abae0e88a9c77bf65df858c442940/src/Microsoft.ApplicationInsights/Extensibility/Implementation/RichPayloadEventSource.cs
+        private static readonly Guid MicrosoftApplicationInsighsDataProvider = new Guid("a62adddb-6b4b-519d-7ba1-f983d81623e0");
```
sharwell: :bulb: Insigh**t**s

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -960,8 +966,8 @@ bool TryProcessDiagnosticSourceStartEvents(TraceEvent data)
             try
             {
                 string taskName = data.PayloadByName("EventName") as string;
-                if (taskName == null)
-                    return false;
+                if (taskName == null) return false;
```
sharwell: ❕ Previous layout was correct

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -987,8 +991,7 @@ bool TryProcessDiagnosticSourceStartEvents(TraceEvent data)
                             string valueStr = value.ToString();
                             if (!string.IsNullOrEmpty(valueStr))
                             {
-                                if (!first)
-                                    sb.Append(' ');
+                                if (!first) sb.Append(' ');
```
sharwell: ❕ Previous layout was correct

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -1006,6 +1009,54 @@ bool TryProcessDiagnosticSourceStartEvents(TraceEvent data)
             return false;
         }
 
+        void FixAndProcessAppInsightsEvents(TraceEvent data)
+        {
+            Debug.Assert(data.ProviderGuid == MicrosoftApplicationInsighsDataProvider);
+            Debug.Assert(data.Opcode == TraceEventOpcode.Start || data.Opcode == TraceEventOpcode.Stop);
+
+            if (data.TaskName.Equals("Request", StringComparison.OrdinalIgnoreCase))
```
sharwell: ❔ Does this really come in different cases? If not, can we use the correct case here and switch to the faster `StringComparison.Ordinal`?

src/TraceEvent/Symbols/SymbolPath.cs
xiaomi7732: Sorry, I over-sighted it. I'll fix it.

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -1006,6 +1009,54 @@ bool TryProcessDiagnosticSourceStartEvents(TraceEvent data)
...
+                        OnStart(data, extraInfo);
+                    }
+                }
+                else
+                {
+                    Debug.Assert(data.Opcode == TraceEventOpcode.Stop);
+                    OnStop(data);
+                }
+            }
+            else if (data.TaskName.Equals("Operation", StringComparison.OrdinalIgnoreCase))
```
sharwell: :memo: Same note here about performance of string comparisons

src/TraceEvent/Computers/StartStopActivityComputer.cs
xiaomi7732: Thanks!

src/TraceEvent/Computers/StartStopActivityComputer.cs
xiaomi7732: :-) Fixed.

src/TraceEvent/Computers/StartStopActivityComputer.cs
xiaomi7732: 👍 Reviewed the code from AI side. I agree, there is little chance the contract being broken. I'll fix it.

**Details:**
https://raw.githubusercontent.com/Microsoft/ApplicationInsights-dotnet/e8f047f6e48abae0e88a9c77bf65df858c442940/src/Microsoft.ApplicationInsights/Extensibility/Implementation/RichPayloadEventSource.cs

xiaomi7732:
I'll squash the commits to merge small details after all code issues are addressed! Do NOT complete this PR before I finish the squash, please.

vancem:
This looks OK.   @xiaomi7732  I can squash the commits as part of the checkin, but I am also happy to wait for you to do that yourself on the branch.

xiaomi7732:
The code is rebased & squashed against the tip of the master branch. Please merge. Thank you all for the review!

PR #595, opened 2018-03-07
Menu for flame graph and some minor UX improvements

adamsitnik:
I just wanted to make context menu working for Flame Graph to make it possible to use all the nice features of PerfView directly from the Flame Graph tab.

But I realized that all the operations exposed by Context Menu in StackWindow are based on the current selected **text**. Which means that if the user selects the wrong cell, the command is not going to work.
I checked the code and verified that every row of every grid available in StackWindow has an `Item` property which exposes `CallTreeNodeBase` or `CallTreeViewNode` (which exposes `CallTreeNodeBase` via the `Data` property). I also have it in Flame Graph for the currently selected box.

So the first big change was to move from `SelectedText` to SelectedNodes.
How it improves the User Experience? No matter which cell is selected users can perform all operations for given row.

The other change was to move away from checking if given command can be executed and printing error in case it's not. I unified all the actions and followed the `CanExecute` pattern. So when user opens context menu/clicks key combination the commands are verified. If the conditions are not met, the menu items are disabled/pressing the key shortcut has no effect.

image

I also found some "hidden gems". It turned out that it's possible to perform some of the operations for Inclusive Samples or Exclusive Samples. The user could click "Drill Into" and if "Exc" column was selected, the operation was being executed for exclusive samples. 
It's not intuitive, without reading the docs/code I would not know about it. So I introduced new commands and now it's explicit: "Drill Into Inclusive Samples" and "Drill Into Exclusive Samples"

image

I have tested all my changes and learned a LOT of useful things about PerfView when I was doing this. This tool is simply amazing!

@vancem  I know that this PR is rather a PITA to merge (a lot of changes, high risk, only UX improved) but I hope that we can somehow work together and merge it.


vancem:
I will take a look but it may take a few days..  Gui stuff is always worrisome because it can be fragile.

jwdj:
This feature (menu in flamegraph) would be very cool to have. Could one of you @vancem or @adamsitnik please take another attempt at merging this PR?
I really like PerfView and I am glad it is open source now so everyone can contribute.

brianrob:

adamsitnik:
@brianrob I was able to solve the merge conflict and get it to the point it builds again. I also did some quick manual testing and it works fine.

However, it's a 7 year old PR, so I please keep in mind that I might not be able to answer all the questions ;)

brianrob:
Thank you very much for the energy around this PR @adamsitnik.  I've done a review pass, and overall it looks good.  My main concern is that there's a good amount of churn in the `StackWindow` code.  We have some test coverage here, but not a ton.

I've done a bit of testing myself and most of it seems OK.  I'm going to use this a bit for the next few days and assuming that goes well, I think we can move forward with this.  I have have some follow-up questions for you as well.  FWIW, this touches an area of PerfView that I've not done much with, so you may know more about it than I do.  :)

brianrob (approved):
LGTM.  Thanks @adamsitnik!

PR #597, opened 2018-03-12
Add Support for Nested Structs to EventPipe

brianrob:
Required to support TraceLogging in EventPipe: https://github.com/dotnet/coreclr/pull/16861.


src/TraceEvent/EventPipe/EventPipeTraceEventParser.cs
```diff
@@ -1,3 +1,4 @@
+#define SUPPORT_V1_V2
```
vancem: We now need this define in two files, we should move it to the TraceEvent.csproj file under DefineConstants.

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -424,7 +424,10 @@ public EventPipeEventMetaData(PinnedStreamReader reader, int length, int fileFor
         public int MetaDataId { get; private set; }
         public string ProviderName { get; private set; }
         public string EventName { get; private set; }
+#if SUPPORT_V1_V2
         public Tuple<TypeCode, string>[] ParameterDefinitions { get; private set; }
+#endif
+        public DynamicTraceEventData Template { get; private set; }
```
vancem: Generally speaking things associated with parsing events (and in particular templates for a TraceEvent belong in the EventPipeTraceEventParser, not here.   

If you move this parsing code to EventPipeTraceEventParser it will avoid the need for this field (which really does not belong here (in the source rather than the parser).

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -424,7 +424,10 @@ public EventPipeEventMetaData(PinnedStreamReader reader, int length, int fileFor
         public int MetaDataId { get; private set; }
         public string ProviderName { get; private set; }
         public string EventName { get; private set; }
+#if SUPPORT_V1_V2
```
vancem: You are leaving behind the NewTemplate() code in EventPipeTraceEventParser (which is why you need these ParameterDefintions), but basically cloning it below in ParseFields.   Given that we will be supporting V1_V2 for a while I would really like to avoid the cloning of this code.   

It seems to me that there are no incompatible differences between V3 and V2 as far as reading fields and thus making a Template, so we should be able to do the same logic for all.    Now maybe I am missing something, but I would at least like to make the attempt.

vancem:
Brian,

First sorry for the delay reviewing this.  
We should probably talk, as I am asking for structural changes that probably have issues with them.   I may back off some of my suggestions.    However I really would to avoid cloning the big switch statement on types.     I am still stewing about what I think about all this...

brianrob:
Thanks for the review @vancem.  I'll come chat.

brianrob:
@vancem, per our conversation, I have made adjustments to this PR.  Please let me know if you have any other concerns.

vancem:
I like this version much better.   Thanks for taking my feedback.   I want to add some comments, but I will do that as a separate checkin

brianrob:
Thanks much @vancem.

PR #599, opened 2018-03-12
Migrate serviceprofiler changes 2

xiaomi7732:
Mainly add switches to allow callers to turn on/off some experimental features.

### DIFF
--- a/src/TraceEvent/Computers/SampleProfilerThreadTimeComputer.cs
+++ b/src/TraceEvent/Computers/SampleProfilerThreadTimeComputer.cs
@@ -41,6 +41,13 @@ namespace Microsoft.Diagnostics.Tracing
         /// If set we compute thread time using Tasks
         /// </summary>
         public bool UseTasks;
+
+        /// <summary>
+        /// Track additional info on like EventName or so.
+        /// Default to true to keep backward compatibility.
+        /// </summary>
+        public bool TrackAdditionalInfo = true;
+
         /// <summary>
         /// Use start-stop activities as the grouping construct. 
         /// </summary>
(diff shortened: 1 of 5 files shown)


xiaomi7732:
Ooops, sorry I did another rebase to merge the conflict and then we lost an important piece of the discussion. Bring it back here:
https://github.com/Microsoft/perfview/commit/a5f5dcfb12d57223a7141493523e3bbed284ab3f#commitcomment-28081622

src/TraceEvent/Computers/SampleProfilerThreadTimeComputer.cs
```diff
@@ -41,6 +41,13 @@ public SampleProfilerThreadTimeComputer(TraceLog eventLog, SymbolReader symbolRe
         /// If set we compute thread time using Tasks
         /// </summary>
         public bool UseTasks;
+
+        /// <summary>
+        /// Track additional info on like EventName or so.
+        /// Default to true to keep backward compatibility.
+        /// </summary>
+        public bool TrackAdditionalInfo = true;
```
vancem: Lets call it IncludeEventSourceEvents -  TrackAdditionalInfo is too generic, the name I am suggesting gives people more information about what to expect.

src/TraceEvent/Computers/SampleProfilerThreadTimeComputer.cs
```diff
@@ -202,17 +209,20 @@ public void GenerateThreadTimeStacks(MutableTraceEventStackSource outputStackSou
 
                 StackSourceCallStackIndex stackIndex = GetCallStack(data, thread);
 
-                // Tack on additional info about the event.
```
vancem: I think you can put the 'TraceEventSource' Boolean around this whole 
```
eventSource.Dynamic.All += .delegate (TraceEvent data)  {
...
}
```
statement above (you can also rip out the TODO FIX NOW line).   It is OK to lose the LogThreadStack update that disappears.   This is what the code looked like before the feature was added.

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -1140,7 +1140,8 @@ void FixAndProcessWindowsASP(TraceEvent data, KeyValuePair<Guid, Guid>[] threadT
                 if (creator == null)        // If we don't have a creator for the ASP.NET event, make one since WCF and others are children of it and we want to see that 
                 {
                     // We create another 
-                    creator = OnStart(data, extraStartInfo, &relatedActivityId, thread, null, "RecHttp", useCurrentActivityForCreatorAsFallback: false);
+                    // It is closer to Start Event than the Send Event, uses &activityID.
+                    creator = OnStart(data, extraStartInfo, &activityID, thread, null, "RecHttp", useCurrentActivityForCreatorAsFallback: false);
```
vancem: I am taking your word that this change makes sense (that for these ASP.NET events using the activityID instead of the relatedActivity is the right thing to do).

src/TraceEvent/Computers/ThreadTimeComputer.cs
```diff
@@ -55,6 +55,11 @@ public ThreadTimeStackComputer(TraceLog eventLog, SymbolReader symbolReader)
         /// </summary>
         public bool UseTasks;
         /// <summary>
+        /// Track additional info on like EventName or so.
+        /// Default to true to keep backward compatibility.
+        /// </summary>
+        public bool TrackAdditionalInfo = true;
```
vancem: Same feedback (change to IncludeEventSourceEvents).

src/TraceEvent/Computers/ThreadTimeComputer.cs
```diff
@@ -300,17 +305,20 @@ public void GenerateThreadTimeStacks(MutableTraceEventStackSource outputStackSou
 
                 StackSourceCallStackIndex stackIndex = GetCallStack(data, thread);
 
-                // Tack on additional info about the event.
```
vancem: Same feedback (you can exclude the whole delegate body)

src/TraceEvent/Symbols/NativeSymbolModule.cs
```diff
@@ -107,24 +107,27 @@ public string FindNameForRva(uint rva, ref uint symbolStartRva)
                 ret = ret.Substring(0, dollarIdx);
 
             // See if we have a Project N map that maps $_NN to a pre-merged assembly name 
-            var mergedAssembliesMap = GetMergedAssembliesMap();
-            if (mergedAssembliesMap != null)
+            if (LookupAssemblyNameForCompiledNative)
```
vancem: I don't really want the LookupAssemblyNameForCompiledNative flag.    In theory only PDBs that come from .NET Native have this section and thus would be affected.   Thus I don't understand why you would need to change this code.

Can you describe more about what fails if this code is in place.   I would hope that what happens is that GetMergedAssemblyMap() returns null and all the code below does not execute, and so you don't care.   

Does that not happen?     

Ideally we don't have to change NativeSymbolModule.cs at all.

xiaomi7732:
@vancem thanks for the review. Please give me some time to work on the question about NativeSymbolModule.

src/TraceEvent/Computers/SampleProfilerThreadTimeComputer.cs
xiaomi7732: Sure. Fix in the next iteration.

src/TraceEvent/Symbols/NativeSymbolModule.cs
xiaomi7732: No, I don't like it more than you do. I didn't have time to look into why symbol match failed in our scenario when these lines of code are on. I'll spend some time to kick the tire again since there have been changes.

PR #603, opened 2018-03-14
Event pipe cleanup

vancem:
There are two commits and it may be useful to view them separately.

The first commit just fiddles with comments, and renames things, and does a small amount of method 
contract simplification.  

The second commit removes a redundant table from the parser (there was already a table that did 
exactly what is needed).


vancem:
cc: @brianrob

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -300,19 +298,20 @@ unsafe public void FromStream(Deserializer deserializer)
     /// Private utility class.
     /// 
     /// An EventPipeEventMetaData holds the information that can be shared among all
```
brianrob: EventPipeEventMetaData --> EventPipeEventMetaDataHeader.

PR #606, opened 2018-03-14
Remove cached etlx file and try again for netperf

xiaomi7732:
When the etlx in the cache is invalid for the current version of the PerfView, delete the etlx in the cache and try to open it again.


vancem:
LGTM

PR #609, opened 2018-03-19
Add CPU ONLY Thread Time if you don't have context switches.

vancem:
You don't get a ThreadTime with task and start-stop activity attributization view if you don't collect with /threadTime.

However you CAN most of this without context switches, so make a (CPU ONLY) version of this view and allow users to use that (and thus get task and activity information) even in the normal default collection mode (which is CPU only, but has all the other events needed).


vancem:
@ilabutin  here is a fixed version.  I recommend looking at the diff without whitespace 

https://github.com/Microsoft/perfview/pull/609/files?w=1

ilabutin:
Tested it now, works like a charm. Thank you, @vancem

PR #618, opened 2018-03-22
Fixes an issue where sometimes native DLLs can't get their symbols re…

vancem:
…solved.

As part of merging ETL files certain Image* events are added to add among other things
the GUID that uniquely identifies the DLL and its cooresponding PDB.
Previous to this change, we required that these Image* events be place BEFORE the
ImageLoad event they are for.

Turns out that is not always the case.   Fix it so that either order is fine.

### DIFF
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -1069,6 +1069,8 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
             DbgIDRSDSTraceData lastDbgData = null;
             ImageIDTraceData lastImageIDData = null;
             FileVersionTraceData lastFileVersionData = null;
+            TraceModuleFile lastTraceModuleFile = null;
+            long lastTraceModuleFileQPC = 0;
 
             kernelParser.ImageGroup += delegate (ImageLoadTraceData data)
             {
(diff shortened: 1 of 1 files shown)


src/TraceEvent/TraceLog.cs
```diff
@@ -1106,13 +1108,11 @@ unsafe private void SetupCallbacks(TraceEventDispatcher rawEvents)
...
-                if (data.Opcode == TraceEventOpcode.DataCollectionStart)
-                    bookKeepingEvent = true;
-                else if (data.Opcode == TraceEventOpcode.DataCollectionStop)
-                    bookKeepingEvent = true;
-                 ***/
+                // Remember this ModuleFile because there can be Image* events after this with 
+                // the same timestamp that have information that we need to put  into it 
+                // (the logic above handles the case when those other events are first).  
+                lastTraceModuleFile = moduleFile;
+                lastTraceModuleFileQPC = data.TimeStampQPC;
```
karpinsn: Yesterday Aaron had mentioned that the QPC timestamps might not be exact. I'm not sure exactly in what scenario that happens though.

vancem: Hmm, that is surprising, can we get someone with more details?   These events are injected after the fact, so it would be really surprising that they were not identical (they HAVE To put some value there, it is hard to believe it would not be the clone of the event they are trying to annotate).   

More generally, I need to know what is the 'right' way of 'knowing' that these two events are meant to be associated.

PR #620, opened 2018-03-27
Bypass getting assembly map on targeting .NetStandard 2.0

xiaomi7732:
Unfortunately, loading DIA is failing when targeting .NET Standard 2.0. Just the same as targeting .NET Standard 1.6. Bypass it to avoid exceptions that break the proper symbol resolving for other cases.


vancem:
While ifdefed code in .NET Core V1.6 it was a hack.  I don't want to extend the hack to 2.0.   

I did a little research and I understand why it is failing now.  The issue is that we were using a foreach, which requires special conversion routines in CustomMarshalers, which is not supported on .net Core.   

It is very easy, however to avoid the foreach (we can use a for(....)), which should work everywhere.  
PR #621 includes this fix.   See 

* https://github.com/Microsoft/perfview/pull/621/files#diff-bfe16fd8ed848d802a238c52165149c7

I am going to close this PR in preference to that one.

PR #625, opened 2018-03-30
Fix PerfViewCollect Breaks

brianrob:
There are a couple of issues that showed up when I tried to capture data on an ARM32 device running Win10:

 - Self-contained publishes of PerfViewCollect fail to start because they seem to select an old version of .NET Core 2.0 that is missing a type needed by PerfViewCollect (ByReference<T>).  This is solved by requiring .NET Core 2.0.5.  This is what my machine was using by default when running against the shared framework.
 - SupportFiles were being unpacked but were not being used for DLL loading.  This is because the loader found TraceEvent and its dependencies next to the app, which is what one would expect for a .NET Core build.  Furthermore, because the files have a different timestamp when they are unpacked, this results in an exception because PerfView tries to make sure that SupportFiles version is used or that it at least matches the version that was used (see https://github.com/Microsoft/perfview/blob/68a35e78d2cd6f6d681727792b419c2f497fe0ad/src/PerfView/App.cs#L166).  I've solved this by (a) not embedding the DLLs into PerfViewCollect.dll and (b) disabling the TraceEvent check for PerfViewCollect.  While it is certainly nice to have everything self-contained in a single EXE, .NET Core doesn't really allow for this, and as such I'm assuming that we'll follow the self-contained semantics for PerfViewCollect.

NOTE: I have left in-place the unpacking of the EULA and users guides so that they can still be used from PerfViewCollect if possible.

### DIFF
--- a/src/PerfViewCollect/PerfViewCollect.csproj
+++ b/src/PerfViewCollect/PerfViewCollect.csproj
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.0</TargetFramework>
-    <RuntimeFrameworkVersion>2.0.5</RuntimeFrameworkVersion>
     <RootNamespace>PerfView</RootNamespace>
   </PropertyGroup>
 
--- a/src/PerfViewCollect/PerfViewCollect.csproj
+++ b/src/PerfViewCollect/PerfViewCollect.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.0</TargetFramework>
+    <RuntimeFrameworkVersion>2.0.5</RuntimeFrameworkVersion>
     <RootNamespace>PerfView</RootNamespace>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


src/PerfViewCollect/PerfViewCollect.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.0</TargetFramework>
+    <RuntimeFrameworkVersion>2.0.5</RuntimeFrameworkVersion>
```
vancem: I guess I don't care enough to actually figure it out, but this should not have been necessary.   I like to believe that PerfViewCollect is a very generic .NET Core 2.0 application, and as such should not need a specific framework version.

vancem: Also if an app works before you publish and breaks afterward, it seems likely it is a runtime issue, not an app issue.

brianrob: What was interesting about this is that the actual coreclr.dll that was used in each scenario (shared framework vs. published) was different.

Fortunately (or unfortunately), I can't reproduce this anymore.  Not sure if it was a restore issue or otherwise, but things seem sane now.  Thus, I have removed the RuntimeFrameworkVersion tag.

Issue #617, opened 2018-03-22
Idea: Make TraceEvent class work with DLR


jazzdelightsme:
TraceEvent could be made more convenient to work with by integrating with the DLR (Dynamic Language Runtime). 

That sounds like a Big Deal, but it just means making TraceEvent derive from [DynamicObject](https://docs.microsoft.com/en-us/dotnet/api/system.dynamic.dynamicobject?view=netframework-4.7.1), or alternatively implementing the [IDynamicMetaObjectProvider](https://docs.microsoft.com/en-us/dotnet/api/system.dynamic.idynamicmetaobjectprovider?view=netframework-4.7.1) interface. (See also: [this PDF](http://download.microsoft.com/download/5/4/B/54B83DFE-D7AA-4155-9687-B0CF58FF65D7/implementing-dynamic-interfaces.pdf))

That would allow changing code like this:

```csharp
TraceEvent te = GetEventSomehow();
object lockName = te.PayloadByName( "LockName" );
```
Into code like this:
```csharp
dynamic te = GetEventSomehow();
object lockName = te.LockName;
```

Additionally, it would allow similarly convenient syntax in other environments that work with the DLR--the one I'm really interested in is PowerShell (@HeathS):

```powershell
# Get the set of all lock names
$edgeCP.EventsInProcess | `
    where ProviderName -eq DManipLockTracer | `
    %{ $PSItem.LockName } | `
    Select -Unique
```

(the DLR interfaces are available in NetStandard2, but not 1.6, which I why I recently suggested the upgrade in the other issue)

I think this would be possible by just overriding two methods (if we went the derive-from-DynamicObject route):

```csharp
public override IEnumerable<string> GetDynamicMemberNames()
{
    return PayloadNames;
}

public override bool TryGetMember( GetMemberBinder binder, out object result )
{
    result = this.PayloadByName( binder.Name );
    return result != null;
}
```
I think this fancy awesomeness is pay-for-play: other than possibly referencing a DLR assembly or something, I don't think it would cause any sort of overhead unless/until DLR machinery is invoked (e.g. casting to `dynamic` in C#).

What do you think?

vancem:
I was initially worried about the dependency on DynamicObject, but it looks pervasive enough that I don't have an objection.  

It sounds like it does not require much code change at all.  If that holds up, I would be OK with merging it.  I would suggest you give it a try and if it stays simple create a pull request and I am open to merging it.

brianrob:
Fixed via #630.

PR #630, opened 2018-04-04
Add DLR support for TraceEvent dynamic properties

jazzdelightsme:
The Dynamic Language Runtime approach won't be as efficient as the
TraceParserGen approach, but it is convenient for prototyping or
accessing event data from PowerShell.

(see the updates to the documentation for a bit more info)

The DLR is supported in .NET Standard 2.0, but not 1.6, so we #if it out
for 1.6.

Addresses #617.


jazzdelightsme:
Check it out, @heaths.

jazzdelightsme:
Looking for a quick way to bring down the count of open Pull Requests? This one is a small, surgical change, complete with appropriate doc and test changes--complete this one for an easy win! :D

vancem:
I agree completely uncontroversial.   Sorry for the delay and thanks for doing this.

PR #634, opened 2018-04-09
Strip c# keywords from manifest task names

andd64:
Some manifest files use keywords as task names, i.e. "new", "object", etc.  Create a dictionary of some keywords, and change the name to title case if found.


vancem:
LGTM.

Thanks for your contribution.
Vance

PR #639, opened 2018-04-21
Fix problem with finding PDBs of some DLLs

vancem:
A previous checkin broke a unsual case where several DLLs all have the same load timestamp (this can happen for DCSTARTs).   The result is that information
about the PDB signature is not attached to the DLL, and thus we can't look that
PDB up on a symbol server.     fixed this.

Also some minor tweeks (fix a logging message, add the QPC time to an event dump).


vancem:
Going to start over with a new branch

PR #640, opened 2018-04-22
Fixed issue where PDBS for DLL were not found.

vancem:
The issue was that a previous change to attribute the PDB
signature to the DLL fixed one case but broke another.
In particualr if several DLL loads have the same timestamp
the logic can get confused.    Fixed this case.

Also added the number of usec for the QPC tick count when
you dump and event (some events use these ticks).

### DIFF
--- a/src/TraceEvent/TraceEvent.cs
+++ b/src/TraceEvent/TraceEvent.cs
@@ -1190,6 +1190,7 @@ namespace Microsoft.Diagnostics.Tracing
             XmlAttrib(sb, "Version", Version);
             XmlAttribHex(sb, "Keywords", (ulong)Keywords);
             XmlAttrib(sb, "TimeStampQPC", TimeStampQPC);
+            sb.Append(" QPC=\"").Append((1000000.0 / Source.QPCFreq).ToString("f3")).Append("us\"");
             sb.AppendLine().Append(" ");
 
             XmlAttrib(sb, "Level", Level);
--- a/src/TraceEvent/TraceEvent.cs
+++ b/src/TraceEvent/TraceEvent.cs
@@ -1190,7 +1190,7 @@ namespace Microsoft.Diagnostics.Tracing
             XmlAttrib(sb, "Version", Version);
             XmlAttribHex(sb, "Keywords", (ulong)Keywords);
             XmlAttrib(sb, "TimeStampQPC", TimeStampQPC);
-            sb.Append(" QPC=\"").Append((1000000.0 / Source.QPCFreq).ToString("f3")).Append("us\"");
+            sb.Append(" QPCTime=\"").Append((1000000.0 / Source.QPCFreq).ToString("f3")).Append("us\"");
             sb.AppendLine().Append(" ");
 
             XmlAttrib(sb, "Level", Level);
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -1957,12 +1957,12 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
                 eventCount, (DateTime.Now - startTime).TotalSeconds);
         }
 
-#if DEBUG
         // Pdbs and DLLs often 'match'.   Use this to insure we have hooked
         // up the PDB to the DLL correctly.    This is heurisitc and only used
         // in testing.  
         static bool RoughDllPdbMatch(string dllPath, string pdbPath)
         {
+#if DEBUG
             string dllName = Path.GetFileNameWithoutExtension(dllPath);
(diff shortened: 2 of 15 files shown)


vancem:
Another important part of the change is that previously I excluded CPU samples from the idle thread (since it does not do anything) however the IDLE thread is used for interrupt service routines and deferred procedure calls.  Thus I include them when this is true.   This is mostly only useful for traces that do alot of work in the OS (typically networking).   This allows you to see this CPU consumption (which is nice).

src/TraceEvent/TraceLog.cs
```diff
@@ -1089,18 +1089,20 @@ unsafe private void SetupCallbacks(TraceEventDispatcher rawEvents)
...
 
                 var moduleFile = this.processes.GetOrCreateProcess(data.ProcessID, data.TimeStampQPC).LoadedModules.ImageLoadOrUnload(data, isLoad, fileName);
-                // TODO FIX NOW review:  is using the timestamp the best way to make the association
+                // TODO review:  is using the timestamp the best way to make the association
                 if (lastDbgData != null && data.TimeStampQPC == lastDbgData.TimeStampQPC)
                 {
                     moduleFile.pdbName = lastDbgData.PdbFileName;
                     moduleFile.pdbSignature = lastDbgData.GuidSig;
                     moduleFile.pdbAge = lastDbgData.Age;
+                    // There is no guarentee that the names of the DLL and PDB match, but they do 99% of the time
```
pharring: Typo: guarantee

src/TraceEvent/TraceLog.cs
```diff
@@ -1125,12 +1127,17 @@ unsafe private void SetupCallbacks(TraceEventDispatcher rawEvents)
...
                 // The ImageIDDbgID_RSDS may be after the ImageLoad
-                if (lastTraceModuleFile != null &&  lastTraceModuleFileQPC == data.TimeStampQPC)
+                if (lastTraceModuleFile != null && lastTraceModuleFileQPC == data.TimeStampQPC && lastTraceModuleFile.pdbName == null)
                 {
                     lastTraceModuleFile.pdbName = data.PdbFileName;
                     lastTraceModuleFile.pdbSignature = data.GuidSig;
                     lastTraceModuleFile.pdbAge = data.Age;
+                    // There is no guarentee that the names of the DLL and PDB match, but they do 99% of the time
+                    // We tolerate the exceptions, because it is a useful check most of the time 
+                    // We tolerate the exceptions, because it is a useful check most of the time
```
pharring: Remove duplicated line.

src/TraceEvent/TraceLog.cs
```diff
@@ -1125,12 +1127,17 @@ unsafe private void SetupCallbacks(TraceEventDispatcher rawEvents)
...
                 noStack = true;
+
                 // The ImageIDDbgID_RSDS may be after the ImageLoad
-                if (lastTraceModuleFile != null &&  lastTraceModuleFileQPC == data.TimeStampQPC)
+                if (lastTraceModuleFile != null && lastTraceModuleFileQPC == data.TimeStampQPC && lastTraceModuleFile.pdbName == null)
                 {
                     lastTraceModuleFile.pdbName = data.PdbFileName;
                     lastTraceModuleFile.pdbSignature = data.GuidSig;
                     lastTraceModuleFile.pdbAge = data.Age;
+                    // There is no guarentee that the names of the DLL and PDB match, but they do 99% of the time
```
pharring: Typo: guarantee

src/TraceEvent/TraceLog.cs
```diff
@@ -1950,6 +1957,33 @@ unsafe private void CopyRawEvents(TraceEventDispatcher rawEvents, IStreamWriter
                 eventCount, (DateTime.Now - startTime).TotalSeconds);
         }
 
+        // Pdbs and DLLs often 'match'.   Use this to insure we have hooked
```
pharring: Typo: ensure

src/TraceEvent/TraceEvent.cs
```diff
@@ -1190,6 +1190,7 @@ public string Dump(bool includePrettyPrint = false, bool truncateDump = false)
             XmlAttrib(sb, "Version", Version);
             XmlAttribHex(sb, "Keywords", (ulong)Keywords);
             XmlAttrib(sb, "TimeStampQPC", TimeStampQPC);
+            sb.Append(" QPC=\"").Append((1000000.0 / Source.QPCFreq).ToString("f3")).Append("us\"");
```
pharring: Nit: Perhaps `QPF=` or `QPCFreq=` would be better since it's the frequency, not the count.

PR #644, opened 2018-04-25
Fix tests that were passing when they should not 

vancem:
…Fix this.

Basically we were turning asserts to throws, which then got caught and swallowed.
Fix it so that we fail the test oin that case.

Also fix these tests to be more uniform about how we unpack the data files we comare to.   improve the test so that it is more helpful when there are failures.

Finally there were problems that were being ignored that needed to be fixed.
In particular, we were assigning two stacks to the same event (becase there were events with the same QPC time).   Fixed this by disambiguating based on processor number (this is the change in TraceLog.cs).

### DIFF
--- a/src/PerfView.TestUtilities/ThrowingTraceListener.cs
+++ b/src/PerfView.TestUtilities/ThrowingTraceListener.cs
@@ -21,6 +21,7 @@ namespace PerfView.TestUtilities
     {
         public override void Fail(string message, string detailMessage)
         {
+            Xunit.Assert.True(false, message + Environment.NewLine + detailMessage);
             throw new DebugAssertFailureException(message + Environment.NewLine + detailMessage);
         }
 
--- a/src/TraceEvent/TraceEvent.cs
+++ b/src/TraceEvent/TraceEvent.cs
@@ -3046,6 +3046,7 @@ namespace Microsoft.Diagnostics.Tracing
             {
                 Debug.WriteLine("Error: exception thrown during callback.  Will be swallowed!");
                 Debug.WriteLine("Exception: " + e.Message);
+                Debug.Assert(false, "Thrown exception " + e.GetType().Name +  " '" + e.Message + "'");
             }
 #endif
         }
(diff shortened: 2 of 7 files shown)


sharwell:
Can you point out the exception handler that was catching this?

src/TraceEvent/TraceEvent.cs
```diff
@@ -3046,6 +3046,7 @@ private void DoDispatch(TraceEvent anEvent)
             {
                 Debug.WriteLine("Error: exception thrown during callback.  Will be swallowed!");
                 Debug.WriteLine("Exception: " + e.Message);
+                Debug.Assert(false, "Thrown exception " + e.GetType().Name +  " '" + e.Message + "'");
```
vancem: Here is the exception handler that was catching exceptions and swallowing them.

vancem:
@sharwell The try that swallows the exception is in TraceEvent.Dispatch().   I have added a comment on the line I changed.

src/PerfView.TestUtilities/ThrowingTraceListener.cs
```diff
@@ -21,6 +21,7 @@ public sealed class ThrowingTraceListener : TraceListener
     {
         public override void Fail(string message, string detailMessage)
         {
+            Xunit.Assert.True(false, message + Environment.NewLine + detailMessage);
```
sharwell: 💭 I would not expect this line to be necessary, but it also shouldn't hurt anything.

vancem: I like this line because it insures that no one can swallow the exception, and the error 'stops' a bit closer to where it happened than if you allow the exception to propagate (and presumably cause the test system to ultimately fail.

sharwell: I'm not sure that's the case. `Assert.True` just throws [a different exception](https://github.com/xunit/assert.xunit/blob/dbb0b75cc8fa9349d63a97d2af453d2cd568faaf/BooleanAsserts.cs#L95) at the same location where we were already throwing one.

vancem: That makes me sad, and seems like a dubious thing for a test harness to do.   It is pretty important that test harnesses don't ignore failures REGARDLESS  of what the code under test does (it may swallow all exceptions, the test harness should not care).   When a Xunit.Assert fails it should be impossible for the test to succeed, period.  It is pretty easy for it to do (it just remembers that it failed).     Something to chase down at some  point...

PR #645, opened 2018-04-25
added mouse wheel zooming to the FrameGraphView

RomanKalachik:


vancem:
+ @adamsitnik 

Unless Adam raises an objection today, I will merge this.

adamsitnik:
@vancem I just cloned @RomanKalachik fork and tested the zooming. 

It looks very promising, however there are few things that need to be improved before we can merge it:

1. When Flame Graph is zoomed-in, the tooltips don't work. If I zoom in, I wan to know what I am looking at. Tooltips are a must have.
2. Zooming has no limit: you can zoom out to infinity. I think it would be nice to limit zoom out to 100% of the original size.
image
3. To move the Flame Graph when it's zoomed in I need to use scrollbars. It would be nice if I could just move it when my left mouse button is pressed and I am moving the mouse.

@RomanKalachik thank you for this PR! Could you please fix it so we can merge it?

RomanKalachik:
@adamsitnik , Thanks for the review! Sure, I will try to fix those issue in the next couple of days.

brianrob:
Closing as we now have mouse wheel zooming on flamegraphs.

Issue #643, opened 2018-04-24
Figure out why GeneralParsing test error is failing in the cloud but not locally. 


vancem:
We seem to have a strange error that only happens on some machines where the GeneralParsing test fails in the CI system but not locally.   This tracks that issue.

The code to remove is in GeneralParsing.cs is 
```
                if (keyValue.Key.Contains("GC/AllocationTick") || keyValue.Key.Contains("Kernel/DiskIO/Read"))
                    continue;
```

PR #650, opened 2018-05-02
Fix stack association logic

karpinsn:
This change fixes stack association logic when ImageID events with opcode 38 and 40 are present. We are seeing these on traces collected with the Diagnostic Tools collector when gathering VS traces. By setting the noStack flag the stacks get properly associated with the Image/Load event.

### DIFF
--- a/src/TraceEvent/Parsers/SymbolTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/SymbolTraceEventParser.cs
@@ -102,13 +102,39 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
                 source.UnregisterEventTemplate(value, DBGID_LOG_TYPE_NONE, ImageIDTaskGuid);
             }
         }
-        // I don't know much about this event at all...
+        // I don't know much about these events, they are injected by KTC
         public event Action<EmptyTraceData> ImageIDOpcode37
         {
             add
             {
                 // action, eventid, taskid, taskName, taskGuid, opcode, opcodeName, providerGuid, providerName
                 source.RegisterEventTemplate(new EmptyTraceData(value, 0xFFFF, 0, "ImageID", ImageIdTaskGuid, 37, "Opcode37", ProviderGuid, ProviderName));
+            }
+            remove
+            {
+                source.UnregisterEventTemplate(value, 37, ImageIdTaskGuid);
+            }
+        }
+
+        public event Action<EmptyTraceData> ImageIDOpcode38
+        {
+            add
+            {
+                // action, eventid, taskid, taskName, taskGuid, opcode, opcodeName, providerGuid, providerName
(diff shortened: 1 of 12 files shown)


vancem:
Thanks for doing this.  I would like to make two improvements.
1) You could achieve the bug with the one line fix in TraceLog.cs

```
            // Symbol parser events never have a stack (but will have a QPC assoicated with the imageLoad) so we want them ignored 
            symbolParser.All += delegate (TraceEvent data) { noStack = true; };
```
You can also remove the ImageIDOpcode37 definition (and indeed the setting of noStack in the other SymbolParser events) since they are not redundant.

2) That is probably the best answer, but if we want to go further we should try to reverse engineer the new events and figure out what they mean, so we can parse them properly.   This is optional.

karpinsn:
Spelunking through Windows source I was able to figure out what these correspond to. Opcode 38 is for DBGID_LOG_TYPE_PPDB which I assume is for portable PDBs and Opcode 40 is for DBGID_LOG_TYPE_DETERM which I assume is for deterministic PDBs. I'll make the changes you suggested in 1. as I think that is a bit more future proof. I can add a comment about Opcode 38 and 40 if you would like.

vancem:
Adding the comments for opcode 38 and 40 are useful.  If you want to actually make an event for them (even if you don't parse the contents, that is even better (and best, look at the dump of the event (in the events view right click -> Dump Event) and reverse-engineer what the layout is (it is probably just ID and a PDB name).   

Do whatever you feel comfortable with...

karpinsn:
I added the events and decoded the payload. The only one that is a bit weird is DbgPPDBTraceData. I looked at how WPA was decoding it and I'm doing the same but the timestamp is "odd", not sure if it is how I am decoding it or the actual event.

karpinsn:
Looks like AppVeyor is failing since I am now decoding Opcode37 as the ILRSDS event. Can we rebaseline it? What will happen now when a user tries to open a file that already has a cached .etlx with the old opcode37 event?

vancem:
To fix failure in the 'GeneralParsing' tests run the tests locally (Tests -> Run -> AllTests) and look at the failure.  You can click to see its output, and in the output is a command you can use to update the baseline.   When you do this it becomes part of your commit, and you can confirm before commit it that it updated just the lines you expected.

karpinsn:
Yeah I saw that reverted the change and reran but it passed. Wasn't sure if they were being parsed so I put them back to how they were generated.

vancem:
@karpinsn submit a commit that has a small whitespace change to force AppVeyor to try again.   The failure in the Debug build is an Access Violation (that is what the -1073741819 (0xC0000005) tells us).   This implies that it was probably the harness (probably the code coverage harness) that failed. 

As mentioned this is probably caused by #643, but we need more information if only to know how to work around the issue.   

My current theory is that some OS state is important and thus depending on exactly which machine runs the tests, you get different behavior.    Still that does not explain the AV.  

So lets just try re-running it and see what we see.

Issue #652, opened 2018-05-02
perfview only reads 50000 samples from linux perf data


jeremymeng:
so perfview doesn't show all the events in a Linux perf.data.txt with more events.

https://github.com/Microsoft/perfview/blob/e6c8766a3ad60fbe8df8ff26954b0ce21cd9e0e1/src/PerfView/OtherSources/Linux/LinuxPerfScriptEventParser.cs#L288

@brianrob should we have a larger limit, or even better a user configurable option?

PR #655, opened 2018-05-03
Fix default on Linux Reads

vancem:
This addresses issue #652


vancem:
I agree that there is no reason to limit the number of events that will be processed to 50K by default.   Making it unlimited by default.

PR #676, opened 2018-05-14
Update TraceEvent parsers to match CoreCLR manifest changes

brianrob:
This update corresponds to https://github.com/dotnet/coreclr/pull/17989.

### DIFF
--- a/src/TraceEvent/Parsers/ClrPrivateTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/ClrPrivateTraceEventParser.cs
@@ -1402,7 +1402,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
             add
             {
                 // action, eventid, taskid, taskName, taskGuid, opcode, opcodeName, providerGuid, providerName
-                RegisterTemplate(new CCWRefCountChangeAnsiTraceData(value, 193, 1, "GC", GCTaskGuid, 40, "CCWRefCountChange", ProviderGuid, ProviderName));
+                RegisterTemplate(new CCWRefCountChangeAnsiTraceData(value, 193, 1, "GC", GCTaskGuid, 40, "CCWRefCountChangeAnsi", ProviderGuid, ProviderName));
             }
             remove
             {
@@ -1420,7 +1420,6 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
             remove
             {
                 source.UnregisterEventTemplate(value, 200, ProviderGuid);
-                source.UnregisterEventTemplate(value, 40, GCTaskGuid);
             }
         }
         public event Action<SetGCHandleTraceData> GCSetGCHandle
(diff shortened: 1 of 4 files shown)


src/TraceEvent/Parsers/ClrPrivateTraceEventParser.cs
```diff
@@ -1397,19 +1397,32 @@ public enum Keywords : long
...
+        {
+            add
+            {
+                // action, eventid, taskid, taskName, taskGuid, opcode, opcodeName, providerGuid, providerName
+                RegisterTemplate(new CCWRefCountChangeTraceData(value, 200, 1, "GC", GCTaskGuid, 40, "CCWRefCountChange", ProviderGuid, ProviderName));
+            }
+            remove
+            {
+                source.UnregisterEventTemplate(value, 200, ProviderGuid);
+                source.UnregisterEventTemplate(value, 40, GCTaskGuid);
```
vancem: Remove the line that removes the opcode 40 GCTaskGuid.   Arguably, it should not even be there for the other version of the event (since it is for pre-Vista compat, and this event does not care about that compat).    However if it is in, you have to have a unique opcode-Task pair, or weird things will happen.     Suffice it to say that less is better.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -1595,7 +1595,20 @@ internal ClrTraceEventParserState State
             add
             {
                 // action, eventid, taskid, taskName, taskGuid, opcode, opcodeName, providerGuid, providerName
-                RegisterTemplate(new MethodJitInliningFailedTraceData(value, 186, 9, "Method", MethodTaskGuid, 84, "InliningFailed", ProviderGuid, ProviderName));
+                RegisterTemplate(new MethodJitInliningFailedTraceData(value, 192, 9, "Method", MethodTaskGuid, 84, "InliningFailed", ProviderGuid, ProviderName));
+            }
+            remove
+            {
+                source.UnregisterEventTemplate(value, 192, ProviderGuid);
+                source.UnregisterEventTemplate(value, 84, MethodTaskGuid);
```
vancem: Same issue here (remove the 84, MethodTaskGuid line).

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -1629,19 +1642,32 @@ internal ClrTraceEventParserState State
...
+        {
+            add
+            {
+                // action, eventid, taskid, taskName, taskGuid, opcode, opcodeName, providerGuid, providerName
+                RegisterTemplate(new MethodJitTailCallFailedTraceData(value, 192, 9, "Method", MethodTaskGuid, 86, "TailCallFailed", ProviderGuid, ProviderName));
+            }
+            remove
+            {
+                source.UnregisterEventTemplate(value, 192, ProviderGuid);
+                source.UnregisterEventTemplate(value, 86, MethodTaskGuid);
```
vancem: Same issue (remove hte 86, MethodTaskGuid line).

vancem:
There is a use of CCWRefCountChangeTraceData in PerfData.cs that needs updating if it is to work with both new and old CCW ref counts.    

It is annoying, but my recommendation is to simply clone the code (it is not huge), and put one after the other.   

Another possibility is to make your ANSI version inherit from the non-ansi.  Slightly tricky, and would need some testing to insure nothing broke, but it would avoid the cloning.

brianrob:
I have confirmed that the inlining events look good.  @luqunl is going to help me get a test case that I can use for the CCW case (hopefully this week).  There does not appear to be anything in core currently.

PR #687, opened 2018-05-29
Adding new overload for PerfView extensions

karpinsn:
Changes from Dan Leviant: added a PerfViewFile.Open overload that can be called by PerfView
extensions like VisualStudioRPS that don't have access to
GuiApp.MainWindow. Also updated the existing Open implementation to
leverage an existing worker, when possible. Diagsession files, which do
a secondary Open operation on the embedded Etl resource leverage this
support.

### DIFF
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -463,11 +463,40 @@ namespace PerfView
             return Path.Combine(dirName, fileName);
         }
 
+        /// <summary>
+        /// for extensions of PerfView that don't have access to the GuiApp.MainWindow
+        /// </summary>
+        /// <param name="doAfter"></param>
+        public void Open(Action doAfter = null)
+        {
+            this.Open(GuiApp.MainWindow, GuiApp.MainWindow.StatusBar, doAfter);
+        }
+
         public override void Open(Window parentWindow, StatusBar worker, Action doAfter = null)
         {
             if (!m_opened)
             {
-                worker.StartWork("Opening " + Name, delegate ()
+                if (worker.IsWorking)
+                {
+                    Action<Action> continuation = OpenImpl(parentWindow, worker);
+
+                    m_opened = true;
+                    FirePropertyChanged("Children");
+
+                    IsExpanded = true;
+                    var defaultSource = GetStackSource();
+                    if (defaultSource != null)
+                        defaultSource.IsSelected = true;
+
+                    if (continuation != null)
+                        continuation(doAfter);
+                    else
+                        doAfter?.Invoke();
+                }
+                else
+                {
+
+                    worker.StartWork("Opening " + Name, delegate ()
                 {
                     Action<Action> continuation = OpenImpl(parentWindow, worker);
                     ExecuteOnOpenCommand(worker);
@@ -487,6 +516,7 @@ namespace PerfView
                             doAfter?.Invoke();
                     });
                 });
+                }
             }
             else
             {
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -476,30 +476,11 @@ namespace PerfView
         {
             if (!m_opened)
             {
-                if (worker.IsWorking)
-                {
-                    Action<Action> continuation = OpenImpl(parentWindow, worker);
-
-                    m_opened = true;
-                    FirePropertyChanged("Children");
-
-                    IsExpanded = true;
-                    var defaultSource = GetStackSource();
-                    if (defaultSource != null)
-                        defaultSource.IsSelected = true;
-
-                    if (continuation != null)
-                        continuation(doAfter);
-                    else
-                        doAfter?.Invoke();
-                }
-                else
-                {
-
                     worker.StartWork("Opening " + Name, delegate ()
                 {
                     Action<Action> continuation = OpenImpl(parentWindow, worker);
                     ExecuteOnOpenCommand(worker);
+
                     worker.EndWork(delegate ()
                     {
                         m_opened = true;
@@ -516,7 +497,6 @@ namespace PerfView
                             doAfter?.Invoke();
                     });
                 });
-                }
             }
             else
             {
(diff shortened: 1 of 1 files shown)


src/PerfView/PerfViewData.cs
```diff
@@ -463,11 +463,40 @@ public static string ChangeExtension(string filePath, string newExtension)
...
+        {
+            this.Open(GuiApp.MainWindow, GuiApp.MainWindow.StatusBar, doAfter);
+        }
+
         public override void Open(Window parentWindow, StatusBar worker, Action doAfter = null)
         {
             if (!m_opened)
             {
-                worker.StartWork("Opening " + Name, delegate ()
+                if (worker.IsWorking)
```
vancem: Please put a comment here that we are trying to support reentrancy (calling Open when processing an Open), and thus can simply do the work inline (since we are already on a worker thread).

vancem: Hmm, the more I think about it the more concerned I get.   Calls like 'Open' are meant to be run on the GUI thread, but if this is a reentrant case, you are probably NOT on the GUI thread (you are on a worker thread).    This is problematic.  

I would like to understand your scenario better.    Seems like you want access to call OpenImpl (which does all the work that can be done on another non-gui thread), but I really don't know


I would like to know more about your scenario.

danlev1: When we open a .diagsession in PerfView, the Open/OpenImpl of the diagsession implementation extracts the resources embedded in the diagsession container and adds them to the Children collection. The VisualStudioRPS extension is then loaded (it implements OnOpen). I have added code to the extension to get at the embedded Etl resource, cast it as an EtlDataFile and Open it. This is done so that PerfView will process the Etl and hydrate its Children so that the VisualStudioRPS extension can then populate the tree view and related Views. Without this change, the Etl file will only be displayed as a node in the treeview and expanding it (dbl-click) will bypass any of the extension code and related visualizations in PerfView (because by that point we have already returned from OnOpen).

The problem was that I couldn't call EtlDataFile.Open(window, statusbar, action) from w/n the extension as I didn't have a way to access those elements. You had previously suggested adding the Open() overload to address this limitation. I had to leverage the existing worker because calling worker.StartWork resulted in a bunch of exceptions. Testing seemed to validate that it works, but certainly there is a risk that I missed some cases.

Here is the call sequence:
 - .diagsession is opened in PerfView
   - calls DiagSessionPerfViewFile.Open
       - calls DiagSessionPerfViewFile.OpenImpl  <-- enumerates the resources in the diagsession container and adds them to the Childrens collection
       - calls PerfViewFile.ExecuteOnOpenCommand  <-- loads extensions that implemented OnOpen
          - calls VisualStudioRPS.OpenEtlFile <-- extension implementation
             -  gets the target Etl resource from the DiagSessionPerfViewFile container
             - casts the Etl resource as ETLPerfViewData
             - calls ETLPerfViewData.Open  <-- in order to initialize the contents of the ETL data file for display

vancem:
Hmm, I think what you are seeing is essentially a bug in the OnOpen functionality in PerfView.   The basic issue is that the 'OnOpened' callback is being made before worker for 'Open' is complete (which is causing the recursion that you don't want.   Also the callback is NOT on the gui thread, but that is a bad conventions because it prevents you from running GUI commands.

As a counter-proposal, I would prefer to simply move the call to ExecuteOnOpenCommand to the end of the 'EndWork' delegate (that is it is the last thing invoked).  (There are two of these one for ETL files and one for DiagnSession files, move both).   

Because this call is in EndWork's delegate, the work is already 'complete', so there is no reentrancy issue, you are on the GUI thread (which is generally speaking a good thing, in particualr you can call 'Open' commands correctly), it  is easier to explain (OnOpen commands run last), and as a bonus is a smaller code change (basically move a line of code in two places (you can continue to expose the new Open() overload).   

There may be unexpected consequences to this change, but I woudl like to try it out (if it works for you scenario, it is likely to work, since you are 'stressing' the OnOpened exensibility more than anyone else.  

Let me know if this makes sense, (and if so if it works).   I woudl be OK merging that change.

danlev1:
Thanks for the suggestion. Unfortunately, this approach did not succeed, and regressed the .Etl.Zip experience as well. The extension's OnOpen implementation gets executed, but none of the delegate operations from the extention (eg: adding a custom SimpleReport the the dataFile's Children collection to represent the 'Scenarios' view) are activated once the delegate returns. In the case of a Diagsession there is the additional failure to populate the Children of the embedded ETL resource.

vancem:
Hmm, Can you try instead simply moving the ExecutOnOpenCommand() to the Beginning of the EndWork delegate?   That should frankly also fix the need for cloning the body of EndWork (since the worker is 'done' but is more compatible (since all the other events fire in exactly the same order as before).

danlev1:
This approach fixed the .Etl.Zip scenario, but the .diagsession scenario remains broken. Now the extention's OnOpen implementation is called twice when opening a .diagsession. The first call the file path is to the .diagsession. The subsequent call is to the .etl file path, which is essentially a no-op as far as the extension is concerned.

src/PerfView/PerfViewData.cs
```diff
@@ -463,14 +463,24 @@ public static string ChangeExtension(string filePath, string newExtension)
...
 
+        /// <summary>
+        /// for extensions of PerfView that don't have access to the GuiApp.MainWindow
+        /// </summary>
+        /// <param name="doAfter"></param>
+        public void Open(Action doAfter = null)
+        {
+            this.Open(GuiApp.MainWindow, GuiApp.MainWindow.StatusBar, doAfter);
+        }
+
```
vancem: Do you actually use this call anymore (since you are now calling OpenWithoutWorker)?     If you are using it, great, just checking.

src/PerfView/PerfViewData.cs
```diff
@@ -463,14 +463,24 @@ public static string ChangeExtension(string filePath, string newExtension)
...
         public override void Open(Window parentWindow, StatusBar worker, Action doAfter = null)
         {
             if (!m_opened)
             {
-                worker.StartWork("Opening " + Name, delegate ()
+                    worker.StartWork("Opening " + Name, delegate ()
```
vancem: nit, I think the original spacing was correct.

src/PerfView/PerfViewData.cs
danlev1: no, not using any more.

src/PerfView/PerfViewData.cs
danlev1: good catch

PR #689, opened 2018-06-04
TraceEvent mods for relogging

andd64:
+Add more template data to each class compiled with TraceParserGen (keywords/providername/providerguid)
+Enable TraceRelogger WriteEvents(...) to overwrite specific ETW data
+Modify EnsureScratchBufferSpace so it copies previously filled data when resizing.
+Add UInt64 to list of parsed payload args


andd64:
@mikepal2 I was not aware that Marshal.ReAllocHGlobal existed, but it makes it a lot easier to use.  Thanks!

P.S. Sorry about the multiple commits, still learning the git process.

mikepal2:
Thank you for commit. Since you are looking into these sources, could you also take a look at string serialisation bug #702

andd64:
@mikepal2  Good call (Y).  Fixed.

PR #691, opened 2018-06-06
Short-circuit DynamicTraceEventParser::GetPayloadValueAt for arrays of bytes

pcwiese:
For reasons I'm not going to get into, we are using the DynamicTraceEventParser class (as opposed to a strongly typed parser from traceparsergen) to retrieve a payload value of type byte array. Its very inefficient; there is already a TODO in the codebase to handle arrays of blittable types more efficiently. I did some brief testing where I emited a single event, captured and opened up that file using the DynamicTraceEventParser, cloned the event, then read a single 2000 byte payload value 10000 times from that clone. Using perfview I can see a 5x reduction in CPU from before and after.

My change looks a little hackish, but seems safe enough until that existing TODO is properly addressed. Existing tests pass.

### DIFF
--- a/src/TraceEvent/DynamicTraceEventParser.cs
+++ b/src/TraceEvent/DynamicTraceEventParser.cs
@@ -570,8 +570,15 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
             {
                 var arrayCount = GetCountForArray(payloadFetch, arrayInfo, ref offset);
 
-                // TODO this is very inefficient for blitable types.   Optimize that.  
+                // TODO this is very inefficient for blitable types. Optimize that.
                 var elementType = arrayInfo.Element.Type;
+
+                // Byte array short-circuit.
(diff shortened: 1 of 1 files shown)


vancem:
I am OK with special casing byte arrays like this.   

Note that if your are trying to simply get at the total payload (that is the payload is just one field that you want returned as a byte array), there is something even more efficient.

First you can call traceEvent.EventData() to return a byte[] for the whole payload area (must like your change will do).   This is about as efficient as what you have.   

However there is another overload
```
 traceEvent.EventData(byte[] targetBuffer, int targetStartIndex, int sourceStartIndex, int length)
```
Which will copy the data into YOUR byte[].   (you can pass traceEvent.EventDataLenght for 'length') so you can reuse one byte[] for all your events.  

This allows you to avoid allocating a byte[] for every object, which avoids the allocation/clearing/GC cost.  

Vance

PR #692, opened 2018-06-07
Fix bug with native symbol resolution

karpinsn:
- Fix bug with native symbol resolution where symbols would not resolve
if the RSDS event came after the module load event


vancem:
LGTM.

PR #693, opened 2018-06-08
Update to support signing Nuget support packages

vancem:
A month ago Nuget requires microsoft packages to be signed.
Update TraceEventPackageSigning to support signing nuget packages that are checked in to NugetSupportFiles.
Also fixed a missing DLL problem (PerfView.csproj).

### DIFF
--- a/src/PerfView/SupportFiles/UsersGuide.htm
+++ b/src/PerfView/SupportFiles/UsersGuide.htm
@@ -4138,8 +4138,8 @@
         </li>
         <li>
             The <strong><a id="AdditionalProvidersTextBox">Additional Providers TextBox</a></strong>
-            - A comma separated list of specifications for providers.&nbsp;&nbsp; This can be specified by using the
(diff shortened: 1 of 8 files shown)


src/PerfView/SupportFiles/UsersGuide.htm
```diff
@@ -4138,8 +4138,8 @@ <h3>Advanced Options</h3>
         </li>
         <li>
             The <strong><a id="AdditionalProvidersTextBox">Additional Providers TextBox</a></strong>
-            - A comma separated list of specifications for providers.&nbsp;&nbsp; This can be specified by using the
-            (the ... button) or by the following textual specification.  Each provider
+            - A comma separated list of specifications for providers cooresponding to the /providers comand line qualifier.
```
pharring: typo: "command"

vancem: I have fix this (actually in a number of places) as part of another checkin.

Issue #554, opened 2018-02-06
Reviving WpfTraceEventParser


heeko:
I'm currently working on consuming ETW events from WPF. In the repository we can find some traces of a "deactivated" WpfTraceEventParser: [TraceLog.cs](https://github.com/Microsoft/perfview/blame/ed9cc8d131564d0a62c2f41a79b51cc67335b17d/src/TraceEvent/TraceLog.cs#L857). 
I suppose it was used back in the days in the WPF performance suite. Is there any plan to revive this parser?

vancem:
I did not realize until just now that WPF does not register its Events by default (you have to explicitly register them).     You can do this by running the following from an admin prompt.
```
wevtutil im C:\Windows\Microsoft.NET\Framework\v4.0.30319\WPF\wpf-etw.man
```
At this point you can use the name Microsoft-Windows-WPF to turn on WPF events like this
```
PerfView /Providers=Microsoft-Windows-WPF collect
```
And perfView (via the RegisterTraceEventParser) will correctly parse the events.  

*********************
As with any ETW provider, if you find that you are processing MANY such events, and/or you want compile time support for fetching fields etc, you can make a parser of your own.   

That is what the TraceParserGen tool does (when you build this repository you build TraceParserGen).   See #336 for more on doing that.

*********************
Finally to answer your direct question, there are not plans to revive the WPF parser in the TraceEvent library.   We cleaned out 'little used' parsers from the library on premise that users who needed them could use TraceParserGen to generate them themselves.

heeko:
Thanks a lot for the detailed answer!

PR #712, opened 2018-07-20
TraceParserGen: fix GetTemplateNameForEvent for empty evnt fields

heeko:
Fixing an issue that happen when the evnt.Fields is not null but empty. Without this fix, if the evnt.Fields is empty, it will generate this kind of event:
```csharp
public event Action<Template_0> AddAnnotationStart
```
instead of:
```csharp
public event Action<EmptyTraceData> AddAnnotationStart
```

This can be easily observe by trying to generate a parser for the **wpf-etw.man**.

I just added a check on the count of evnt.Fields. Let me know if you prefer the null conditional form `if (evnt.Fields?.Count > 0 )`, I can change that.

### DIFF
--- a/src/TraceParserGen/TraceParserGen.cs
+++ b/src/TraceParserGen/TraceParserGen.cs
@@ -674,7 +674,7 @@ class TraceParserGen
     private static string GetTemplateNameForEvent(Event evnt, string eventName)
     {
         var ret = "EmptyTraceData";
-        if (evnt.Fields != null)
+        if (evnt.Fields != null && evnt.Fields.Count > 0)
         {
             ret = TraceParserGen.ToCSharpName(evnt.TemplateName);
             if (ret == null || ret.StartsWith("tid_"))


idg10:
What may not be obvious to casual observer is that this fixes a bug that causes the tool to generate uncompilable code for a fairly widely used ETW provider manifest: the one for the `Microsoft-Windows-WPF` provider. So the tool cannot currently be used to generate a working parser for the consuming ETW events from WPF.

(Apparently, at one point there was a pre-generated parser class for this provider, which may be why people haven't reported this before, but the current advice is now not to use that old class, and to generate a parser directly from the manifest using `TraceParserGen`: https://github.com/Microsoft/perfview/issues/554 So people will now start to run into this. I was about to submit a PR for this until I realised @heeko is ahead of me, so people are definitely running into this bug.)

So in the hope of getting this PR merged sooner, I'm adding this explanation. (I'm 100% certain that @heeko already knows all this. I just suspect it's not entirely obvious to anyone who hasn't just spend a big chunk of time diagnosing this problem.)

The manifest for the WPF provider, `wpf-etw.man`, has an empty template called `Template_0`. The code in `GetTemplateNameForEvent` (prior to the fix in this PR) decides that it needs to use a generated `Template_0` class for all events that refer to this empty template. But the code that generates the classes corresponding to templates decides that it does not need to generate a `Template_0` class.

This problem occurs because two different but interconnected parts of the code apply incompatible policies here: the `GetTemplateNameForEvent` and `GenerateEventPayloadClass` methods need to agree on whether or not a particular event will use a generated custom payload class. Unfortunately, with the code as it is, `GetTemplateNameForEvent` decides that a custom class is required whenever an event uses a template (even an empty one like `Template_0`), which causes lots of code that refers to this custom class to be generated. But `GenerateEventPayloadClass` then decides that because the template defines no fields, no such custom class is required.

`GenerateEventPayloadClass` performs this check at https://github.com/Microsoft/perfview/blob/master/src/TraceParserGen/TraceParserGen.cs#L339

The `allFields.Count <= 0` test is the one that decides that no custom class is required when a template defines no fields.

@heeko's proposed modifies `GetTemplateNameForEvent` to apply the same policy. It would also be possible to fix this by bringing `GenerateEventPayloadClass` into line with the policy applied by `GetTemplateNameForEvent`. However, while that would get rid of the compiler errors, it @heeko's fix is better because it avoids generating an unneeded class.

Issue #724, opened 2018-08-13
ETW EnableKernelProvider is not working on docker Windows Container (Windows Server Core)


Pellared:
Environment:
- Microsoft.Diagnostics.Tracing.TraceEvent.2.0.19
- docker: Version 18.06.0-ce-win72 (19098)
- image: repository: microsoft/windowsservercore; tag: ltsc2016; iamge ID: 7d89a4baf66c

I have created a console application that monitors Kernel Events.
When I run it inside a docker container using powershell then I get the following exception:

```
System.UnauthorizedAccessException: Error Starting ETW:  Access Denied (Administrator rights required to start ETW)
   at Microsoft.Diagnostics.Tracing.Session.TraceEventSession.InsureStarted(EVENT_TRACE_PROPERTIES* properties)
   at Microsoft.Diagnostics.Tracing.Session.TraceEventSession.EnableKernelProvider(Keywords flags, Keywords stackCapture)
   at XYZ.<StartAsync>b__38_0()
   at System.Threading.Tasks.Task.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at XYZ.<StartAsync>d__38.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at XYZ.UnhandledExceptionHandler.<ExecuteAsync>d__6.MoveNext()
```

Add the beginning there of the program there is check:
```
if (TraceEventSession.IsElevated() != true)
{
   Logger.Log("Must be elevated(Admin) to run this program.");
   Environment.Exit(1);
}
```
passing. I have also added a double check in powershell:
```
[...]

PR #726, opened 2018-08-14
Fixing bug for Windows Container support

Pellared:
Should fix https://github.com/Microsoft/perfview/issues/724


Pellared:
Take into consideration that it is not tested (however the bug looks obvious) on Thursday I am going to test if it fixes the issue. I will simply change this one line using https://github.com/0xd4d/dnSpy.

vancem:
as mentioned in #724, Process | Thread | ImageLoad | Profile | ContextSwitch are the events that are ALLOWED in a container (not surprising, they are the most fundamental).   Thus this is not the appropriate fix.  It seems that there is an issue at the OS level.

PR #732, opened 2018-08-27
Enable Heap Snapshot Analysis via EventPipe

brianrob:
Allow PerfView to consume heap snapshots captured via EventPipe.


vancem:
LGTM

jzabroski:
Just wanted to say this is super exciting. Thanks for working on this.

brianrob:
Thanks @vancem and thanks @jzabroski!

src/PerfView/PerfViewData.cs
```diff
@@ -8359,6 +8367,7 @@ protected override Action<Action> OpenImpl(Window parentWindow, StatusBar worker
             bool hasGC = false;
             bool hasJIT = false;
             bool hasAnyStacks = false;
+            bool hasDotNetHeapDumps = true;
```
mattwarren: Erm, apologies if I'm missing the point completely, but shouldn't this default to `hasDotNetHeapDumps = false`, then set it to `true` only if the event starts with `GC/BulkNode`? 

Other wise it'll always try and add a `PerfViewHeapSnapshots`?

Issue #300, opened 2017-07-06
Display long text of the cell - right align content of the status bar


ilabutin:
PerfView displays content of the string in the status bar and one can even scroll/select text there to see full content. This is very useful when investigating deep call stacks where things are getting further to the right in the tree.

![image](https://user-images.githubusercontent.com/1869103/27909705-911fbefe-625a-11e7-8b48-9e215acff7cc.png)

But I often find myself clicking the status bar and scrolling it to the end with mouse to look at the end of the line. Mostly because generic parameters are usually there and it's the way to get understanding what kind of overload was called at that node.

Suggestion is to modify status bar to show rightmost part of the cell by default. Like that:

![image](https://user-images.githubusercontent.com/1869103/27909762-e92cadc8-625a-11e7-8600-71729c25dd39.png)

Could be that additionally it would be good to show some visual indication in the bottom left corner that something is there on the left.

Additional bonus would be to have a tooltip shown for the grid cell with full content and also some hotkey to scroll the status bar to the right. Most likely the latter would not be obvious, so it might be enough to add ability to press End in the status bar or just have a small button on the right to quickly show end of the text.

brianrob:
@ilabutin thanks for the feedback.  I am hesitant to make the bar the bottom show the last bit of text as opposed to the first bit.  I could imagine this would be confusing for most users who aren't expecting this, and haven't ever seen it in another app.  That said, I do think that the idea of a tooltip is good, and would be useful to implement.  I realize this is an old issue, but would you be interested in submitting a PR for this?

ilabutin:
Actually tooltip won't help much unless it shows the rightmost part of the cell.
But what I tried to implement instead is horizontal scrolling of the grid itself. So that there is a horizontal scrollbar under the grid which scrolls the content of the "Name" column of the grid.

My two previous pull requests were not merged for various reasons, but I personally use the version of PerfView with this functionality and it feels OK.

If you feel this might be interesting, then I can try to prepare another PR for this change so that you can evaluate.
@brianrob

PR #739, opened 2018-09-07
Added horizontal scrolling of PerfGrids

ilabutin:
As an attempt to make better solution for #300 I've implemented horizontal scrolling within first column of the grid.
The scrollbar appears below grid when necessary.

Works fine for me for few months already. There are some issues (like resetting to the leftmost position after expanding/collapsing nodes, but it does not affect overall better UI experience).

GitHub shows .xaml file contains lots of changes, but this is primarily because of changed nesting, thus indentation.

It is better to see diff ignoring whitespace differences.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,3 +3,7 @@ root=true
 [*.cs]
 indent_style = space
 indent_size = 4
+
+[*.xaml]
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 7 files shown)


vancem:
Sorry for the delay, I don't have much time to spend on PerfView these days.

The general philosphy behind perfview was to NOT be very innovative about GUI stuff.   Ideally it only used 'stock' things and spend most of its effort simply organizing the data in interesting ways (that is the model that GUI is displaying).     

My concern with this PR is that there are 400+ lines of new code to implement what is basically a GUI feature (scrollable Perf Grids).   This is a fair bit of complexity.    

I will need time to look at it.

ilabutin:
WPF is sometimes too verbose. I do not insisting on merging, since I'm using my custom version anyway :)

PR #740, opened 2018-09-10
Initial checkin if EventRegister code.

vancem:
EventRegister is a tool for registering EventSources.   It did not have a open-source home, and I am suggesting that we put it here in the PerfView repo in the 'related' subdirectory.

This code is a copy of the latests code (but it had not changed in over a year).   I did just enough to make it
build in the current repository (which included turning off any signing).

You can do an 'msbuild' in the EventRegister directory to build it (or launch EventRegister.sln and build in VS).

To make this 'real' we woudl need to make a nuget package spec for it, and have it be built and signed officially.


vancem:
@brianrob

PR #745, opened 2018-09-13
Fix dev-time load failure loading KernelTraceControl.

vancem:
OSExtentions needs to be 'next' to the architecture directories where KernelTraceControl.dll lives for it to load
them properly.   This works for published .NET Core apps, but by default 'dotnet run' runs nuget DLLs out of
the nuget cache (it does not copy them to the app directory).   This breaks the loading of KernelTraceControl.dll.

We already manually copy the native files and some interop files to the output directory in the TraceEvent nuget
package.   Here we simply add OSExtentions.dll to that list so it is loaded from the same place even in the dev-time
scenario.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -22,8 +22,8 @@
 
   <PropertyGroup>
     <!-- These are the versions of the things we are CREATING in this repository -->
-    <PerfViewVersion>2.0.23</PerfViewVersion>
-    <TraceEventVersion>2.0.23</TraceEventVersion>
+    <PerfViewVersion>2.0.25</PerfViewVersion>
+    <TraceEventVersion>2.0.25</TraceEventVersion>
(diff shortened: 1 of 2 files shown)


vancem:
@adamsitnik - this is my counter-proposal for fixing the issue you ran into.    Basically it is you option 2 (fighting with nuget to get the DLL in the right place).   It just so happens, we had done this fighting in the past, so it was pretty easy to add this.    It is also the least hacky way to solve the problem (TraceEvent code should not know about Nuget packaging conventions).

vancem:
@adamsitnik  you can try this out yourself, but I was able to reproduce your failure, and this fixed it.

adamsitnik:
@vancem thanks a lot! I have verified it and the fix solves the problem.

And I have learned new NuGet trick from this PR ;)

PR #747, opened 2018-09-14
Fixes in TraceLoadedDotNetRuntime.SetupCallback

chrisnas:
- Provide the generation in GCStart event (instead of always 0)
- don't forget to count the suspension time between the start of BGC gen 2 and inner gen 0/1

### DIFF
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -669,7 +669,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                     {
                         TraceGC _gc = new TraceGC(stats.GC.m_stats.HeapCount);
                         Debug.Assert(0 <= data.Depth && data.Depth <= 2);
-                        // _event.GCGeneration = data.Depth;   Old style events only have this in the GCStop event.  
+                        _gc.Generation = data.Depth;
                         _gc.Reason = data.Reason;
                         _gc.Number = data.Count;
                         _gc.Type = data.Type;
@@ -692,6 +692,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                         if (isEphemeralGCAtBGCStart || _gc.Reason == GCReason.PMFullGC)
                         {
                             _gc.PauseStartRelativeMSec = data.TimeStampRelativeMSec;
+
                             if (_gc.Reason == GCReason.PMFullGC)
                             {
                                 TraceGC lastGC = TraceGarbageCollector.GetCurrentGC(stats);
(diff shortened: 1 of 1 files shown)


Maoni0:
I just looked at the code to refresh my memory since I wrote this code years ago. So the initial suspension time was indeed counted *somewhere* - it’s counted as part of the BGC’s SuspensionDurationMSec in the GCStart handler –

```
_gc.SuspendDurationMSec = stats.GC.m_stats.suspendEndTimeRelativeMSec - stats.GC.m_stats.suspendTimeRelativeMSec;
```
But this was not counted towards BGC’s pause time which is supposed to include the suspension time. So instead of counting it as part of the EphemeralGCAtBGCStart pause time, I will count this as part of BGC’s pause since it’s already counted as part of its suspension time; we should be consistent.

I will take this change in your PR:

```_gc.Generation = data.Depth;```

As this is useful for real time processing. You can just get rid of this comment 

```// _event.GCGeneration = data.Depth;   Old style events only have this in the GCStop event.```

since now we always have this field in GCStart (I suppose we just don’t support the older versions that didn’t have this field anymore). You can even get rid of setting this in the GCStop handler if you are already setting it in GCStart.

Would you like to revise the PR and resubmit?

File: src/TraceEvent/Computers/TraceManagedProcess.cs [hunk 1]
```diff
@@ -691,7 +692,16 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
                         Debug.Assert(stats.GC.m_stats.suspendTimeRelativeMSec != -1);
                         if (isEphemeralGCAtBGCStart || _gc.Reason == GCReason.PMFullGC)
                         {
-                            _gc.PauseStartRelativeMSec = data.TimeStampRelativeMSec;
+                            if (isEphemeralGCAtBGCStart)
+                            {
+                                // in that case, we don't want to lose the suspension time elapsed since the BGC started
+                                _gc.PauseStartRelativeMSec = stats.GC.m_stats.suspendTimeRelativeMSec;
```
Maoni0: instead of counting it we should initialize the BGC pause a few lines below
```
                        _gc.StartRelativeMSec = data.TimeStampRelativeMSec;
                        stats.GC.GCs.Add(_gc);
                        if (_gc.Type == GCType.BackgroundGC)
                        {
                            stats.GC.m_stats.currentBGC = _gc;
                            // For BGC, we need to add the suspension time so far to its pause so we don't miss including it.
                            // If there's an ephemeral GC happening before the BGC starts, AddConcurrentPauseTime will not
                            // add this suspension time to GC pause as that GC would be seen the ephemeral GC, not the BGC.
                            _gc.PauseDurationMSec = _gc.SuspendDurationMSec;
                            _gc.ProcessCpuAtLastGC = stats.GC.m_stats.ProcessCpuAtLastGC;
                        }
```

chrisnas: You're right: sounds more consistent to me

Maoni0:
thanks for doing this work!

PR #756, opened 2018-10-05
Added CTRL+L command to focus the Directory (HistoryComboBox).

### LINKED ISSUE #132: pressing control+L should bring focus to the current directory text box
When you press Control + L in a browser or in File Explorer focus is brought to the URL or current directory. PerfView should do this to.

ppozeti:
Fixes #132 
The only issue is when you focus the RichTextBox. As it has its own commands and bindings, the CTRL+L will not work. You would have to press TAB to lose focus and then CTRL+L should work again.
I tested adding Focused=False on RichTextBox to avoid firing up the commands. As this RTB is always disabled, no need to listen to the editing commands, but, another issue I found is after clicking on a link inside RTB. It somehow gives the RTB focus again and CTRL+L doesn't work.

If adding focused false is fine, let me know so I can add that change as well.

### DIFF
--- a/src/PerfView/MainWindow.xaml
+++ b/src/PerfView/MainWindow.xaml
@@ -39,7 +39,11 @@
         <CommandBinding Command="{x:Static src:MainWindow.ILSizeCommand}" Executed="DoILSize"/>
         <CommandBinding Command="{x:Static src:MainWindow.HeapSnapshotFromDumpCommand}" Executed="DoTakeHeapShapshotFromProcessDump"/>
         <CommandBinding Command="Help" Executed="DoHyperlinkHelp"/>
+        <CommandBinding Command="{x:Static src:MainWindow.FocusDirectoryCommand}" Executed="DoFocusDirectory"/>
     </Window.CommandBindings>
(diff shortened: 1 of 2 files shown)


vancem:
Thanks for doing this @ppozeti.  The code you have written looks good.   But to be really useful it needs to be discoverable by users.   Thus a couple more things would be good to do

1. Add documentation to the help documentation that the 'Tips' hyperlink goes to.  (Search for Main View Tips) in usersGuide.htm.
2. I like it if EVERYTHING that is a keyboard shortcut also has a menu item of some sort as well.  That way there is a place to put the keyboard shortcut so people can learn them.   If the main view had a context menu, I woudl put it on that, but since it does not, I woudl add a entry after 'User Command' on the 'File' menu (and put the shortcut there).

kerdrek:
@vancem since @ppozeti  is not responding I took the liberty of adding the requested discoverability of CTRL+L in the following PR https://github.com/microsoft/perfview/pull/937

brianrob:
Closing as #937 includes this change.

Issue #680, opened 2018-05-17
Perfview switches to Monitor 1 even if perfview is on Monitor 2


nibubt:
Steps to repro...

1. Move perfview to Monitor 2/3
2. Load up an ETL trace trace
3. Dbl Click on any item for analysis, for e.g. "Exception stacks"
4. Perfview process list window opens up on monitor 1 instead of Monitor 2/3
5. Its becoming tedious to keep dragging around windows. 

Great tool BTW :-)

vancem:
This behavior is from the underlying WPF routines (PerfView mostly does not set the position of windows).   However it could (and thus keep new windows 'near' old ones.    It is a straightforward change.   If I have tome I will do it, but if other wish to jump in that is great.

You can see logic where PerfVIew remembers the position where it last opened a stack window in the StackWindow constructor in StackWindow.xaml.cs.   

To do a proper job of this we want this to work for not just stack windows but for the dialog boxes and the logic that saves the position from run to run needs to work with multiple monitors (right now it is broken for multiple monitors).

jquaglio:
@vancem  Can we create a `Window` base class that sets the `Owner` property to the parent window and `WindowStartupLocation` to `CenterOwner`, and update all the xaml to use this new base class?

vancem:
I am open to people doing SMALL changes to fix small issues like this.   Your suggestion may work out and as long as the number of lines of code changed is small and/or really simple, I am OK with it.    I am not likely to get to this, so it is up to someone else to take a wack at it.

jquaglio:
I can do it! Could you please assign it to me?

vancem:
GitHub won't seem to let me assign it to you, but anyone else who wants this should check here first, so there should not be a conflict.

jquaglio:
@vancem, Maybe you need to add me as a contributor on this repo, to be able to assign me to this.

vancem:
Fixed by #758

vancem:
Thanks again for actually providing the fix.

nibubt:
Thanks!

PR #758, opened 2018-10-05
Add a feature to open new window based on parent window location.

jquaglio:
A proposed fix for issue [#680](https://github.com/Microsoft/perfview/issues/680).


vancem:
Sorry for the delay.   The large number of files changed did put me off. 

However, the modifications are conceptually simple (you are changing the base for every window), and do solve the problem completely (every window now will stay near its parent and thus on the same monitor).   

Thanks for doing this.

PR #759, opened 2018-10-06
Grouping: [group CLR/OS entries]: Added System.Xaml namespace to be grouped with WPF

otmopo3:
Grouping option `[group CLR/OS entries] \Temporary ASP.NET Files\->;v4.0.30319\%!=>CLR;v2.0.50727\%!=>CLR;mscoree=>CLR;\mscorlib.*!=>LIB;\System.*!=>LIB;Presentation%=>WPF;WindowsBase%=>WPF;system32\*!=>OS;syswow64\*!=>OS;{%}!=> module $1` does not consider that `System.Xaml` namespace refers to WPF, but is displayed as `LIB` currently which can lead to extra lines in treeview like WPF->LIB(System.Xaml)->WPF.
Fixed that behaviour in this PR. Added `;\System.Xaml.*!=>WPF` to grouping expression `[group CLR/OS entries] ...`
perfview-wpf-wrong-grouping

### DIFF
--- a/src/PerfView/Extensibility.cs
+++ b/src/PerfView/Extensibility.cs
@@ -1108,7 +1108,7 @@ namespace PerfViewExtensibility
             else
             {
                 ret.FilterGuiState.GroupRegEx.Value =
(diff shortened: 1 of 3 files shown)


vancem:
Looks good.

PR #760, opened 2018-10-08
Fix Null Ref Exception

vancem:


vancem:
This is not a complete fix.   Abandoning this fix for another.

PR #765, opened 2018-10-15
Added text editor path variable option

cirojr:
Added the menu item "User Config" under "File" menu, moving there the other variable settings dialogs.
Now, when comes the time to open the text editor, PerfView can look at this variable to check if it's any path configured there to open the source code, instead of the built-in text editor or the fixed notepad.
I've also included some entries for the new variable in "User Guide".


src/PerfView/App.cs
```diff
@@ -729,6 +729,30 @@ public static string SourcePath
                 App.ConfigData["_NT_SOURCE_PATH"] = value;
             }
         }
+        public static string TextEditorPath
+        {
+            get
+            {
+                if (m_TextEditorPath == null)
+                {
+                    var symPath = new SymbolPath(Environment.GetEnvironmentVariable("_NT_TEXTEDITOR_PATH"));
```
vancem: I web searched it and _NT_TEXTEDITOR_PATH seems to be a new environment variable.   
It is true that PerfVIew uses _NT_SOURCE_PATH and _NT_SYMBOL_PATH, but only because those are already well established conventions by the operating system.  Thus PerfView is simply 'playing along' (not establishing its own standard).  

My recommendation is to rename it to be PERFVIEW_TEXT_EDITOR_COMMAND.

It is also reasonably important that we support editors that can go to a particular line, because you tend to be lost without it.   (We should unconditionally log to the status bar (StatusBar.Log($"[opening  file {file}:{line}"]"))  (notice the [] that forces the log gin to show up on the status bar).  that line number we wanted to jump to in case the editor does not support doing this from the command line so that users can do it themselves.

The the variable should be the command line not the editor's path.   It should have %f and %l special symbols that expand to the file name and line number.    

The logic here should only check the environment variable it it App.ConfigData does not already have a command.   (That way if you override in the GUI, that takes prescidence).   

If no environment variable is set and notepad2 is on the path then it should default to 
 
* notepad2 /g %l %f

Everything above is subject to negotiation if you don't like it, but I have reasons for most of it...

As an aside, Ideally we fix the built in editor to know how to go to a line number.  

Thanks for helping to make PerfView better.

cirojr: Alright. All sugestions make sense. Thanks!

brianrob:
@cirojr, I'm going to close this for now.  Feel free to re-open with @vancem's feedback addressed.  Thank you.

PR #768, opened 2018-10-18
Fixing Very Fragmented Ephemeral condemn reason name and description

### LINKED ISSUE #616: "Very Fragmented Ephemeral" condemn reason is incorrect
In _Condemned reasons for GCs_ section of _GCStats_ view, there is a column named _"Very Fragmented Ephemeral"_ with an additional _"Ephemeral generations are VERY fragmented, doing a full GC"_ description.

This is however incorrect. This reason internally refers to `gen_max_high_frag_e_p` condition, which is set if fragmentation of generation 2 (`max_generation`, see `dt_high_frag_p ` below) is big enough to fit generation 1:

```
if (dt_low_ephemeral_space_p (check_only_p ? tuning_deciding_full_gc : tuning_deciding_condemned_gen))
{
	...
	local_condemn_reasons->set_condition (gen_low_ephemeral_p);
	...
	
	//It is better to defragment first if we are running out of space for
	//the ephemeral generation but we have enough fragmentation to make up for it
	//in the non ephemeral generation. Essentially we are trading a gen2 for 
	// having to expand heap in ephemeral collections.
	if (dt_high_frag_p (tuning_deciding_condemned_gen, 
						max_generation - 1, 
						TRUE))
	{
		high_fragmentation = TRUE;
		local_condemn_reasons->set_condition (gen_max_high_frag_e_p);
	}
}

BOOL 
gc_heap::dt_high_frag_p (gc_tuning_point tp, int gen_number, BOOL elevate_p)
{
	BOOL ret = FALSE;
	switch (tp)
	{
		case tuning_deciding_condemned_gen:
		{
			dynamic_data* dd = dynamic_data_of (gen_number);
			float fragmentation_burden = 0;
			if (elevate_p)
			{
				ret = (dd_fragmentation (dynamic_data_of (max_generation)) >= dd_max_size(dd));
			}
			...
		}
	}
	return ret;
}
```

I suggest changing the column name and description to the "_Ephemeral Low with Very Fragmented Gen2_" and "_We are running low on the ephemeral segment but gen2 is fragmented enough to compensate that, doing a full GC_".

kkokosa:
This PR resolves #616 - by providing more accurate name and description of the condemn reason named so far "Very Fragmented Ephemeral".

### DIFF
--- a/src/PerfView/GcStats.cs
+++ b/src/PerfView/GcStats.cs
@@ -1045,7 +1045,7 @@ namespace Stats
             new string[] {"Ephemeral<BR/>Low", "We are running low on the ephemeral segment, GC needs to do at least a gen1 GC"},
             new string[] {"Expand<BR/>Heap", "We are running low in an ephemeral GC, GC needs to do a full GC"},
             new string[] {"Fragmented<BR/>Ephemeral", "Ephemeral generations are fragmented"},
-            new string[] {"Very<BR/>Fragmented<BR/>Ephemeral", "Ephemeral generations are VERY fragmented, doing a full GC"},
+            new string[] {"Ephemeral Low<BR/>with Very<BR/>Fragmented Gen2", "We are running low on the ephemeral segment but gen2 is fragmented enough to compensate that, doing a full GC"},
             new string[] {"Fragmented<BR/>Gen2", "Gen2 is too fragmented, doing a full blocking GC"},
             new string[] {"High<BR/>Memory", "We are in high memory load situation and doing a full blocking GC"},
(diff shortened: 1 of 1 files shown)


vancem:
@Maoni0  Please take a quick look.   You seemed to agree with his analysis in  issue #616.  If you are OK with it feel free to merge it.

Maoni0:
I would rename the header to "Low Ephemeral<BR>Fragmented Gen2". now that I read the text I am not sure how people would interpret "but gen2 is fragmented enough to compensate (for) that", perhaps it's better to make it more clear by saying "but gen2 is fragmented enough so a full GC would avoid expanding the heap"

kkokosa:
I've finally fixed it according to @Maoni0 suggestions.

PR #772, opened 2018-10-24
Fix non-working URL

TDabasinskas:
For some reason, `Tutorial.cs` link was pointing to `Tutorial.cs.txt` file, which returned `404: Not Found`. Updated the link to point to `Tutorial.cs` instead.


vancem:
Hmmm,  I can't reproduce your issue (I can click on the Tutorial.cs hyperlink and it loads it.   The URL needs to end in TXT so that it loads a text view for the file (it does not think it needs to load Visual Studio ...).   

Note that the way this is supposed to work is that the Users guide as well as the tutorial.txt both live in a directory $HOMEPATH$\AppData\Roaming\PerfView\VER.* where * is a timestamp associated with building the PerfView.exe.    You should see both a  UsersGuide.htm and Tutorial.cs and Tutorial.cs.txt in that directory.  

I am closing this but feel free to reopen if you still have problems.

Issue #777, opened 2018-10-30
New tread created for every GC/Start event


sharwell:
https://github.com/Microsoft/perfview/blob/0f77be5938135af9a64c2f928c44b19dc7dbc048/src/TraceEvent/TraceLog.cs#L1994

The **GC/Start** event is resulting in `isThreadCreateEvent` getting set to true.

Assigning to @vancem since this was likely introduced by #763.

vancem:
Closing fixed by #778.   \\clrmain\tools\PerfView.exe has the fixed copy.  The offical  github copy was never broken (the break postdated the last release).

Issue #766, opened 2018-10-17
INVALID_HANDLE_VALUE constants has wrong value


RubenHesselbaek:
The INVALID_HANDLE_VALUE constants has the value -1, which means INVALID_HANDLE_VALUE, but according to the documentation on https://docs.microsoft.com/en-us/windows/desktop/etw/starttrace
It's mention that invalid handles have the value 0.
This among other thing leads to closing+ reopening realtime session, when all you want is to attach and most likely other stuff where the session is deemed invalid, but is not.

vancem:
I have fixed it so that after calling StartTrace we insure that on failure the handle is INVALID_HANDLE_VALUE.    That brings the StartTrace API in line with the expecations of the higher level code.

If StartTrace returns INVALID_HANDLE_VALUE (that is -1), to represent a valid trace, then we would still have problems, but I don't think this is the case.   

I am closing this issue, feel free to reopen it if you feel there is still something to fix.

PR #778, opened 2018-10-31
Fixes

vancem:
Small fixes, in particular for issue #777

Fixed issue #777 (Excessive creation of TraceThreads)

Improved logging for finding PDBs slightly (now prints SOMETHING even if you hit a PDB cache).

Fixed a bug where a PDB was not being found because the last lookup had different PDB
lookup options set.

Responded to issue #766 about the fact that StartTrace does not set the handle to invalid if it fails.

Do a bit better at categorizing threads a .NET ThreadPool threads (or I/O threadpool threads).

@sharwell

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -22,8 +22,8 @@
 
   <PropertyGroup>
     <!-- These are the versions of the things we are CREATING in this repository -->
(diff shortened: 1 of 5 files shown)


vancem:
I accidentally included TraceEvent.cs, but it has only whitespace changes.  Ignore it.

sharwell:
👍

PR #785, opened 2018-11-06
Revert artificial limitations in RecursionGuard

### LINKED ISSUE #784: "Stack overflow" prevents analysis of real-world heaps
The recursion guard was modified to artificially limit the ability of PerfView to analyze large heaps. This change breaks critical functionality and should be reverted.

https://github.com/Microsoft/perfview/blob/21ad23f98636fbdd7994f7d658be9c2de2e3d30b/src/TraceEvent/Stacks/RecursionGuard.cs#L29

sharwell:
Fixes #784

### DIFF
--- a/src/TraceEvent/Stacks/RecursionGuard.cs
+++ b/src/TraceEvent/Stacks/RecursionGuard.cs
@@ -1,11 +1,7 @@
-﻿using System;
-
-namespace Microsoft.Diagnostics.Tracing.Stacks
+﻿namespace Microsoft.Diagnostics.Tracing.Stacks
 {
     /// <summary>
     /// This structure provides a clean API for a lightweight recursion stack guard to prevent StackOverflow exceptions
-    /// We do ultimately do a stack-overflow to prevent infinite recursion, but it is now under our
-    /// control and much larger than you may get on any one thread stack.  
     /// </summary>
     internal struct RecursionGuard
     {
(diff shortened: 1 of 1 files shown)


vancem:
We have a policy issue to resolve.    Having completely open-ended recursion is problematic because it can make perfView effectively hang depending on input, which is unacceptable (you don't get to say ANY of your data).     Also the viewer is not very happy with 10K nestings.    

The original change tried to address this by dealing with some simple recursion cases in a way that did not cause recursion, and limiting recursion to an arbitrary depth for the rest, (yes, it is inaccurate, but you get to see it, and it tells you that it gave up).   

It seems you have scenarios where this is also unacceptable.   Please give me more details so we can devise an appropriate response.

sharwell:
@vancem A substantial percentage of heaps coming in for low-VM analysis have object graphs deeper than the artificial limits here. If we need to be able to avoid a hang for situations like this, an approach that works for everyone is to make the affected operation cancellable. I'm willing to run the analysis overnight if it comes back with usable results; it's not ideal but at least it doesn't give up and leave me with nothing.

vancem:
It sounds like heap analysis (as opposed to call stack / causality analysis), is your scenario.    Is that correct?

Also you suggest that the results are unusable if they are not complete.   My understanding is that in what will happen if we max out, is that you will end up with 'orphans'  (which I agree require more sophistication on the part of the user to interpret).

While making the operation cancelable is a good step, I don't think having the default be to take however long it takes is a good default (since it is not pay for play, and certainly not very discoverable what your options are).   

For what it is worth: I am OK with the following resolution: 
1. Making sure it is obvious that we have given up.   The idea is that there should be a node in the graph (ideally for both fragments), that indicate that we gave up.   
2. Have a command line option (which can also be part of user config state, so that users can set it 'permanently') which can turn off the limit.   

That way you get 'preliminary' results, and if the analysis shows that it is important they can re-run the analysis with the limit removed.    People like yourself can set the default and not have to think about it again.

sharwell:
@vancem What about just the second part of the resolution? I'm already used to that approach because I always use `/ShowUnknownAddresses`.

vancem:
Sure, you can do just the part you care about.

sharwell:
I looked into implementing a `/NoRecursionGuard` option today, but didn't see a way to do it without introducing non-trivial overhead for performance-sensitive paths.

brianrob:
I added a comment in #784 on how we might proceed on this.  For now, closing this PR.

PR #786, opened 2018-11-06
Increase StackInfoCacheSize to 128K

sharwell:
Reduces the number of cases where `AddSample` falls back to searching for nodes.


brianrob:
@sharwell this change makes sense, but I would like to understand its impact.  Do you have data to show how much better this is?

brianrob:
I'm going to go ahead and close this one.  I think that this would help with heap dump viewing, but will likely hurt other scenarios.  I've seen cases lately where large amounts of memory have been taken up by stack caches, and I want to avoid making this worse.  If we do need this for memory analysis, then we should look into making this configurable.

Issue #779, opened 2018-11-05
Clarification on Microsoft.Diagnostics.Tracing.TraceEvent 2.0.29 license


joshfree:
Posting on behalf of a user who emailed the support link on https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.TraceEvent/ :

> Hello!
> Licensing question here - I see that Trace Event is licensed under the .Net license on the download page (yet has no .js files), but the LICENSE.TXT file contains the MIT license.
> So how is Trace Event licensed?
> Thank you!

cc: @vancem

vancem:
TraceEvent should work just like all the other .NET libraries.    We will fix this.

vancem:
PR #792 harmonizes the nuget package and the source code.   They now both use https://github.com/Microsoft/perfview/blob/master/LICENSE.TXT which is consistent with how licensing works in the https://github.com/dotnet/corefx repo.

vancem:
This will show up in version 2.0.31 or later of the TraceEvent package.

PR #792, opened 2018-11-12
Harmonize the license in the source and in the nuget packages.

vancem:
Fixes issue #779.   This makes us consistent with the rest of .NET Core Framework (https://github.com/dotnet/corefx).

Like the rest of corefx, we ship under the MIT license.

### DIFF
--- a/LICENSE.TXT
+++ b/LICENSE.TXT
@@ -1,6 +1,8 @@
 The MIT License (MIT)
 
-Copyright (c) 2015 Microsoft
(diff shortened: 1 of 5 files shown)


vancem:
@joshfree  @terrajobst  FYI,   Making the PerfView/TraceEvent license consistent with what we do in corefx.  
This puts it under the MIT license.     We had this license in the source, but the nuget package pointed at https://www.microsoft.com/net/dotnet_library_license.htm.

PR #795, opened 2018-11-15
determine which architecture a dump is without relying on exceptions

ranamauro:
determine which architecture a dump is without relying on ApplicationException and falling back to 32 bit

### DIFF
--- a/src/PerfView/memory/HeapSnapshot.cs
+++ b/src/PerfView/memory/HeapSnapshot.cs
@@ -85,19 +85,16 @@ namespace PerfView
             // Determine if we are on a 64 bit system.
             if (Environment.Is64BitOperatingSystem)
             {
-                // TODO FIX NOW.   Find a way of determing which architecture a dump is
-                try
+                bool isDump64 = DumpReader.IsDump64(processDumpFile).GetValueOrDefault();
+                if (isDump64)
                 {
-                    log.WriteLine("********** TRYING TO OPEN THE DUMP AS 64 BIT ************");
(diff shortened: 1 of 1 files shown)


vancem:
Thanks for looking into this, and sorry for the delay in getting back to you.  

I think that this may be a case where the cure is worse than the disease.    This edit basically drags several hundred lines of code about process dumps into the PerfView exectuable, that don't otherwise need to be there.    I would like to keep the loose coupling.    

As you may know, the code for looking at CLR data structures is architecture specific (at least today).  Thus you need to be in a 32 bit process to look at 32 bit dumps and 64 bit process to look at 64 bit dumps.   Because of this PerfView has two version of Heapdump.exe that it launches and it is this subprocess that does all the work.      HeapDump.exe already has logic that can detect if it is in the wrong process type, and my recommendation is that HeapDump.exe should simply return a special error code that indicates this so you can only retry when it is appropriate.   

Here is a commit that gives you the idea that I am going after.

https://github.com/vancem/perfview/commit/4a2bb3fe53942ad034b62a957218cd67748acb0d

I tried testing it but the CLR DAC logic fails before the logic that throws the HeapDumpException is thrown.   Things need to be re-jiggered (maybe putting your logic in there, but hopefully we can reuse existing logic).    I ran out of time to do more, but this is the basic idea if you are still interested in pursuing fixing this.

brianrob:
@ranamauro, thanks for offering this PR.  Based on feedback from @vancem above, I'm going to close this PR.  Please feel free to bring a new PR along the lines of what @vancem suggests, or open an issue if you'd like to discuss further.  Thanks.

PR #807, opened 2018-11-29
Read cpu number only when it's available in Linux event

### LINKED ISSUE #806: Incorrect parsing of Linux cpu-clock event when cpu number is missing
In some configuration `perf` doesn't have `cpu` attribute associated with the `cpu-clock` event, so the merged `perf.data.txt` has events like below

```
dotnet    10/1273  610737.797495:    1001001 cpu-clock:
```

in comparison, another event with cpu number looks like

```
WebAPIvNext 16828/16999 [003]  5620.842122: cpu-clock:
```
Note the cpu number of `[003]`

The code to read cpu number doesn't handle the case where cpu number is missing, thus following values are parsed into incorrect fields.

https://github.com/Microsoft/perfview/blob/ebc8046a8a70853fa1276089f2e0a07ec14f14fa/src/PerfView/OtherSources/Linux/LinuxPerfScriptEventParser.cs#L348-L351

jeremymeng:
Fixes #806

### DIFF
--- a/src/PerfView/OtherSources/Linux/LinuxPerfScriptEventParser.cs
+++ b/src/PerfView/OtherSources/Linux/LinuxPerfScriptEventParser.cs
@@ -346,9 +346,13 @@ namespace Diagnostics.Tracing.StackSources
 
                 // CPU
                 source.SkipWhiteSpace();
-                source.MoveNext(); // Move past the "["
-                int cpu = source.ReadInt();
-                source.MoveNext(); // Move past the "]"
+                int cpu = -1;
+                if (source.PeekString(1) == "[")
+                {
+                    source.MoveNext(); // Move past the "["
+                    cpu = source.ReadInt();
+                    source.MoveNext(); // Move past the "]"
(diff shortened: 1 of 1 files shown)


src/PerfView/OtherSources/Linux/LinuxPerfScriptEventParser.cs
```diff
@@ -346,9 +346,13 @@ private IEnumerable<LinuxEvent> NextEvent(Regex regex, FastStream source)
 
                 // CPU
                 source.SkipWhiteSpace();
-                source.MoveNext(); // Move past the "["
-                int cpu = source.ReadInt();
-                source.MoveNext(); // Move past the "]"
+                int cpu = -1;
+                if (source.PeekString(1) == "[")
```
jeremymeng: Maybe there is a better way so that this is only checked once, instead of on every cpu-clock event.  I assume if one doesn't have cpu number, the rest won't have it either.

PR #810, opened 2018-12-06
Add ReadyToRun Flags to Module and Method Events

brianrob:
@AndyAyersMS added new ReadyToRun flags to the module and method events that will allow us to understand if a module is a ReadyToRun module and if a method was JIT-compiled because its ReadyToRun code was rejected.  These were implemented in https://github.com/dotnet/coreclr/pull/15801.

This PR adds these flags to the appropriate flags enums so that they are resolved in the Events view.

I've also added a few more flags in this PR to match the values present in CoreCLR.


vancem:
Brian I have not checked, but I suspect that the testing failed because there are events that set some of the bits you just defined, and thus some of the 'events' lines now have symbolic names when they did not before.   

You just need to copy the new baselines (run the tests locally and they should fail, and then follow the directions in the failure message to copy the new baseline).    

Including these baseline updates should fix things.

brianrob:
You are correct - just pushed a fix for these.

PR #815, opened 2018-12-12
Fix missing nuget dependency (Microsoft.Win32.Registry)

chrisnas:
If a TraceEvent client is disposing a `TraceEventSession`, a missing assembly exception is thrown when `Close `is calling `ResetWindowsHeapTracingFlags`.
Otherwise, all clients have to add Microsoft.Win32.Registry by hand in their project


chrisnas:
Fixed in 2.0.31

Issue #823, opened 2018-12-16
FocusProcess: expose via UI


ilabutin:
With the recent addition of FocusProcess command line option it would be great to have it exposed in Collection dialog as well.
I've prepared small PR for that. I hope you find it OK to expose this option (disabled by default).

PR #824, opened 2018-12-16
Added UI for the FocusProcess option

ilabutin:
This is simple UI to expose FocusProcess command line option in UI.
Addresses #823

### DIFF
--- a/src/PerfView/SupportFiles/UsersGuide.htm
+++ b/src/PerfView/SupportFiles/UsersGuide.htm
@@ -3831,8 +3831,9 @@
         </li>
         <li>
             The <strong><a id="FocusProcessTextBox">Focus Process TextBox</a></strong> - This is only
-            active for the Collect command.&nbsp; This allows you to only turn on non-Kernel events
-            for a particular process, and thus cut the overhead / size of the collection when there are many
+            active for the Collect command.&nbsp; It is either a decimal process ID or a name of an executable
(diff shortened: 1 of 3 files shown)


src/PerfView/SupportFiles/UsersGuide.htm
```diff
@@ -3827,7 +3827,18 @@ <h2>
         <li>
             The <strong><a id="CommandToRunTextBox">Command TextBox</a></strong> - This is only
             active for the Run command.&nbsp; It is the command line to run after collection
-            has been turned on.&nbsp;&nbsp; This textbox is inactive for the Collect command.&nbsp;
+            has been turned on.&nbsp;&nbsp; This textbox is hidden for the Collect command.&nbsp;
+        </li>
+        <li>
+            The <strong><a id="FocusProcessTextBox">Focus Process TextBox</a></strong> - This is only
```
vancem: Please state before anything else that the focus process is either a decimal process ID or a name of an executable without the Directory part  but WITH the suffix.  (e.g. MyProgram.exe).

Also please put validation in that checks that it is one of these things  and produce an error immediately (rather than simply failing to collect anything which is what would otherwise happen).   

Other than that, I think it looks good.  

Thanks for doing this.

ilabutin: Requested changes are done and pushed here..

vancem:
Thanks for doing this @ilabutin

PR #835, opened 2019-01-02
Various fixes in the generation of manifest files and C# code from them

OfekShilon:
After running the manifest + C# generation on all ~1100 OS-registered providers, it seems ~20% fail to generate manifests, and of those that succeed ~30% fail to generate valid C# code.
This is a start of an attempt to improve the state of these (mighty useful!) features.  All solving real generation problems.
Will add xUnit tests in a future PR.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -43,7 +43,6 @@
     <AzureCoreVersion>1.36.0</AzureCoreVersion>
     <AzureIdentityVersion>1.10.4</AzureIdentityVersion>
     <MicroBuildCoreVersion>0.3.1</MicroBuildCoreVersion>
-    <MicrosoftNETCorePortableCompatibilityVersion>1.0.2</MicrosoftNETCorePortableCompatibilityVersion>
     <MicrosoftIdentityClientVersion>4.58.1</MicrosoftIdentityClientVersion>
     <MicrosoftIdentityClientExtensionsMsalVersion>4.58.1</MicrosoftIdentityClientExtensionsMsalVersion>
     <MicrosoftIdentityModelTokensVersion>7.1.2</MicrosoftIdentityModelTokensVersion>
--- a/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
+++ b/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
@@ -30,7 +30,6 @@
 
     <dependencies>
       <group targetFramework=".NETStandard2.0">
-        <dependency id="Microsoft.NETCore.Portable.Compatibility" version="$MicrosoftNETCorePortableCompatibilityVersion$" />
         <dependency id="Microsoft.Win32.Registry" version="$MicrosoftWin32RegistryVersion$" />
         <dependency id="System.Collections.Immutable" version="$SystemCollectionsImmutableVersion$" />
         <dependency id="System.Diagnostics.Process" version="$SystemDiagnosticsProcessVersion$" />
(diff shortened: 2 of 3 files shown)


OfekShilon:
Is it possible to add a PR with new tests  that do not pass?  (I think it is a per-repository policy in github)

src/TraceParserGen/TraceParserGen.cs
```diff
@@ -1327,6 +1304,10 @@ private class StructInfo : FieldInfo
     /// </summary>
     private SortedDictionary<string, List<Event>> m_eventsByName;
 
-    private static Dictionary<string, string> reservedKeywords = new Dictionary<string, string> { { "object", "Object" }, { "new", "New" }, { "protected", "Protected" } };
+    /// <summary>
+    ///  Used in the generation of C#-valid identifiers
+    /// </summary>
+    static private Microsoft.CSharp.CSharpCodeProvider CS_provider = new Microsoft.CSharp.CSharpCodeProvider();
```
sharwell: 💡 

```suggestion
    private static readonly Microsoft.CSharp.CSharpCodeProvider CS_provider = new Microsoft.CSharp.CSharpCodeProvider();
```

src/TraceParserGen/TraceParserGen.cs
```diff
@@ -945,20 +926,16 @@ private string GetType(List<FieldInfo> versions)
...
-            char c = input[i];
-            if (!Char.IsLetter(c) && !Char.IsDigit(c) && c != '_')
-            {
-                return Regex.Replace(input, @"[^\w\d_]", "");          //Simply remove them.    
-            }
-        }
-        return input;
+        string output = Regex.Replace(input, @"[^\w\d_]", "");
+
+        if (Char.IsDigit(output[0]))
```
sharwell: 💡 

```suggestion
        if (char.IsDigit(output[0]))
```

src/TraceParserGen/TraceParserGen.cs
```diff
@@ -945,20 +926,16 @@ private string GetType(List<FieldInfo> versions)
     /// </summary>
     public static string ToCSharpName(string input)
     {
-        if (input == null)
-        {
-            return null;
-        }
+        if (String.IsNullOrEmpty(input))
```
sharwell: 💡 

```suggestion
        if (string.IsNullOrEmpty(input))
```

src/TraceParserGen/TraceParserGen.cs
sharwell: ❓ Will this map `object` to `Object` or to `@object`?

NN---: It is enough to use HashSet here and just add '@' character in front of reserved name.

src/TraceParserGen/TraceParserGen.cs
OfekShilon: The rest of the code uses Char.IsDigit. Is there a reason to prefer char here?

OfekShilon: Sorry, i see now it is mixed. If you guys prefer this convention I'll stick to it.

src/TraceParserGen/TraceParserGen.cs
OfekShilon: Again - the existing code is mixed, but I'll gladly adhere to this convention.

src/TraceParserGen/TraceParserGen.cs
OfekShilon: In fact it maps `object` to `_object`

brianrob:
@OfekShilon thank you for the contribution and apologies for the delay in responding to this.  Can you please rebase your PR and then see which of your changes are still needed?  #950 also includes changes in this space and it looks like some of them overlap.  Thanks!

PR #822, opened 2018-12-15
Add a userCommand NetPerfToJson

vancem:
This converts a .NetPerf file to a *.netperf.json file that should be very easy to render in tools like speedscope.

You can convert a file with the command
```
    perfView userCommand NetPerfToJson  FILE.netperf
```
It will generate a FILE.netperf.json file.

This format is still a bit experimental.  It will probably be tweeked.

PR #842, opened 2019-01-06
SpeedScopeExporter

adamsitnik:
@vancem this is the initial version, WIP

Sample exported file: https://gist.github.com/adamsitnik/db8f40f4e84fd9c2975b7ce7b8b468c0 (you need to  click the raw button and save it as file and then upload to https://www.speedscope.app/)


src/TraceEvent/Stacks/SpeedScopeExporter.cs
```diff
@@ -0,0 +1,260 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Tracing.Stacks
+{
+    public static class SpeedScopeExporter
+    {
+        internal struct Sample
```
vancem: I would like code to follow the following convention. 
1. Public APIs some first, and generally in an order that makes sense from a consumers' point of view (e.g. constructors first, most important/popular APIs first ...)
2. All public APIs have C# XML doc comments, at least the 'summary' part.  You don't need to do the arguments / return elements of doc comments if the summary sufficiently explains them (you will see that mostly the doc comments only have the summary part).
3. All private/internal things are in a #region private , so that they are collapsed aggressively by VS.  Thus when you look at the API you only see the public stuff (and their comments), which is what you want)
4. When possible (the fields are private)  state (fields) come last in the class (typically in the private region.   This lets you see what is 'in' a class quickly.  
5. Types that are private to one class are nested in that class (in the private region.

Can you rearrange your code to this standard?   Thanks

adamsitnik: > Can you rearrange your code to this standard?

Done

src/TraceEvent/Stacks/SpeedScopeExporter.cs
```diff
@@ -0,0 +1,271 @@
...
+using System.Globalization;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Tracing.Stacks
+{
+    public static class SpeedScopeExporter
+    {
+        /// <summary>
+        /// exports provided StackSource to a https://www.speedscope.app/ format
```
vancem: Can you also include the links to the documentation on the format that you put in an e-mail to me.

src/TraceEvent/Stacks/SpeedScopeExporter.cs
```diff
@@ -0,0 +1,271 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Tracing.Stacks
+{
+    public static class SpeedScopeExporter
```
vancem: Call it SpeedScopeStackSourceWriter to be consistent with XmlStackSourceWriter.

vancem:
Go ahead and make a user command 'NetPerfToSpeedScope' (like 'NewPerfToJson in https://github.com/Microsoft/perfview/pull/822/files) so that  we have something immediately that we can use for demos.  

I am OK with merging this with minimal testing (what you have is enough), as long as there are not known issues.   Do you know of any issues?

vancem:
Let me know when you are ready.   Things that are left are 

1. Change namespace to ....Stacks.Formats
2. Add PerfView user command 

For 2. you can just paste this into the UserCommands.cs file 
```
        /// <summary>
        /// Save Thread stacks from a NetPerf file into a *.speedscope.json file.
        /// </summary>
        /// <param name="netPerfFileName">The ETL file to convert</param>
        public void NetperfToSpeedScope(string netPerfFileName)
        {
            string outputName = Path.ChangeExtension(netPerfFileName, ".speedscope.json");

            string etlxFileName = TraceLog.CreateFromEventPipeDataFile(netPerfFileName);
            using (var eventLog = new TraceLog(etlxFileName))
            {
                var startStopSource = new MutableTraceEventStackSource(eventLog);
                // EventPipe currently only has managed code stacks.
                startStopSource.OnlyManagedCodeStacks = true;

                var computer = new SampleProfilerThreadTimeComputer(eventLog, App.GetSymbolReader(eventLog.FilePath));
                computer.GenerateThreadTimeStacks(startStopSource);

                SpeedScopeExporter.WriteStackViewAsJson(startStopSource, outputName);

                LogFile.WriteLine("[Converted {0} to {1}  Use https://www.speedscope.app/ to view.]", netPerfFileName, outputName);
            }
        }
```

adamsitnik:
@vancem I have fixed all issues, it's ready for final review and merge (if it passes the review ;) )

vancem:
LGTM

Thanks Adam for doing this!   This gives us a viewing 'story' on Linux and Mac for netperf files.

PR #843, opened 2019-01-06
Add PerfViewJS and delete bitrotted TraceEventAPIServer and HtmlJs

mjsabby:
This brings in PerfViewJS into PerfView.

PerfViewJS is a web viewer frontend for TraceEvent data, which also makes it an ETL and NetPerf data viewer.

PerfViewJS is a reincaration of HtmlJS and TraceEventAPIServer (both of which are being deleted in this PR), which itself is an open-source extraction of Bing.com's internal production profiler viewer (BPerf). 

With PerfViewJS we see a viable cross-platform webviewer that gets PerfView's most important features:

* Stack Viewer with all the filtering capabilities, including Drill Into Exclusive and Inclusive Samples.
* Event Viewer with filtering capabilities of the original viewer.

The next steps are to develop it in the open. The first few next steps are:

* Make it a global tool so a developer can do dotnet perfview myETLFile and it launches the browser
* Implement right-click support
* Basic integration with Chromium Embedded Framework to make it into a client-side app.
* Get it out there for Windows & Linux usage to see what are the most annoying parts.

NOTE: This is not meant to replace PerfView on Windows. PerfView has years of little things that make the experience a delight that will take time to replicate. That said, on Linux there are clear advantages in using a web tool because often Linux users are analyzing server perf, so this tool will allow to serve their server data in a convenient web first manner.


mjsabby:
@vancem All the changes are in HtmlJS (deleted), TraceEventAPIServer (deleted), and new directory (PerfViewJS). I wanted to get in so others could play with it and then we can see how much work is needed to get it polished.

I don't expect you to actually review the contents since it's overwhelming, suffice to say that it is very similar to the code Kaelin had imported from BPerf, with bitrot and various bug fixes from the past 2 years.

PR #844, opened 2019-01-07
Code changes to enable use of Segmented Lists

mpeyrotc:
These changes permit us to use Segmented Lists in place of Growable Arrays to reduce the chance of running out of memory when using the MemoryGraph class. In order to not break the existing API of Perfview I made GrowableArray and SegmentedList share a common interface. When you create a Graph or MemoryGraph, you can use the overloaded constructor to specify that it should store the internal data in a SegmentedList instead of the GrowableArray. This way I do not require any client to change its current programs that use PerfView. The other classes I create are used to support the use of streams with segmented lists because they run into the same problem.

We added the sln and csproj files to HeapDump to enable the building of
the dll. Since it is only consumed by VS it only targets net45, since
net standard is missing some of its dependencies.


src/FastSerialization/GrowableArray.cs
```diff
@@ -18,7 +37,7 @@ namespace System.Collections.Generic
 #if GROWABLEARRAY_PUBLIC
     public
 #endif
-    struct GrowableArray<T>
+    struct GrowableArray<T> : IMutableList<T>
```
sharwell: ⚠️  This one is a possible problem. If `GrowableArray<T>` gets boxed, it could get corrupted. Implementing an interface makes it much easier for this to occur.

src/MemoryGraph/MemoryGraph.cs
```diff
@@ -146,7 +161,7 @@ void IFastSerializable.FromStream(Deserializer deserializer)
 
         // This array survives after the constructor completes
         // TODO Fold this into the existing blob. Currently this dominates the Size cost of the graph!
-        protected GrowableArray<Address> m_nodeAddresses;
+        protected IMutableList<Address> m_nodeAddresses;
```
sharwell: 💡 I believe performance would be improved by simply using `SegmentedList<Address>` here and removing the ability to use `GrowableList<Address>`

src/MemoryGraph/graph.cs
```diff
@@ -141,6 +141,10 @@ public virtual NodeType AllocTypeNodeStorage()
         /// </summary>
         public int TotalNumberOfReferences { get { return m_totalRefs; } }
 
+        protected int m_segmentSize;
```
sharwell: ❔ Can we inline a constant here? It doesn't seem especially advantageous to make it configurable.

src/MemoryGraph/graph.cs
```diff
@@ -141,6 +141,10 @@ public virtual NodeType AllocTypeNodeStorage()
...
+
+        protected bool m_isUsingSegmentedList = false;
```
sharwell: 💡 Consider inlining this as `true`.

src/FastSerialization/SegmentedMemoryStreamWriter.cs
```diff
@@ -0,0 +1,106 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+
+namespace FastSerialization
+{
+    public class SegmentedMemoryStreamWriter : IMemoryStreamWriter
```
sharwell: ❔ Can we avoid adding this for now, in favor of using the one from #307 based on memory-mapped files? I would send a separate pull request for the updated streams.

src/MemoryGraph/graph.cs
mpeyrotc: I have been using a size of 131_072 and it has worked fine, however this value does affect performance and the amount of fragmentation that occurs, that is why I thought it best to be configurable.

src/MemoryGraph/graph.cs
mpeyrotc: If we only use one stream class and one internal data structure this flag will be removed.

src/FastSerialization/GrowableArray.cs
sharwell: For the specific locations where `SegmentedList` was introduced, I don't believe a performance problem will occur. In addition, the segmentation algorithm used by this type is quite fast.

src/MemoryGraph/MemoryGraph.cs
sharwell: I'm not convinced the indirection is observably more costly than the used of interface dispatch. 😄

src/MemoryGraph/graph.cs
sharwell: We should consider a value which avoids the creation of pages in the Large Object Heap. I believe the largest such value would be 8192.

mpeyrotc: I was looking into this and I found that the LOH threshold is 85,000 bytes (According to [here](https://blogs.msdn.microsoft.com/dotnet/2011/10/03/large-object-heap-improvements-in-net-4-5/)). In that case limiting the size to 8192 elements would make it not enter LOH. Is that what you were referring to? Avoid entering the LOH?

src/MemoryGraph/graph.cs
```diff
@@ -140,6 +140,10 @@ public virtual NodeType AllocTypeNodeStorage()
         /// The number of references (arcs) in the graph
         /// </summary>
         public int TotalNumberOfReferences { get { return m_totalRefs; } }
+        /// <summary>
+        /// Each segment added to the segmented list used in this class will have a size as specified by this constant.
+        /// </summary>
+        protected static readonly int m_segmentSize = 8_192;
```
vancem: Make it a const.   I don't think it is interesting to expose this configuration parameter (and we can do so easily later if we ever care to).    Also put a rational for this choice in the comment  (that because T is 8 bytes, it keeps all the segments  ~64K and thus out of the large object heap).  

When you make it const, rename it to be SegmentSize (since it is no longer an member field)

src/Utilities/Utilities.csproj
```diff
@@ -18,7 +18,11 @@
 
   <ItemGroup>
     <!-- *** SourceLink Support *** -->
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All"/>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
+  </ItemGroup>
+  <ItemGroup>
+    <WCFMetadata Include="Connected Services\" />
+    <PackageReference Include="Microsoft.Diagnostics.Runtime" Version="$(MicrosoftDiagnosticsRuntimeVersion)" />
   </ItemGroup>
```
vancem: This does not seem needed.  Also I don't think you are changing anything in Utilities, so you should not need even the AllowUnsafeBlocks change below either.  I would recommend undoing any changes to this file.

vancem:
LGTM

PR #851, opened 2019-01-15
Memory mapped streams

sharwell:


src/FastSerialization/StreamReaderWriter.cs
```diff
@@ -567,6 +587,661 @@ protected override void Dispose(bool disposing)
...
+            _offset = (int)offset;
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                var view = _view;
+                _view = null;
+                view?.Dispose();
```
sharwell: 💡 

```csharp
Interlocked.Exchange(ref _view, null)?.Dispose();
```

src/FastSerialization/StreamReaderWriter.cs
```diff
@@ -567,6 +587,661 @@ protected override void Dispose(bool disposing)
...
+
+                var file = _file;
+                _file = null;
+                file?.Dispose();
```
sharwell: 💡 

```csharp
Interlocked.Exchange(ref _file, null)?.Dispose();
```

src/HeapDump/Debugger/DumpDebugging/utility.cs
```diff
@@ -100,14 +99,10 @@ public CorDebugPlatform GetPlatform()
...
-            if (bytesRead == 0)
-            {
-                throw new System.Runtime.InteropServices.COMException("Could not read memory requested at address " + address + ".");
-            }
-
-            return bytesRead;
+            uint result = m_reader.ReadPartialMemory(address, buffer, bytesRequested);
+            byte[] data = new byte[bytesRequested];
+            Marshal.Copy(buffer, data, 0, (int)bytesRequested);
+            return result;
```
sharwell: 💭 It seems like all of the changes in this file are unnecessary and can be omitted.

src/HeapDump/GCHeapDumper.cs
```diff
@@ -296,9 +297,113 @@ private CollectionMetadata DumpHeapFromProcessDump(string processDumpFile)
...
+        }
+    }
+
     private void InitializeClrRuntime(string processDumpFile, out DataTarget target, out ClrRuntime runtime)
     {
-        target = DataTarget.LoadCrashDump(processDumpFile);
+        var reader = (IDataReader)Activator.CreateInstance(typeof(DataTarget).Assembly.GetType("Microsoft.Diagnostics.Runtime.DbgEngDataReader"), processDumpFile);
+        var debugInterface = reader.GetType().GetProperty("DebuggerInterface", BindingFlags.Instance | BindingFlags.NonPublic).GetValue(reader);
+        var wrapper = new DataReaderWrapper(reader);
+        target = (DataTarget)typeof(DataTarget).GetMethod("CreateFromReader", BindingFlags.Static | BindingFlags.NonPublic).Invoke(null, new[] { wrapper, debugInterface });
```
sharwell: 😟 Maybe there are new public APIs to avoid this reflection?

sharwell: On further review, it appears these changes were part of a separate investigation, and not required for this pull request.

sharwell:
@vancem @mpeyrotc This should now be ready for review. Please feel free to ask questions about anything.

src/FastSerialization/FastSerialization.cs
```diff
@@ -1126,6 +1127,17 @@ public IFastSerializable GetEntryObject()
...
 
         // For FromStream method bodies.  
+        public void Read(byte[] buffer, int offset, int length)
+        {
+#if DEBUG
+            StreamLabel label = reader.Current;
+#endif
+            reader.Read(buffer, offset, length);
+#if DEBUG
+            Log("<ReadByte Value=\"" + "[...]" + "\" StreamLabel=\"0x" + label.ToString("x") + "\"/>");
```
vancem: ReadByte -> Read

```suggestion
            Log("<Read Value=\"" + "[...]" + "\" StreamLabel=\"0x" + label.ToString("x") + "\"/>");
```

src/FastSerialization/FastSerialization.cs
```diff
@@ -153,6 +153,7 @@ interface IStreamWriter : IDisposable
 #endif
     interface IStreamReader : IDisposable
     {
+        void Read(byte[] data, int offset, int length);
```
vancem: Can you put this physically at the end of the interface?   I would consider this an 'advanced' API and I like to the simpler more common APIs first.      

Put a doc comment on this, and also indicate that the 'length' is likely to be restricted by the internal buffer size in the particular StreamReader.   Values over 8K need closer coordination with the reader. 

I don't know where you use Read(), but I made something called 'PinnedStreamReader' that tries to solve the same problem (getting blobs from the stream) It has 'GetPointer(length)' API.     It avoids the buffer copy, which is nice.  

That was in the days before Span<byte>.   Now I would be tempted to implement 'Span<byte> GetBytes(length) which is safe.    

I understand that that is a bigger change, but we are changing an interface here, and such contracts live forever.  (we could write a Extension method that implements Read() in terms of it, so you would only need to change the implementers.  

Hmm, something to think about...

sharwell:
@vancem Please review the change from `byte[]` to `Span<byte>` before I split the new class out to its own file. Let me know if that's the direction you had in mind.

src/FastSerialization/FastSerialization.cs
```diff
@@ -174,6 +174,10 @@ interface IStreamReader : IDisposable
         /// </summary>
         string ReadString();
         /// <summary>
+        /// Read a span of bytes from the stream.
+        /// </summary>
+        void Read(Span<byte> span);
```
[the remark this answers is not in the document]
vancem: Let me know how you want to handle it.   As mentioned, I am OK with your approach above.   Split out the new Mapped Reader/writers into their own file(s), and make sure that Read() works if the size of the target read is larger than the buffer size of the reader/writer (there is a loop until done).   After that I am OK with it.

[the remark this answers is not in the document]
sharwell: > An alternative is `Span<byte> GetData(int length)`, where the stream still owns the buffer and the returned Span is only valid until the next operation on the Stream.

I'm not a fan of this approach because the usable life of the `Span<byte>` is difficult to work with. It seems more likely to be a source of bugs without a clear performance need.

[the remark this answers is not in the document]
vancem: I am OK with Read(Span).

vancem:
The only thing left is splitting out the MemoryMapped classes into their own files, which can be optional.   Let me know if you want to do that or not.    Then one way or the other I will merge.

sharwell:
@vancem I am going to fix the merge conflict and split the source file. Please confirm that you are also OK with the added reference to System.Memory (I assume so since you want `Span<T>`).

src/FastSerialization/StreamReaderWriter.cs
```diff
@@ -567,6 +588,653 @@ protected override void Dispose(bool disposing)
     }
 #endif
 
+    public class MemoryMappedFileStreamReader : IStreamReader
```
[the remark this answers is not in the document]
sharwell: These are now split. I'll split the rest of the types in **StreamReaderWriter.cs** as a separate pull request.

vancem:
LGTM.   Thanks Sam.

Issue #845, opened 2019-01-08
Typo in inlining code


davmason:
I was looking at the source to see how perfview constructs a name out of the TraceData classes, and I think this is a typo:
https://github.com/Microsoft/perfview/blob/d9cde4ef13b80b72e05ea3a4d64790a13806a33c/src/TraceEvent/Computers/TraceManagedProcess.cs#L1398-L1407

Shouldn't it be data.**InlineeNamespace** on line 1405?

vancem:
Yes, that is a typo.  I have it fixed locally and will resolve when I check this in.
Thanks for alerting us.

vancem:
Fixed by #854

PR #854, opened 2019-01-16
Small fixes

vancem:
Fix logic that allows EventSources to skip the * syntax (e.g. /ProviderName=XXX will assume XXX is
and eventSource (with a warning in the log), if it could not find it in the OS table).    This was the
intent of code added long ago but it was broken.

Renamed Insure->Ensure

Small fix in inlining processing logic.


vancem:
Fixes #845

Issue #780, opened 2018-11-05
Retrieve type of unmanaged HeapAlloc events


kayle:
PerfView currently supports tracking native allocations/stacks.

I investigated what it would take to output the types of allocations as well (e.g. int[], MyStruct, etc). I believe Visual Studio is using the same ETW events (https://docs.microsoft.com/en-us/visualstudio/profiling/memory-usage?view=vs-2017), but it is non-obvious to me how to get from AllocAddress + pdb to a c++ type.

vancem:
Indeed it does sound like VS knows something that PerfView does not with respect to native allocation instrumentation (since it can get the type of the allocation, and PerfView can't).   

I would start by making a very simple demo (that say allocate a lot of just a couple paraticular types in native code).   Just having a PerfView CPU trace (use /SessionName=MySession and a PerfView Ver > 2.0.27 to avoid conflicts over the kernel session), will be useful to determine what they are doing.    If we can look at their ETL file even better.  

Most likely there is either 
1. more information being logged at allocation time. OR
2. It is deducing it from the allocation stack (possibly by having addition PDB data).

kayle:
There is an etl file created by Visual Studio, and it shows HeapAlloc events +stacks as the primary info logged, so I believe (2) the pdb is supplying the missing type information and not (1). 

This header file mentioned [here](https://blogs.msdn.microsoft.com/vcblog/2016/05/25/tracking-custom-memory-allocations-with-visual-studio-15-preview-2/) allows collecting the same info for a custom heap:
https://msdnshared.blob.core.windows.net/media/2016/05/VSCustomNativeHeapEtwProvider.h

Is there someone familiar with the c++ pdb/visual studio interaction that could chime in?

kayle:
I should've read the previously linked page more closely. It specifies exactly how it works:

> At each call to an allocator function, the address of the call site, the size of the call instruction, and the typeid of the new object are emitted into a new S_HEAPALLOCSITE symbol inside the PDB.  When the Windows heap code emits an ETW event with a call stack for an allocation, the memory tool walks the call stack looking for a return address matching an S_HEAPALLOCSITE symbol.  The typeid in the symbol determines the runtime type of the allocation.

Using the data referenced from https://github.com/KirillOsenkov/Dia2Dump/blob/master/PrintSymbol.cpp
```
////////////////////////////////////////////////////////////
// Print a callsite symbol info: SEG:OFF, RVA, type
//
[...]

PR #857, opened 2019-01-17
track type of native allocations when available

kayle:
This PR demonstrates how to retrieve the type of native allocations as described in #780 

This would be useful functionality to have in the TraceEvent nuget package to allow for scenarios like https://github.com/dotnet/BenchmarkDotNet/issues/457

There's a few questions in the PR on where certain functionality belongs.

### DIFF
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -5713,6 +5713,9 @@ table {
                 var heapParser = new HeapTraceProviderTraceEventParser(eventSource);
                 Dictionary<Address, StackSourceSample> lastHeapAllocs = null;
 
+                Dictionary<TraceModuleFile, List<HeapAllocationSite>> allocationSites = null;
+                var allocationTypeNames = new Dictionary<StackSourceCallStackIndex, string>();
+
                 Address lastHeapHandle = 0;
 
                 float peakMetric = 0;
@@ -5729,10 +5732,27 @@ table {
                                         allocs = GetHeap(data.HeapHandle, heaps, ref lastHeapAllocs, ref lastHeapHandle);
                                     }
 
+                                    if (allocationSites == null)
+                                    {
+                                        var symReader = GetSymbolReader(log, SymbolReaderOptions.CacheOnly);
+                                        allocationSites = new Dictionary<TraceModuleFile, List<HeapAllocationSite>>();
+                                        foreach (var item in data.Process().LoadedModules)
+                                        {
+                                            if (symReader.FindSymbolFilePathForModule(item.FilePath) is string pdb &&
+                                                symReader.OpenNativeSymbolFile(pdb) is NativeSymbolModule symbolFile)
+                                            {
+                                                var sites = symbolFile.GetHeapAllocationSites().ToList();
+                                                if (sites.Any())
+                                                    allocationSites[item.ModuleFile] = sites;
+                                            }
+                                        }
+                                    }
+
                                     sample.TimeRelativeMSec = data.TimeStampRelativeMSec;
                                     sample.Metric = data.AllocSize;
-                                    var nodeIndex = stackSource.Interner.FrameIntern(GetAllocName((uint)data.AllocSize));
-                                    sample.StackIndex = stackSource.Interner.CallStackIntern(nodeIndex, stackSource.GetCallStack(data.CallStackIndex(), data));
+                                    var callerIndex = stackSource.GetCallStack(data.CallStackIndex(), data);
+                                    var nodeIndex = stackSource.Interner.FrameIntern(GetAllocationType() ?? GetAllocName((uint)data.AllocSize));
+                                    sample.StackIndex = stackSource.Interner.CallStackIntern(nodeIndex, callerIndex);
                                     var addedSample = stackSource.AddSample(sample);
                                     allocs[data.AllocAddress] = addedSample;
 
@@ -5744,6 +5764,27 @@ table {
                                     }
                                     sumCumMetric += cumMetric;
                                     cumCount++;
+
+                                    string GetAllocationType()
+                                    {
+                                        if (!allocationTypeNames.TryGetValue(callerIndex, out var typeName))
+                                        {
+                                            var cs = data.CallStack();
+                                            while (cs?.CodeAddress.ModuleFile is TraceModuleFile module)
+                                            {
+                                                if (allocationSites.TryGetValue(module, out var heapAllocationSites) &&
+                                                    heapAllocationSites.FirstOrDefault(x => x.Rva == cs.CodeAddress.Address - module.ImageBase)?.TypeName is string name)
+                                                {
+                                                    typeName = name;
+                                                }
+                                                cs = cs.Caller;
+                                            }
+
+                                            allocationTypeNames[callerIndex] = typeName;
+                                        }
+
+                                        return typeName;
+                                    }
                                 };
 
                 heapParser.HeapTraceFree += delegate (HeapFreeTraceData data)
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -5713,7 +5713,7 @@ table {
                 var heapParser = new HeapTraceProviderTraceEventParser(eventSource);
                 Dictionary<Address, StackSourceSample> lastHeapAllocs = null;
 
-                Dictionary<TraceModuleFile, NativeSymbolModule> loadedModules = new Dictionary<TraceModuleFile, NativeSymbolModule>();
+                var loadedModules = new Dictionary<TraceModuleFile, NativeSymbolModule>();
                 var allocationTypeNames = new Dictionary<CallStackIndex, string>();
                 var symReader = GetSymbolReader(log);
 
@@ -5750,11 +5750,17 @@ table {
                                     sumCumMetric += cumMetric;
                                     cumCount++;
 
+                                    // Performs a stack crawl to match the best typename to this allocation. 
+                                    // Returns null if no typename was found.
+                                    // This updates loadedModules and allocationTypeNames. It reads symReader/eventLog.
                                     string GetAllocationType(CallStackIndex csi)
                                     {
                                         if (!allocationTypeNames.TryGetValue(csi, out var typeName))
                                         {
-                                            for (var current = csi; current != CallStackIndex.Invalid; current = eventLog.CallStacks.Caller(current))
+                                            const int frameLimit = 25; // we'll search the first 25 frames for the best match
+
+                                            int frameCount = 0;
+                                            for (var current = csi; current != CallStackIndex.Invalid && frameCount < frameLimit; current = eventLog.CallStacks.Caller(current), frameCount++)
                                             {
                                                 var module = eventLog.CodeAddresses.ModuleFile(eventLog.CallStacks.CodeAddressIndex(current));
                                                 if (module == null)
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -5713,8 +5713,9 @@ table {
                 var heapParser = new HeapTraceProviderTraceEventParser(eventSource);
                 Dictionary<Address, StackSourceSample> lastHeapAllocs = null;
 
-                Dictionary<TraceModuleFile, List<HeapAllocationSite>> allocationSites = null;
-                var allocationTypeNames = new Dictionary<StackSourceCallStackIndex, string>();
+                Dictionary<TraceModuleFile, NativeSymbolModule> loadedModules = new Dictionary<TraceModuleFile, NativeSymbolModule>();
+                var allocationTypeNames = new Dictionary<CallStackIndex, string>();
+                var symReader = GetSymbolReader(log);
 
                 Address lastHeapHandle = 0;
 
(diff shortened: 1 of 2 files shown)


src/PerfView/PerfViewData.cs
```diff
@@ -5729,10 +5732,27 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                                         allocs = GetHeap(data.HeapHandle, heaps, ref lastHeapAllocs, ref lastHeapHandle);
                                     }
 
+                                    if (allocationSites == null)
+                                    {
+                                        var symReader = GetSymbolReader(log, SymbolReaderOptions.CacheOnly);
```
kayle: The addresses/type names need to be extracted from each pdb before HeapAlloc events are processed. This implies that if a native pdb is resolved later, we need to process the trace again. I'm not sure how this should be handled. Maybe this should be in LoadSymbolsForModule?

src/PerfView/PerfViewData.cs
```diff
@@ -5744,6 +5764,27 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
+                                    {
+                                        if (!allocationTypeNames.TryGetValue(callerIndex, out var typeName))
+                                        {
+                                            var cs = data.CallStack();
+                                            while (cs?.CodeAddress.ModuleFile is TraceModuleFile module)
+                                            {
+                                                if (allocationSites.TryGetValue(module, out var heapAllocationSites) &&
+                                                    heapAllocationSites.FirstOrDefault(x => x.Rva == cs.CodeAddress.Address - module.ImageBase)?.TypeName is string name)
+                                                {
+                                                    typeName = name;
```
kayle: Note that this currently resolves most allocations from STL containers as type "void" because the std::allocator<T> implementation uses void* as the return type. I think the Visual Studio implementation of the feature may special case this to extract the T, but I wanted to keep this simple for a first pass.

src/TraceEvent/Symbols/NativeSymbolModule.cs
```diff
@@ -26,6 +26,30 @@ namespace Microsoft.Diagnostics.Symbols
     /// </summary>
     public unsafe class NativeSymbolModule : ManagedSymbolModule
     {
+        /// <summary>
+        /// Retrieves the S_HEAPALLOCSITE information from the pdb as described here:
+        /// https://docs.microsoft.com/en-us/visualstudio/profiling/custom-native-etw-heap-events?view=vs-2017
```
pharring: Nit: Remove the "/en-us" part of the URL and the "?view=vs-2017" query param. The URL will still work and it will redirect to any localized version and future VS version as necessary.

sharwell:
```suggestion
        /// https://docs.microsoft.com/visualstudio/profiling/custom-native-etw-heap-events 
```

src/PerfView/PerfViewData.cs
```diff
@@ -5713,6 +5713,9 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                 var heapParser = new HeapTraceProviderTraceEventParser(eventSource);
                 Dictionary<Address, StackSourceSample> lastHeapAllocs = null;
 
+                Dictionary<TraceModuleFile, List<HeapAllocationSite>> allocationSites = null;
```
vancem: Making that returned value a List is inefficient (you end up scanning it later on).   There seems to be no reason for it to not be a Dictionary<uint, string> (mapping the RVA to the type name),   Simply add all the entries returned from  GetHeapAllocationSites  to that dictionary).

src/PerfView/PerfViewData.cs
vancem: I think that the lookup of the AllocationSite info can be lazy. (as you encounter a new TraceModuleFile you add that information to 'allocationSites'.

src/PerfView/PerfViewData.cs
```diff
@@ -5729,10 +5732,27 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                                         allocs = GetHeap(data.HeapHandle, heaps, ref lastHeapAllocs, ref lastHeapHandle);
                                     }
 
+                                    if (allocationSites == null)
+                                    {
+                                        var symReader = GetSymbolReader(log, SymbolReaderOptions.CacheOnly);
+                                        allocationSites = new Dictionary<TraceModuleFile, List<HeapAllocationSite>>();
+                                        foreach (var item in data.Process().LoadedModules)
```
vancem: Remove this foreach and instead add the allocation sites lazily below.

src/PerfView/PerfViewData.cs
```diff
@@ -5744,6 +5764,27 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                                     }
                                     sumCumMetric += cumMetric;
                                     cumCount++;
+
+                                    string GetAllocationType()
```
vancem: I don't actually like local functions that get too much of their state implicitly via variable capture).   In particular at least pass things that represent non-contextual state (in this case the callerIndex).

src/PerfView/PerfViewData.cs
```diff
@@ -5744,6 +5764,27 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
+                                    {
+                                        if (!allocationTypeNames.TryGetValue(callerIndex, out var typeName))
+                                        {
+                                            var cs = data.CallStack();
+                                            while (cs?.CodeAddress.ModuleFile is TraceModuleFile module)
+                                            {
+                                                if (allocationSites.TryGetValue(module, out var heapAllocationSites) &&
+                                                    heapAllocationSites.FirstOrDefault(x => x.Rva == cs.CodeAddress.Address - module.ImageBase)?.TypeName is string name)
```
vancem: This can now be a simple lookup by Rva.

src/PerfView/PerfViewData.cs
```diff
@@ -5744,6 +5764,27 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
                                     cumCount++;
+
+                                    string GetAllocationType()
+                                    {
+                                        if (!allocationTypeNames.TryGetValue(callerIndex, out var typeName))
+                                        {
+                                            var cs = data.CallStack();
+                                            while (cs?.CodeAddress.ModuleFile is TraceModuleFile module)
+                                            {
+                                                if (allocationSites.TryGetValue(module, out var heapAllocationSites) &&
```
vancem: Here is where I would put the lazy initialization of the heapAllocationSites (if you can't find the module file)  then populated it (a present but null value means its was populated as empty).

src/PerfView/PerfViewData.cs
```diff
@@ -5744,6 +5764,27 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                                     }
                                     sumCumMetric += cumMetric;
                                     cumCount++;
+
+                                    string GetAllocationType()
+                                    {
+                                        if (!allocationTypeNames.TryGetValue(callerIndex, out var typeName))
+                                        {
+                                            var cs = data.CallStack();
```
vancem: Generally I don't use CallStack() which creates a object for each frame), but rather use data.CallStackIndex() and use eventlog.CallStacks.CodeAddress(CallStackIndex) to fetch things.   This avoids the extra overhead of making (and deleting) the call stack object for every frame of every stack that you need to look up.

src/PerfView/PerfViewData.cs
```diff
@@ -5729,10 +5732,27 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
+                                    if (allocationSites == null)
+                                    {
+                                        var symReader = GetSymbolReader(log, SymbolReaderOptions.CacheOnly);
+                                        allocationSites = new Dictionary<TraceModuleFile, List<HeapAllocationSite>>();
+                                        foreach (var item in data.Process().LoadedModules)
+                                        {
+                                            if (symReader.FindSymbolFilePathForModule(item.FilePath) is string pdb &&
+                                                symReader.OpenNativeSymbolFile(pdb) is NativeSymbolModule symbolFile)
+                                            {
+                                                var sites = symbolFile.GetHeapAllocationSites().ToList();
```
vancem: Consider:   Maybe you just have NativeSymbolModule implement GetTypeForHeapAllocationSite(uint rva), (and a bool HasHeapAllocationSiteInfo)   Then you don't need to enumerate the tuples.  Either you can use the DIA interface to look up the allocation sites by RVA (it probably offers that, or at worse, you make a NativeSymbolModule hold the RVA->string dictionary.    That avoids even needing to define the HeapAllocation struct (which is just a copy anyway).

vancem:
Thanks for following up to make this PR.   I like the general ideal, and commented on some of the details.

src/PerfView/PerfViewData.cs
kayle: yes, Dictionary<uint, string> is all I need here. I started with a list because I wasn't getting direct matches on rva until I figured out that the sym.length needed to be added in.

src/PerfView/PerfViewData.cs
```diff
@@ -5713,6 +5713,10 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                 var heapParser = new HeapTraceProviderTraceEventParser(eventSource);
                 Dictionary<Address, StackSourceSample> lastHeapAllocs = null;
 
+                Dictionary<TraceModuleFile, NativeSymbolModule> loadedModules = new Dictionary<TraceModuleFile, NativeSymbolModule>();
+                var allocationTypeNames = new Dictionary<CallStackIndex, string>();
+                var symReader = GetSymbolReader(log);
```
kayle: Is this the appropriate way to get a SymbolReader? The CacheOnly argument can skip pdbs, which results in not finding any typenames for allocations from those modules. Even if I later load symbols for a module, I don't see a way to refresh the entire stackSource.

src/PerfView/PerfViewData.cs
```diff
@@ -5744,6 +5749,39 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
+                                                var module = eventLog.CodeAddresses.ModuleFile(eventLog.CallStacks.CodeAddressIndex(current));
+                                                if (module == null)
+                                                {
+                                                    continue;
+                                                }
+
+                                                if (!loadedModules.TryGetValue(module, out var symbolModule))
+                                                {
+                                                    loadedModules[module] = symbolModule =
+                                                        (module.PdbSignature != Guid.Empty
```
kayle: I expected there to be a direct call to get a SymbolModule from a TraceModuleFile, but I didn't find one. This is my best guess based on some other areas of the code.

src/TraceEvent/Symbols/NativeSymbolModule.cs
```diff
@@ -1175,12 +1203,173 @@ public byte[] GetTypeMDTokenMap()
         private ManagedSymbolModule m_managedPdb;
         private bool m_managedPdbAttempted;
 
-        internal SymbolReader m_reader;
-        internal IDiaSession m_session;
-        private IDiaDataSource3 m_source;
-        private IDiaEnumSymbolsByAddr m_symbolsByAddr;
+        internal readonly IDiaSession m_session;
+        private readonly SymbolReader m_reader;
```
kayle: A few style questions for this area:
1. Is it good to switch to constructors to allow for readonly fields?
2. Any reason for the bool/data field combination instead of just using Lazy<T>
3. Should the m_heapAllocationSites initializer be embedded in constructor or extracted to static method?

kayle:
Thanks for the feedback. I left a few more questions as comments in the updated files.

src/TraceEvent/Symbols/NativeSymbolModule.cs
```diff
@@ -1175,12 +1203,173 @@ public byte[] GetTypeMDTokenMap()
...
+        private readonly IDiaDataSource3 m_source;
+        private readonly IDiaEnumSymbolsByAddr m_symbolsByAddr;
+        private readonly Lazy<IReadOnlyDictionary<uint, string>> m_heapAllocationSites; // rva => typename
 
         #endregion
+
+        /// <summary>
```
vancem: Please put this in the private region before the private fields (I want fields to be last in the class so they are easy to find).

src/PerfView/PerfViewData.cs
```diff
@@ -5744,6 +5749,39 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                                     }
                                     sumCumMetric += cumMetric;
                                     cumCount++;
+
+                                    string GetAllocationType(CallStackIndex csi)
```
vancem: Give this a method comment, if only so it is clear that this code is a nested function.   mention any local variables that it is mutating (in particular loadedModules and allocationTypeNames).

src/PerfView/PerfViewData.cs
```diff
@@ -5744,6 +5749,39 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
...
+                                    {
+                                        if (!allocationTypeNames.TryGetValue(csi, out var typeName))
+                                        {
+                                            for (var current = csi; current != CallStackIndex.Invalid; current = eventLog.CallStacks.Caller(current))
```
vancem: Is there a reasonable bound we can put on number of frames of the stack crawl?  (e.g. are the number of frames below the alloc allways less than say 10?)   That would be nice as it cuts down search by potentially a large number (stacks are often 50 or more, and we are searching every unique stack).

src/PerfView/PerfViewData.cs
```diff
@@ -5713,6 +5713,10 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                 var heapParser = new HeapTraceProviderTraceEventParser(eventSource);
                 Dictionary<Address, StackSourceSample> lastHeapAllocs = null;
 
+                Dictionary<TraceModuleFile, NativeSymbolModule> loadedModules = new Dictionary<TraceModuleFile, NativeSymbolModule>();
```
vancem: I won't require this, but you can turn loadedModules into an array of NativeSymbolModule.   Every TraceModuleFile has a Index property which is a small integer whose bound is given by TraceModuleFiles.Count.   Thus you can turn the dictionary into an array.

vancem:
Only a few minor things (mentioned in line comments) left.

src/PerfView/PerfViewData.cs
kayle: Yes, but unfortunately I think 10 is a bit too small. From my limited testing, 15 frames was the longest chain I saw, so I'm thinking of starting with a limit of 20.

Name
 wchar_t
+ wow64!Wow64SystemServiceEx
 + wow64cpu!ServiceNoTurbo
  + wow64cpu!BTCpuSimulate
   + wow64!RunCpuSimulation
    + wow64!Wow64LdrpInitialize
     + ntdll!LdrpInitializeProcess
      + ntdll!??_LdrpInitialize
       + ntdll!LdrpInitialize
        + ntdll!LdrInitializeThunk
         + ntdll!NtTraceEvent
          + ntdll!RtlpLogHeapAllocateEvent
           + ntdll!??RtlpAllocateHeapInternal
            + ntdll!RtlAllocateHeap
             + combase!utInitModuleNameCache

vancem: 20 is fine, indeed 25 is fine (that still avoids 1/2 the cost for almost nothing).

src/PerfView/PerfViewData.cs
kayle: It's a bit awkward to keep track of initialization with an array. Would you use something like Optional<NativeSymbolModule>[] or (NativeSymbolModule, bool)[]?

adamsitnik:
@vancem what is the status of this PR? I can see that @kayle has pushed some code cleanup commit after your review.

vancem:
@adamsitnik  - I will take a look later today or tomorrow.

vancem:
I looked this over.  It still has the big non-pay-for-play issue (basically it will try to download every PDB in the trace (often 100s since it includes every OS dll from every process).    Thus the window will not open for typically 10s of seconds  (maybe 100s of seconds).   

But I took some time to implement the work-around that I suggested in the PR.  Thus I will check this in and immediately follow up with my PR to bring back pay-for-play.

vancem:
A big thank you to @kayle  for implementing this.  It is very nice to have the C++ types available and it was pretty non-trivial to track down the new information in the PDBS that enabled this and then learn enough about PerfView to incorporate it.   Thanks!

kayle:
Thanks @vancem for getting this in and adding the fix to allow refreshing after loading symbols.

PR #858, opened 2019-01-18
Fix bug where PerfView fails to launch.

vancem:
We added code to give windows owners (so that windows end up on a related screen when using multiple screens)
However it turns out you can't give a window an owner if the owner was never display, and we do precisely this
when we pop a EULA dialog on startup.  This causes a silent crash (because the normal windows to display errors
also fail).    Fix this by catching errors and proceeding without setting owners.

Also updated to use a newer SupportFiles (this just adds a PDB), and update the PerfView version number.

### DIFF
--- a/src/PerfView/GuiUtilities/WindowBase.cs
+++ b/src/PerfView/GuiUtilities/WindowBase.cs
@@ -17,7 +17,7 @@ namespace PerfView
         {
             if (parentWindow != null)
             {
-                // Try to set the owner, but it might fail (if the window has never been displayed)
+                // Try to set the owner, but it might fail (e.g. if parentWindow has never been displayed)
                 // give up setting the owner in that case (it is not critical to have an owner)  
                 try
                 {
--- a/src/PerfView/GuiUtilities/WindowBase.cs
+++ b/src/PerfView/GuiUtilities/WindowBase.cs
@@ -15,8 +15,17 @@ namespace PerfView
 
         public WindowBase(Window parentWindow)
         {
-            Owner = parentWindow;
-            WindowStartupLocation = WindowStartupLocation.CenterOwner;
+            if (parentWindow != null)
+            {
+                // Try to set the owner, but it might fail (if the window has never been displayed)
+                // give up setting the owner in that case (it is not critical to have an owner)  
+                try
+                {
+                    Owner = parentWindow;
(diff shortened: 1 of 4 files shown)


src/PerfView/GuiUtilities/WindowBase.cs
```diff
@@ -15,8 +15,17 @@ public class WindowBase : Window
...
+            if (parentWindow != null)
+            {
+                // Try to set the owner, but it might fail (if the window has never been displayed)
+                // give up setting the owner in that case (it is not critical to have an owner)  
+                try
+                {
+                    Owner = parentWindow;
+                    WindowStartupLocation = WindowStartupLocation.CenterOwner;
+                }
+                catch (System.Exception) { }
```
sharwell: ❓ Can we be more specific about the exception type?

src/PerfView/GuiUtilities/WindowBase.cs
```diff
@@ -15,8 +15,17 @@ public class WindowBase : Window
 
         public WindowBase(Window parentWindow)
         {
-            Owner = parentWindow;
-            WindowStartupLocation = WindowStartupLocation.CenterOwner;
+            if (parentWindow != null)
+            {
+                // Try to set the owner, but it might fail (if the window has never been displayed)
```
sharwell: ❔ Is `window` here referring specifically to `parentWindow`? If so, it may help to do this:

```suggestion
                // Try to set the owner, but it might fail (e.g. if parentWindow has never been displayed)
```

src/PerfView/GuiUtilities/WindowBase.cs
vancem: The exception that is thrown in the case that caused the issue is an InvalidOperationException.  However the logic goes for any error.    I do understand that assuming that other exceptions don't corrupt state is a risk, but most of that risk is mitigated by the runtime treating 'corruption state exceptions' as uncatchable.   There is risk either way, and I guess I expecting that the likelihood of a truly corrupting state when all you are doing is setting the owner is unlikely (but the possibility of other exceptions (like null reference exceptions) being higher likelihood.

vancem:
I did look around for somethign that woudl have said 'no yet opened' or something like that.   I did not see anything.  It may be there, but I did not find it.

PR #870, opened 2019-02-05
Remove Span to avoid adding Memory.dll dependency

mpeyrotc:
Could not build Perfview locally. Making PR to get review.

### DIFF
--- a/src/EtwClrProfiler/ETWClrProfilerX64.vcxproj
+++ b/src/EtwClrProfiler/ETWClrProfilerX64.vcxproj
@@ -26,7 +26,7 @@
     <CharacterSet>Unicode</CharacterSet>
   </PropertyGroup>
   <PropertyGroup>
-    <PlatformToolset>v142</PlatformToolset>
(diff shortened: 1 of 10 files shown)


mpeyrotc:
We understand that System.Memory.dll is present in Roslyn and this would make Span<T> available in regular VS. However, the reason we don't want to add it is that Roslyn is not available in the remote debugger, hence the dependency would need to be added. At this current stage that would be a big issue for us.

vancem:
Sam, I am OK with putting Read back to use an array, offset, length.    It is actually kind of 'par for the course' that existing libraries can't aggressively use new data types because of dependency issues'.   From a Pay for play perspective, Read() for StreamReaders is a VERY rare API, and so to take a dependency on a new DLL for it is got poor pay-for-play characteristics (everyone pays for the dependency, very few benefit from the API being there).   The alternative (very few people need to use a slightly less elegant API, but we avoid any dependency issues), has good pay-for-play characteristics.

PR #871, opened 2019-02-06
Enable user-defined path conversion

nonmanager:
TraceEvent automatically converts kernel paths found in ETW events
to user paths. Depending on the available input, the conversion can
end up being inaccurate. For example, in real-time ETW sessions,
the conversion will fail for events originating from a drive inserted
after the session started (such as a USB key).

The patch enables the user to specify explicitly how path conversion
should be done by setting KernelPathToUserPathMapper. For instance,
if set to the identity function, no conversion is done and the
original kernel paths from the events are exposed.


vancem:
Looks good to me.

PR #872, opened 2019-02-06
To avoid large allocations create SegmentedStreams and use in graph. 

vancem:
This mostly pulls in code that Marco wrote a while back (some small tweeks).   Bascially instead 
of using MemoryStream we use a new SementedMemoryStream. 

Marco. Please build this locally and test that it performs well and does what you want.  

Note that we don't use MemoryMappedFileStreamWriter anymore so I have #ifdefed it out for now.
Also to avoid dependency of MemoryMappedFileStreamReader on MemoryMappedFileStreamWriter I cloned the BlockCopyCapacity constant into  MemoryMappedFileStreamReader 

@mpeyrotc  @sharwell

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -22,8 +22,8 @@
 
   <PropertyGroup>
     <!-- These are the versions of the things we are CREATING in this repository -->
-    <PerfViewVersion>2.0.37</PerfViewVersion>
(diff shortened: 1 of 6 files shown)


vancem:
@mpeyrotc telle me this passes testing

PR #877, opened 2019-02-14
Changes window ownership for better z-ordering

### LINKED ISSUE #865: Owner windows are never shown on top of owned windows
#758, the fix for #680, has the unfortunate side effect that once you open a window, you can't alt-tab back to the main PerfView window easily. This is beacuse "owned" windows in WPF can never cover their "owner", as documented [here](https://docs.microsoft.com/en-us/dotnet/api/system.windows.window.owner?view=netframework-4.7.2).

Would a reasonable compromise be to set ```WindowStartupLocation``` to ```WindowStartupLocation.CenterScreen``` and not set the Owner? This will center the window on the screen with the mouse cursor. This would mean that someone navigating aggressively with the keyboard might still end up with a misplaced window, but I would bet a much more common flow is to click on something and hit one key (ctrl-d, etc.)

billwert:
Previous #758 fixed window creation so that child windows would be on the same monitor as their parents. This change has the side effect of making it impossible to put parent windows on top of child windows.

This change instead puts the window on the screen where the mouse cursor is at the time the window is created. This isn't perfect; if you double click something and quickly move the mouse you can wind up with the window on the "wrong" monitor. I think this is a better compromise than the current situation.

Fixes #865

### DIFF
--- a/src/PerfView/GuiUtilities/WindowBase.cs
+++ b/src/PerfView/GuiUtilities/WindowBase.cs
@@ -21,8 +21,7 @@ namespace PerfView
                 // give up setting the owner in that case (it is not critical to have an owner)  
                 try
                 {
-                    Owner = parentWindow;
-                    WindowStartupLocation = WindowStartupLocation.CenterOwner;
+                    WindowStartupLocation = WindowStartupLocation.CenterScreen;
                 }
                 catch (System.Exception) { }
             }


vancem:
I have noticed the same annoying Z behavior so I am happy to have a fix.    Thanks!

PR #881, opened 2019-02-22
Remove destructive behavior from copy

mpeyrotc:
CopyRangeTo method had destructive behavior to improve memory usage, further discussion agreed this is not a behavior a list should have.


vancem:
I did not look at that the first time around.   I agree that having the list have any state that feels like a cursor  (that is state over and above the items in the list), is a bad idea.  I like that they are removed.

PR #883, opened 2019-02-25
Some very small changes

vancem:
Removed some workarounds in PerfVIew.csproj, which are the only thing that might cause grief.

### DIFF
--- a/src/PerfView/PerfView.csproj
+++ b/src/PerfView/PerfView.csproj
@@ -462,21 +462,6 @@
     <Compile Update="**\*.xaml.cs" SubType="Designer" DependentUpon="%(Filename)" />
   </ItemGroup>
 
-  <!-- Work around CoreCompile doesn't exist error -->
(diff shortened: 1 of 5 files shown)


src/PerfView/PerfView.csproj
```diff
@@ -462,21 +462,6 @@
...
-
-  <!--
-     Work around to fix Intellisense file generation for XAML projects
-     https://github.com/dotnet/project-system/issues/2488
-  -->
-  <Target Name="WorkaroundForXAMLIntellisenseBuildIssue" AfterTargets="_CheckCompileDesignTimePrerequisite">
-    <PropertyGroup>
-      <BuildingProject>false</BuildingProject>
-    </PropertyGroup>
-  </Target>
```
sharwell: ❗️ These changes will break the WPF support. I can update to MSBuild.Sdk.Extras to fix it without breaking IDE support.

Issue #886, opened 2019-02-26
Suggest add System.Memory as a dependent package for TraceEvent NuGet


xiaomi7732:
**Issue:**

In the latest TraceEvent library (2.0.36), there is an hidden dependency for .NET Fx of System.Memory.

**Repro steps:**

* Create a .NET Fx (4.6.1 in this case) project and reference TraceEvent.
* Run the following code:

```csharp
  string etlx = TraceLog.CreateFromEventPipeDataFile(@"A-NetPerf-File.netperf");
  Console.WriteLine($"eltx: {etlx}");
```

**Expected:**
The netperf file being converted to etlx correctly.

**Actual:**
System.IO.FileNotFoundException: 'Could not load file or assembly 'System.Memory, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.'

**Work around**
Add package reference to `System.Memory` in the user's project.

```xml
    <PackageReference Include="System.Memory">
      <Version>4.5.2</Version>
    </PackageReference>
```

But the exception won't be found until runtime.

**Proposal for the fix**
In the nuspec for TraceEvent, claim System.Memory as its dependency.

pharring:
Wasn't this already addressed with #870 by removing the use of `Span<T>`?

vancem:
Yes, #870 should remove the dependency.   Note that we have not published a new package (although we have built them).   You can try out \\vancem1\public\Microsoft.Diagnostics.Tracing.TraceEvent.2.0.38.nupkg.  If it looks OK, I will push it to nuget.    @mpeyrotc are you happy with the lastest TraceEvent package?

xiaomi7732:
@vancem / @pharring yes, I didn't notice the dependency is removed by #870. Just built the tip of the code and verified the change and it worked without the dependency of System.Memory.

I'll close this issue and abandon the PR. Thanks for pointing it out.

Issue #793, opened 2018-11-14
Microsoft.Diagnostics.Tracing.TraceEvent NuGet package is not properly being installed for .NET Framework 4.5


rpajak-solarwinds:
Steps to reproduce:
1. Create a console project targeting .NET Framework 4.5.2
2. Get https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.TraceEvent/2.0.30 from NuGet
3. Run (maybe it is required to paste some ETW sample code, but probably not)

Actual results:
Could not find dependent assembly 'System.Runtime, Version=4.1.0'

Executed result:
Works :)

Notes:
Most likely this issue was introduced when TraceEvent started to support .NET Standard 1.6. Probably it was https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.TraceEvent/2.0.22 (but you need to double check if you want to be sure.
It is probably caused by Microsoft.Diagnostics.Tracing.TraceEvent.props which references always the .NET Standard 1.6 libraries.
The workaround for the issue is to manually edit the .csproj where the NuGet is used. Remove the Import of Microsoft.Diagnostics.Tracing.TraceEvent.props and add the items manually.
The biggest issue is that you need to do this workaround each time you want to upgrade the version of Microsoft.Diagnostics.Tracing.TraceEvent using NuGet.

PR #887, opened 2019-02-27
Add package dependencies for TraceEvent NuGet pkg

xiaomi7732:
Address #886 & #793 .


pharring:
Is the System.Memory reference still needed? I thought we removed it with #870 ?

xiaomi7732:
This is not needed anymore since #870 did it.

PR #888, opened 2019-02-28
Add configure to bypass assert dialogs

xiaomi7732:
Set the value to false will allow build debug binaries while not showing the assertion failure dialogs. Just to make the debugging easier.

### DIFF
--- a/src/PerfView/app.config
+++ b/src/PerfView/app.config
@@ -1,5 +1,9 @@
 <?xml version="1.0"?>
 <configuration>
+  <system.diagnostics>
+    <!-- Set the value to false to avoid poping up dialogs when assert fails. -->
+    <assert assertuienabled="true"/>
+  </system.diagnostics>
   <startup>
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5"/>
   </startup>


vancem:
If I understand this, this does not actually change the default (which is 'true') and thus asserts will pop a dialog box.   It also obviously will not affect a release build (which has no asserts). but simply makes this useful option more discoverable) and easier to turn on (you don't have to remember the exact name).  

I am conflicted about this because it is frankly not too likely to be discovered (how will users know to look there.  Only those people who know about this switch will think to look).  and frankly turning off asserts is probably not a good idea (they are there for a reason).   I know you can redirect to a file, but that is not what this encourages, and frankly if it is not in your face it is not doing its job.  

In my own usage, when I hit asserts that I don't want to investigate, I simply comment it out (that way all the others are still on).   There used to be an option on the dialog for this.   

This to me is reasonably 'obvious' (If I keep hitting an assert I don't care about comment it out).  

If you still feel strongly that this adds value, let me know in antoher comment on this PR and I will check it in (since it is a no-op and will be valuable to at least those people like yourself that know about it), but I guess I am trying to convince you that commenting out is a better strategy, and we should not bother.   If you agree that it is not worth it, just close this.

I am OK either way.

xiaomi7732:
* Yes, this doesn't change the default behavior.
* There is scenario every once a while that I need it. Since it is not frequently used, I find myself searching for it again and again. I know what I want, but I can't remember the schema.

I'll appreciate it if this gets in, then I'll save the time for searching the schema when I ran into that scenario again. And hopefully, this could save someone else's time as well.

Bottom line, this should have no side effect by default. :-)

PR #894, opened 2019-03-19
Fix issue where activity paths were not being shown as paths.

vancem:
Basically the issue was that to look up an activity path you need to know that process ID.    We have a mechanism for doing this without the process ID, but it is does not work for process IDs > 64K.   We were running into traces that exceeded this.  

I did two things. 
1.  Fixed it so that we always use the process ID.
2. As extra insurance, even when we don't allow the 'guess' to work with process IDs that fit in 20 bits (rather than 16).

### DIFF
--- a/src/TraceEvent/Computers/StartStopActivityComputer.cs
+++ b/src/TraceEvent/Computers/StartStopActivityComputer.cs
@@ -653,7 +653,7 @@ namespace Microsoft.Diagnostics.Tracing
             if (processID == 0)
             {
                 // We guess that the process ID is < 20 bits and because it was xored
-                // with the lower bits, the upper 20 bits should be independent of the
+                // with the lower bits, the upper 12 bits should be independent of the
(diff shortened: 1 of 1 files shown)


src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -652,11 +652,11 @@ public static unsafe bool IsActivityPath(Guid guid, int processID)
             uint sum = uintPtr[0] + uintPtr[1] + uintPtr[2] + 0x599D99AD;
             if (processID == 0)
             {
-                // We guess that the process ID is < 16 bits and because it was xored
-                // with the lower bits, the upper 16 bits should be independent of the
+                // We guess that the process ID is < 20 bits and because it was xored
+                // with the lower bits, the upper 20 bits should be independent of the
```
pharring: Should that 2nd "20" be "12" ?

PR #896, opened 2019-03-20
Fixes for support for showing types in the Net OS Heap Alloc view.

vancem:
This improves https://github.com/Microsoft/perfview/pull/857.
In particular it does not look up every PDB aggressively.


vancem:
If you look at the diff, turn off whitespace diffing, as I reformatted the file.

PR #905, opened 2019-03-29
Fix bug where after relogging process IDs -1 on CSWITCHES.

vancem:


vancem:
If you look at the diff, turn off whitespace diffs.  The change is very small.

PR #906, opened 2019-03-31
Update the docs to refer to BufferSizeMB

felipepessoto:
Update the docs to refer to BufferSizeMB, instead of the old convention without the units


vancem:
Thanks for the update.

PR #909, opened 2019-04-02
Add R2RGetEntryPoint event to ClrTraceEventParser

brianrob:
This allows PerfView and TraceEvent to understand the instrumentation added to CoreCLR by https://github.com/dotnet/coreclr/pull/23662.


brianrob:
Thanks @mjsabby!

Issue #919, opened 2019-04-24
Add ability to set startTime, endTime and advapi32.dll location in ProcessTrace


mjsabby:
We're using a custom file format (not ETL but like it) that is generated by BPerf and used extensively in Bing. We'd like to be able to provide to TraceEvent the abililty to locate our methods for OpenTraceW, ProcessTrace and CloseTrace.

In addition to that we'd like to be able to supply Start Time and End Time to ProcessTrace, since in our format we're able to skip and only do I/O for the events we care about.

mjsabby:
cc @brianrob -- I'll supply the change. It should be fairly minimal and backwards compatible.

PR #920, opened 2019-04-24
Implement custom ETW module support

mjsabby:
Fixes https://github.com/Microsoft/perfview/issues/919.

### DIFF
--- a/src/TraceEvent/ETWTraceEventSource.cs
+++ b/src/TraceEvent/ETWTraceEventSource.cs
@@ -50,8 +50,11 @@ namespace Microsoft.Diagnostics.Tracing
         /// If type == Session this is the name of real time session to open.</param>
         /// <param name="type"></param>
         // [SecuritySafeCritical]
-        public ETWTraceEventSource(string fileOrSessionName, TraceEventSourceType type)
+        public ETWTraceEventSource(string fileOrSessionName, TraceEventSourceType type, string etwModulePath = default(string), IntPtr startTime = default(IntPtr), IntPtr endTime = default(IntPtr))
         {
+            this.etwModulePath = etwModulePath;
+            this.startTime = startTime;
+            this.endTime = endTime;
             Initialize(fileOrSessionName, type);
         }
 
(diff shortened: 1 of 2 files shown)


mjsabby:
The scenario here is that we have a file format that is a list of EVENT_RECORDs serialized in a special way. So it's very much like ETL, but has a few optimizations. All our tooling is built on TraceEvent, and we'd like to be able to use it to analyze these files.

src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -383,6 +386,23 @@ public ImageData(int ProcessID, string DllName, Address BaseAddress, int Size)
             public int Size;
         }
 
+        private IntPtr GetFunctionPointerForEtwLibrary(string procName)
+        {
+            IntPtr loadLibraryHandle = TraceEventNativeMethods.LoadLibrary(this.etwModulePath);
+            if (loadLibraryHandle == IntPtr.Zero)
+            {
+                throw new Exception($"LoadLibrary failed. Please check if the file (${this.etwModulePath}) exists, and is of the correct bitness.");
```
pharring: Use DllNotFoundException.
Also, you have string interpolation inside string interpolation. Remove the inner $

src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -449,7 +469,17 @@ private void Initialize(string fileOrSessionName, TraceEventSourceType type)
...
             {
-                handles[i] = TraceEventNativeMethods.OpenTrace(ref logFiles[i]);
+                if (string.IsNullOrEmpty(this.etwModulePath))
+                {
+                    handles[i] = TraceEventNativeMethods.OpenTrace(ref logFiles[i]);
+                }
+                else
+                {
+                    var procAddress = this.GetFunctionPointerForEtwLibrary("OpenTraceW");
+                    var openTrace = (TraceEventNativeMethods.OpenTraceDelegate)Marshal.GetDelegateForFunctionPointer(procAddress, typeof(TraceEventNativeMethods.OpenTraceDelegate));
```
pharring: You should cache this delegate in a field so you don't have to LoadLibrary+GetProcAddress each time around the loop.

mjsabby:
Thanks @pharring. Updated.

brianrob:
@mjsabby thanks for the PR.  I've spent some time thinking about this, and I'd like to propose an alternative design here.  Specifically, rather than building this into TraceEvent, I'd like to propose an interface such as IETWTraceProcessor, an instance of which can be passed to ETWTraceEventSource instead of the 3 new parameters you pass in this PR.  The interface can have methods for opening, closing, and processing the trace just like you do here. - This does a few things:

1. Formalizes the interface in managed code.
2. Allows for more flexibility in the custom handler should you need to do something else (you don't necessarily have to change TraceEvent).
3. You can plumb the start and end time directly through to your custom module instead of having to have TraceEvent know about it (and any future parameters you might like to add).

What do you think?

src/TraceEvent/TraceEventNativeMethods.cs
```diff
@@ -464,6 +468,16 @@ internal unsafe struct EVENT_FILTER_EVENT_ID
 
         #endregion
 
+        #region Library functions
+
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode)]
+        internal static extern IntPtr LoadLibrary([MarshalAs(UnmanagedType.LPWStr)]string lpFileName);
+
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode)]
```
pharring: There is no GetProcAddressW. It's ANSI only. So it should be `Charset.ANSI`, `UnmanagedType.LPStr` and you probably want to add `ExactSpelling = true` for good measure (so the CLR doesn't try to find the A/W version)

mjsabby:
@brianrob Sounds good to me.

PR #921, opened 2019-04-24
Display error message when PerfView64 wrapper can't load PerfView.exe

Quppa:
If you're dumb like me and download the `PerfView64.exe` wrapper and run it without `PerfView.exe` being present, it will silently error (though it will leave an error in the Windows Event Viewer).

This change means an error message will be displayed when `PerfView.exe` is missing.

2019-04-24 PerfView64 Error

### DIFF
--- a/src/PerfView64/App.cs
+++ b/src/PerfView64/App.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Diagnostics;
+using System.IO;
+using System.Windows;
 
 namespace PerfView64
 {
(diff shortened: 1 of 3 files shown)


brianrob:
@Quppa, thank you for the contribution.  I'm on-board with such a change, but would like to understand a bit more about the implementation.  Can you tell me what the app manifest is required for?

Quppa:
Sorry, should have called that out - the app manifest enables Common Controls v6, without which visual styles would be disabled for the error message box (i.e. it would use the [pre-Windows XP style](https://stackoverflow.com/questions/5289328/wpf-messagebox-window-style)).

While I'm at it, the reason for moving the `Main` method call into its own method is that the try/catch won't catch the exception if the call is done directly.

brianrob:
@Quppa thanks for the explanation.  The fact that you had to make separate method in order to catch the exception feels like a bug to me, but if it is a bug, it feels like a runtime bug and not one in PerfView.  I'll follow up on this, but in the meantime, let me merge this.

Thanks again for contributing!

brianrob:
To close the loop on this, it is by design that you had to move the call of PerfView.App.Main into a helper function in order to get the catch handler to execute.  The reason for this is that the the call to PerfView.App.Main was causing the load of PerfView.exe to occur before PerfView64's Main method could be jitted completely, and so the catch handler didn't exist yet and we hadn't even entered the Main method yet.  Moving it to a helper method allowed for PerfView64's Main method to be jitted and execute before the load of PerfView.exe starts.

PR #923, opened 2019-04-28
Add BPerf Trace Log support to PerfView

mjsabby:
This change adds the capability to analyze BTL (BPerf Trace Log) files directly in PerfView. BTL files are used in many production services like Bing. This is because they are compressed, don't need to be read from beginning to end and can be analyzed even if the file is not completely written.

The BTL format is quite simple:

`int compressedSize;`
`int uncompressedSize;`
`byte[compressedSize];`

The `byte[compressedSize]` is a list of a `EVENT_RECORD` sequentially laid out.

@LakshanF @brianrob


brianrob:
@mjsabby this looks a bit different than our last discussion.  I'd like to think on this and also sync with @LakshanF on what the goals are for this.  Let me do that and then get back to you.  Thanks.

mjsabby:
@brianrob Yes it is, but it's more in line with what has been done in PerfView for additional file formats that PerfView knows about it. BTL is an important internal format that we'd like to keep first class support for.

brianrob:
I think that's the difference.  I did not understand your goal was for PerfView to understand this format.  The original change you proposed would not have done this without further changes.

The proposal of the interface was so that you could re-use all of the ETW handling logic but with your own file handling logic.  Then, the app would just reference both the TraceEvent package and your interface when consuming the file.  This change puts everything directly into TraceEvent and modifies PerfView to know about the file format.  I do want to think about whether or not this belongs in a plugin.  I'll follow-up via e-mail as well to get a bit more context.

src/TraceEvent/BPerf/BPerfEventSource.cs
```diff
@@ -0,0 +1,312 @@
...
+
+        public override bool Process()
+        {
+            this.ProcessInner();
+            return true;
+        }
+
+        internal override string ProcessName(int processID, long time100ns)
+        {
+            if (!this.processNameForID.TryGetValue(processID, out var ret))
```
brianrob: Does this format handle multiple processes?  If so, this won't work if PIDs are re-used.

src/TraceEvent/BPerf/BPerfEventSource.cs
```diff
@@ -0,0 +1,312 @@
...
+            return offset;
+        }
+
+        private static int AlignUp(int num, int align)
+        {
+            return (num + (align - 1)) & ~(align - 1);
+        }
+
+        [DllImport("ntdll.dll")]
+        private static extern unsafe uint RtlDecompressBufferEx(ushort compressionFormat, byte* uncompressedBuffer, int uncompressedBufferSize, byte* compressedBuffer, int compressedBufferSize, out int finalUncompressedSize, byte* workSpace);
```
brianrob: This means that this source will only work on Windows.  I take it that is OK?

src/TraceEvent/BPerf/BPerfEventSource.cs
```diff
@@ -0,0 +1,312 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using Microsoft.Diagnostics.Tracing.Parsers;
+using Microsoft.Diagnostics.Tracing.Parsers.Kernel;
+
+namespace Microsoft.Diagnostics.Tracing
+{
+    public sealed class BPerfEventSource : TraceEventDispatcher, IDisposable
```
brianrob: Can you add some comments to describe what's happening in this file?  Would love to have a class-level comment at a minimum that describes the overall algorithm here.

src/TraceEvent/BPerf/BPerfEventSource.cs
```diff
@@ -0,0 +1,312 @@
...
+            : this(btlFilePath, 0)
+        {
+        }
+
+        public BPerfEventSource(string btlFilePath, long fileOffset)
+            : this(btlFilePath, fileOffset, new byte[BufferSize], new byte[BufferSize], new byte[BufferSize])
+        {
+        }
+
+        public BPerfEventSource(string btlFilePath, long fileOffset, byte[] workspace, byte[] uncompressedBuffer, byte[] compressedBuffer)
```
brianrob: Can you describe the uses of the different constructors here?  Specifically, why would you or would you not specify the buffers?

src/TraceEvent/BPerf/BPerfEventSource.cs
```diff
@@ -0,0 +1,312 @@
+﻿//     Copyright (c) Microsoft Corporation.  All rights reserved.
+// This file is best viewed using outline mode (Ctrl-M Ctrl-O)
+//
+// This program uses code hyperlinks available as part of the HyperAddin Visual Studio plug-in.
+// It is available from http://www.codeplex.com/hyperAddin 
+// 
+
+using System;
```
brianrob: Can you please add a test case for BTL files to make sure that subsequent TraceEvent/PerfView changes don't break this support?  You can follow the pattern that is used for ETL and EventPipe traces in https://github.com/Microsoft/perfview/tree/master/src/TraceEvent/TraceEvent.Tests.

src/TraceEvent/BPerf/BPerfEventSource.cs
mjsabby: Yes, this is Windows only.

src/TraceEvent/BPerf/BPerfEventSource.cs
mjsabby: Yup will do.

src/TraceEvent/BPerf/BPerfEventSource.cs
mjsabby: Yup will do. The reason to specify buffers is to reduce allocations in a server context.

src/TraceEvent/BPerf/BPerfEventSource.cs
mjsabby: Yes I'll add a small BTL file (~4MB)

src/TraceEvent/BPerf/BPerfEventSource.cs
brianrob: Is there a reason to specify them to the constructor vs. allocating them as byte arrays that are private to the source?  In both cases, they are byte[] and so the same number of allocations will need to occur either way.  Do the sizes vary?

src/TraceEvent/BPerf/BPerfEventSource.cs
brianrob: Sounds great.

src/TraceEvent/BPerf/BPerfEventSource.cs
mjsabby: Because one use case of this will be tools that already have these workspace buffers allocated associated with some context. And yes the sizes are fixed.

src/TraceEvent/BPerf/BPerfEventSource.cs
mjsabby: It does but the calls to this are made in such a way that the dictionary can only have a single PID at any given time.

brianrob:
Thanks for the changes Mukul!

Issue #924, opened 2019-04-28
TraceEvent GC stats break when using Sample Profiler


noahfalk:
If you use EventPipe in its default tracing configuration both on .Net Core 3.0 preview builds, both sample profiling and GC events will be enabled. However when you open the GCStats view in PerfView or look at the underlying API in TraceEvent many GCs will be missing and Suspend times will be incorrect.

The underlying issue is that the parser which tracks EESuspendStart, EESuspendStop, and EERestartStop is expecting a linear sequence of events, presumably from a single thread. It expects events that look like this:

```
EESuspendStart
EESuspendStop
GCStart
GCStop
EERestartStart
EERestartStop
```

However when you enable the sample profiler, the events interlace in various patterns such as:

```
EESuspendStart (Sampler)
EESuspendStart (GC)
EESuspendStop (Sampler)
EESuspendStop (GC)
EERestartStart (Sampler)
EERestartStop (Sampler)
GCStart
GCStop
EERestartStart (GC)
EERestartStop (GC)
```

The parser gets confused quite confused and incorrectly infers that the Sampler EESuspendStop and EERestartStop are paired with the GC EESuspendStart, which means the GC isn't inside of a suspended region, which causes the GC to be filtered out of the results. Other orderings generate different forms of incorrect results.

noahfalk:
Fixed by #925

PR #925, opened 2019-04-28
Fix GC suspension parser to coexist with stack sampler

noahfalk:
This fixes issue #924. The parser expects to see only a single threaded sequence of SuspendStart, SuspendStop, RestartStart, RestartStop but using GC and stack sampling simultaneously can interleave that pattern on two different threads. This confuses the parser and breaks the GC stats.

The issue is fixed by filtering out all SuspendStart events that don't have a GC reason and then filtering out further suspension events if they are not on the same thread as the SuspendStart event.

### DIFF
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -512,7 +512,6 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                 {
                     var process = data.Process();
                     var mang = currentManagedProcess(data);
-                    mang.GC.m_stats.lastSuspendReason = data.Reason;
                     switch (data.Reason)
                     {
                         case GCSuspendEEReason.SuspendForGC:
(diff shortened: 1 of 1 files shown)


noahfalk:
cc @brianrob @billwert @Maoni0

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -522,9 +521,11 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
                             mang.GC.m_stats.suspendThreadIDBGC = data.ThreadID;
                             break;
                         default:
-                            mang.GC.m_stats.suspendThreadIDOther = data.ThreadID;
```
Maoni0: > mang.GC.m_stats.suspendThreadIDOther = data.ThreadID; [](start = 27, length = 54)

when you showed me the code I didn't search for the usage of suspendThreadIDOther but now I see there's only one usage and since you are not assigning a value to suspendThreadIDOther anymore that only usage will never be used; so might as well get rid of it all together. I think when I added this there was some peculiar thing wrt GCStart being fired multiple times (when it shouldn't be) but honestly I don't remember the detail. 

otherwise LGTM. thanks for fixing!

noahfalk: oops! I meant to revert this line too so that the IDOther would still get initialized in this code. Would you prefer that or you still want me to eliminate it?

Maoni0: ahh...then I would prefer to revert :)

noahfalk: Should be fixed now. I also needed to add a check to EERestartStop to ensure that we clear threadIDOther since it no longer reaches the bottom of the function as it did before.

sergiy-k:
Is this ready to be merged?

Maoni0:
yes I'm happy with the change. please merge. thanks for fixing this!

PR #929, opened 2019-05-05
Fix an issue caused by unexpected stop of trace sessions when TraceEventSession is just created and attached to check if a session is active.

ZhaminXu:
Fix an issue caused by unexpected stop of trace sessions when TraceEventSession is just created and attached to check if a session is active.

### DIFF
--- a/src/TraceEvent/TraceEventSession.cs
+++ b/src/TraceEvent/TraceEventSession.cs
@@ -867,7 +867,14 @@ namespace Microsoft.Diagnostics.Tracing.Session
                 if (m_StopOnDispose)
                 {
                     m_StopOnDispose = false;
-                    Stop(true);
+
+                    // Only dispose session and source when required.
+                    // For session just attached to check if it's active, we must not call dispose.
+                    // Otherwise, it will caused unexpected stop of trace sessions.
+                    if (m_Create)
+                    {
+                        Stop(true);
+                    }
                 }
 
                 // TODO need safe handles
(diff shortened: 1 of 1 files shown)


ZhaminXu:
We encountered an issue when trying to use latest TraceEvent nuget package. Trace sessions are closed unexpected due to GC-ed TraceEventSession objects. So propose this fix, as current logic will caused trouble to users whoever uses GetActionSession method right after creating the real trace session.

brianrob:
@ZhaminXu, thanks for the contribution.  Do you know which line(s) in ```Dispose()``` are causing the attached session to be closed?  I'm concerned that we might leak resources if ```Dispose``` doesn't get called at all even in the attach case.

ZhaminXu:
> @ZhaminXu, thanks for the contribution.  Do you know which line(s) in Dispose() are causing the attached session to be closed?  I'm concerned that we might leak resources if Dispose doesn't get called at all even in the attach case.
Yes, it's Stop(true) method. I also checked more code there. If TraceEventSession is created and attached, m_sessionHandle is invliad, m_source and m_kernelSession are not used. So I think it's safe to skip the entire dispose method. But after thinking carefully, I think your concern is valid. let me try to just skip the stop method.

src/TraceEvent/TraceEventSession.cs
```diff
@@ -867,7 +867,14 @@ public void Dispose()
                 if (m_StopOnDispose)
                 {
                     m_StopOnDispose = false;
-                    Stop(true);
+
+                    // Only dispose session and source when required.
+                    // For session just attached to check if it's active, we must not call dispose.
+                    // Otherwise, it will caused unexpected stop of trace sessions.
+                    if (m_Create)
```
brianrob: Can you update the comment here to state that we only stop the session when we were the original creator of it and not for cases where we attach?  It looks like this is the original comment, but it's no longer correct.  Thanks.

brianrob: Otherwise, this looks good to me.  Thanks!

ZhaminXu: done. Thanks for reviewing!

PR #939, opened 2019-05-16
Add some perf events/data for tiered compilation (PerfView)

kouvel:
- Depends on https://github.com/dotnet/coreclr/pull/24607
- Added tiering events and updated existing events from the above PR
- Added switch `-ShowOptimizationTiers`. When enabled:
  - Stack frames include the optimization tier and it's possible to see stacks separately for each tier of a method
  - The JITStats page shows the optimization tier in the JIT events table
  - Optimization tiers include `ReadyToRun` so that it's possible to see time spent in R2R pregenerated code
- When CPU stacks and optimization tiers are available, and `-ShowOptimizationTiers` is not enabled, added a CPU stacks view that shows optimization tiers to the advanced group.
- Fixes https://github.com/dotnet/coreclr/issues/23222

Fixes dotnet/coreclr#23222:
- Took optimization tiers from the Linux perf trace to include the optimization tier in `StackFrame`s
- Updated the XML stack view writer to include the optimization tier as an attribute of the `Frame` element
- Updated the XML stack view reader (for the XML stack source) to include the optimization tier in frame names similarly to how they show in traces taken on Windows, when the `-ShowOptimizationTiers` switch is enabled


kouvel:
Test runs tier 0 and tier 1 versions of a method that continuously increments a static variable:

```c#
    private static void Main()
    {
        Thread t;
        ThreadStart runForever = () => Run(true);

        // Run tier 0 version of Run() continuously
        t = new Thread(runForever);
        t.IsBackground = true;
        t.Start();
        s_runStarted.WaitOne();

        // Tier-up Run()
        Run(false);
        Thread.Sleep(500);
        for (int i = 1; i < 100; ++i)
            Run(false);
        Thread.Sleep(500);

        // Run tier 1 version of Run() continuously
        t = new Thread(runForever);
        t.IsBackground = true;
        t.Start();
        s_runStarted.WaitOne();
        Thread.Sleep(500);

        // Should have reached steady state perf by now
        Thread.Sleep(2000);
    }

    [MethodImpl(MethodImplOptions.NoInlining)]
    private static void Run(bool forever)
    {
        s_runStarted.Set();
        do
        {
            ++s_value;
        } while (forever);
    }

    private static int s_value;
    private static AutoResetEvent s_runStarted = new AutoResetEvent(false);
```

New events:
- `Settings: Flags` - Event is sent only when tiered compilation is enabled. `Flags` indicate whether some suboptions of tiering are enabled.
- `Resume: NewMethodCount` - Number of methods called for the first time while tiering was paused
- `BackgroundJitStart/BackgroundJitStop: PendingMethodCount` - Number of methods currently scheduled for background JIT
- `BackgroundJitStop: JittedMethodCount` - Number of methods jitted in the background since the previous `BackgroundJitStart` event on the same thread

```
Event Name                                                            Time MSec  Rest
Microsoft-Windows-DotNETRuntime/TieredCompilation/Settings              737.137  ThreadID="10,736" ... Flags="QuickJit, QuickJitForLoops"
Microsoft-Windows-DotNETRuntime/TieredCompilation/Pause                 796.593  ThreadID="10,736" ...
Microsoft-Windows-DotNETRuntime/TieredCompilation/Resume              1,007.163  ThreadID="13,984" ... NewMethodCount="122"
Microsoft-Windows-DotNETRuntime/TieredCompilation/BackgroundJitStart  1,319.822  ThreadID="13,984" ... PendingMethodCount="7"
Microsoft-Windows-DotNETRuntime/TieredCompilation/BackgroundJitStop   1,322.385  ThreadID="13,984" ... PendingMethodCount="0" JittedMethodCount="7"
Microsoft-Windows-DotNETRuntime/TieredCompilation/Pause               4,322.383  ThreadID="14,016" ...
```

Existing method JIT events were updated to include the optimization tier. With app not R2R'ed:

```
Event Name                                          Time MSec  Rest
Microsoft-Windows-DotNETRuntime/Method/LoadVerbose    813.692  ... ThreadID="10,736" ... MethodSize="527" ... MethodNamespace="Test" MethodName="Main" MethodSignature="void  ()" ClrInstanceID="5" ReJITID="0" OptimizationTier="QuickJitted"
Microsoft-Windows-DotNETRuntime/Method/LoadVerbose    813.764  ... ThreadID="10,736" ... MethodSize="70" ... MethodNamespace="Test+<>c" MethodName=".cctor" MethodSignature="void  ()" ClrInstanceID="5" ReJITID="0" OptimizationTier="QuickJitted"
Microsoft-Windows-DotNETRuntime/Method/LoadVerbose    813.846  ... ThreadID="10,736" ... MethodSize="30" ... MethodNamespace="Test+<>c" MethodName=".ctor" MethodSignature="instance void  ()" ClrInstanceID="5" ReJITID="0" OptimizationTier="QuickJitted"
Microsoft-Windows-DotNETRuntime/Method/LoadVerbose    814.477  ... ThreadID="10,736" ... MethodSize="72" ... MethodNamespace="Test" MethodName=".cctor" MethodSignature="void  ()" ClrInstanceID="5" ReJITID="0" OptimizationTier="QuickJitted"
Microsoft-Windows-DotNETRuntime/Method/LoadVerbose    814.561  ... ThreadID="15,180" ... MethodSize="31" ... MethodNamespace="Test+<>c" MethodName="<Main>b__0_0" MethodSignature="instance void  ()" ClrInstanceID="5" ReJITID="0" OptimizationTier="QuickJitted"
Microsoft-Windows-DotNETRuntime/Method/LoadVerbose    814.691  ... ThreadID="15,180" ... MethodSize="115" ... MethodNamespace="Test" MethodName="Run" MethodSignature="void  (bool)" ClrInstanceID="5" ReJITID="0" OptimizationTier="QuickJitted"
Microsoft-Windows-DotNETRuntime/Method/LoadVerbose  1,320.075  ... ThreadID="13,984" ... MethodSize="47" ... MethodNamespace="Test" MethodName="Run" MethodSignature="void  (bool)" ClrInstanceID="5" ReJITID="1" OptimizationTier="Optimized"
```

With app R2R'ed:

```
Event Name                                          Time MSec  Rest
Microsoft-Windows-DotNETRuntime/Method/LoadVerbose  1,176.261  ... ThreadID="9,124" ... MethodSize="47" ... MethodNamespace="Test" MethodName="Run" MethodSignature="void  (bool)" ClrInstanceID="5" ReJITID="1" OptimizationTier="Optimized"
```

With the new `-ShowOptimizationTiers` switch to PerfView, stacks frames include the optimization tier and time spent at each optimization tier of a method can be seen. With app not R2R'ed:

```
Name                                                                                                  Exc %    Exc
 system.private.corelib![ReadyToRun]System.Threading.ThreadHelper.ThreadStart_Context(System.Object)    0.0      0
+ test![QuickJitted]Test+<>c.<Main>b__0_0()                                                             0.0      0
 + test![QuickJitted]Test.Run(bool)                                                                    15.0    915
 |+ coreclr!JIT_GetSharedNonGCStaticBase_SingleAppDomain                                               42.2  2,576
 + test![Optimized]Test.Run(bool)                                                                      40.9  2,494
```

With app R2R'ed:

```
Name                                                                                                  Exc %    Exc
 system.private.corelib![ReadyToRun]System.Threading.ThreadHelper.ThreadStart_Context(System.Object)    0.0      0
+ test![ReadyToRun]Test+<>c.<Main>b__0_0()                                                              0.0      0
 + test![ReadyToRun]Test.Run(Boolean)                                                                  57.3  3,512
 + test.il![Optimized]Test.Run(bool)                                                                   40.8  2,497
```

Methods can be grouped by optimization tier with the pattern "*![{%}]->[$1]" to see how much time is spent overall at each optimization tier. With app not R2R'ed:

```
Name           Exc %    Exc
[Optimized]     40.9  2,494
[QuickJitted]   15.0    915
[ReadyToRun]     0.0      0
```

With app R2R'ed:

```
Name           Exc %    Exc
[ReadyToRun]    57.3  3,512
[Optimized]     40.8  2,497
[QuickJitted]    0.0      0
```

With the new switch, the JITStats page also shows the optimization tier. Some methods with app not R2R'ed:

Start MSec | JitTime MSec | ThreadID | IL Size | Native Size | OptimizationTier | MethodName | Trigger
--: | --: | --: | --: | --: | --: | :-- | :--
812.746 | 0.946 | 10736 | 156 | 527 | QuickJitted | Test.Main() | FG
813.713 | 0.052 | 10736 | 11 | 70 | QuickJitted | Test+<>c..cctor() | FG
813.781 | 0.065 | 10736 | 7 | 30 | QuickJitted | Test+<>c..ctor() | FG
814.408 | 0.068 | 10736 | 12 | 72 | QuickJitted | Test..cctor() | FG
814.502 | 0.059 | 15180 | 7 | 31 | QuickJitted | Test+<>c.<Main>b__0_0() | FG
814.578 | 0.113 | 15180 | 27 | 115 | QuickJitted | Test.Run(bool) | FG
1319.849 | 0.226 | 13984 | 27 | 47 | Optimized | Test.Run(bool) | TC

kouvel:
It might be better to have a checkbox in the stack view. Otherwise multiple options like these would create a large matrix of views for different combinations of options. I'll look into what it would take to add a check box for the option.

brianrob:
We're already there somewhat right now - we do have a good number of views.  However, they only show up when you add the data so it tends not to hurt too much.

I would prefer not to add a checkbox to the stack viewer UI.  This is not something that folks are used to seeing and the stack viewer is used pretty much everywhere.

Also, this is likely to be difficult because stacks are computed and interned before the view comes up.  So this would mean that you'd have to change the order of operations in a pretty meaningful way.

kouvel:
So far it looks like adding a checkbox would be fairly straightforward, though it needs a bit more investigation. I don't want to duplicate each of the many stack views just to modify how it displays managed frames. It's understandable that there are several stack views currently, as they are pre-filtered to show certain types of data for some types of investigations. The option being added is not really a new view but rather would apply to most of those existing stack views, with a change to how managed stacks are displayed. I think something like a check box or an option in a menu of a stack view window would be more appropriate, and it would combine well with a command-line argument that would affect the option's default value for new stack windows.

In any case in the interest of prioritization I'll leave it as is for now and it can be refined later.

brianrob:
Stepping back a bit, let me briefly explain the design behind the set of views that we have today.  PerfView has a set of views ranging from general purpose views like the CPU stacks view all the way down to very specific views like CCW Ref Count stacks.  The design is that views that are more general purpose live on the main menu within PerfView whereas more specific views live in the Advanced or other view drop downs.  We do this to make sure that folks are presented with the most generic views first to try and solve their problem, and then they go and search out other views depending on the issues that they find in the more generic views.  When I think about tiering, I'd like to believe that the number of folks using these views is going to be relatively small - mostly internal people who are looking specifically at tiered compilation or codegen behavior.  As such, these views would fall into the "more specific" category and I'd like them to live in the Advanced drop down.  Having a checkbox means that users are presented with more options in the generic views such as CPU stacks, and I don't want folks to have to learn about more features unless/until they need to.

Do you believe that a significant segment of the population using PerfView is going to need to annotate their stacks with tiering information to make forward progress on performance investigations?

kouvel:
The design and layout as you described makes sense to me, but as I described above it's not a new view that is being added, instead it's modifying how existing views would display data. Take a look at the `ShowUnknownAddresses` option, which I followed for the new switch that was added. It works on every stack view because the intention is to change an general option that applies to all stack views, not to add a new version of each stack view. For the reasons I mentioned above, I don't think replicating the stack views would fit well into the current design.

I think the new switch would be enabled mainly for perf investigations, so it is disabled by default.

kouvel:
@brianrob do you have any other feedback? From our discussions on your suggestion I don't think we have reached consensus. I believe those things can be refined later, I'd like to unblock this PR from that aspect of discussion.

Let me know if anyone has any other feedback. If there's no further feedback I'll go ahead and merge this tomorrow.

kouvel:
Looks like I don't have merge permissions anyway. I'll just open another PR on top of this.

kouvel:
@brianrob, not having this change committed is starting to quickly become problematic. Considering that time spent on your suggestion is perhaps not more important than having a solution to the problem, is there any possibility that you could convince yourself that this PR could be merged as-is, even if it were in temporary fashion? If you or others have the cycles to implement your suggestion then I would certainly not oppose, but I feel that this needs to have some traction soon and I feel your reasons for blocking this PR are inadequate.

kouvel:
Perhaps a convincing factor could be that often no single person would know best. While I appreciate our difference in opinion, there are more dire needs than either of our opinions. As far as time/budget is concerned in making the changes you requested, that's not a simple matter in my case and perhaps not in your case either, so the appreciation of that reality perhaps would enable you to yield results to customers in this matter.

kouvel:
cc @sergiy-k

kouvel:
There appear to be many ways of creating stack sources and opening stack views. Just a couple of examples:
- Diffing two profiles through a menu option
- Opening a stack source from a filtered portion from the GC heap analyzer

There are also many varieties of stack views that would be useful for different types of investigations.

With the many ways of stack sources being used, I would still like to have a global option that always separates stack frames by code version / optimization tier. I don't think this would not be useful, perf by code version is a more accurate representation of where time is going, merging frames of different code versions may be ok for the most part so it may make sense to do that by default to reduce confusion.

If it would be useful to add a few stack views somewhere under Advanced (I'm thinking under Advanced / Optimization tiers / ...), then I think that would be doable, but under my proposal that would be in addition to the global config such that those stack views would be available by default without having to use the global config.

Thoughts?

brianrob:
@kouvel, thanks for bring this PR back to the forefront.  I like your latest proposal to leave the switch in and also have some views under Advanced.  I think my biggest thing is that I want it to be clear that the data is present (or not) in a given trace without having to re-launch PerfView with the switch.  At the same time, there definitely could be value in having this information available for more views than I currently see value for.

If you can add a view for CPU time with optimization tiers under advanced, I think that would most definitely be sufficient to move this PR along.  Also, there is a test failure on this PR.  It is probably that an event was modified and so it doesn't match the previous version of the event.  You probably just need to update the baseline for that test file.  If you run the test locally, you'll see how the baseline files get created and you can diff them and copy the new file over the existing baseline if the change is intentional.

How does that sound?

kouvel:
Sounds good to me, thanks @brianrob. I'll fix the test issue as well, haven't verified the tests after rebasing.

brianrob:
Great.  Thanks @kouvel.

kouvel:
As discussed above, when CPU stacks and optimization tiers are available, added a CPU stacks view to the advanced group when `-ShowOptimizationTiers` is not enabled. Folded https://github.com/microsoft/perfview/pull/947 into this PR. @brianrob can you take a look?

src/TraceEvent/TraceLog.cs
```diff
@@ -8622,6 +8636,10 @@ void IFastSerializable.FromStream(Deserializer deserializer)
                     codeAddressInfo.moduleFileIndex = (ModuleFileIndex)deserializer.ReadInt();
                     codeAddressInfo.methodOrProcessOrIlMapIndex = deserializer.ReadInt();
                     deserializer.Read(out codeAddressInfo.InclusiveCount);
+
+                    // Version 1
+                    codeAddressInfo.optimizationTier = (OptimizationTier)deserializer.ReadByte();
```
brianrob: The one thing that I'm slightly concerned about is that this is a breaking change for older etlx files.  This won't hurt PerfView as it separates etlx files by version, but it could hurt TraceEvent users if they use TraceLog and keep the etlx files around or transport them as a format.

It's not clear to me that there is an easy way to address this either though, as it's a per-method field, so you won't be able to use Deserializer.WriteTagged and TryReadTagged.

Let's leave this as is, but just be aware that if this becomes an issue we may need to address it.

src/PerfView/OtherSources/XMLStackSource.cs
```diff
@@ -8,6 +8,8 @@
 using System.Xml;
 using System.Diagnostics;
 
+using OptimizationTier = Microsoft.Diagnostics.Tracing.Parsers.Clr.OptimizationTier;
```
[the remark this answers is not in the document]
kouvel: For Linux perf traces, XML is exported in place of the etlx on Windows, such that the conversion and export to XML/etlx are done only once per trace file and reused thereafter (or until the exported temp file is deleted). So the exported XML/etlx has to contain all of the info from the original trace since it is used as the source for everything else thereafter. The goal here was to always include the optimization tier in the XML, and when loading it (which is what happens when opening a Linux perf trace), optimization tiers will be shown based on config or stack view chosen.

src/TraceEvent/TraceLog.cs
kouvel: I was thinking about using a tagged value instead, I suppose the etlx would either always have or always not have optimization tiers for code addresses, so maybe it could work, but I was inclined to keep the size overhead as small as possible since there would be a lot of code addresses, though I didn't see much noticeable effect on size in traces over relatively larger apps compared to baseline whether using the current method or tagging. I could switch it to tagging if you think it might save some grief.

src/PerfView/OtherSources/XMLStackSource.cs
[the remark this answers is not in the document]
kouvel: Another option which would give the best of both of the above options is to read/write a tagged value before code address infos to indicate a code address info version, and use that to look for or skip optimization tiers for each code address info. I'll give it a try.

src/TraceEvent/TraceLog.cs
brianrob: I'm not sure you actually can tag it because tagging really only adds value at a per-type level, where the type implements IFastSerializable.  The class in question here is basically a list of them, so I don't think you can do this without restructuring things, and thus having a breaking change anyway.  I wouldn't worry about it at this point.

src/PerfView/OtherSources/XMLStackSource.cs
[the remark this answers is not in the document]
brianrob: That might work since I believe that if it fails to read the tagged value it will back the stream up.  Let me know if that causes a bunch of grief and we can re-evaluate.

On the XML, that makes sense.  Thanks for the explanation.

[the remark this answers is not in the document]
kouvel: Sorry responded to the wrong comment, I meant that for the other comment about tagging, will continue about that there

src/TraceEvent/TraceLog.cs
kouvel: > Another option which would give the best of both of the above options is to read/write a tagged value before code address infos to indicate a code address info version, and use that to look for or skip optimization tiers for each code address info. I'll give it a try.

> That might work since I believe that if it fails to read the tagged value it will back the stream up. Let me know if that causes a bunch of grief and we can re-evaluate.

I remember now that there were also other issues with serialization compatibility that were causing old etlx files to be incompatible anyway and I hadn't found an easy way to solve those, so I hadn't bothered to maintain compatibility here. I don't remember the details now, will do a scan.

brianrob: Ok, sounds good.

kouvel: Looks like this is the only issue currently. Fixed with the latest commit, it can now load etlx files created before this change.

brianrob:
FYI, @billwert.

billwert:
this is awesome!

PR #947, opened 2019-06-05
[WIP] Take optimization tiers from the Linux perf trace and show them in the UI to see stacks separately for each code version (PerfView)

kouvel:
- Related to dotnet/coreclr#24967
- Built on top of and depends on #939
  - This PR actually begins with commit titled "Take optimization tiers reported in Linux perf maps and show them in the UI (PerfView)"

Fixes dotnet/coreclr#23222:
- Took optimization tiers from the Linux perf trace to include the optimization tier in `StackFrame`s
- Updated the XML stack view writer to include the optimization tier as an attribute of the `Frame` element
- Updated the XML stack view reader (for the XML stack source) to include the optimization tier in frame names similarly to how they show in traces taken on Windows, when the `-ShowOptimizationTiers` switch is enabled


kouvel:
@brianrob @noahfalk

brianrob:
@kouvel, is this PR dependent on #939?

kouvel:
I don't remember if it has a strict dependency but softly yes in the sense I had only tested this change on top of the other PR and I had intended for this to go in only after the other PR. Marked it as WIP for now.

kouvel:
Folded into https://github.com/microsoft/perfview/pull/939

PR #948, opened 2019-06-05
Support for the Nettrace file format

noahfalk:
@brianrob 

These changes allow TraceEvent to parse the new nettrace format while maintaining full compat for the previous netperf format. No API changes are necessary for consumers, it should be a drop-in replacement/upgrade.

I spent a while trying to do fancier file format things, but ultimately I talked myself out of it - the complexity was  getting out of hand. The new format is only barely more extensible than the previous one (some variable size block headers), but it does tackle many of the concrete issues we know about right now. I am still doing some testing and the runtime side doesn't compile on Linux yet but I'm not expecting major changes at this point so I was hoping you could take a look. 

I do plan to amend this PR with updated docs on the new file format, hopefully in the next 24 hours, but there is some cursory information in the commit descriptions and from past discussions. The corresponding runtime changes are in https://github.com/dotnet/coreclr/pull/24853


src/FastSerialization/StreamReaderWriter.cs
```diff
@@ -656,6 +656,41 @@ public override void Goto(StreamLabel label)
...
+            if (length > bytes.Length)
+            {
+                int cacheBytes = endPosition - position;
+                Buffer.BlockCopy(bytes, position, data, offset, cacheBytes);
+                int bytesRead = cacheBytes;
+                while (bytesRead < length)
+                {
+                    int count = inputStream.Read(data, offset + bytesRead, length - bytesRead);
+                    inputStreamBytesRead += (uint)count;
+                    if (count == 0)
```
brianrob: I think it's possible here to get into a state where the stream doesn't return the requested number of bytes but does return more than 0.  In this case, wouldn't inputStream.Read return a number > 0 but < count?  If so, then I think you would want to update the Goto line to be Goto(Current.Add(bytesRead));

src/TraceEvent/EventPipe/EventCache.cs
```diff
@@ -0,0 +1,282 @@
...
+                    }
+                    else
+                    {
+                        // When a thread id is recycled the sequenceNumber can abruptly reset to 1 which
+                        // makes droppedEvents go negative
+                        Debug.Assert(droppedEvents == 0 || sequenceNumber == 1);
+                    }
+                    thread.SequenceNumber = sequenceNumber;
+                }
+                cursor += 12;
```
brianrob: Lots of 12s around here.  It might be good to turn this into a constant so that it's clear what the 12s mean.  I'm assuming its the size of the sequence point block?

src/TraceEvent/EventPipe/EventCache.cs
```diff
@@ -0,0 +1,282 @@
...
+            }
+
+            // If the app creates and destroys threads over time we need to flush old threads
+            // from the cache or memory usage will grow unbounded. AddThread handles the
+            // the thread objects but the storage for the queue elements also does not shrink
+            // below the high water mark unless we free it explicitly. Although not unbounded
+            // growth our current runtime policy reads ahead up to 10,000 events ahead per-thread
+            // * 5000 threads * 24 bytes = ~1GB. It would be nice not to leave that much memory
+            // laying around probably mostly unused.
+            foreach(Queue<EventMarker> q in threadQueues)
```
brianrob: Don't block on this change, but we should do some testing of reasonable size profiles to make sure that we don't OOM on 32-bit PerfView.

src/TraceEvent/EventPipe/EventCache.cs
```diff
@@ -0,0 +1,282 @@
...
+            // To ensure we don't have unbounded growth we evict old threads to make room
+            // for new ones. Evicted threads can always be re-added later if they log again
+            // but there are two consequences:
+            // a) We won't detect lost events on that thread after eviction
+            // b) If the thread still had events pending dispatch they will be lost
+            // We pick the thread that has gone the longest since it last logged an event
+            // under the presumption that it is probably dead, has no events, and won't
+            // log again.
+            //
+            // In the future if we had explicit thread death notification events we could keep
```
brianrob: I believe there is an event for this, but probably not good to depend upon it unless we make it explicitly required for operation of EventPipe.

src/TraceEvent/EventPipe/EventCache.cs
```diff
@@ -0,0 +1,282 @@
...
+            if(headerSize < 20 || headerSize > eventBlockData.Length)
+            {
+                Debug.Assert(false, "Invalid EventBlock header size");
+                return;
+            }
+            ushort flags = BitConverter.ToUInt16(eventBlockData, 2);
+            bool useHeaderCompression = (flags & (ushort)EventBlockFlags.HeaderCompression) != 0;
+
+            // parse the events
+            PinnedBuffer buffer = new PinnedBuffer(eventBlockData);
```
brianrob: I'm wondering if you need to pin using a handle or if you can pin using the fixed keyword.  I see one place where the buffer gets handed out, but not sure if it must be pinned for that operation.  Unlikely to hurt for smaller traces, but for very large traces, I could see this having an impact.

src/TraceEvent/EventPipe/StackCache.cs
```diff
@@ -0,0 +1,79 @@
...
+            {
+                stackBytesSize = 0;
+                stackBytes = IntPtr.Zero;
+                return false;
+            }
+        }
+
+        public void ProcessStackBlock(byte[] stackBlock)
+        {
+            PinnedBuffer buffer = new PinnedBuffer(stackBlock);
```
brianrob: Same here - wondering if we can replace the pinned handle with fixed keyword.

src/TraceEvent/TraceEvent.Tests/EventPipeTestBase.cs
```diff
@@ -18,7 +18,8 @@ protected EventPipeTestBase(ITestOutputHelper output)
         }
 
         private static IEnumerable<string> TestEventPipeZipFiles
-            => Directory.EnumerateFiles(TestDataDir, "*.netperf.zip");
+            => Directory.EnumerateFiles(TestDataDir, "*.netperf.zip")
+            .Union(Directory.EnumerateFiles(TestDataDir, "*.nettrace.zip"));
```
brianrob: Thank you for adding a new test case!  It looks like the baseline.txt file for the new test is empty, so something went wrong.

src/TraceEvent/EventPipe/EventCache.cs
noahfalk: Are you using 'reasonable' to mean 'typical' or 'big without being crazy'? In my test scenario that does a lot of memory allocation with verbose GC + stack sampling with parallel tasks 30 sec execution creates about 5MB of data (~80,000 events). 'Typical' might be running that for a few minutes (~30MB), 'big without being crazy' might be running that 24 hours (14.4GB)?
The typical case probably won't tax the cache even if we are doing it badly because the file won't be big enough. I haven't eliminated the 32 bit position offsets in the deserializer yet so anything over 4GB will fail at this point, but we could certainly generate a 3.9GB trace.

Also thanks for reminding me about this comment because the story has actually changed a bit and I should update it. The new story is that the stream includes sequence points and nothing other than event metadata is cached across a sequence point (the reader code implements this, its just the comment that is outdated). The runtime triggers the insertion of a sequence point for every X amount of buffer space that is allocated (current 10MB, could become configurable). In the ideal world this would translate to at most 10MB of data in the stream and then 10MB of data cached in the reader. However in the hypothetical worst case the runtime could have a huge number of threads, each of those threads could have already allocated a 1MB buffer which they all fill before any of them needs to make a new allocation. Generating anything close to that pattern accidentally seems unlikely but I am happy to put that assertion to the test on some sample workloads.

src/TraceEvent/EventPipe/EventCache.cs
noahfalk: Agreed. I wasn't planning to take a dependency on any runtime events at this layer, I want to treat them as opaque. For the thread death notification we'd probably want some properties that don't come for free with generic runtime events:

a) Guaranteed emission in the stream, regardless of memory pressure on the buffer pool
b) No events come after the notification on the same thread
c) Every thread which has ever logged an event to the stream sends the notification (regardless of whether it ran managed code or has an entry in the runtime ThreadStore).

My assumption if we did it is that it would be a dedicated top-level object in the stream similar to SequencePointBlock and it would have dedicated code in EventPipe that emits it.

src/TraceEvent/EventPipe/EventCache.cs
noahfalk: We could pin/un-pin the event buffer + stack buffer on every dispatch, but I didn't expect that to perform better than pinning the buffers once up-front. If performance profiling revealed it was an issue and fixed performed better I certainly have no qualms switching, its just a matter of doing investigative work in the first place. The perf criteria for file load speed so far has been 'manual testing doesn't show it is clearly slower than other trace formats on modest input sizes.' So nothing overly rigorous : )

src/TraceEvent/EventPipe/StackCache.cs
noahfalk: I think this one is in the same bucket as above. We could use fixed to pin/unpin on every dispatch, but its not clear it would be a perf win and/or if the current scheme is a perf problem. If you think it needs to be investigated let me know, otherwise I wasn't planning to optimize read performance until we had some perf data/customer feedback that said read performance was too slow.

src/TraceEvent/TraceEvent.Tests/EventPipeTestBase.cs
noahfalk: Well that very conveniently explains why it never passed in CI : )

src/FastSerialization/StreamReaderWriter.cs
noahfalk: I assume where you said 'count' you meant (length - bytesRead) which is the number of bytes we request to read from the stream.
If it returns a value > 0 and less than (length - bytesRead) then we execute: bytesRead += count, loop back to while(bytesRead < length) and start another iteration of the loop because the condition is still true. We should only exit the while loop if either:

a) we throw the exception that we read past end of stream
b) bytesRead >= length (and bytesRead should never go bigger than length unless the stream has a bug and returned more bytes than we requested)


When we execute the Goto, bytesRead == length, unless I am still missing something?

src/TraceEvent/EventPipe/EventCache.cs
noahfalk: There are actually two things which are both 12. The sequence point block has a 12 byte header:
8 byte Timestamp + 4 byte count of threads

And a 12 byte per-thread entry:
8 byte thread id + 4 byte sequence number

I added some const ints to better describe the magic numbers.

src/TraceEvent/EventPipe/EventCache.cs
noahfalk: From a separate angle I'd be curious to hear from you how hard it might be to get rid of the pinning + raw pointer passing in this code path entirely? Certainly we can harden it and still use pointers, but it is far more error prone vs using byte[] + index or Span where the runtime will range check everything automatically. The part I haven't investigated is that currently the contract for all Sources is that they return an EVENT_RECORD*, and then there are more pointers inside of that structure. The API contract would have to change, as would the upstream callers.

src/TraceEvent/TraceEvent.Tests/EventPipeTestBase.cs
noahfalk: Debugging it I realize part of what happened. Locally I worked around a bug in TraceEvent where the ContentionData for the ClrParser is out-of-sync with the event definition in the runtime. I just hand-editted it to make it match because I think you said hand-editting was necessary in an email a while back, but let me know if that wasn't the right way to fix this.

With that out of the way now I am checking in a shiny new baseline.

src/TraceEvent/EventPipe/EventPipeFormat.md
```diff
@@ -4,8 +4,8 @@ EventPipe is the name of the logging mechanism given to system used by the .NET
 runtime to log events in a OS independent way.   It is meant to serve roughly the same
 niche as ETW does on Windows, but works equally well on Linux. 
 
-By convention files in this format are call *.netperf files and this can be thought
-of as the NetPerf File format.   However the format is more flexible than that.  
+By convention files in this format are call *.nettrace files and this can be thought
```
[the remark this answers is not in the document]
noahfalk: I think both the reader and writer have equal stake in defining a format. I think we'd need a good reason to reverse the already established precedent, that the docs are in the TraceEvent repo and I don't think we have one. I agree that we can link from one side, I'd just suggest it is runtime repo which has the link and TraceEvent keeps the pre-existing doc.

src/FastSerialization/StreamReaderWriter.cs
brianrob: Yes, you're right.  Not sure what I missed there.

src/TraceEvent/EventPipe/EventCache.cs
brianrob: I've not looked into it, but I expect that we could Span-ify this code.  I'm not sure how easy it would be to switch to byte[] + index as there are some places where we operate on native buffers and wouldn't want to do the copy.  This just hasn't been Span-ified because it hasn't come to the front of the priority queue.

In general, fixed doesn't have a cost runtime until there is a GC and that code is on the stack, so that's why it's preferrable from a performance POV.  If the creation/destruction of the handle doesn't happen very often, then it's not interesting at all.  I think I'm concerned about this more as a scalability thing than anything else.  I don't see this as blocking, but figured that if it was super-cheap to switch to fixed then it could be worthwhile.  I'm also OK deferring this unless/until we need it.  Out of curiosity, how large are these blocks?

src/TraceEvent/EventPipe/EventCache.cs
brianrob: Awesome.  Thanks.

src/TraceEvent/EventPipe/EventCache.cs
brianrob: These sound like very reasonable sizes and so I'm not worried about them.  I mostly wanted to make sure that we weren't going to have to route lots of people to PerfView64.  I'm not super-concerned about pathological cases or 24 hours of trace.  In those cases with ETW, we already force folks to seek because PerfView can't handle loading traces that big into memory at once.

src/TraceEvent/EventPipe/EventCache.cs
brianrob: Yup, makes sense.

src/TraceEvent/TraceEvent.Tests/EventPipeTestBase.cs
brianrob: Awesome.  Yes, hand editing is required at this point.

src/TraceEvent/EventPipe/EventCache.cs
noahfalk: Current limit is 100KB on the blocks, but they could be much smaller if the stream is being flushed frequently and a low volume of events are flowing through.

PR #952, opened 2019-06-18
Support NetTrace >4GB streams, Processor Numbers and a few bugs

noahfalk:
Changes:
1) The NetTrace file format now has a 32 bit processor number header field available
2) StreamReader/Writer now support >4GB stream by internally tracking StreamLabel with 64bit values. When a StreamLabel is persisted into the stream it remains 32 bit to preserve file format compatibility. This means >4GB streams should not persist stream labels and NetTrace follows this rule.
3) StreamReader now always maintains its internal positionInStream state as an 8 byte aligned value. Previously Goto() would move the stream forward to a non aligned position and then Fill() would try to back up the cursor again to align it. Backing up in the underlying stream isn't allowed when streaming so either we need logic to verify the cursor is only moving in a portion of the stream we haven't read yet, or we eliminate the case by ensuring we don't need to back up. I did the latter.
4) Fixed a bug in the recently added StreamReader.Read() path for reads > cache size where an unaligned read left positionInStream at an unaligned location. Large reads now do a read that terminates at an aligned location and then if necessary fill the cache to complete the remainder of the read that is unaligned.
5) Fixed some missing parsing of header fields in the case where NetTrace is using uncompressed headers (a case I don't expect we'll use in practice)
6) Fixed a doc bug - we were missing the sequence number in the uncompressed header

PTAL @brianrob


brianrob:
@noahfalk, this looks reasonable to me.  Thanks for adding tests as well.  For much of this, I am depending on the test coverage to confirm that things work, and I could believe that we might get bit by corner cases.  It is possible that we'll see a bug tail from this as we put this into PerfView, but we can address that if/when it happens.

Issue #958, opened 2019-06-27
Unable to Open Some Traces


brianrob:
Opening some traces fails and the PerfView UI just acts as if the trace hasn't been opened.  Under the debugger, an exception is thrown:

```
'An attempt was made to move the file pointer before the beginning of the file.

 	mscorlib.dll!System.IO.__Error.WinIOError(int errorCode, string maybeFullPath)	Unknown
 	mscorlib.dll!System.IO.FileStream.SeekCore(long offset, System.IO.SeekOrigin origin)	Unknown
 	mscorlib.dll!System.IO.FileStream.Seek(long offset, System.IO.SeekOrigin origin)	Unknown
>	Microsoft.Diagnostics.FastSerialization.dll!FastSerialization.IOStreamStreamReader.Fill(int minimum) Line 778	C#
 	Microsoft.Diagnostics.FastSerialization.dll!FastSerialization.MemoryStreamReader.ReadInt32() Line 88	C#
 	Microsoft.Diagnostics.FastSerialization.dll!FastSerialization.MemoryStreamReader.ReadLabel() Line 153	C#
 	Microsoft.Diagnostics.FastSerialization.dll!FastSerialization.Deserializer.ResolveForwardReference(FastSerialization.ForwardReference reference, bool preserveCurrent) Line 1516	C#
 	Microsoft.Diagnostics.Tracing.TraceEvent.dll!Microsoft.Diagnostics.Tracing.Etlx.TraceLog.FastSerialization.IFastSerializable.FromStream(FastSerialization.Deserializer deserializer) Line 3599	C#
 	Microsoft.Diagnostics.FastSerialization.dll!FastSerialization.Deserializer.ReadObjectDefinition(FastSerialization.Tags tag, FastSerialization.StreamLabel objectLabel) Line 1905	C#
 	Microsoft.Diagnostics.FastSerialization.dll!FastSerialization.Deserializer.ReadObjectDefintion() Line 1826	C#
 	Microsoft.Diagnostics.FastSerialization.dll!FastSerialization.Deserializer.GetEntryObject() Line 1109	C#
 	Microsoft.Diagnostics.Tracing.TraceEvent.dll!Microsoft.Diagnostics.Tracing.Etlx.TraceLog.InitializeFromFile(string etlxFilePath) Line 3354	C#
[...]

PR #963, opened 2019-06-28
Fix parsing of ETLX files larger than 2GB

noahfalk:
This fixes issue #958 

StreamLabels serialized in the FastSerialization format are 4 byte unsigned integers, but the FastSerialization library reads them using ReadInt32() so values above 2GB appear negative. Prior to PR #952 StreamLabel was a uint so (StreamLabel)ReadInt32() restored the correct unsigned semantics to the value. In PR #952 StreamLabel became a long which meant values above 2GB remained negative. Attempting to Goto() a negative StreamLabel put the reader in a bad state which lead to the observed error later in Fill().

The fix casts the int first to a uint, then assigns to the StreamLabel as before. The problematic trace was only manifesting an issue at StreamReaderWriter.cs:153, but I searched for all locations where an int was being assigned to a StreamLabel in case any of them were following the same convention of using a signed type to represent unsigned data. SegmentedMemoryStreamReader.cs:135 and graph.cs:579 strongly appear to follow that pattern, I think it is unlikely in the other cases but the uint cast is a no-op in that case and does no harm.

The issue wasn't caught in testing because we don't have any ETW tests with sample inputs larger than 2GB. I added a unit test for parsing StreamLabel with a test case larger than 2GB to prevent regressions.

### DIFF
--- a/src/FastSerialization/SegmentedMemoryStreamReader.cs
+++ b/src/FastSerialization/SegmentedMemoryStreamReader.cs
@@ -132,7 +132,7 @@ namespace FastSerialization
         /// </summary>
         public StreamLabel ReadLabel()
         {
-            return (StreamLabel)ReadInt32();
+            return (StreamLabel)(uint)ReadInt32();
         }
         /// <summary>
         /// Implementation of IStreamReader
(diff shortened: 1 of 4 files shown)


mjsabby:
Thanks! I was losing my mind figuring out what went wrong ... our traces are usually ~3.5GB

noahfalk:
@mjsabby - I take it this fix resolved the issue you were having? Let us know if not.

PR #969, opened 2019-07-10
WIP: Improvements for use by dotnet-gc-infra

andy-ms:
* `GetCurrentGC` used to only return the current GC between the GC/Start and GC/Stop events. But some events would happen slightly before and after those.
  It will now return the current GC in between GC/SuspendEEStart and GC/RestartEEEnd (+1ms due to some events coming out slightly later).

* So we are now creating the GC in `source.Clr.GCSuspendEEStart +=`. We are still filling in most of it in `GCStart` though.

* Commented out `Debug.Assert(_event.PauseDurationMSec == 0);` that was firing.

* New public properties:
	* Adds ExpandMechanisms and CompactMechanisms to GCPerHeapHistory
	* Adds ThreadId and IsBGCThread to GcJoin

### DIFF
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -508,8 +508,12 @@ namespace Microsoft.Diagnostics.Tracing.Analysis
                     }
                 };
 
+                bool isEESuspended = false;
+
                 source.Clr.GCSuspendEEStart += delegate (GCSuspendEETraceData data)
                 {
+                    isEESuspended = true;
+
                     var process = data.Process();
                     var mang = currentManagedProcess(data);
                     switch (data.Reason)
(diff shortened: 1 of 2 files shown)


src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -567,6 +567,30 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
                             }
                         }
                     }
+
+                    if (data.Reason == GCSuspendEEReason.SuspendForGC)
+                    {
+                        TraceGC gc = new TraceGC(mang.GC.m_stats.HeapCount);
+                        gc.Index = mang.GC.GCs.Count;
+                        mang.GC.GCs.Add(gc);
```
Maoni0: > mang.GC.GCs.Add(gc); [](start = 23, length = 21)

this looks like the only place you are adding to the GCs list. we can have a GC that happens without the SuspendEEStart event because it's the GC do we before we do the BGC, how do you handle that?

brianrob:
Closing as @andy-ms is no longer at MSFT.

PR #972, opened 2019-07-22
Update the MethodJitTailCallFailed event ID

jorive:
This change maps the event to the ID in https://github.com/dotnet/coreclr/blob/2a8529c6130f6c70488e8bace26404b319eed00f/src/vm/ClrEtwAll.man#L3175
The current ID maps the MethodJitInliningFailed event ID, which is wrong.

### DIFF
--- a/src/TraceEvent/Parsers/ClrTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/ClrTraceEventParser.cs
@@ -1691,7 +1691,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
             add
             {
                 // action, eventid, taskid, taskName, taskGuid, opcode, opcodeName, providerGuid, providerName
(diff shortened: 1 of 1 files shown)


src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -1691,7 +1691,7 @@ internal ClrTraceEventParserState State
             add
             {
                 // action, eventid, taskid, taskName, taskGuid, opcode, opcodeName, providerGuid, providerName
-                RegisterTemplate(new MethodJitTailCallFailedTraceData(value, 192, 9, "Method", MethodTaskGuid, 86, "TailCallFailed", ProviderGuid, ProviderName));
+                RegisterTemplate(new MethodJitTailCallFailedTraceData(value, 191, 9, "Method", MethodTaskGuid, 86, "TailCallFailed", ProviderGuid, ProviderName));
```
brianrob: Please also update the remove clause to match the correct event ID.

PR #980, opened 2019-08-05
add "join all threads" to the list of pre-defined GroupPats, implement Select Process Window for LTTng traces

adamsitnik:
Recently I've been profiling a lot of parallel benchmarks that do the same job on every thread (an example is PlainText TechEmpower benchmark).

I find `Thread -> AllThreads` very useful when working with FlameGraph in such cases:

Before:

obraz

After:

obraz

my proposal is to add "[join all threads]            Thread -> AllThreads" to the list of predefined GroupPats. Just to make it more accessible for new PerfView users who are not familiar with all grouping capabilities.

### DIFF
--- a/src/PerfView/StackViewer/StackWindow.xaml.cs
+++ b/src/PerfView/StackViewer/StackWindow.xaml.cs
@@ -516,6 +516,7 @@ namespace PerfView
                 }
 
                 GroupRegExTextBox.Text = justMyApp;
+                GroupRegExTextBox.Items.Add(@"[join all threads]            Thread -> AllThreads");
             }
 
 
--- a/src/PerfView/StackViewer/StackWindow.xaml.cs
+++ b/src/PerfView/StackViewer/StackWindow.xaml.cs
@@ -516,7 +516,7 @@ namespace PerfView
                 }
 
                 GroupRegExTextBox.Text = justMyApp;
-                GroupRegExTextBox.Items.Add(@"[join all threads]            Thread -> AllThreads");
+                GroupRegExTextBox.Items.Add(@"[fold threads]            Thread -> AllThreads");
             }
 
 
--- a/src/PerfView/StackViewer/StackWindow.xaml.cs
+++ b/src/PerfView/StackViewer/StackWindow.xaml.cs
@@ -478,7 +478,7 @@ namespace PerfView
 
             // Get tbe name and create the 'justMyApp pattern. 
             string justMyApp = null;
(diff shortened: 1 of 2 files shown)


src/PerfView/StackViewer/StackWindow.xaml.cs
```diff
@@ -516,6 +516,7 @@ private void FixupJustMyCodeInGroupPats(StackSource stackSource)
                 }
 
                 GroupRegExTextBox.Text = justMyApp;
+                GroupRegExTextBox.Items.Add(@"[join all threads]            Thread -> AllThreads");
```
brianrob: Why do you need to add the item twice?  The first one in PerfViewData.cs I understand, but I'm not clear on why you have to add it here as well.

adamsitnik:
> Can you change the name of the grouping to [fold threads] to match the wording that PerfView uses for this?

sure, done!

> Why do you need to add the item twice? The first one in PerfViewData.cs I understand, but I'm not clear on why you have to add it here as well.

For the `.etl` trace files PerfView recognizes the list of processes and asks the user to choose one. After doing so it adds the process to `IncPats` textbox.

For the PerfCollect trace files, it does not recognize the list of processes and does not ask the user to choose one hence `IncPats` is empty. Then this line returns null:

https://github.com/microsoft/perfview/blob/e8307efc8d903e9a87784861563819baed799451/src/PerfView/StackViewer/StackWindow.xaml.cs#L481

and we end up with single `GroupPats` available for Linux trace files:

https://github.com/microsoft/perfview/blob/e8307efc8d903e9a87784861563819baed799451/src/PerfView/StackViewer/StackWindow.xaml.cs#L515

This is why I had to add it in both places. I am going to take a look if it would be possible to get the list of processes and show the Select Process Window

adamsitnik:
I am able to get the process name, process id is missing.

This is how it looks like for LTTng trace file after my local fixes:

obraz

@brianrob do you like it? If yes, I am going to push it to this PR

adamsitnik:
ok @brianrob I've taken one more look and fixed the source of the problem

the new LTTng trace experience:

1. User can select the process from a list:

obraz

2. GroupPats and IncPats are now populated with the right values without duplication:

obraz

adamsitnik:
>  can you please update your comments to refer to perf_events instead of LTTng, as the limitations (PID missing) are from perf_events and not LTTng

done. Thanks for the code review!

brianrob:
Thanks @adamsitnik!

PR #981, opened 2019-08-05
Support the 2 types of .NET Core 3.0 counters (Sum and Mean)

chrisnas:
FIX: no counters where recognized with regex
FIX: show the counter display name if available

### DIFF
--- a/src/PerfView/EventViewer/EventWindow.xaml.cs
+++ b/src/PerfView/EventViewer/EventWindow.xaml.cs
@@ -9,7 +9,6 @@ using System.ComponentModel;
 using System.Diagnostics;
 using System.Globalization;
 using System.IO;
-using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
@@ -517,50 +516,7 @@ namespace PerfView
             string templatePath = Path.Combine(SupportFiles.SupportFileDir, "EventCounterVisualization.html");
             string template = File.ReadAllText(templatePath);
 
-            double t = 0;
-           
-            var counters = new Dictionary<string, List<Tuple<double, double>>>();
-            string pattern = @"Name:""([^""]*)"", Mean:([^,]*).*IntervalSec:([^}]*)";
-            m_source.ForEach(delegate (EventRecord event_)
-            {
-                string rest = event_.Rest;
-                if (rest == null)
-                {
-                    return false;
-                }
-                MatchCollection matches = Regex.Matches(rest, pattern);
-                if (matches.Count != 1 || matches[0].Groups.Count != 4)
-                {
-                    return false;
-                }
-
-                string namePart = matches[0].Groups[1].Captures[0].Value;
-                string meanPart = matches[0].Groups[2].Captures[0].Value;
-                string intervalSecPart = matches[0].Groups[3].Captures[0].Value;
-
-                double mean;
-                double intervalSec;
-                if (!double.TryParse(meanPart, out mean))
-                {
-                    return false;
-                }
-                if (!double.TryParse(intervalSecPart, out intervalSec))
-                {
-                    return false;
-                }
-
-                List<Tuple<Double, Double>> points;
-                if (!counters.TryGetValue(namePart, out points))
-                {
-                    points = new List<Tuple<double, double>>();
-                    counters.Add(namePart, points);
-                }
-                points.Add(Tuple.Create(t, mean));
-
-                t += intervalSec;
-
-                return true;
-            });
+            var counters = BuildCounters(m_source);
 
             var firstCounter = true;
             var sb = new StringBuilder();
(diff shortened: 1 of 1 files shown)


src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -605,7 +561,143 @@ private void DoShowEventCounterGraph(object sender, ExecutedRoutedEventArgs e)
...
+        {
+            var next = pos;
+
+            // a numeric field is stored in the payload as:
+            //    <token><value>
+            // note that <token> has the following format: <field>:
+            //
+            next = payload.IndexOf(field, next);
+            if (next == -1)
+                return null;
```
brianrob: Your code works for netcoreapp3.0, but not for desktop.  The reason is that IntervalSec is the last entry in the payload and so there isn't a ',' at the end of it.  With this change, your code works for both netcoreapp3.0 and desktop.

```suggestion
            {
                end = payload.IndexOf(" }", next);
                if (end == -1)
                    return null;
            }
```

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -605,7 +561,143 @@ private void DoShowEventCounterGraph(object sender, ExecutedRoutedEventArgs e)
...
+        private const string MeanToken = "Mean:";
+        private const string IncrementToken = "Increment:";
+        private const string IntervalToken = "IntervalSec:";
+
+        private Dictionary<string, List<Tuple<double, double>>> BuildCounters(EventSource source)
+        {
+            // look for events from "EventCounters"
+            // i.e. within Payload={...}, need to find Name, DisplayName and IntervalSec fields
+            // however, two counter types exist:
+            //  - Mean: Min, Max, Mean fields (all with the same current value with the 3.0 P7 implementation)
```
noahfalk: FYI: 'all with same current value' is only true for PollingCounter. If you use EventCounter with more than one call to WriteMetric in a sampling interval you will observe Count > 1 and Min/Max could be different than Mean. The counters in System.Runtime don't wind up using EventCounter, but others could.

src/PerfView/EventViewer/EventWindow.xaml.cs
chrisnas: here is the required change (i.e. not at next level but to compute end)

            var end = payload.IndexOf(',', next);
            // handle the case of the last numeric value of the payload 
            // i.e. look for " }" instead of ","
            if (end == -1)
            {
                end = payload.IndexOf(" }", next);
                if (end == -1)
                    return null;
            }

I've tested it and it works

src/PerfView/EventViewer/EventWindow.xaml.cs
chrisnas: Thanks a lot @noahfalk!
I think I've finally understood what you meant (thanks to @brianrob test code that is using EventCounter ;^)
I'll change the comments and also remove "Core" everywhere since it applies to both Framework and Core

src/PerfView/EventViewer/EventWindow.xaml.cs
brianrob: Ah, yes.  You're right.  I put this in the wrong spot.

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -605,7 +561,149 @@ private void DoShowEventCounterGraph(object sender, ExecutedRoutedEventArgs e)
             string uri = "file:///" + html.Replace('\\', '/').Replace(" ", "%20");
             Process.Start(uri);
         }
-            
+
+        private const string PayloadToken = "Payload=\"{";
```
felipepessoto: For .NET Core, the Payload format is different, PayloadToken should be "Payload:{"

Example:

HasStack="True" ThreadID="3,120" ProcessorNumber="0" ="{ Payload:{ Name:"cpu-usage", DisplayName:"CPU Usage", Mean:0, StandardDeviation:0, Count:1, Min:0, Max:0, IntervalSec:0.0010742, Series:"Interval=1000", CounterType:"Mean", Metadata:"", DisplayUnits:"" } }"

chrisnas: I don't know from where you got this payload @felipepessoto but it is not what is parsed by perfview to generate the graphs. 
Click one of the */EventCounters event and look at its content with ALT+D/right click | Dump Event and you'll see that Payload= is present also for .NET Core 3.0.
You could also set a breakpoint at the change position and look at the string being parsed.

felipepessoto: @chrisnas, I got it from dotnet-trace:

dotnet-trace collect --process-id <PID> --providers System.Runtime:ffff:5:EventCounterIntervalSec=1,Microsoft-Windows-DotNETRuntime:0:1,Microsoft-DotNETCore-SampleProfiler:0:1

PR #984, opened 2019-08-05
Update profiler to use Win10 SDK

brianrob:


brianrob:
Thanks @billwert.  Yup, tested on Win 8.1 successfully.

PR #1006, opened 2019-10-07
Fix handling of SegmentedList initial capacity

sharwell:


brianrob:
@sharwell can you share more details about this change?

src/FastSerialization/SegmentedList.cs
```diff
@@ -69,9 +69,13 @@ public SegmentedList(int segmentSize, int initialCapacity)
 
             if (initialCapacity > 0)
             {
-                this.items = new T[(initialCapacity + this.segmentSize - 1) >> this.segmentShift][];
-                initialCapacity = Math.Min(this.segmentSize, initialCapacity);
```
sharwell: 📝 Prior to this change, if `initialCapacity` was greater than `segmentSize`, then the `SegmentedList<T>` was not allocated with the initial capacity set to the desired value. For 32-bit heap dump analysis, the incorrectly-sized initial list forced the list to be resized during the heap extraction, which also happens to be a time where the system is under extreme memory pressure. I observed a substantial increase in processing failures due to out of memory exceptions. Pre-allocating the list allows the system to reserve the necessary large arrays and avoid memory problems later in processing.

sharwell:
@brianrob Sure! See https://github.com/microsoft/perfview/pull/1006#discussion_r335124131

PR #1007, opened 2019-10-10
Move the PacketCapture parser to TraceEvent

lupino3:
This should make the Microsoft-Windows-NDIS-PacketCapture parser
available to whoever is using the Microsoft.Diagnostics.Tracing library.

I fixed some visibility errors and ran unit tests, they all pass.

I don't know if this actually does what I think it does, please let me know if I am missing something.

Thanks!

### DIFF
--- a/src/TraceEvent/Parsers/Microsoft-Windows-NDIS-PacketCapture.cs
+++ b/src/TraceEvent/Parsers/Microsoft-Windows-NDIS-PacketCapture.cs
@@ -76,7 +76,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
         }
 
         static private volatile TraceEvent[] s_templates;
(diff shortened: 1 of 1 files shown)


brianrob:
@lupino3, thanks for contributing.  Can you tell me what your ultimate goal is with this change?  Thanks.

lupino3:
Thanks for the reply @brianrob. The goal is to make the Microsoft-Windows-NDIS-PacketCapture parser available to people importing the nuget package Microsoft.Diagnostics.Tracing.TraceEvent (https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.TraceEvent/).

Today if people import the aforementioned Nuget package the parser class for packet captures is not available; there is a such a parser in another part of the PerfView source code, therefore I am moving this parser in the appropriate directory to make sure it gets included in the nuget package.

Makes sense?

Thanks,
Andrea

PR #1012, opened 2019-10-16
Fix IndexOfSetBit for the case of a negative pow2 input parameter.

PeterSolMS:
Before the fix, the while loop would not execute at all because pow2 was negative, and thus the return value was 0.

The fix is to remove the test for pow2 > 0 from the while condition, but test for the special case pow2 == 0 before going into the loop, otherwise an infinite loop would result.

The call sites never call this routine with a 0 parameter, but the routine should be robust to a 0 input parameter nonetheless.


brianrob:
Thanks @PeterSolMS for the fix.  It looks like the issue here is that there was an assumption in the previous implementation that the input value would not be negative.  However, the failing test cases pass negative values, for example `0x‭FFFFFFFFC0000001‬`, which with the previous implementation returned 0, but with the proposed implementation returns 1.

I want to make sure that the negative values are expected and that this isn't a matter of missing a cast to a smaller data type.  I don't expect that with your change it will impact things either way, but I want to make sure that there isn't another bug lurking here.

As far as the failing tests, these are in-fact from your change, but they are expected.  You are changing the output results of processing a trace and these TraceEvent tests are making sure that nothing changes.  To fix this, you'll just need to update the baselines for these tests and make sure that they are correct.  If you run the tests locally in VS, they will show you where the baseline and actual txt file results live and you can diff them and/or copy the actual txt file over top of the baseline and commit it.

Maoni0:
the usage of this method means it will always take a negative number because if it's none 0 it means by definition the MSB is set. so this method never worked 'cause it always returned 0. returning 1 is correct in this case. 

as far as whether to handle the 0 input, it's ok to have it return -1, or perhaps a bit better - you could hoist the not_specified value for various enums and have it return that instead, that way we don't need to filter out 0 first before calling this method. it requires a bit more code change so it's up to you.

brianrob:
Ok, great.  Thanks @Maoni0.  I'm ok with either option you propose.

PeterSolMS:
I removed the checking from the callsites and integrated it into IndexOfSetBit, and I updated the baseline files.

PR #1015, opened 2019-10-17
Reduced allocations

sharwell:
Reduces allocations in PerfView when opening certain memory views. Implemented as part of a solution to a 7,000+ second load time for a recent OOM heap.

⚠️ **Please do not rebase or squash this pull request during the merge.** ⚠️

### DIFF
--- a/src/TraceEvent/Stacks/CallTree.cs
+++ b/src/TraceEvent/Stacks/CallTree.cs
@@ -161,7 +161,7 @@ namespace Microsoft.Diagnostics.Tracing.Stacks
         /// <summary>
         /// Cause the children of each CallTreeNode in the CallTree to be sorted (accending) based on comparer
         /// </summary>
-        public void Sort(Comparison<CallTreeNode> comparer)
+        public void Sort(IComparer<CallTreeNode> comparer)
         {
             m_root.SortAll(comparer);
         }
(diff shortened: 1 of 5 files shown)


src/TraceEvent/Stacks/CallTree.cs
```diff
@@ -297,8 +297,11 @@ private CallTreeNode FindTreeNode(StackSourceCallStackIndex stack, RecursionGuar
         {
             if (recursionGuard.RequiresNewThread)
             {
+                var capturedThis = this;
```
sharwell: 📝 I was surprised to find this explicit capture was required in order to avoid a closure allocation for the method.

/cc @jaredpar @agocke

agocke: Looks right to me. `this` is a parameter, scope is the start of the method.

src/TraceEvent/Stacks/CallTree.cs
```diff
@@ -431,8 +436,12 @@ private void AccumulateSumByID(CallTreeNode treeNode, Dictionary<int, CallTreeNo
...
+                var capturedTreeNode = treeNode;
+                var capturedCallersOnStack = callersOnStack;
+                var capturedRecursionGuard = recursionGuard;
```
brianrob: @sharwell, for each of these blocks, can you please add a comment as to why you're doing this?  It was not obvious to me until I read your comment above.

sharwell: Fixed

sharwell:
> can you provide an example of how much benefit comes from this change?

For the processing of the heap dump in [DevDiv 968672](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/968672), the improvements reduced allocations by more than 20GiB/minute for a heap that takes over 2hrs to open the stacks.

PR #1023, opened 2019-10-26
Setting font-family from docs.microsoft.com on all .htm files

amithegde:
- Serif font makes it hard to read and make the docs _less approachable_. Updated all .htm files to use sans-serif font family (actually copied from docs.microsoft.com).
- style for `hr` tag modified to double lines
- PerfView download link updated on UserGuide.htm

### DIFF
--- a/src/PerfView/SupportFiles/EventCounterVisualization.html
+++ b/src/PerfView/SupportFiles/EventCounterVisualization.html
@@ -1,10 +1,23 @@
 ﻿
 <!DOCTYPE html>
-<html>
-
+<html lang="en">
 <head>
-    <meta charset="utf-8">
+    <meta charset="UTF-8">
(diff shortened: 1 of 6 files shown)


sharwell:
I'm neutral on the change here. Long term I would be in favor of converting all these documents to Markdown.

amithegde:
> I'm neutral on the change here (leaving decision to the primary maintainers). Long term I would be in favor of converting all these documents to Markdown, but that is a larger change that would need to be approved before it could proceed.

Converting to markdown would be good as well. I wanted to make minimal changes just to make the docs go easy on eyes.

brianrob:
Having these docs in markdown is also fine, but there is benefit to having them accessible inside of PerfView so I wouldn't want to lose that.

That being said, I am OK with this change as well, and have confirmed that it works on Win7 and Win10.  Thanks for contributing @amithegde.

PR #1027, opened 2019-10-28
Parse EventPipe payloads for DiagnosticSourceEventSource events

noahfalk:
NetPerf and NetTrace v1 format don't support serializing array typed parameters.
DiagnosticSourceEventSource declares well known events that use array typed parameters
and the runtime serializes EventPipe metadata which claims the events have no parameters.
This is a targetted fix so that EventPipeEventSource will recognize these well-known events
and ignore the empty parameter metadata provided in the stream, treating the events
as if the runtime had provided the correct parameter metadata.

I am concurrently working on a runtime fix and updated file format revision which can
correctly encode these parameter types. However for back-compat with older runtimes we
need this.

### DIFF
--- a/src/TraceEvent/EventPipe/EventPipeEventSource.cs
+++ b/src/TraceEvent/EventPipe/EventPipeEventSource.cs
@@ -414,17 +414,81 @@ namespace Microsoft.Diagnostics.Tracing
             }
             else
             {
-                classInfo = new DynamicTraceEventData.PayloadFetchClassInfo()
+                classInfo = CheckForWellKnownEventFields(eventMetaDataHeader);
+                if (classInfo == null)
                 {
-                    FieldNames = new string[0],
-                    FieldFetches = new DynamicTraceEventData.PayloadFetch[0]
-                };
+                    classInfo = new DynamicTraceEventData.PayloadFetchClassInfo()
+                    {
+                        FieldNames = new string[0],
+                        FieldFetches = new DynamicTraceEventData.PayloadFetch[0]
+                    };
+                }
             }
 
             template.payloadNames = classInfo.FieldNames;
             template.payloadFetches = classInfo.FieldFetches;
 
             return template;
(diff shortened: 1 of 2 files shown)


noahfalk:
@brianrob @josalem @sywhang - code review please : )

cc @karpinsn @davidmikh - This is the TraceEvent change that should unblock your VS 16.4 work once checked in.

davidmikh:
We're still seeing some issues with our scenario, for our nettrace file at least we still aren't able to get the parameter metadata. From what I was able to see we don't seem to be hitting any of the code that was changed here when opening a nettrace file that has the DiagnosticSource events we need, though I didn't dig deep enough to figure out why.

I sent an email with additional information and the actual nettrace file I used to test this change with but figured I'd post it here too

noahfalk:
Talked offline with @davidmikh - the issue he mentioned above was resolved, no change to the PR needed.

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -427,6 +431,66 @@ private DynamicTraceEventData ReadEventParametersAndBuildTemplate(EventPipeEvent
             return template;
         }
 
+        // The NetPerf and NetTrace V1 file formats were incapable of representing some event parameter types that EventSource and ETW support.
+        // This works around that issue without requiring a runtime or format update for well-known EventSources that used the indescribable types.
+        // Going forward the goal is to update the runtime and file format so these special cases aren't needed, but for backwards compat this is
+        // all we can do.
+        private DynamicTraceEventData.PayloadFetchClassInfo CheckForWellKnownEventFields(EventPipeEventMetaDataHeader eventMetaDataHeader)
```
brianrob: We should only be calling this for existing versions of the file, so we're going to need another change here to make sure that we don't invoke this code once you make the runtime change.

PR #1030, opened 2019-10-30
Fix Caching Issues in TraceCodeAddresses

ninedan:
There are two issues with TraceCodeAddresses: 

1) resizing for codeAddressObjects uses linear growth (+ 16) leading to lots of LOH allocations, This is O(N^2) algorithm. Replacing with sparse array.

2) In Name(CodeAddressIndex), there is a cache stored in string array, but no string is ever stored back, causing lots of duplicating string generation. Add write back.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -204,3 +204,4 @@ src/TraceEvent/EventCounterHandler.cs
 OSExtensions.cs
 password.txt
 documentation/internal/
+*.userosscache
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -7572,9 +7572,15 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
     public sealed class TraceCodeAddresses : IFastSerializable, IEnumerable<TraceCodeAddress>
     {
         /// <summary>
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -204,3 +204,4 @@ src/TraceEvent/EventCounterHandler.cs
 OSExtensions.cs
 password.txt
 documentation/internal/
+*.userosscache
```
brianrob: Is this intentional?  I've not seen this type file show up anywhere.

src/TraceEvent/TraceLog.cs
```diff
@@ -7587,7 +7593,8 @@ public string Name(CodeAddressIndex codeAddressIndex)
                 names = new string[Count];
```
pshelton-skype: Can we update the other usages of `names` to `this.names` for consistency in this method?

.gitignore
brianrob: @ninedan, can you comment on the gitignore change?  Thanks.

ninedan: Yes, the .userosscache file seems to be generated by Visual Studio.

brianrob: Thanks much.

PR #1034, opened 2019-10-31
Update to the latest version

paulomorgado:
Update download link to the latest version.

### DIFF
--- a/documentation/Downloading.md
+++ b/documentation/Downloading.md
@@ -14,9 +14,9 @@ care about specific bug fixes or features go there.
 ## Shortcut to Download the Latest PerfView.exe
 
 In the common case, you only need one file, PerfView.exe, to use the tool.  The most recent copy of
-this file can be downloaded here
+this file can be downloaded here:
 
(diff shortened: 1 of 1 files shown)


brianrob:
@sharwell that's a great idea.  @paulomorgado do you want to change this to use an auto-updating badge?

paulomorgado:
Sure, @microsoft.

Is that what you posted, @sharwell?

Or is it more like this?

    [GitHub release (latest SemVer)](https://github.com/microsoft/perfview/releases/download/P2.0.46/PerfView.exe)

Still needs a manual change for the download link.

Maybe this would be enough:

    [GitHub release (latest SemVer)](https://github.com/microsoft/perfview/releases/latest)
[GitHub release (latest SemVer)](https://github.com/microsoft/perfview/releases/latest)

It points to the page of the latest release.

brianrob:
@paulomorgado, lets use the latter.  That way, when we add new releases, this doesn't have to be modified.  I always forget.  :)

paulomorgado:
Done!

Issue #825, opened 2018-12-17
Exception when rendering flame graph


grisharav:
> StackTrace:
System.ArgumentException: Width and Height must be non-negative.
   at System.Windows.Rect..ctor(Double x, Double y, Double width, Double height)
   at PerfView.FlameGraphDrawingCanvas.Draw(IEnumerable`1 boxes)
   at PerfView.StackWindow.RedrawFlameGraph()
   at PerfView.StackWindow.RedrawFlameGraphIfVisible()
   at PerfView.StackWindow.FlameGraphCanvas_SizeChanged(Object sender, SizeChangedEventArgs e)
   at System.Windows.SizeChangedEventArgs.InvokeEventHandler(Delegate genericHandler, Object genericTarget)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(Object target, RoutedEventArgs routedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.UIElement.RaiseEventImpl(DependencyObject sender, RoutedEventArgs args)
   at System.Windows.UIElement.RaiseEvent(RoutedEventArgs e)
   at System.Windows.FrameworkElement.OnRenderSizeChanged(SizeChangedInfo sizeInfo)
   at System.Windows.ContextLayoutManager.fireSizeChangedEvents()
   at System.Windows.ContextLayoutManager.UpdateLayout()
   at System.Windows.UIElement.UpdateLayout()
   at System.Windows.Controls.TabItem.OnPreviewGotKeyboardFocus(KeyboardFocusChangedEventArgs e)
   at System.Windows.UIElement.OnPreviewGotKeyboardFocusThunk(Object sender, KeyboardFocusChangedEventArgs e)
   at System.Windows.Input.KeyboardFocusChangedEventArgs.InvokeEventHandler(Delegate genericHandler, Object genericTarget)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(Object target, RoutedEventArgs routedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
[...]

Issue #873, opened 2019-02-07
Crashing when trying to view the flame graph of diff between two  gc dumbs


kshyju:
Thanks for building this tool. I am following the channel 9 tutorials and found a bug with the gc dump diff feature.

**Steps to replicate**

1. Open perfview -> Memory -> Take heap snapshot
2.  Select the process (my case, i chose `dotnet.exe` where I was running my asp net core app)
3. Enable "Freeze" checkbox
4. Click "Dumb GC Heap"
This will create the dotnet.gcdump file.

5. Run step 1 to 4 one more time. This will create a second one. 
6. Open "Heap Stacks" from both snapshots
7. Pick one and select "Diff" from menu bar, Choose the other one as baseline.
8.The diff window will open.
9. Click on Flame graph tab.

This will cause the exception. Here the content of the window

```
An unhanded exception occured.

At this point you can opt to continue, however it is possible that the aborted computation will cause additional failures.   Because PerfView generally only opens files for reading, there is no danger of corrupting  files, so it generally does not hurt to try.   However be on guard for unusual/incorrect behavior going forward.

You can of course exit and restart PerfView to be completely safe.

The fact that this exception went unhanded is a programmer error.   The fact that this failure occured has been logged so it can be fixed.  You don't need to take any action.

StackTrace:
System.ArgumentException: Width and Height must be non-negative.
   at System.Windows.Rect..ctor(Double x, Double y, Double width, Double height)
   at PerfView.FlameGraphDrawingCanvas.Draw(IEnumerable`1 boxes)
   at PerfView.StackWindow.RedrawFlameGraph()
   at PerfView.StackWindow.RedrawFlameGraphIfVisible()
   at PerfView.StackWindow.FlameGraphCanvas_SizeChanged(Object sender, SizeChangedEventArgs e)
   at System.Windows.SizeChangedEventArgs.InvokeEventHandler(Delegate genericHandler, Object genericTarget)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(Object target, RoutedEventArgs routedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.UIElement.RaiseEventImpl(DependencyObject sender, RoutedEventArgs args)
   at System.Windows.UIElement.RaiseEvent(RoutedEventArgs e)
[...]

PR #1044, opened 2019-11-07
fix flamegraph exception with negative metrics

rifatx:
In flamegraph view, X coordinates of stack boxes are calculated using `CallTreeNode.InclusiveMetric` property, and this can have negative values in diff stack views. In those cases, a negative width might be passed to `Rect` constructor, causing an `ArgumentException`.

This PR handles the negative metric values by using their absolute values and applying an aqua color theme for the related box. This way, such boxes in flamegraph can be interpreted as a gain in memory.

This PR fixes issues #825 and #873.


brianrob:
@adamsitnik, PTAL.  Thanks.

adamsitnik:
> This PR handles the negative metric values by using their absolute values and applying an aqua color theme for the related box. This way, such boxes in flamegraph can be interpreted as a gain in memory.

I must admit that it's a very smart way of solving the problem. :+1: 

However, we still need to explain the users what the aqua color boxes mean. Could you please add a short note (most probably a new line) to the tooltip information (for the special boxes only)?
 https://github.com/microsoft/perfview/blob/0838fd3606a1007dffc0ce6b43f796a23dd89057/src/PerfView/StackViewer/FlameGraph.cs#L26

And describe it shortly in the docs?

https://github.com/microsoft/perfview/blob/f35dccc88a2e6d467d26411bb09705375b823b9d/src/PerfView/SupportFiles/UsersGuide.htm#L3281-L3294

Adding a sample screen shot would be great!

Thank you!

rifatx:
You're welcome! Pls see commit 68cc176cdfee8c4c17f70858e443fcf822e48f54.

PR #1046, opened 2019-11-12
pause mode and compact/expand mechanisms

Maoni0:
currently if you look at pause mode it just shows an int; this adds an enum so it'll show something readable. 
also get compact/expand mechanisms in GCPerHeapHistory so we can process them in gc perf infra.


Maoni0:
@andy-ms PTAL

Maoni0:
@brianrob the baseline needs to be updated for this - do you have instructions how to?

brianrob:
@Maoni0 when you run the tests locally in VS, the failing test will point you to the baseline file as well as the actual file.  You'll just want to replace the baseline with the actual file and commit the change.

Maoni0:
can this be merged? I seem to have lost merge permission.

PR #1060, opened 2019-11-25
Add Tiered Compilation events to the CLR TraceEvent parser

brianrob:
Adds new tiered compilation events to the CLR parser so that they will display properly in the Events view and AnyStacks view.

Depends on https://github.com/dotnet/coreclr/pull/27942


brianrob:
The runtime changed how it does call counting which makes this PR obsolete.

PR #1061, opened 2019-11-26
Get Perfview.csproj building in VS2019 and F5

WardenGnaw:
These changes find the correct MSBuildToolsPath to use in order to find the Microsoft.CSharp.Targets file with only VS2019 installed. 

Also has been validated to build and launch in VS2017by @mpeyrotc


WardenGnaw:
Was using an old branch.

PR #1075, opened 2019-12-19
Decoding the reason why we change the generation to condemn during joined_generation_to_condemn() from the event stream

cshung:
This PR has a brother PR in the runtime repo [here](https://github.com/dotnet/runtime/pull/1042) so that the information recorded here is decoded. That is already merged so we can proceed with this one now.

The PR is meant for fixing the decoding aspect of https://github.com/dotnet/coreclr/issues/26665

When the GC decides which generation to condemn, it uses `generation_to_condemn(...)`. However, after that, we have another call to `joined_generation_to_condemn()` which could potentially veto the decision because of various reasons, this PR is meant to record them so that we can use that information to diagnose the decision better.

Since I am adding about 15 more reasons, the existing format for showing each reason as a column becomes really ugly, so I eliminated the columns with no entries, and eliminated the whole table if there are no columns that have any entries.

@maoni0, @vsadov, @ivdiazsa


Maoni0:
making this one column would get rid of some nice benefits like being able to easily see how often a reason is hit (much harder to see when you list them in a single cell) and heap index for server GC.

one way you could shrink down the # of columns is you could get rid of all empty columns 'cause usually many of them are empty; that should free up some columns for reasons from joined_generation_to_condemn (obviously also do the same thing for JGTC reasons which means you'll get a lot fewer than ~15).

brianrob:
@cshung thanks for submitting this.  I agree with @Maoni0 that it would be good to keep and shrink the table.  Usually you have many GCs and so having a single column will make it very difficult to look at the data in aggregate.

src/PerfView/GcStats.cs
```diff
@@ -830,6 +830,45 @@ private static void PrintEventCondemnedReasonsTable(TextWriter writer, TraceProc
...
+                    }
+                    if (condemnedReasonRow[j] != 0)
+                    {
+                        hasAnyContent = true;
+                        columnHasContent[j] = true;
+                    }
+                }
+            }
+
+            if (!hasAnyContent)
```
Maoni0: we'd still want to display the text that says ""Condemned reasons for GC" from below when there's no condemned reasons because this is saying "we do have the events to decode the condemned reasons but there's none" vs. "we don't have the events to decode the condemned reasons and we will not even display the text".

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -3124,7 +3145,21 @@ public enum CondemnedReasonGroup
         Too_Small_For_BGC = 12,
         Ephemeral_Before_BGC = 13,
         Internal_Tuning = 14,
-        Max = 15,
+        Almost_max_alloc = 15,
```
Maoni0: nit - could we please following the style here for cases like the existing code?

src/PerfView/GcStats.cs
```diff
@@ -1051,7 +1104,21 @@ private static string PrintCondemnedReasonsToHtml(TraceGC gc)
             new string[] {"Compacting<BR/>Full<BR/>GC", "Last GC we trigger before we throw OOM"},
             new string[] {"Small<BR/>Heap", "Heap is too small for doing a background GC and we do a blocking one instead"},
             new string[] {"Ephemeral<BR/>Before<BR/>BGC", "Ephemeral GC before a background GC starts"},
-            new string[] {"Internal<BR/>Tuning", "Internal tuning"}
+            new string[] {"Internal<BR/>Tuning", "Internal tuning"},
+            new string[] { "Almost_max_alloc", "Almost_max_alloc"},
```
Maoni0: let's chat about this on Monday to have more meaningful text here. the text should help folks to understand what this reason is for. eg in this [SO issue](https://stackoverflow.com/questions/59130015/condemned-generation-is-not-garbage-collected) the final generation is different from generation and the "Avoid_unproductive" reason should have text that explains this exact issue.

src/PerfView/GcStats.cs
```diff
@@ -1051,7 +1104,21 @@ private static string PrintCondemnedReasonsToHtml(TraceGC gc)
             new string[] {"Compacting<BR/>Full<BR/>GC", "Last GC we trigger before we throw OOM"},
             new string[] {"Small<BR/>Heap", "Heap is too small for doing a background GC and we do a blocking one instead"},
             new string[] {"Ephemeral<BR/>Before<BR/>BGC", "Ephemeral GC before a background GC starts"},
-            new string[] {"Internal<BR/>Tuning", "Internal tuning"}
+            new string[] {"Internal<BR/>Tuning", "Internal tuning"},
+            new string[] {"Max Generation Budget", "We are in high memory load situation and have consumed enough max generation budget, so we decided to do a full GC"},
+            new string[] {"Avoid unproductive<BR>Full GC", "This happens when the GC detects previous attempts to do a full compacting GC is not making progress and therefore reduce to Gen1"},
```
Maoni0: if you look above you'll notice that all words in the title start with a capital letter, can we please follow that convention? also I would use "HardLimit" instead of "hard limit" in the titles (not in the description) just because this is what you set with the config.

otherwise LGTM.

cshung:
@brianrob Can you please take a look?

cshung:
@brianrob Can you please take a look?

brianrob:
@cshung, overall, looks good.  A couple of things:

1. It would be good to detect when there is no table, and either remove the section entirely, or have a note that states that there wasn't any data to display the table (or something like that).  Otherwise, you just see this:
image
2. There might be a bug in the logic, as I saw this with one of my test traces:
image

cshung:
> @cshung, overall, looks good. A couple of things:
> 
> 1. It would be good to detect when there is no table, and either remove the section entirely, or have a note that states that there wasn't any data to display the table (or something like that).  Otherwise, you just see this:
>    image
> 2. There might be a bug in the logic, as I saw this with one of my test traces:
>    image

@brianrob, thanks for the review, the issues should be fixed.

PR #1076, opened 2019-12-19
Simplify symbol resolution work for BTL files

mjsabby:
Because BTL files support reading from an offset into the file, symbol resolution gets complicated since it needs process state information.

This complicated code actually doesn't work in the case where the cpu time is spent inside the process of interest but is in kernel code because our support code which logs symbolic information in the sidecar database can only do so for user-mode modules.

Also additional testing revealed that when we're looking for symbolic resolution in BTL files, i.e. we're doing an investigation that'll need stacks it is often time acceptable that we iterate the whole file in search for process state information. However, we don't want to lose the focusing on a specific part of the file (one of the key benefits of BTL) so the compromise is that we iterate the whole file (in search for info needed for symbolic resolution) but only do analysis of events within the time range requested.

### DIFF
--- a/src/TraceEvent/Symbols/SymbolReader.cs
+++ b/src/TraceEvent/Symbols/SymbolReader.cs
@@ -5,6 +5,8 @@ using System.Diagnostics;
 using System.IO;
 using System.Net;
 using System.Net.Http;
+using System.Security.Cryptography;
+using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
@@ -147,7 +149,7 @@ namespace Microsoft.Diagnostics.Symbols
         /// for the PDB file.</param>
         /// <param name="fileVersion">This is an optional string that identifies the file version (the 'Version' resource information.  
         /// It is used only to provided better error messages for the log.</param>
-        public string FindSymbolFilePath(string pdbFileName, Guid pdbIndexGuid, int pdbIndexAge, string dllFilePath = null, string fileVersion = "")
+        public string FindSymbolFilePath(string pdbFileName, Guid pdbIndexGuid, int pdbIndexAge, string dllFilePath = null, string fileVersion = "", bool portablePdbMatch = false)
         {
             m_log.WriteLine("FindSymbolFilePath: *{{ Locating PDB {0} GUID {1} Age {2} Version {3}", pdbFileName, pdbIndexGuid, pdbIndexAge, fileVersion);
             if (dllFilePath != null)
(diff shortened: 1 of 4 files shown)


mjsabby:
cc @LakshanF @brianrob

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -1901,11 +1912,21 @@ protected virtual string GetSourceFromSrcServer()
...
+                }
+
                 HttpResponseMessage response = httpClient.GetAsync(url).Result;
 
                 response.EnsureSuccessStatusCode();
                 Stream content = response.Content.ReadAsStreamAsync().Result;
-                string cachedLocation = GetCachePathForUrl(url);
+                string cachedLocation = Path.Combine(
+                    _symbolModule.SymbolReader.SourceCacheDirectory,
+                    BitConverter.ToString(SHA256.Create().ComputeHash(Encoding.UTF8.GetBytes(url.ToUpperInvariant())))
```
brianrob: We should probably be caching the result of ```SHA256.Create``` as this is going to be called once per symbol file.

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -1901,11 +1912,21 @@ protected virtual string GetSourceFromSrcServer()
             if (url != null)
             {
                 HttpClient httpClient = new HttpClient();
```
brianrob: Since you're already in here, can you please put a using block around the construction of the ```HttpClient```?

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -428,6 +435,10 @@ public string SymbolCacheDirectory
             }
         }
         /// <summary>
+        /// Authorization header to be ued when making requests to source server (only for SourceLink)
+        /// </summary>
+        public string AuthorizationHeaderForSourceLink { get; set; }
```
brianrob: I think this is OK for now.  If we run into more of these, I might like to remove this and provide a callback for modifying the ```HttpClient``` instead.

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -229,6 +231,11 @@ public string FindSymbolFilePath(string pdbFileName, Guid pdbIndexGuid, int pdbI
                         }
 
                         pdbPath = GetFileFromServer(element.Target, pdbIndexPath, Path.Combine(cache, pdbIndexPath));
+
+                        if (pdbPath == null && portablePdbMatch)
+                        {
+                            pdbPath = GetFileFromServer(element.Target, pdbSimpleName + @"\" + pdbIndexGuid.ToString("N").ToUpper() + "FFFFFFFF" + @"\" + pdbSimpleName, Path.Combine(cache, pdbIndexPath));
```
brianrob: Can you please add a comment here on what the concatenated string path should look like here?

PR #1077, opened 2019-12-31
Include the BulkType event for GC profiling

sharwell:
This change ensures the GC finalization table contains type information alongside the existing object counts.

### DIFF
--- a/src/PerfView/CommandLineArgs.cs
+++ b/src/PerfView/CommandLineArgs.cs
@@ -444,7 +444,7 @@ namespace PerfView
                 KernelEvents = KernelTraceEventParser.Keywords.Process | KernelTraceEventParser.Keywords.Thread | KernelTraceEventParser.Keywords.ImageLoad | KernelTraceEventParser.Keywords.VirtualAlloc;
                 ClrEvents = ClrTraceEventParser.Keywords.GC | ClrTraceEventParser.Keywords.GCHeapSurvivalAndMovement | ClrTraceEventParser.Keywords.Stack |
                             ClrTraceEventParser.Keywords.Jit | ClrTraceEventParser.Keywords.StopEnumeration | ClrTraceEventParser.Keywords.SupressNGen |
-                            ClrTraceEventParser.Keywords.Loader | ClrTraceEventParser.Keywords.Exception;
+                            ClrTraceEventParser.Keywords.Loader | ClrTraceEventParser.Keywords.Exception | ClrTraceEventParser.Keywords.Type | ClrTraceEventParser.Keywords.GCHeapAndTypeNames;
                 TplEvents = TplEtwProviderTraceEventParser.Keywords.None;
 
                 // This is not quite correct if you have providers of your own, but this covers the most important case.  
@@ -462,7 +462,7 @@ namespace PerfView
                 // TODO this logic is cloned.  We need it in only one place.  If you update it do the other location as well
                 // The process events are so we get process names.  The ImageLoad events are so that we get version information about the DLLs 
                 KernelEvents = KernelTraceEventParser.Keywords.Process | KernelTraceEventParser.Keywords.ImageLoad;
-                ClrEvents = ClrTraceEventParser.Keywords.GC | ClrTraceEventParser.Keywords.Exception;
+                ClrEvents = ClrTraceEventParser.Keywords.GC | ClrTraceEventParser.Keywords.Type | ClrTraceEventParser.Keywords.GCHeapAndTypeNames | ClrTraceEventParser.Keywords.Exception;
                 ClrEventLevel = TraceEventLevel.Informational;
                 TplEvents = TplEtwProviderTraceEventParser.Keywords.None;
                 NoRundown = true;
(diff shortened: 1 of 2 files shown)


src/PerfView/CommandLineArgs.cs
```diff
@@ -462,7 +462,7 @@ private void SetupCommandLine(CommandLineParser parser)
                 // TODO this logic is cloned.  We need it in only one place.  If you update it do the other location as well
                 // The process events are so we get process names.  The ImageLoad events are so that we get version information about the DLLs 
                 KernelEvents = KernelTraceEventParser.Keywords.Process | KernelTraceEventParser.Keywords.ImageLoad;
-                ClrEvents = ClrTraceEventParser.Keywords.GC | ClrTraceEventParser.Keywords.Exception;
+                ClrEvents = ClrTraceEventParser.Keywords.GC | ClrTraceEventParser.Keywords.Type | ClrTraceEventParser.Keywords.GCHeapAndTypeNames | ClrTraceEventParser.Keywords.Exception;
```
brianrob: Having this for GCOnly makes sense, but I don't think we'd want this for GCCollectOnly.  @Maoni0, what do you think?

Maoni0: we definitely wouldn't want this in GCCollectOnly. GCCollectOnly is supposed to be very lightweight and can be on all the time. GCHeapAndTypeNames is extremely heavy weight and artificially increases the GC pause time by a lot so should not be on for GCCollectOnly. I don't think it should be on for GCOnly either 'cause it instantly changes the perf characteristics of GC but since I'm not aware anyone actually uses GCOnly I don't have a strong opinion there.

brianrob: Thanks for clarifying @Maoni0.  Based on this, I think that we should not make this change.  Is the type information just missing for /GCCollectOnly and /GCOnly?  Specifically, does the type information appear when using the default collection parameters?

sharwell: I never use /GCCollectOnly so I don't have an opinion about the behavior there. I frequently use /GCOnly, which is where the notable functionality gap was observed (/GCOnly provides information for the GC Stats view, except the finalized objects information is missing).

Maoni0: which finalizer object info? we fire this event for finalized objects:
```
                    <event value="29" version="0" level="win:Verbose"  template="FinalizeObject"
                           keywords ="GCKeyword"
                           opcode="FinalizeObject"
                           task="GarbageCollection"
                           symbol="FinalizeObject" message="$(string.RuntimePublisher.FinalizeObjectEventMessage)"/>
```
and that's on the GC keyword verbose level

sharwell: @Maoni0 either that event is not sufficient, or PerfView does not know how to interpret it. Here is the GCStats prior to this change:

image

Here it is after this change:

Maoni0:
@vsadov can you take a look please?

VSadov:
Will take a look.

VSadov:
Here is what I found:
`FinalizeObject` is a "Verbose" event. It is _not_ logged when doing "GCCollectOnly". Therefore adding type information to "GCCollectOnly", at least from the finalization details view, is pointless.

"GCCollect" reports at erbose level and does log `FinalizeObject`. It is however accompanied only by methodtable. To map that to actual types an additional information is needed. - That is why right now you can only see the number of finalized objects, but cannot see it tallied by the object type. 

To enable type names both `ClrTraceEventParser.Keywords.Type`  and `ClrTraceEventParser.Keywords.GCHeapAndTypeNames`   must be enabled

I can see how this adds value to `FinalizeObject` reporting.

VSadov:
Basically - we should not add this to "GCCollectOnly", but adding to "GCCollect" is useful. 

There is some expense to that though.

Once `ClrTraceEventParser.Keywords.Type` is enabled, we will try to report type info for _every_ finalized object. There is a way to report only once per type, but `SendFinalizeObjectEvent` disables that to avoid hashtable lookup (under lock).

I still think enabling type names in `GCOnly` could be worth it. 

It may be worth to explore whether hashtable lookup for already reported type could make this cheaper. It might, assuming that the number of finalized objects can be much bigger than the number of finalizable types used by the app.

Maoni0:
by "GCCollect" I presume you meant "GCOnly".

@VSadov I'm unclear if you are saying both ClrTraceEventParser.Keywords.Type and ClrTraceEventParser.Keywords.GCHeapAndTypeNames need to be enabled or just the 1st one (your 1st msg seems to indicate the former while the 2nd one seems to indicate the latter). 

the GCAllocationTick event does not need these events to get the type name. is there a way to make the finalize event work the same way? however, I'm not opposed to adding it to GCOnly. I'm just curious whether we could make it so that we don't need these events.

sharwell:
> I'm unclear if you are saying both ClrTraceEventParser.Keywords.Type and ClrTraceEventParser.Keywords.GCHeapAndTypeNames need to be enabled or just the 1st one (your 1st msg seems to indicate the former while the 2nd one seems to indicate the latter).

It needs to be both if you want to see the table for finalized objects.

VSadov:
Right - FinalizeObject needs both - one to trigger type information, but that would omit names if  `GCHeapAndTypeNames` is not set.

Not sure how `GCAllocationTick` gets around needing the `GCHeapAndTypeNames`. Perhaps it calls something else. 
I will take a look if that can be used here. Otherwise we can just add these keywords to `GCOnly`

sharwell:
I'll update this PR to apply for GCOnly, but not apply for GCCollectOnly.

sharwell:
This PR is now updated.

PR #1078, opened 2019-12-31
Improve handling of OutOfMemoryException

sharwell:
* Don't continue without roots on OutOfMemoryException
* Treat ERROR_NOT_ENOUGH_MEMORY as OutOfMemoryException


brianrob:
@sharwell do I read this as an attempt to be more explicit about cases where things used to succeed but give you bad data silently?

sharwell:
When an OOM occurs, PerfView resizes the data structures to a slightly smaller set and tries again. This pull request fixes a couple cases where PerfView was proceeding with incomplete results rather than starting over with smaller data structures.

brianrob:
Gotcha.  Thanks!

PR #1080, opened 2020-01-02
Allows opt out nested activity id reducing

xiaomi7732:
For .NET Core 3.0 trace analysis, we need to have an option to disable application insights activity id filter.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -22,8 +22,8 @@
 
   <PropertyGroup>
     <!-- These are the versions of the things we are CREATING in this repository -->
-    <PerfViewVersion>2.0.49</PerfViewVersion>
-    <TraceEventVersion>2.0.49</TraceEventVersion>
+    <PerfViewVersion>2.0.48</PerfViewVersion>
+    <TraceEventVersion>2.0.48</TraceEventVersion>
   </PropertyGroup>
 
   <!-- versions of dependencies that more than one project use -->
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -22,8 +22,8 @@
 
   <PropertyGroup>
     <!-- These are the versions of the things we are CREATING in this repository -->
-    <PerfViewVersion>2.0.48</PerfViewVersion>
-    <TraceEventVersion>2.0.48</TraceEventVersion>
+    <PerfViewVersion>2.0.49</PerfViewVersion>
+    <TraceEventVersion>2.0.49</TraceEventVersion>
   </PropertyGroup>
 
   <!-- versions of dependencies that more than one project use -->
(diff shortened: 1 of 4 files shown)


sharwell:
Is there a way you can show a screenshot either of the problem or a comparison with the difference?

xiaomi7732:
Hey @sharwell , thank you for the quick turn around. There's quite a story about this. I have actually been working with @brianrob on the details. Please allow me some time before I response with details.

src/Directory.Build.props
```diff
@@ -22,8 +22,8 @@
 
   <PropertyGroup>
     <!-- These are the versions of the things we are CREATING in this repository -->
-    <PerfViewVersion>2.0.48</PerfViewVersion>
-    <TraceEventVersion>2.0.48</TraceEventVersion>
+    <PerfViewVersion>2.0.49</PerfViewVersion>
+    <TraceEventVersion>2.0.49</TraceEventVersion>
```
brianrob: Let's make this a separate change, since it's not logically part of this fix.

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -708,7 +709,7 @@ internal static unsafe string CreateActivityPathString(Guid guid)
             {
                 uint nibble = (uint)(*bytePtr >> 4);
                 bool secondNibble = false;              // are we reading the second nibble (low order bits) of the byte.
-                NextNibble:
+            NextNibble:
```
brianrob: Nit: Can you remove this whitespace diff please?

src/TraceEvent/Computers/SampleProfilerThreadTimeComputer.cs
```diff
@@ -48,6 +48,12 @@ public SampleProfilerThreadTimeComputer(TraceLog eventLog, SymbolReader symbolRe
         /// </summary>
         public bool GroupByStartStopActivity;
 
+        /// <summary>
+        /// Reduce nested application insights requests by using related activity id.
+        /// </summary>
+        /// <value></value>
+        public bool ReduceNestedApplicationInsightsRequests { get; set; } = true;
```
brianrob: @xiaomi7732 I'm not sure I understand the name "Reduce" here.  Looks like you use the name ignoreApplicationInsightsRequestsWithRelatedActivityId also which makes it more clear to me what's happening.  Can we use IgnoreApplicationInsightsRequestsWithRelatedActivityId instead for the property name?

brianrob:
@sharwell, I spoke with @xiaomi7732 about this early last month.  The issue is basically that on .NET Core 3.0 there are some changes to telemetry events that result in AI events having a related activity ID when in Core 2.2 and before they did not.  The result is that TraceEvent makes an assumption that is no longer true.  This change allows folks who are impacted by this to fix the issue without making a wider behavior change, as it's not clear whether or not this change is OK for everyone to take (AI events can be used in a number of different ways).

src/TraceEvent/Computers/StartStopActivityComputer.cs
xiaomi7732: Sure.

src/Directory.Build.props
xiaomi7732: Sure.

PR #1083, opened 2020-01-03
Use a full size user mode session buffer to avoid lost GC events

sharwell:


src/PerfView/CommandProcessor.cs
```diff
@@ -545,11 +545,13 @@ public void Start(CommandLineArgs parsedArgs)
...
-                    // Note that you don't need the rundown 300Meg if you are V4.0.   
+
+                    // Note that you don't need the rundown 300Meg if you are V4.0.
                     if (parsedArgs.CircularMB != 0)
                     {
-                        // Typically you only need less than 1/5 the space + rundown 
-                        userModeSession.CircularBufferMB = Math.Min(parsedArgs.CircularMB, parsedArgs.CircularMB / 5 + 300);
+                        // Typically you only need less than 1/5 the space + rundown. However, some scenarios primarily
+                        // use the user mode session so we keep it the full size.
+                        userModeSession.CircularBufferMB = parsedArgs.CircularMB + 300;
```
brianrob: Did the `+300` get left in by accident?

sharwell: I left it in case the buffer was smaller than 300 for some reason. I wasn't sure if it would be safe to remove but either way wouldn't affect my scenarios.

brianrob: I expect it's fine.  Let's leave it.

PR #1084, opened 2020-01-03
Prune GCBulkMovedObjectRanges from the ETLX

sharwell:
This event is large, but unused. Removing it allows for a substantial increase in the amount of GC allocation data that can be analyzed in one trace.


brianrob:
@sharwell, I don't think that we should remove this event from the resulting etlx.  The general pattern here is that any event whose data is captured in the resulting TraceLog side tables during creation can then be removed.  This rationale for this pattern is that many of these events are rundown events that might be present later in the trace but needed for analysis of earlier events.  The GCBulkMovedObjectRanges event is not encapsulated in the resulting TraceLog and is needed at processing time by any tools that might keep track of the heap state of the process.

sharwell:
I may have understated the impact of this event. This event is 75+% of the resulting ETLX, while the GC events are _much_ smaller. By omitting this event, I can analyze ~10x the GC data I would otherwise be able to.

sharwell:
Another option is finding a way to exclude this event from the "GC Only" scenario.

brianrob:
When you collect with /GCOnly, what type of analysis are you doing?  For most GC analysis, we recommend /GCCollectOnly which only tells you about collections and should not contain heap information events such as GCBulkMovedObjectRanges.

sharwell:
I'm using GC Heap Alloc Ignore Free for scenarios that allocate 200-500GiB.

brianrob:
Are you using sampled allocations, or are you using the profiler API based .NET Alloc support?

sharwell:
I'm using whatever the default options are for the "GC Only" check box.

sharwell:
image

brianrob:
Right, what I mean is which of the views are you using?  I suspect that there is a configuration you can use to get what you want.

sharwell:
I use **GCStats** and **GC Heap Alloc Ignore Free (Coarse Sampling) Stacks**

sharwell:
@VSadov are you able to help with this one?

VSadov:
I am not sure I will have time to look at this in the next few days, but filtering out collected events seems unusual.
CC @Maoni0

sharwell:
@VSadov equally acceptable would be an easier way to prevent this costly but unused event from getting collected.

Maoni0:
> This event is large, but unused.

this event is used to construct heap graphs. 

if you want to prevent it from being collected for the /GCOnly commandline, then yes I'd agree with that (in fact I've advocated this before but since very few people actually use the /GCOnly cmdline we never actually did anything). but if you mean someone already collected them in the .etl and you want to get rid of them when we write out the .etlx that doesn't seem reasonable 'cause obviously you can use them and if someone intentionally collected them they should keep them.

sharwell:
@Maoni0 yes, as soon as I know how to update `/GCOnly` to avoid this event, I will update this pull request to implement it that way. The problem is I don't know how/why this event is being enabled.

Maoni0:
this event is under the `GCHeapSurvivalAndMovement` keyword; please see https://github.com/microsoft/perfview/blob/master/src/PerfView/CommandLineArgs.cs#L446

sharwell:
Awesome! Are there any other events enabled by that keyword that I wouldn't want to remove?

sharwell:
Also @Maoni0, I'd like to turn this off for the default thread time and CPU sampling scenarios, in addition to /GCOnly. Obviously the event would be enabled for /GCCollectOnly. This leaves the following questions:

1. Do we need an easy way to turn this event back on for scenarios other than /GCCollectOnly?
2. If the answer is 'yes', which check box would it be tied to? My guesses are:
    1. The existing **.NET Alloc** check box, which is not checked by default
    2. A new check box, possibly called **GC Heap**

Maoni0:
> Obviously the event would be enabled for /GCCollectOnly.

typo? would NOT be enabled for /GCCollectOnly.

I think the bulk move event should only be enabled for options that actually make used of them. and I'm unfamiliar which options in the UI do that. @brianrob would know more.

sharwell:
I found a different way to achieve this goal

PR #1086, opened 2020-01-03
Implement 'Merge and Zip All'

sharwell:


src/PerfView/MainWindow.xaml.cs
```diff
@@ -1386,6 +1387,39 @@ private void DoZip(object sender, RoutedEventArgs e)
...
+        {
+            var unmergedFiles = new List<PerfViewFile>();
+            foreach (var file in TreeView.Items.OfType<PerfViewFile>())
+            {
+                if (file.FilePath.EndsWith(".etl"))
+                    unmergedFiles.Add(file);
+            }
+
+            List<Action> actions = new List<Action>();
+            foreach (var file in TreeView.Items.OfType<PerfViewFile>().Reverse())
```
brianrob: Why do you need to call `Reverse` here?

sharwell: It's constructing continuations. In another life this would have been async/await code, but here we need to know what runs after an action in order to construct the action:

1. Construct an action that executes the last item
1. Construct an action that executes the last two items, by executing the second to last item followed by the action from the previous step
1. Construct an action that executes the last there items, by executing the third to last item followed by the action from the previous step
1. Continue until you have an action that executes all items
1. Call the action that executes everything

brianrob: Ah, gotcha.  Thanks for the explanation.

PR #1088, opened 2020-01-05
Update `tha` to `the` in the xml for GCEnd

pgermishuys:
While working on replacing the performance counters in our application, I noticed the xml for GCEnd is referring to `tha` instead of `the`.


sharwell:
Thank you!

PR #1102, opened 2020-02-03
Improve SpeedScope exporter

adamsitnik:
Each sample contains `RelativeTime` and `Metric`.

For some reason, the trace files very often contain overlapping samples. An example: 

* Main: `RelativeTime=0` and `Metric=1`
* Main: `RelativeTime=0` and `Metric=2`

In the initial implementation of the exporter, I assumed that overlapping samples should be modified (see `MakeSureSamplesDoNotOverlap` method) so they just don't overlap.

Last week I've been converting a LOT of TechEmpower trace files to Speedscope and I was hitting this exception a lot:

https://github.com/adamsitnik/perfview/blob/abd56ddbc09106f7be2be32d1861522fdf5b784f/src/TraceEvent/Stacks/SpeedScopeStackSourceWriter.cs#L206-L207

What turned out to be a good solution it to sort the samples by relative time, then metric and later let `GetAggregatedOrderedProfileEvents` method merge them into continuous events. For the example above we just get an event that starts at time 0 and lasts 2.

@brianrob PTAL

### DIFF
--- a/src/TraceEvent/Stacks/SpeedScopeStackSourceWriter.cs
+++ b/src/TraceEvent/Stacks/SpeedScopeStackSourceWriter.cs
@@ -26,9 +26,6 @@ namespace Microsoft.Diagnostics.Tracing.Stacks.Formats
         {
             var samplesPerThread = GetSortedSamplesPerThread(source);
 
-            foreach (var samples in samplesPerThread.Values)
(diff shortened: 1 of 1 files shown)


brianrob:
@adamsitnik apologies for the delay.  LGTM!

adamsitnik:
@brianrob thanks!

PR #1104, opened 2020-02-06
Fix unaligned memory access errors in TraceEvent

swift-kim:
This fixes unaligned read errors (SIGBUS) caused by unmanaged pointer operations on ARM platforms. The [`Marshal.Read*()`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.readint64?view=netcore-3.1) operations are portable and have little performance impacts.

### DIFF
--- a/src/TraceEvent/TraceEvent.csproj
+++ b/src/TraceEvent/TraceEvent.csproj
@@ -68,6 +68,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles" Version="$(MicrosoftDiagnosticsTracingTraceEventSupportFilesVersion)" />
     <PackageReference Include="System.Reflection.Metadata" Version="1.5.0" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" />
 
   <!-- *** SourceLink Support *** -->
   <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
--- a/src/PerfView/PerfView.csproj
+++ b/src/PerfView/PerfView.csproj
@@ -254,6 +254,13 @@
       <Link>System.Reflection.Metadata.dll</Link>
       <Visible>False</Visible>
     </EmbeddedResource>
+    <EmbeddedResource Include="..\TraceEvent\$(OutDir)System.Runtime.CompilerServices.Unsafe.dll">
+      <Type>Non-Resx</Type>
+      <WithCulture>false</WithCulture>
+      <LogicalName>.\System.Runtime.CompilerServices.Unsafe.dll</LogicalName>
+      <Link>System.Runtime.CompilerServices.Unsafe.dll</Link>
+      <Visible>False</Visible>
+    </EmbeddedResource>
     <EmbeddedResource Include="..\TraceEvent\$(OutDir)Microsoft.Diagnostics.Tracing.TraceEvent.xml">
       <Type>Non-Resx</Type>
       <WithCulture>false</WithCulture>
--- a/src/TraceEvent/TraceEvent.cs
+++ b/src/TraceEvent/TraceEvent.cs
@@ -4368,19 +4368,21 @@ namespace Microsoft.Diagnostics.Tracing
         }
         internal static unsafe double ReadDouble(IntPtr pointer, int offset)
         {
-            return *((double*)((byte*)pointer.ToPointer() + offset));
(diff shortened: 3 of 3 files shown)


sharwell:
The JIT output seems very different ([reference](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8YUCAAcAymIBuvMDFwBuGjWIBmJqQYBhBgG8aDE0y3dc2AGYwGAEwgdgAG1sBZcgAoRGAAoYoBgkIXlEoNAZQjAYIKytcGAwASkNjU3TiAHYGACpPTwcnVxyk/OAATwwYEuCosRYAFQhfELDPFIBqGLiE5KTVanSAXzVB000GC2tbQpd3Um8ef0DasIio7vjElKMx9JNZbEDD5wYAXgY3I9wAC2xndhhsOx9UT1Wq8M3e/rT9pmyeVmxSSADITgNhjQhkA==)):

```csharp
public unsafe double M1(IntPtr pointer, int offset) {
    return *((double*)((byte*)pointer.ToPointer() + offset));
}
```

```
C.M1(IntPtr, Int32)
    L0000: vzeroupper
    L0003: movsxd rax, r8d
    L0006: vmovsd xmm0, qword [rdx+rax]
    L000b: ret
```

```csharp
public unsafe double M2(IntPtr pointer, int offset) {
    var val = Marshal.ReadInt64(pointer, offset);
    return *(double*)&val;
}
```

```
C.M2(IntPtr, Int32)
    L0000: sub rsp, 0x28
    L0004: vzeroupper
    L0007: mov rcx, rdx
    L000a: mov edx, r8d
    L000d: call System.Runtime.InteropServices.Marshal.ReadInt64(IntPtr, Int32)
    L0012: mov [rsp+0x20], rax
    L0017: vmovsd xmm0, qword [rsp+0x20]
    L001d: add rsp, 0x28
    L0021: ret
```

sharwell:
This option seems better:

```csharp
public unsafe double M3(IntPtr pointer, int offset) {
    return Unsafe.ReadUnaligned<double>(IntPtr.Add(pointer, offset).ToPointer());
}
```

```
C.M3(IntPtr, Int32)
    L0000: vzeroupper
    L0003: movsxd rax, r8d
    L0006: add rax, rdx
    L0009: vmovsd xmm0, qword [rax]
    L000d: ret
```

swift-kim:
@sharwell Thanks for the review and introducing a good tool.

- I didn't actually consider Unsafe.* because it required an additional package dependency. If you don't mind, I'll add it.

- I replaced `IntPtr.Add(pointer, offset).ToPointer()` → `(byte*)pointer.ToPointer() + offset` for consistency with other methods. They are equivalent according to the IL compiler you referenced.

src/TraceEvent/TraceEvent.csproj
```diff
@@ -71,6 +71,7 @@
 
   <!-- *** SourceLink Support *** -->
   <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
+  <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" />
```
sharwell: 💡  This should appear after line 70 (so it's not listed as part of SourceLink support)

sharwell: @brianrob @adamsitnik Please review this added dependency

sharwell: ❗️ I believe this new dependency needs one (or more if it has transitive dependencies) new blocks like this so the distribution is still complete:

https://github.com/microsoft/perfview/blob/b2f8e0ebd0434fac7e40960dc5521ce121c26a72/src/PerfView/PerfView.csproj#L250-L256

swift-kim: Thanks. I didn't check.

swift-kim: Done.

swift-kim:
For reminder, @sharwell @brianrob, could you take a look?

brianrob:
@swift-kim, thanks for contributing this.  This looks OK to me - just want to confirm that you've confirmed that the issue is fixed with the latest version of the PR, since you switched APIs?

swift-kim:
@brianrob Yes, I've confirmed dotnet-counters.dll now works in our production system. I'll report if I ever find any other issue. Thank you.

brianrob:
Thanks @swift-kim!

PR #1065, opened 2019-11-28
WIP: Improvements for use by GC infra

andy-ms:
* `GetCurrentGC` used to only return the current GC between the GC/Start and GC/Stop events. But some events would happen slightly before and after those.
  It will now return the current GC in between GC/SuspendEEStart and GC/RestartEEEnd (+1ms due to some events coming out slightly later).
  - So we are now creating the GC in `source.Clr.GCSuspendEEStart +=`. We are still filling in most of it in `GCStart` though.

* Adds code for correctly categorizing each join event to the correct GC and thread time. We can now handle join events that come out before the beginning or after the end of a GC.

* Adds ThreadId and IsBGCThread to GcJoin

* I am using the [Obsolete] annotation to mark things that are only "public" for the sake of the code in the performance repo.

* There are a lot of debug asserts that probably just never worked.
  (They won't affect normal PerfView users if they are using a release build.)
  I've commented out asserts that were failing for me.

PR #1107, opened 2020-02-20
Improvements for use by GC infra

VSadov:
Continuation of https://github.com/microsoft/perfview/pull/1065

### DIFF
--- a/src/TraceEvent/ETWTraceEventSource.cs
+++ b/src/TraceEvent/ETWTraceEventSource.cs
@@ -900,7 +900,7 @@ namespace Microsoft.Diagnostics.Tracing
         // Used to give TraceLogging events Event IDs. 
         private TraceLoggingEventId traceLoggingEventId;
 
-        [Obsolete]
+        [Obsolete("This is experimental, you should not use it yet for non-experimental purposes.")]
         public override string ProcessName(int processID, long time100ns)
         {
             string ret;
--- a/src/TraceEvent/ETWTraceEventSource.cs
+++ b/src/TraceEvent/ETWTraceEventSource.cs
@@ -900,7 +900,8 @@ namespace Microsoft.Diagnostics.Tracing
         // Used to give TraceLogging events Event IDs. 
         private TraceLoggingEventId traceLoggingEventId;
 
-        internal override string ProcessName(int processID, long time100ns)
+        [Obsolete]
+        public override string ProcessName(int processID, long time100ns)
         {
             string ret;
             if (!processNameForID.TryGetValue(processID, out ret))
--- a/src/TraceEvent/ETWTraceEventSource.cs
+++ b/src/TraceEvent/ETWTraceEventSource.cs
@@ -900,7 +900,6 @@ namespace Microsoft.Diagnostics.Tracing
         // Used to give TraceLogging events Event IDs. 
         private TraceLoggingEventId traceLoggingEventId;
 
-        [Obsolete("This is experimental, you should not use it yet for non-experimental purposes.")]
         public override string ProcessName(int processID, long time100ns)
         {
             string ret;
--- a/src/TraceEvent/Parsers/KernelTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/KernelTraceEventParser.cs
@@ -7,10 +7,13 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
 using Utilities;
 using Address = System.UInt64;
+using ProcessID = System.Int32;
+using ThreadID = System.Int32;
 
 /* This file was generated with the command */
 // traceParserGen /needsState /merge /renameFile KernelTraceEventParser.renames /mof KernelTraceEventParser.mof KernelTraceEventParser.cs
@@ -24,6 +27,47 @@ using Address = System.UInt64;
 // TODO I have low confidence in the TCP headers, especially for Versions < 2 (how much do we care?)
 namespace Microsoft.Diagnostics.Tracing.Parsers
 {
+    [Obsolete]
+    public interface IThreadIDToProcessID
+    {
+        ProcessID? ThreadIDToProcessID(ThreadID threadID, long timeQPC);
+        IEnumerable<ThreadIDAndTime> ProcessIDToThreadIDsAndTimes(ProcessID processID);
+    }
+
+    [Obsolete]
+    public struct ThreadIDAndTime
+    {
+        public readonly ThreadID ThreadID;
+        public readonly long TimeQPC;
+
+        public ThreadIDAndTime(ThreadID threadID, long timeQPC)
+        {
+            ThreadID = threadID;
+            TimeQPC = timeQPC;
+        }
+    }
+
+    [Obsolete]
+    internal class ThreadIDToProcessIDImpl : IThreadIDToProcessID
+    {
+        private readonly KernelTraceEventParserState state;
+        public ThreadIDToProcessIDImpl(KernelTraceEventParserState state)
+        {
+            this.state = state;
+        }
+
+        public ProcessID? ThreadIDToProcessID(ThreadID threadID, long timeQPC)
+        {
+            ProcessID res = state.ThreadIDToProcessID(threadID, timeQPC);
+            return res == -1 ? (ProcessID?) null : res;
+        }
+
+        public IEnumerable<ThreadIDAndTime> ProcessIDToThreadIDsAndTimes(ProcessID processID) =>
+           from entry in state.threadIDtoProcessID.Entries
+           where entry.Value == processID
+           select new ThreadIDAndTime((ThreadID) entry.Key, entry.StartTime);
+    }
+
     /// <summary>
     /// The KernelTraceEventParser is a class that knows how to decode the 'standard' kernel events.
     /// It exposes an event for each event of interest that users can subscribe to.
@@ -247,6 +291,14 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
         }
 
         public KernelTraceEventParser(TraceEventSource source) : this(source, DefaultOptionsForSource(source)) { }
+
+        [Obsolete]
+        public IReadOnlyHistoryDictionary<int> ThreadIDToProcessID =>
+            State.threadIDtoProcessID;
+
+        [Obsolete]
+        public IThreadIDToProcessID ThreadIDToProcessIDGetter =>
+            new ThreadIDToProcessIDImpl(State);
 
         public KernelTraceEventParser(TraceEventSource source, ParserTrackingOptions tracking)
             : base(source)
@@ -4450,7 +4502,15 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.Kernel
         // public int PreviousCState { get { return GetByteAt(10); } }
         // public int SpareByte { get { return GetByteAt(11); } }
 
-        public ThreadWaitReason OldThreadWaitReason { get { return (ThreadWaitReason)GetByteAt(0xc); } }
+        public ThreadWaitReason OldThreadWaitReason {
+            get {
+                int res = GetByteAt(0xc);
+                // Assert it's a valid reason 
+                // TODO: this fails all the time!!
+                // Debug.Assert(res <= (int) ThreadWaitReason.Unknown);
+                return (ThreadWaitReason)res;
+            }
+        }
         public ThreadWaitMode OldThreadWaitMode { get { return (ThreadWaitMode)GetByteAt(0xd); } }
         public ThreadState OldThreadState { get { return (ThreadState)GetByteAt(0xe); } }
         public int OldThreadWaitIdealProcessor { get { return GetByteAt(15); } }
@@ -7204,7 +7264,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.Kernel
         {
             Debug.Assert(!(Version == 0 && EventDataLength != 20));
             Debug.Assert(!(Version == 1 && EventDataLength < HostOffset(28, 1)));   // TODO fixed by hand
-            Debug.Assert(!(Version > 1 && EventDataLength < HostOffset(28, 1)));
+            //Debug.Assert(!(Version > 1 && EventDataLength < HostOffset(28, 1)));
             Action(this);
         }
         public override StringBuilder ToXml(StringBuilder sb)
@@ -7373,7 +7433,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.Kernel
         }
         protected internal override void Dispatch()
         {
-            Debug.Assert(!(Version == 2 && EventDataLength != HostOffset(36, 1)));
+            //Debug.Assert(!(Version == 2 && EventDataLength != HostOffset(36, 1)));
             Debug.Assert(!(Version > 2 && EventDataLength < HostOffset(36, 1)));
             Action(this);
         }
@@ -7481,7 +7541,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.Kernel
         }
         protected internal override void Dispatch()
         {
-            Debug.Assert(!(Version == 2 && EventDataLength != HostOffset(44, 1)));
+            //Debug.Assert(!(Version == 2 && EventDataLength != HostOffset(44, 1)));
             Debug.Assert(!(Version > 2 && EventDataLength < HostOffset(44, 1)));
             Action(this);
         }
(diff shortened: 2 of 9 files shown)


Maoni0:
I don't see any reason for marking ProcessName obsolete, do you?

Maoni0:
>         private static void MaybePrintEvent(TraceEvent te)

I would just move the check for DEBUG_PRINT_GC to PrintEvent instead of having another method called MaybePrintEvent. 

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:1643 in c36a12a. [](commit_id = c36a12a7b993251a0c1597664da9c36b7f12cb52, deletion_comment = False)

Maoni0:
> using Microsoft.Diagnostics.Tracing.Parsers.Tpl;

what are we using from Tpl?

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:16 in c36a12a. [](commit_id = c36a12a7b993251a0c1597664da9c36b7f12cb52, deletion_comment = False)

Maoni0:
>             // Not true, there may be that many heaps!

we should have this kind of comments checked in... this is completely temporary.

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:4360 in c36a12a. [](commit_id = c36a12a7b993251a0c1597664da9c36b7f12cb52, deletion_comment = False)

VSadov:
I am not sure about the whole "Obsolete" deal. Is that to avoid documenting public APIs ?

---
In reply to: [590551806](https://github.com/microsoft/perfview/pull/1107#issuecomment-590551806) [](ancestors = 590551806)

Maoni0:
I don't think we should be adding Obsolete annotations in these files (we do have Obsolete annotations in GC related stuff in TraceManagedProcess.cs so it's obvious which ones are "officially supported" and which ones are experimental - it's really a way to indicate experimental, not obsolete) - @brianrob can comment on the more general parts in TraceEvent (TraceEvent.cs and etc)

src/TraceEvent/Parsers/KernelTraceEventParser.cs
```diff
@@ -4450,7 +4507,15 @@ public enum ThreadWaitMode
         // public int PreviousCState { get { return GetByteAt(10); } }
         // public int SpareByte { get { return GetByteAt(11); } }
 
-        public ThreadWaitReason OldThreadWaitReason { get { return (ThreadWaitReason)GetByteAt(0xc); } }
+        public ThreadWaitReason OldThreadWaitReason {
+            get {
+                int res = GetByteAt(0xc);
+                // Assert it's a valid reason 
+                // TODO: this fails all the time!!
```
VSadov: I've not seen this assert fail after enabling, so I will enable it and remove TODO

src/TraceEvent/Parsers/KernelTraceEventParser.cs
```diff
@@ -7204,7 +7269,7 @@ protected internal override void Dispatch()
         {
             Debug.Assert(!(Version == 0 && EventDataLength != 20));
             Debug.Assert(!(Version == 1 && EventDataLength < HostOffset(28, 1)));   // TODO fixed by hand
-            Debug.Assert(!(Version > 1 && EventDataLength < HostOffset(28, 1)));
+            //Debug.Assert(!(Version > 1 && EventDataLength < HostOffset(28, 1)));
```
VSadov: Enable this. I do not see it triggered.

VSadov:
>             Debug.Assert(!(Version == 2 && EventDataLength != HostOffset(16, 1)));

This one actually fails frequently.

---
Refers to: src/TraceEvent/Parsers/KernelTraceEventParser.cs:9910 in 8c119ed. [](commit_id = 8c119ed7aa6d05a2887107886601e796d573870d, deletion_comment = False)

src/TraceEvent/Parsers/KernelTraceEventParser.cs
VSadov: Most other commented asserts do not seem triggered either.

---
In reply to: [384846815](https://github.com/microsoft/perfview/pull/1107#discussion_r384846815) [](ancestors = 384846815)

src/TraceEvent/TraceUtilities/HistoryDictionary.cs
```diff
@@ -193,7 +249,8 @@ internal HistoryValue(long startTime100ns, Address key, T value)
             internal Address key;
             internal long startTime;
             internal T value;
-            internal HistoryValue next;
+            //TODO:INTERNAL
+            public HistoryValue next;
```
VSadov: this does not need to be public

src/TraceEvent/TraceUtilities/HistoryDictionary.cs
```diff
@@ -131,25 +146,66 @@ public IEnumerable<HistoryValue> Entries
...
+            HashSet<E> set = new HashSet<E>();
+            foreach (E x in i)
+            {
+                if (!set.Contains(x))
+                {
+                    yield return x;
+                    set.Add(x);
+                }
+            }
+        }
```
VSadov: This is just a handwritten `Distinct` operator. Just call the operator.

VSadov:
>             // Not true, there may be that many heaps!

Yes. And check for the join type first.

---
In reply to: [590569873](https://github.com/microsoft/perfview/pull/1107#issuecomment-590569873) [](ancestors = 590569873)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:4360 in c36a12a. [](commit_id = c36a12a7b993251a0c1597664da9c36b7f12cb52, deletion_comment = False)

VSadov:
> using Microsoft.Diagnostics.Tracing.Parsers.Tpl;

"Remove/Sort Usings" removed these  and couple more.

---
In reply to: [590562804](https://github.com/microsoft/perfview/pull/1107#issuecomment-590562804) [](ancestors = 590562804)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:16 in c36a12a. [](commit_id = c36a12a7b993251a0c1597664da9c36b7f12cb52, deletion_comment = False)

VSadov:
>         private static void MaybePrintEvent(TraceEvent te)

Making "PrintEvent" conditional on  DEBUG_PRINT_GC would be even more appropriate

---
In reply to: [590554542](https://github.com/microsoft/perfview/pull/1107#issuecomment-590554542) [](ancestors = 590554542)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:1643 in c36a12a. [](commit_id = c36a12a7b993251a0c1597664da9c36b7f12cb52, deletion_comment = False)

VSadov:
>                 // TODO: clear entries in this dictionary when we're sure a GC is done (to save memory)

Remove corresponding join states on GCEnd

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:972 in 8c119ed. [](commit_id = 8c119ed7aa6d05a2887107886601e796d573870d, deletion_comment = False)

src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -900,7 +900,8 @@ private bool TraceEventBufferCallback(IntPtr rawLogFile)
         // Used to give TraceLogging events Event IDs. 
         private TraceLoggingEventId traceLoggingEventId;
 
-        internal override string ProcessName(int processID, long time100ns)
+        [Obsolete("This is experimental, you should not use it yet for non-experimental purposes.")]
+        public override string ProcessName(int processID, long time100ns)
```
brianrob: I think it is fine to make these public without the [Obsolete] attribute.

brianrob:
> using HeapID = System.Int32;

I've briefly looked through this file, but the changes are fairly significant.  Can you share an overall description for these changes?

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:23 in 1410320. [](commit_id = 141032008316318d2472ce08721d5c0da78c4b3f, deletion_comment = False)

Maoni0:
>                         // the final GCRestartEEStop after the end of the GC will be the final value of PauseEndRelativeMSec.

this is actually not used. do you really want to keep it?

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:734 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

Maoni0:
>                     foreach (ServerGcHistory hp in gc.ServerGcHeapHistories)

you could easily avoid doing this if all the heaps' thread id is already filled...instead of doing this over and over again for each GC

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:750 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

Maoni0:
>         internal TraceGC currentOrFinishedBGC = null;

what is the purpose of this...it just gets assigned at the same time as currentBGC?

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:5754 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

Maoni0:
>             double timeStampRelativeMSec,

this is also unused

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:2355 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

Maoni0:
>                         bool haveEnoughGCs = stats.GC.GCs.Count >= 3;

we can't do this... sometimes we get a trace with just one complete GC if we are collecting so many events (CSwitch is very verbose), we need to be able to analyze that.

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:1009 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

VSadov:
>                         bool haveEnoughGCs = stats.GC.GCs.Count >= 3;

This was in the original PR. I do not know the history of this requirement. 

I actually did try relaxing this requirement and that resulted in asserts and failures. Do we want to change this before merging this or as a separate item? This may not be a simple change.

---
In reply to: [594239621](https://github.com/microsoft/perfview/pull/1107#issuecomment-594239621) [](ancestors = 594239621)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:1009 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

Maoni0:
well, this is a consideration you make when you are writing tooling code so it's not specifically spelled out as a requirement for this particular PR. if the tooling code is often used to handle traces with very few GCs then you must handle them. collecting traces with CSwitch, since CSwitch is such a voluminous event, could easily mean you only get traces with one or a couple of GCs.

VSadov:
Yes it may take long time to get to 3+ GCs.  That is why I tried to relax the requirement in the original code. It is not trivial though. We seem to depend on this in other places. So I was not sure if the limitation is actually expected and/or whether the work to reduce the limit should happen in this PR or as a separate item.

---
In reply to: [594964902](https://github.com/microsoft/perfview/pull/1107#issuecomment-594964902) [](ancestors = 594964902)

src/TraceEvent/ETWTraceEventSource.cs
VSadov: I will make these public then. Without the attribute.

---
In reply to: [385882429](https://github.com/microsoft/perfview/pull/1107#discussion_r385882429) [](ancestors = 385882429)

VSadov:
>             double timeStampRelativeMSec,

removed. Here  and in all places that pass and propagate the relative time

---
In reply to: [594237717](https://github.com/microsoft/perfview/pull/1107#issuecomment-594237717) [](ancestors = 594237717)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:2355 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

VSadov:
>         internal TraceGC currentOrFinishedBGC = null;

Yes. It gets assigned at the same time, but later `currentBGC` may be assigned to null if we detect that it has ended. This is used to keep track of what _was_ the last BGC if we do not have one currently.

---
In reply to: [594237176](https://github.com/microsoft/perfview/pull/1107#issuecomment-594237176) [](ancestors = 594237176)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:5754 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

VSadov:
>                         // the final GCRestartEEStop after the end of the GC will be the final value of PauseEndRelativeMSec.

It is used in an assert in `GetCurrentGCForJoin`. It is easy to remove though, since it is debug-only.

---
In reply to: [594216006](https://github.com/microsoft/perfview/pull/1107#issuecomment-594216006) [](ancestors = 594216006)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:734 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

VSadov:
>                         // the final GCRestartEEStop after the end of the GC will be the final value of PauseEndRelativeMSec.

It is public though and used by the analysis though, so it cannot be removed.

---
In reply to: [594984933](https://github.com/microsoft/perfview/pull/1107#issuecomment-594984933) [](ancestors = 594984933,594216006)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:734 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

VSadov:
>                         // the final GCRestartEEStop after the end of the GC will be the final value of PauseEndRelativeMSec.

It is assigned here even though it will be assigned again on GCStop. I am guessing this is the best value that we know, so we are assigning it - in case GCStop event is missing, we will still have some valid and reasonable value.

---
In reply to: [594986880](https://github.com/microsoft/perfview/pull/1107#issuecomment-594986880) [](ancestors = 594986880,594984933,594216006)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:734 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

VSadov:
I will try removing the `3+ GCs` requirement. 

---
In reply to: [594976219](https://github.com/microsoft/perfview/pull/1107#issuecomment-594976219) [](ancestors = 594976219,594964902)

VSadov:
>                     foreach (ServerGcHistory hp in gc.ServerGcHeapHistories)

It looks like this is done mostly for the Assert in the else block.

---
In reply to: [594218265](https://github.com/microsoft/perfview/pull/1107#issuecomment-594218265) [](ancestors = 594218265)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:750 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

VSadov:
>                     foreach (ServerGcHistory hp in gc.ServerGcHeapHistories)

I am not sure the `FinishUpGC` is needed at all. We will know all thread/heap associations by the time we set up histories and will not know any better later. This can be removed.

---
In reply to: [595966614](https://github.com/microsoft/perfview/pull/1107#issuecomment-595966614) [](ancestors = 595966614,594218265)

---
Refers to: src/TraceEvent/Computers/TraceManagedProcess.cs:750 in 1513d0f. [](commit_id = 1513d0f6d16f5ac319a01cc9b215c75e1555c5aa, deletion_comment = False)

VSadov:
I think I am done with the last round of requests. 
I cannot resolve comments that are not mine in the CodeFlow (I guess that is because of the permissions on the repo), but I looked through all of them.

In this iteration:
-	Relaxed 3+ GC requirement.   
it was really a hack to compensate for tying heap count calculation with EE suspension events, which could be lost or not in the range of collected trace, so we waited for 3 GCs.
-	Changing when we calculate heap count and thread to heap mapping also made FinishUpGC unnecessary. 
That resolved concerns about the method

-	Fixed a bug where we could detect more heaps than really exists if trace collection started after process starts.
I.E. I saw 18 heaps detected on my machine where it should be 12.
That was a common case for failures for me locally. Typical reason is expecting more join ends than threads and thus getting into impossible states.
-	A few minor bugs in the analysis.
-	Made a few changes to make it more stable in situations that could be caused by missing events. 
This is common when getting CPU samples. On large traces it is almost certain and we were too eager to throw exceptions. 
The whole thing is much more reliable now. Please take a look.

brianrob:
@VSadov, since this is going to be re-architected, can you please close this until you're ready to bring it back for review?  Thanks.

VSadov:
@brianrob  No problem. Thanks!

PR #1108, opened 2020-02-21
Add start/stop and kernel tracing to perfcollect

ooooolivia:
- Added start/stop to perfcollect
- Added noperf option to turn off perf when necessary
- Enabled lttng tracing of kernel events 

New usage would be:
```./perfcollect collect start <trace_name> -noperf``` to start a lttng-only trace 
```./perfcollect collect stop <trace_name>``` to stop a trace
```./perfcollect collect start <trace_name> -noperf -events kernel``` to start lttng-only kernel trace (currently only process exec/exit events, will add more)

### DIFF
--- a/src/perfcollect/perfcollect
+++ b/src/perfcollect/perfcollect
@@ -1814,7 +1814,7 @@ DoCollect()
         # Pass trace directory and session name to the stop process.
         popd > /dev/null
         usePerf=0
-        declare -p | grep -e lttngTraceDir -e lttngSessionName -e tempDir -e usePerf -e useLttng > $collectInfoFile
+        declare -p | grep -e lttngTraceDir -e lttngSessionName -e tempDir -e usePerf -e useLttng -e logFile > $collectInfoFile
     else
         if [ "$usePerf" == "1" ]
         then
@@ -1944,7 +1944,10 @@ fi
 EnsurePrereqsInstalled
 
 # Initialize the log.
-InitializeLog
+if [ "$1" != "stop" ]
+then
+    InitializeLog
+fi
 
 # Process arguments.
 ProcessArguments $@
--- a/src/perfcollect/perfcollect
+++ b/src/perfcollect/perfcollect
@@ -15,13 +15,21 @@
 # How to collect a performance trace:
 # 1. Prior to starting the .NET process, set the environment variable COMPlus_PerfMapEnabled=1.
 #    This tells the runtime to emit information that enables perf_event to resolve JIT-compiled code symbols.
+# 
 # 2. Setup your system to reproduce the performance issue you'd like to capture.  Data collection can be
 #    started on already running processes.
-# 2. Run this script: sudo ./perfcollect collect samplePerfTrace
-#    This will start data collection.
-# 3. Let the repro run as long as you need to capture the performance problem.
-# 4. Hit CTRL+C to stop collection.
-#    When collection is stopped, the script will create a trace.zip file matching the name specified on the
+# 
+# 3. [Usage #1] use "collect" command
+#    - Run this script: sudo ./perfcollect collect samplePerfTrace. This will start data collection.
+#    - Let the repro run as long as you need to capture the performance problem.
+#    - Hit CTRL+C to stop collection.
+#
+#    [Usage #2] use "start" and "stop" command
+#    - Run this script: sudo ./perfcollect start samplePerfTrace. This will start data colletion.
+#    - Let the repro run as long as you need to capture the performance problem.
+#    - Run: sudo ./perfcollect stop samplePerfTrace. This will stop collection.
+#
+# 4. When collection is stopped, the script will create a trace.zip file matching the name specified on the
 #    command line.  This file will contain the trace, JIT-compiled symbol information, and all debugging
 #    symbols for binaries referenced by this trace that were available on the machine at collection time.
 #
@@ -1687,13 +1695,19 @@ PrintUsage()
     echo "For detailed collection and viewing steps, view this script in a text editor or viewer."
     echo ""
     echo "./perfcollect <action> <tracename>"
-    echo "Valid Actions: collect view livetrace install"
+    echo "Valid Actions: collect start/stop view livetrace install"
     echo ""
     echo "collect options:"
     echo "By default, collection includes CPU samples collected every ms."
     echo "    -pid          : Only collect data from the specified process id."
     echo "    -threadtime       : Collect context switch events."
     echo "  -hwevents         : Collect (some) hardware counters."
+    echo ""
+    echo "start:"
+    echo "    Start collection, but with Lttng trace ONLY. It needs to be used with 'stop' action."
+    echo ""
+    echo "stop:"
+    echo "    Stop collection if 'start' action is used."
     echo ""
     echo "view options:"
     echo "    -processfilter      : Filter data by the specified process name."
--- a/src/perfcollect/perfcollect
+++ b/src/perfcollect/perfcollect
@@ -655,16 +655,14 @@ LogAppend()
 
 RunSilent()
 {
-    #HACK 
-    echo $*
-    # if (( $logEnabled == 1 ))
-    # then
-    #     echo "Running \"$*\"" >> $logFile
-    #     $* >> $logFile 2>&1
-    #     echo "" >> $logFile
-    # else
-    #     $* > /dev/null 2>&1
-    # fi
+    if (( $logEnabled == 1 ))
+    then
+        echo "Running \"$*\"" >> $logFile
+        $* >> $logFile 2>&1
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/perfcollect/perfcollect [hunk 1]
```diff
@@ -1171,13 +1180,25 @@ ProcessArguments()
...
+
     # Set the data file.
     inputTraceName=$2
-    if [ "$inputTraceName" == "" ]
+    if [ "$inputTraceName" == "" ] 
     then
         FatalError "Invalid trace name specified."
     fi
 
+    args=( "$@" )
```
ooooolivia: Will remove this in the next commit

src/perfcollect/perfcollect
```diff
@@ -1171,13 +1180,25 @@ ProcessArguments()
         FatalError "Invalid action specified."
     fi
 
+    # Validate operation name.
+    if [ "$action" == "collect" ]
+    then
```
brianrob: Overall, this seems pretty clean.  I would like to ask for a few changes, which I'll add here in a few comments.  The most structural request is this: Rather than having the command be "collect start" or "collect stop" can you just make it "start" or "stop"?  Then you can leave "collect" as is.  This is what we do in PerfView, so the behavior will extend well to PerfCollect.

src/perfcollect/perfcollect
```diff
@@ -607,6 +607,11 @@ declare -a DotNETRuntimePrivate_DynamicTypeUsageKeyword=(
     DotNETRuntimePrivate:ObjectVariantMarshallingToManaged
 )
 
+declare -a Kernel_Default=(
```
brianrob: Can you name this something like "LTTng_Kernel_Default"?

src/perfcollect/perfcollect
```diff
@@ -1280,53 +1304,61 @@ CreateLTTngSession()
...
-        EnableLTTngEvents ${DotNETRuntimePrivate_GCPrivateKeyword[@]}
-        EnableLTTngEvents ${DotNETRuntime_JitKeyword[@]}
-        EnableLTTngEvents ${DotNETRuntime_LoaderKeyword[@]}
-        EnableLTTngEvents ${DotNETRuntime_ExceptionKeyword[@]}
-    elif [ "$gcWithHeap" == "1" ]
-    then
-        usePerf=0
-        EnableLTTngEvents ${DotNETRuntime_GCKeyword[@]}
-        EnableLTTngEvents ${DotNETRuntime_GCHeapSurvivalAndMovementKeyword[@]}
+        RunSilent "lttng add-context --kernel -t pid -t procname"
```
brianrob: Rather than treating the kernel events differently in SetupLTTngSession, it would be great to have them interleaved into the rest of the code.  It is safe to run add-context even if no events come through.  So you should be OK to just add context to both the kernel and user space channels unconditionally.  The one question is whether or not it's safe to call lttng add-context --kernel if there is no kernel space channel (e.g. if the LTTng kernel modules aren't loaded).  Can you check this?

Then, you should just be able to add another if statement on the same level as if [ "$events" == "" ] and just call EnableLTTngKernelEvents.

brianrob:
> EnableLTTngEvents()

Can you rename this to EnableLTTngUserEvents?

---
Refers to: src/perfcollect/perfcollect:1387 in 0b8fe17. [](commit_id = 0b8fe177930fbcd26f098c607e4550a42c2a906e, deletion_comment = False)

src/perfcollect/perfcollect
```diff
@@ -1373,7 +1414,6 @@ ProcessCollectedData()
     local directoryName=$traceName$traceSuffix
     mkdir $directoryName
 
-    # Save LTTng trace files.
```
brianrob: I think this deletion was unintentional?

src/perfcollect/perfcollect
```diff
@@ -1625,6 +1665,15 @@ CTRLC_Handler()
 
 EndCollect()
 {
+    # The user must either use "collect stop" or CTRL+C to stop collection.
+    if [ "$operation" == "stop" ]
+    then 
+        # Recover trace info in the previous program.
+        source $collectInfoFile  # TODO: exit and dispose upon missing file
```
brianrob: Agreed that it's important to handle the case where the file isn't there.  Also, it's useful to handle the case where someone calls start two times in a row.

brianrob: It looks like the collection info file lives in the current working directory.  It should live in the temp directory to avoid cases where you can't write to the current working directory.

src/perfcollect/perfcollect
```diff
@@ -1765,16 +1814,21 @@ DoCollect()
     then
         WriteStatus "Collection started. Collection will automatically stop in $duration second(s).  Press CTRL+C to stop early."
     else
-            WriteStatus "Collection started. Press CTRL+C to stop."
+        WriteStatus "Collection started."
```
brianrob: Please make sure to leave the CTRL+C message here if the user runs perfcollect collect so that they know what to do to stop collection.

src/perfcollect/perfcollect
```diff
@@ -634,6 +639,9 @@ declare logFile
 logFilePrefix='/tmp/perfcollect'
 logEnabled=1
 
+# Collect info to pass between processes
+collectInfoFile='collectinfo'
```
brianrob: Can you rename this to something like "perfcollect.sessioninfo" so that it's clear if someone runs across this what it is?

src/perfcollect/perfcollect
ooooolivia: Having "start/stop" only sounds good to me! Will change.

src/perfcollect/perfcollect
ooooolivia: Yes I'll revert it

src/perfcollect/perfcollect
ooooolivia: I was thinking about this scenario --- currently it seems we don't handle when two collections are on simultaneously? My assumption is we allow this to happen and they just different sessions run in parallel.

brianrob: This could work for LTTng, but I'm not sure it would work for perf.  Either way, I think the key si that we don't want to settings file to get clobbered as a result of multiple runs in different consoles.

ooooolivia: Thought about the same thing but the purpose of collection info file is to locate where the temp dir is. After exiting from the previous program, we totally lost this info.

brianrob: Oh, right - we create a random directory.  You should still be able to find the actual temp directory (/tmp) and put the file in there.

ooooolivia: You're right I forgot we can use ```/tmp```

src/perfcollect/perfcollect
ooooolivia: Good question, could you clarify  " (e.g. if the LTTng kernel modules aren't loaded). " ? I assume we always add context before channel is created? 
Based on the original code, we add context first, and the default channel will be created when we enable the events.

src/perfcollect/perfcollect
```diff
@@ -1280,10 +1292,12 @@ CreateLTTngSession()
 
 SetupLTTngSession()
 {
+    
     # Setup per-event context information.
     RunSilent "lttng add-context --userspace --type vpid"
     RunSilent "lttng add-context --userspace --type vtid"
     RunSilent "lttng add-context --userspace --type procname"
+    RunSilent "lttng add-context --kernel -t pid -t procname"
```
billwert: Does this taking a pid imply we've started the target process before we start tracing?

brianrob: No, this is telling LTTng to add the pid and process name to every event.  "lttng add-context" allows you to add additional context information to each event that comes from a particular channel (user or kernel).

src/perfcollect/perfcollect
```diff
@@ -1327,6 +1341,11 @@ SetupLTTngSession()
         elif [ "$events" == "threading" ]
         then
             EnableLTTngEvents ${DotNETRuntime_ThreadingKeyword[@]}
+        elif [ "$events" == "kernel" ]
```
brianrob: Hmm... your comment here makes a good point.  Let's change this to "processlifetime" (and adjust the other names like "LTTng_Kernel_Default" to "Lttng_Kernel_ProcessLifetime") so that we don't define the kernel name.

brianrob:
Looking good.  I think just one more thing and then this is likely ready to merge.

src/perfcollect/perfcollect
ooooolivia: good point!

brianrob:
@ooooolivia this looks good.  One final request - can you please add start, stop to the list of valid verbs for the usage message.  Other than that, LGTM and we should be ready to merge this.

PR #1110, opened 2020-03-05
Add Linux Kernel Parser to TraceEvent

ooooolivia:
- Add Linux Kernel Parser to TraceEvent parsers and make existing ctf parsing tools usable on Linux kernel events. 
- This is just a prototype, which includes only Linux Process Start/Stop events. More events and other details will be added on demand.


src/TraceEvent/Ctf/CtfMetadataLegacyParser.cs
```diff
@@ -39,6 +39,8 @@ public CtfMetadataLegacyParser(Stream stream)
         public override IEnumerable<CtfMetadataDeclaration> Parse()
         {
             string metadata = GetMetadata();
+            // add new line before string in case some metadata doesn't start with \n (kernel metadata)
+            metadata = "\n" + metadata;
```
brianrob: Can you describe why this is needed?

src/TraceEvent/Ctf/CtfReader.cs
```diff
@@ -90,14 +90,19 @@ public IEnumerable<CtfEventHeader> EnumerateEventHeaders()
...
+            CtfInteger vtid = null;
             CtfArray processName = null;
             string lastProcessName = "";
             int processLen = 0;
             CtfStruct eventContext = _streamDefinition.EventContext;
             if (eventContext != null)
             {
-                pid = (CtfInteger)eventContext.GetField("_vpid")?.Type;
-                tid = (CtfInteger)eventContext.GetField("_vtid")?.Type;
+                vpid = (CtfInteger)eventContext.GetField("_vpid")?.Type;
```
brianrob: At this point, do we know if this is a kernelspace or userspace event?  I'd like to be able to deterministically select between vpid and vtid or pid and tid depending upon the channel (kernel or user) rather than depending upon the existence of one or the other.  It is technically possible for usermode events to contain context records for vpid, vtid, pid, and tid at the same time.

src/TraceEvent/Ctf/CtfTraceEventSource.cs
```diff
@@ -113,6 +113,11 @@ select int.Parse(processor)
         {
             Dictionary<string, ETWMapping> result = new Dictionary<string, ETWMapping>();
 
+            // Kernel events
```
brianrob: Can you make it clear in the comment here that these are Linux Kernel events?

src/TraceEvent/Parsers/LinuxKernelEventParser.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        #endregion
+    }
+}
+
+namespace Microsoft.Diagnostics.Tracing.Parsers.LinuxKernel
+{
+    public sealed class ProcessStartTraceData : TraceEvent
+    {
+/*      public int ProcessID { get { return GetInt32At(0); } }
```
brianrob: I'm assuming that these are commented out because these properties exist in the `TraceEvent` base class?  You can probably get away with not having these fields here at all because they are present in the event context associated with each event if you use `lttng add-context`.  Are there other fields beyond these that we want to capture, such as command line?

src/TraceEvent/Ctf/CtfMetadataLegacyParser.cs
ooooolivia: In line 60 ``` int start = metadata.Substring(0, open).LastIndexOf('\n') + 1;``` the parser always looks for a new line before parsing a metadata declaration. This doesn't work for kernel metadata, which doesn't start with a new line and the first declaration will not be found by this code. I need to do more tests to know if my assumption that "only kernel metadata begins without a new line'' but it is more safe to add a new line before the metadata string for either case. Please let me know if there's better approach than this hack...

src/TraceEvent/Parsers/LinuxKernelEventParser.cs
ooooolivia: Yes these properties already exist in the base class and sometimes the parsing doesn't work properly. I plan to add other fields after this prototype works.

src/TraceEvent/Ctf/CtfReader.cs
ooooolivia: Good point! I assumed we only turn on vpid and vtid for user mode. I need to investigate if user/kernel mode is known at this point. Also my question would be, under the scenario that both vpid and pid exist, we should probably store both fields, and which one should we use to match pids in startup? I assume vpid only exists if we use a container. (Didn't find good documentation online...)

ooooolivia:
I would prefer merge the prototype to validate the change of the ctf parser and fix possible bugs earlier. If this code won't be used immediately I'm fine with keeping working on this PR.

src/TraceEvent/Ctf/CtfReader.cs
ooooolivia: Just checked, we do know the domain information. It exists in the metadata passed to the reader. So by default, we will use vpid/vtid for user mode and pid/tid for kernel mode.

brianrob:
Spoke to @ooooolivia offline.  She will test locally and then we'll get this merged once local testing is complete and feedback is addressed.

src/TraceEvent/Ctf/CtfTracing.Tests/CtfTraceTests.cs
```diff
@@ -133,5 +138,33 @@ public void LTTng_GCStartStopEvents()
...
+                string path = Path.Combine(TestDataDirectory, file);
+                using (CtfTraceEventSource ctfSource = new CtfTraceEventSource(path))
+                {
+                    var kernelParser = new LinuxKernelEventParser(ctfSource);
+
+                    var procStartList = new List<TraceEvent>();
+                    var gcList = new List<TraceEvent>();
+                    kernelParser.ProcessStart += delegate (ProcessStartTraceData data)
+                    {
+                        procStartList.Add(data.Clone());
```
brianrob: This looks great.  Can I recommend that instead of cloning the items and adding them to a list that you count the number of each event and then assert that it matches the expected count at the end?  This adds just a bit more robustness.

src/TraceEvent/Parsers/LinuxKernelEventParser.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        public static ulong GetKeywords() { return 1; }
+        public static string GetProviderName() { return "Linux-Kernel"; }
+        public static Guid GetProviderGuid() { return new Guid("4da4eb17-cd7d-4b13-b807-22908d2996b5"); }
+        private event Action<ProcessStartTraceData> Action;
+        #endregion
+    }
+    public sealed class ProcessStopTraceData : TraceEvent
+    {
+/*      public int ProcessID { get { return GetInt32At(0); } }
```
brianrob: Can you please add some comments around the commented out sections of code in this file?  I think most things are commented out because you're actually getting the data from the context stream instead of the event payload, which is fine.  I just want it to be clear to readers in the future.  If you think any of the commented-out code should never be present then let's remove it.

brianrob:
@ooooolivia a few more small comments, but overall this looks good.  Once you have these fixed up, we can merge this if you're ready.

src/TraceEvent/Ctf/CtfTracing.Tests/CtfTraceTests.cs
ooooolivia: I was thinking about what asserts it can do and this sounds like a good one!

src/TraceEvent/Parsers/LinuxKernelEventParser.cs
```diff
@@ -162,6 +164,8 @@ protected internal override void Dispatch()
         }
         protected internal override void Validate()
         {
+            // TODO: write new Validate for this TraceEvent
```
brianrob: Looks great!  Is this TODO still applicable?  If I understand correctly, it can be removed.

ooooolivia: I'm actually not sure about this one.  I commented out the automatically generated validate method because the assert doesn't pass. The original one looks very confusing to me, and I wonder if we need to write new ``` Validate()``` for this traceevent. If so, my plan is to leave it to the next PR.

brianrob: I think you can get rid of this TODO - as you pointed out to me offline the data in the event isn't being parsed properly.  That would need to be fixed before a new Validate method can be written.

ooooolivia: Seems more complicated than I thought. I'll resolve it for now and we can talk about this later offline.

brianrob: Sounds good.  Is this ready for merge?

ooooolivia: yes I'll merge it

ooooolivia: Yes, just realized I can't merge it.

PR #1112, opened 2020-03-06
move ParallelLinuxPerfScriptStackSource to TraceEvent, fixes #1103

### LINKED ISSUE #1103: How to open a PerfCollect trace file using TraceEvent
I want to extend the `dotnet trace` tool with the possibility to convert trace files captured with PerfCollect to Speedscope file format. Just to be able to run perf investigation on Linux machine without copying the trace file to Windows and converting it to SpeedScope using PerfView.

I was hoping that it's going to be a matter of replacing [this](https://github.com/dotnet/diagnostics/blob/42f505ef918f7aabccad7de4e4e7aab7318f5de1/src/Tools/dotnet-trace/TraceFileFormatConverter.cs#L66) line of code with:

```cs
// before
var etlxFilePath = TraceLog.CreateFromEventPipeDataFile(fileToConvert, null, new TraceLogOptions() { ContinueOnError = continueOnError } );
// after
var etlxFilePath =
    fileToConvert.EndsWith(".trace.zip")
        ? TraceLog.CreateFromLttngTextDataFile(fileToConvert, null, new TraceLogOptions() { ContinueOnError = continueOnError })
        : TraceLog.CreateFromEventPipeDataFile(fileToConvert, null, new TraceLogOptions() { ContinueOnError = continueOnError });
```

However, for all the trace files I have I am getting following exception:

```log
System.IO.EndOfStreamException: No CTF Information found in ZIP file.
   at Microsoft.Diagnostics.Tracing.CtfTraceEventSource..ctor(String fileName)
   at Microsoft.Diagnostics.Tracing.Etlx.TraceLog.CreateFromLttngTextDataFile(String filePath, String etlxFilePath, TraceLogOptions options)
```

The exception is thrown here:

https://github.com/microsoft/perfview/blob/2fb635bcde289b2995570001334c08934371003c/src/TraceEvent/Ctf/CtfTraceEventSource.cs#L83-L86

Since PerfView was able to open this trace file I've debugged it and it turned out that the exception is handled by the `GetTraceLog` method:

https://github.com/microsoft/perfview/blob/b2f8e0ebd0434fac7e40960dc5521ce121c26a72/src/PerfView/PerfViewData.cs#L8463-L8466

and later [ParallelLinuxPerfScriptStackSource](https://github.com/microsoft/perfview/blob/master/src/PerfView/OtherSources/Linux/LinuxPerfScriptStackSource.cs) is used to create a `StackSource` from the trace file. The problem is that `ParallelLinuxPerfScriptStackSource` is not part of the `TraceEvent` library so I can't use it outside of PerfView.

Have the LTTng trace file format changed and the throwing code should be updated?
Would it be OK to move `ParallelLinuxPerfScriptStackSource` to `TraceEvent` library? (I could send a PR ofc)

/cc @brianrob

adamsitnik:
cc @brianrob


adamsitnik:
Please excuse me, but for some reason the build succeded on my machine while it should have failed.

It turns out that .NET Standard 1.6 is missing some very basic APIs and types like `ApplicationException` or `Stream.Close`:

* https://apisof.net/catalog/System.ApplicationException
* https://apisof.net/catalog/System.IO.Stream.Close()
* https://apisof.net/catalog/System.Threading.Thread.CurrentCulture
* https://apisof.net/catalog/System.Text.Encoding.Default
* https://apisof.net/catalog/System.Diagnostics.Contracts.Contract.Requires(Boolean,String)

@brianrob I've added a rule to the project file to not compile these types for .NET Standard 1.6. I hope that it's OK.

brianrob:
> namespace Diagnostics.Tracing.StackSources

As part of the move, let's change the namespace for these classes to be Microsoft.Diagnostics.Tracing.StackSources.

---
Refers to: src/TraceEvent/Stacks/Linux/LinuxPerfScriptEventParser.cs:15 in e3658a4. [](commit_id = e3658a44869cf778926d14c51d2b7dcdbd3ec563, deletion_comment = False)

brianrob:
> namespace PerfView.Utilities

Let's put this one into Microsoft.Diagnostics.Tracing.Utilities.

---
Refers to: src/TraceEvent/Utilities/FastStream.cs:6 in e3658a4. [](commit_id = e3658a44869cf778926d14c51d2b7dcdbd3ec563, deletion_comment = False)

brianrob:
@adamsitnik I think it will be fine to not compile these types for the netstandard1.6 target.  I believe this is around for users of older runtimes (e.g. older desktop and .NET Core 1.x).  It's actually probably safe to remove the netstandard1.6 target entirely, but we don't need to do that now.

adamsitnik:
@brianrob thanks for the review! I've changed the namespaces (which was a good idea BTW)

PR #1113, opened 2020-03-06
Chromium trace file format exporter

adamsitnik:
While working on improving our results in TechEmpower I realized that we don't have any way of visualizing the work of all CPUs on a single screen using Linux. Speedscope is great, but it has a per-thread view and it's hard to diagnose threading issues. I was basically missing VS Concurrency Visualizer on Linux.

I've reminded myself about the work @josalem did in https://github.com/dotnet/diagnostics/issues/447#issuecomment-550555045 to implement an exporter to Chromium Trace Event File format.

By using it I was able to quickly identify one major bottleneck and [improve](https://github.com/dotnet/runtime/pull/2346) almost all TechEmpower benchmarks by 2-10% RPS (including the super-optimized Plaintext). 

obraz

obraz

How to use it:

1. In PerfView, choose "File => Save View As => Chromium Trace Event" and export the trace file.

obraz

2. Go to chromium and type `chrome://tracing/`  (or `edge://tracing/` if you are using latest Edge)

3. Choose "Load" from the top left corner, select the exported trace file and click open

obraz

4. Enjoy

obraz

An alternative: Go to Chromium, press F12 (to open Developer Tools) and choose "Load profile" from top left corner:

obraz

The [Trace Event Format](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU) is very similar to speedscope and I was able to reuse 95% of speedscope exporter code and just extend it with the missing 5%. 

This PR combined with #1112 is going to allow me to extend `dotnet trace` with the possibility to convert `.nettrace` and `.trace.zip` files to the new file format and improve our user experience on Linux.

/cc @brianrob @tmds @noahfalk


josalem:
This is phenomenal @adamsitnik!  I'm happy to see that experiment prove helpful 😃

brianrob:
@adamsitnik this is great.  Does the chromium view show you stacks?

adamsitnik:
>  Does the chromium view show you stacks?

Yes, when I select something from the graph (every sample is represented as event and it's clickable) in the `edge://tracing/` it shows a simple call stack: 

obraz

The "DevTools View" (F12) can also show it, but it requires the user to expand the call stack manually by clicking on the arrow ;/ 

obraz

Also, if you zoom in the time you can see method names:

obraz

brianrob:
That's awesome!  Let me take a look at the code diff here and get back to you.

src/TraceEvent/Stacks/ChromiumStackSourceWriter.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// </summary>
+        public static void WriteStackViewAsJson(StackSource source, string filePath, bool compress)
+        {
+            if (File.Exists(filePath))
+                File.Delete(filePath);
+
+            using (var writeStream = File.CreateText(filePath))
+                Export(source, writeStream, Path.GetFileNameWithoutExtension(filePath));
+
+            if (compress)
```
brianrob: It looks like PerfView doesn't ever compress the file, but for apps that do, would it make sense to not unconditionally write out the file and then open it and read it back in to compress it?  Presumably, Export could take a compressed stream (or FileStream) and just write to it.

adamsitnik: that was a very good suggestion, #done

PR #1117, opened 2020-03-10
Fix CTF Tests

brianrob:
The CTF tests have been disabled for some time, and so now the files that are present in the repo are no longer usable as breaking changes were made some time back.  These changes were made intentionally, and so the fix is to capture new traces on newer runtimes.  This change captures a 2.2 and a 3.1 trace and runs them against the tests.


brianrob:
@ooooolivia, FYI, this change will re-enable the CTF tests.  Note that it also skips over the kernel metadata document if present because the document doesn't contain any trace information.  This is likely because I'm not collecting any trace data.  So, you may need to modify this as part of your changes to support kernel events to conditionally look at the kernel metadata if kernel events are present.  Let me know if you have questions and we can work through the details.

PR #1120, opened 2020-03-12
Removed all compiler warnings.

HMartyrossian:
Removed warnings reported by C# compiler. Fixes issue #1121.


brianrob:
@HMartyrossian thanks for the contribution.  For the most part this looks good.  The one thing that I would like to change or revert is the overriding of the properties in ClrTraceEventParser and WpfTraceEventParser.  Both of these properties match the manifest used to produce the events, but happen to overlap with the base class property Source and the base class field source, which are used to provide access to the underlying TraceEventSource that produced the event.  The fact that the TraceEventSource isn't accessible without a cast is a bug, and so rather than making this change to make the warning go away, I would like to modify the code to make the TraceEventSource accessible.  I think that you can do this reasonably in two steps:

1. Rename `TraceEvent.source` since it is internal to something like `TraceEvent.traceEventSource`.
2. Rename `KnownPathProbedTraceData.Source' to 'KnownPathProbedTraceData.PathSource'.

I took a quick look at the CI failure as well, but that doesn't appear related to your change so hopefully re-triggering it will address the failure.

PR #1124, opened 2020-03-14
[TraceEvent]: Fix CreateDisposition in CreateFile - it's the KM enum

NightRa:
In the CreateFile event, the Disposition byte consists of the KM enum for the disposition, not the UM one passed to CreateFileW.

This PR changes the values of the enum in a source-compatible way.

For future reference, the data in the ETW event is taken from the _FLT_IO_PARAMETER_BLOCK.Create struct, and probably another union branch for other file events.

### DIFF
--- a/src/TraceEvent/Parsers/KernelTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/KernelTraceEventParser.cs
@@ -6455,27 +6455,35 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.Kernel
 
     public sealed class FileIOCreateTraceData : TraceEvent
     {
+        // Pointer to fltmgr!_FLT_CALLBACK_DATA
         public Address IrpPtr { get { return GetAddressAt(0); } }
+        // Pointer to nt!_FILE_OBJECT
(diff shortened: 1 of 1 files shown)


NightRa:
The test failures seem unrelated to the PR

NightRa:
Pushed without changes to re-run tests. Now they pass.

brianrob:
@NightRa, thanks for contributing this.  I have looked for some documentation to confirm your change, but haven't been able to find any.  Do you have a pointer to documentation for this?

NightRa:
@brianrob Hi. It's from https://docs.microsoft.com/en-us/windows-hardware/drivers/ifs/irp-mj-create:

> IrpSp->Parameters.Create.Options Bitmask of flags that specify the options to be applied when creating or opening the file, as well as the action to be taken if the file already exists.
> 
> The high 8 bits of this parameter correspond to the Disposition parameter to IoCreateFileSpecifyDeviceObjectHint.

There seems to be no where documenting that the specific ETW event takes it's data from `_FLT_IO_PARAMETER_BLOCK.Create`, but this was derived from a long session of kernel debugging in order to understand why the `Disposition` field showed the wrong information, and finding the kernel code writing the ETW event.
After the fix it does indeed show the right information.

PR #1139, opened 2020-03-27
Enable perfcollect to take a list of keywords

ooooolivia:
- Add ```LTTng_Kernel_ContextSwitchKeyword``` 
- Enable ```-events``` to take a list of keywords declared in the file
  usage example:  
```./perfcollect start <tracename> -events LTTng_Kernel_ProcessLifetimeKeyword,DotNETRuntime_GCKeyword,<otherkeyword>```

### DIFF
--- a/src/perfcollect/perfcollect
+++ b/src/perfcollect/perfcollect
@@ -615,9 +615,13 @@ declare -a DotNETRuntimePrivate_DynamicTypeUsageKeyword=(
     DotNETRuntimePrivate:ObjectVariantMarshallingToManaged
 )
 
-declare -a LTTng_Kernel_ProcessLifetime=(
+declare -a LTTng_Kernel_ProcessLifetimeKeyword=(
     sched_process_exec
     sched_process_exit
+)
+
+declare -a LTTng_Kernel_ContextSwitchKeyword=(
+    sched_switch
 )
 
 ######################################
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/perfcollect/perfcollect [hunk 1]
```diff
@@ -1339,11 +1347,25 @@ SetupLTTngSession()
...
+            for (( i=0; i<${#events[@]}; i++ ))
+            do 
+                local event=${events[$i]}
+                if [ "${!event}" == "" ]
+                then
+                    echo Invalid keyword $event, skipped...
+                    continue
+                fi
+                local -a 'keywords=("${'"$event"'[@]}")'
+                if [ "${event#"LTTng_Kernel_"}" != "$event" ]
```
brianrob: Seems completely reasonable.  One thing that would be good here is to make sure that the array exists as well so that EnableLTTngEvents doesn't blow up, or worse, just silently fail.

src/perfcollect/perfcollect
```diff
@@ -615,11 +615,15 @@ declare -a DotNETRuntimePrivate_DynamicTypeUsageKeyword=(
...
 
-declare -a LTTng_Kernel_ProcessLifetime=(
+declare -a LTTng_Kernel_ProcessLifetimeKeyword=(
     sched_process_exec
     sched_process_exit
 )
 
+declare -a LTTng_Kernel_ContextSwitchKeyword=(
+    sched_switch
+)
```
brianrob: As we spoke about offline, there are a bunch of other events that you'll need to include here as well, but I've not had good luck getting a clear picture of thread time from these events yet.  May be best to not include this yet until we know the full set of events and how to interpret them.

src/perfcollect/perfcollect
ooooolivia: That's right, will check if the array is empty instead of checking if the array variable exists.

src/perfcollect/perfcollect
ooooolivia: Yes, thanks for pointing it out!

PR #1143, opened 2020-04-03
Add all payload fields to Linux Kernel Parser

ooooolivia:
Add more payload fields and their debug validations to Linux Kernel Parser


brianrob:
@ooooolivia, overall looks good.  I think this warrants a new test to validate that this code is able to parse the events properly.  Can you add a new test along with a trace that contains these kernel events and validates that they are correct?

ooooolivia:
@brianrob Sure! Thanks for the review.

PR #1147, opened 2020-04-07
Revert "Introduce compiled shared instances of commonly used regex"

brianrob:
This reverts commit f4b3490205e060ed1a4de4daf95f049962caecf9.

### DIFF
--- a/src/PerfView/EtwEventSource.cs
+++ b/src/PerfView/EtwEventSource.cs
@@ -646,8 +646,6 @@ namespace PerfView
 
             #region private
 
-            private static readonly Regex specialCharRemover = new Regex(" *[\r\n\t]+ *", RegexOptions.Compiled);
-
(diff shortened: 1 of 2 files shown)


brianrob:
@phduck, there is something about your change that the CI doesn't like, and so it's been failing since I merged your change.  Interestingly though, the CI run for #1144 succeeded, so I can't explain that.  I'm going to revert this for now, and then try and figure out what happened.

PhDuck:
Sure. 

I don't really see why it should fail tho'. Does the CI give any information about what it is failing on?

sharwell:
super strange...

brianrob:
It's very weird - it fails to nuget restore PerfView.sln and says it's malformed.  I would not have believed that your change would touch this, but it seems to have some impact.  I very much can't explain it.

PR #1149, opened 2020-04-08
expose ExtraGen0Commit field in perheap history in TraceGC

Maoni0:
I needed to use this field in the gc perf infra.


Maoni0:
@cshung @brianrob PTAL.

PR #1152, opened 2020-04-09
Allow RecursionGuard Constants to be Configured

brianrob:
There have been some cases recently of extremely deep stack traces that have resulted in `RecursionGuard` throwing a `StackOverflowException`  This change allows the maximum values that trigger a `StackOverflowException` to be configured, and also increases the maximum number of trampoline operations from 20 to 40.

### DIFF
--- a/src/TraceEvent/Stacks/RecursionGuard.cs
+++ b/src/TraceEvent/Stacks/RecursionGuard.cs
@@ -2,6 +2,32 @@
 
 namespace Microsoft.Diagnostics.Tracing.Stacks
 {
+    public static class RecursionGuardConfiguration
+    {
+        private static ushort _singleThreadRecursionLimit = 400;
+        private static ushort _maxResets = 40;
+
+        /// <summary>
+        /// For recursive methods that need to process deep stacks, this value defines the limit for recursion within
+        /// a single thread. After reaching this limit, methods need to trampoline to a new thread before continuing to
+        /// recurse.
+        /// </summary>
+        public static ushort SingleThreadRecursionLimit
+        {
+            get { return _singleThreadRecursionLimit; }
+            set { _singleThreadRecursionLimit = value; }
+        }
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


brianrob:
@xiaomi7732 can you confirm that this configuration API will work for you?

sharwell:
We should just take #785 instead, and stop paying a performance penalty for artificially-limited functionality.

src/TraceEvent/Stacks/RecursionGuard.cs
```diff
@@ -2,31 +2,45 @@
...
-    /// <summary>
-    /// This structure provides a clean API for a lightweight recursion stack guard to prevent StackOverflow exceptions
-    /// We do ultimately do a stack-overflow to prevent infinite recursion, but it is now under our
-    /// control and much larger than you may get on any one thread stack.  
-    /// </summary>
-    internal struct RecursionGuard
+    public static class RecursionGuardConfiguration
     {
+        private static ushort _singleThreadRecursionLimit = 400;
+        private static ushort _maxResets = 80;
```
sharwell, line 8: If we use configuration, only `_maxResets` needs to be configurable. This value is only used on the "slow" path, where `_singleThreadRecursionLimit` is used on the "fast" path.

xiaomi7732:
@brianrob I will take a look at it next Monday. Could you please clarify which one are you going to go with? This one or #785?

src/TraceEvent/Stacks/RecursionGuard.cs
brianrob: You're right that only `_maxResets` needs to be configured.  We've not hit an issue with the single thread recursion limit.  I'll make that change.

I spent a couple of days thinking about your proposal.  In the end, I still very much want a "backstop" to prevent infinite recursion, even in release builds where TraceEvent gets used in automated data pipelines.  I would much rather folks hit an exception and address it than end up with infinite recursion and have to detect it.  With that said, I appreciate your point, and thanks for the review @sharwell.

PR #1156, opened 2020-04-15
Add "Open in Browser" option for HTML reports in the TreeView

ryandle:
Scenario: I was opening some pretty large HTML reports in perfview and noticed that the default WebBrowser control was struggling to open large HTML files generated. (I had a particular /GCCollectOnly trace that had captured 180k GCs.)

To be fair, a modern browser also isn't perfect with an HTML file that big, but it is better.

So to address this I added a small feature to the right-click menu on the main TreeView that lets you open the report in your browser instead of the WPF control. Here's a demo:

OpenInBrowser

To do this I had to fix a second issue - when you right-click an item in the tree view it is not automatically selected before you choose an item in the right-click menu. This meant that you could not reliably use TreeView.SelectedItem in the callback of the command that was chosen in the right-click menu. I fixed this by hooking in to the right-button-down event and navigating the visual tree to find the TreeViewItem related to the visual that was clicked.

Here's the before (notice how CPU Stacks stays selected):

before

and after (TraceInfo is selected):

after

### DIFF
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -1116,6 +1116,10 @@ namespace PerfView
             }
         }
 
+        /// <summary>
+        /// Generates an HTML report and opens it using the machine's default handler .html file paths.
+        /// </summary>
+        /// <param name="worker">The StatusBar that should be updated with progress.</param>
         public void OpenInExternalBrowser(StatusBar worker)
         {
             TraceLog trace = GetTrace(worker);
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -894,7 +894,7 @@ namespace PerfView
             if (caller == StackSourceCallStackIndex.Invalid)
             {
                 string topCallStackStr = stackSource.GetFrameName(stackSource.GetFrameIndex(callStack), true);
-                
+
                 if (GetProcessForStackSourceFromTopCallStackFrame(topCallStackStr, out ret))
                 {
                     processes.Add(ret);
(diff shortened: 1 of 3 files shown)


src/PerfView/PerfViewData.cs
```diff
@@ -1185,8 +1116,110 @@ public override void Open(Window parentWindow, StatusBar worker, Action doAfter)
...
+                {
+                    Process.Start(reportFileName);
+                });
+            });
+        }
+
         public override void Close() { }
         public override ImageSource Icon { get { return GuiApp.MainWindow.Resources["HtmlReportBitmapImage"] as ImageSource; } }
+
+        private TraceLog GetTrace(StatusBar worker)
```
ryandle, line 1141: This method was extracted from above using VS refactor.

src/PerfView/PerfViewData.cs
```diff
@@ -1185,8 +1116,110 @@ public override void Open(Window parentWindow, StatusBar worker, Action doAfter)
...
+                    {
+                        trace = eventPipeDataFile.GetTraceLog(worker.LogWriter);
+                    }
+                }
+            }
+
+            return trace;
+        }
+
+        private string GenerateReportFile(StatusBar worker, TraceLog trace)
```
ryandle, line 1169: This method was extracted from above using VS refactor.

ryandle:
@brianrob I would guess you would be the person to review/consider this? No rush, but I wanted to mention you in case you don't get notifications for new PRs.

brianrob:
@ryandle, thanks for contributing this.  I like the idea - let me take a look at the code diff and get back to you.

brianrob:
@ryandle thanks for your patience.  This is really great, and is likely to be a huge win for big files like GCStats, as you mentioned.  Thanks for doing this!

cc: @Maoni0

Maoni0:
thanks @ryandle and hope to see many more contributions from you!

PR #1158, opened 2020-04-17
Improvements around MinSecForTrigger

### LINKED ISSUE #799: Feature request: Performance counter trigger - make trigger interval configurable
Having triggers from performance counters is really useful.  Sadly there is undocumented behavior which means triggers only activate if above the threshold for 3 consecutive seconds.

Sadly I'm often asked to investigate issues which are long over within 3 secs of any symptoms showing.

It would be really useful to override this parameter so that I can trigger after a single (or more) results over a threshold or test for triggers more (or less) frequently that 1000ms.

Thanks. Ping me if that needs clarifying further or would like to discuss.

dennisjlee:
Summary:
* Allows the MinSecForTrigger command-line flag to apply to
StartOnPerfCounter (previously only worked with StopOnPerfCounter)
* Improves logging and documentation around StopOnPerfCounter and
StartOnPerfCounter
* Improve log messages around perf counters to be clearer and make
sense for both "greater than" and "less than" scenarios.

Fixes #799

### DIFF
--- a/src/PerfView/CommandLineArgs.cs
+++ b/src/PerfView/CommandLineArgs.cs
@@ -97,7 +97,7 @@ namespace PerfView
         public int StopOnGCOverMsec;
         public int StopOnBGCFinalPauseOverMsec; // Stop on a BGC whose final pause is over this many ms
         public float DecayToZeroHours;          //causes 'StopOn*OverMSec' timeouts to decay to zero over this time period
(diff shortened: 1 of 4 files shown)


dennisjlee:
I'm currently double checking with my company to ensure there is no problem with me signing the CLA.

PR #1176, opened 2020-05-19
Add array type support for EventPipe to TraceEvent

davmason:
Companion PR to https://github.com/dotnet/runtime/pull/36242. This PR adds support for array type parsing to EventPipeEventSource.


davmason:
@noahfalk I apparently don't have the permissions to add you as reviewer

src/TraceEvent/TraceEvent.Tests/EventPipeTestBase.cs
```diff
@@ -32,54 +32,57 @@ public static IEnumerable<object[]> StreamableTestEventPipeFiles
...
         protected void PrepareTestData()
         {
-            Assert.True(Directory.Exists(TestDataDir));
-            TestDataDir = Path.GetFullPath(TestDataDir);
-            Assert.True(Directory.Exists(OriginalBaselineDir));
-            OriginalBaselineDir = Path.GetFullPath(OriginalBaselineDir);
-
-            // This is atomic because this method is synchronized.  
-            if (!s_fileUnzipped)
+            lock (s_fileLock)
```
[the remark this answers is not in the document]
davmason: I don't think I explained anywhere in the PR, but the issue was that this method had MethodImplOptions.Synchronized, which locks on the instance of the class. But xunit creates a new class for every test that runs so the lock effectively did nothing. Changing to locking on a static object fixed it.

src/TraceEvent/TraceEvent.Tests/EventPipeParsing.cs
```diff
@@ -276,6 +276,102 @@ public void CanReadV4EventPipeTraceBiggerThan4GB()
...
+                            char[] testEvent3Array = new char[] { (char)59299, (char)13231, (char)38541, (char)7407, (char)35812 };
+                            if (!Enumerable.SequenceEqual((char[])traceEvent.PayloadValue(0), testEvent3Array)) { throw new Exception(String.Format(tracePayloadValueValidationMessage, testEvent3Array, traceEvent.PayloadValue(0), "TestEventSource0", "TestEvent3", "arg0")); }
+
+                            ++successCount;
+                            return;
+                        }
+                    }
+                };
+
+                var privateClr = new ClrPrivateTraceEventParser(traceSource);
```
brianrob, line 365: Doesn't look like the private parser gets used at all.  Do you need it?

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -844,39 +968,59 @@ protected unsafe override void ReadBlockContents(PinnedStreamReader reader)
         }
     }
 
+    internal enum EventPipeMetaDataVersion
+    {
+        LegacyV1 = 1, // Used by NetPerf version 1
+        LegacyV2 = 2, // Used by NetPerf version 2
+        NetTraceV1 = 3, // Used by NetPerf version 3 and NetTrace version 1 onwards
+    internal enum EventPipeMetaDataVersion
```
brianrob, line 976: Looks like there is a merge error here as EventPipeMetadataVersion is nested inside another definition of EventPipeMetadataVersion.

src/TraceEvent/TraceEvent.Tests/EventPipeParsing.cs
davmason: No, I don't need it and removed it

src/TraceEvent/EventPipe/EventPipeEventSource.cs
davmason: There were a couple of noah's suggestions that caused build breaks, those should all be fixed now

davmason:
Theres one remaining failure that is only showing up in CI, I believe I know what the issue is and I'm working on it.

@noahfalk you probably want to at least glance again, the changes are big enough I don't consider them just updates

davmason:
Ok, nevermind. It looks like this is a pre-existing failing test

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -490,157 +557,40 @@ private DynamicTraceEventData.PayloadFetchClassInfo CheckForWellKnownEventFields
             return null;
         }
 
-        private DynamicTraceEventData.PayloadFetchClassInfo ParseFields(PinnedStreamReader reader, int numFields)
+        private DynamicTraceEventData.PayloadFetchClassInfo ParseFields(PinnedStreamReader reader, int numFields, EventPipeMetaDataVersion encodingVersion,
```
noahfalk, line 560: metadataVersion isn't needed (it is only used to pass to ParseType, and ParseType only uses it to pass back to ParseFields)

brianrob:
Overall, this looks good to me as well.  @noahfalk, @davmason are you both ready for merge?

davmason:
I'm done with my changes and ready to merge

PR #1182, opened 2020-05-27
update for loop in perfcollect

ooooolivia:
Current for loop expression will allow any possible inner loop to change ```$i``` randomly and result in early-exit or infinite loop. ( If there's no specific reason to use ```$i``` for loop,  recommending to get rid of this value completely to avoid similar bugs in the future )

### DIFF
--- a/src/perfcollect/perfcollect
+++ b/src/perfcollect/perfcollect
@@ -1371,9 +1371,8 @@ SetupLTTngSession()
         else     
             # Enable other keywords
             events=(${events//","/" "})
-            for (( i=0; i<${#events[@]}; i++ ))
+            for event in ${events[@]}
             do 
-                local event=${events[$i]}
                 if [ "${!event}" == "" ] || [ "$( echo `declare -p $event | grep -- -a` )" == "" ]
                 then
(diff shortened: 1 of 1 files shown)


brianrob:
@ooooolivia did you run into a specific issue here, or is this an improvement?

ooooolivia:
@brianrob  This causes specific issue. For command ```./perfcollect start <trace name> -events DotNETRuntimePrivate_StartupKeyword,LTTng_Kernel_ProcessLifetimeKeyword```, it won't enable the kernel keyword because the inner loop for enabling the startup keyword changes ```$i``` in ```for (( i=0; i<${#events[@]}; i++ ))``` and causes it to exit early.

ooooolivia:
@brianrob  This causes specific issue. For command ```./perfcollect start <trace name> -events DotNETRuntimePrivate_StartupKeyword,LTTng_Kernel_ProcessLifetimeKeyword```, it won't enable the kernel keyword because the inner loop for enabling the startup keyword changes ```$i``` in ```for (( i=0; i<${#events[@]}; i++ ))``` and causes it to exit early.

brianrob:
Got it.  Thanks!

PR #1184, opened 2020-06-01
Fix aggregation of samples in the SpeedScope/Chromium exporter

### LINKED ISSUE #1178: Export to speedscope sometimes has asymmetrical open/close frame event sequencing
Hi! I received a bug report to speedscope (https://github.com/jlfwong/speedscope/issues/272) which I'm fairly sure is caused by the event sequencing of the speedscope export of nettrace being incorrect. There's more in-depth discussion of the problem and relevant files in that issue.

If someone does plan on patching this issue, it would also be helpful to update this line:

https://github.com/microsoft/perfview/blob/9ca3b2e803d862935fd29789d550fe97a3bc9b4e/src/TraceEvent/Stacks/SpeedScopeStackSourceWriter.cs#L55

The intent of that field in the file format is to indicate what program exported the file so I can figure out if it's a bug in speedscope's export format or not. In this particular case, it was not, but if it weren't for @swift-kim being very thorough in their bug report, I might've gone down a rabbit hole trying to figure out where the bug was.

I'd prefer if that line was instead:

```
 writer.Write("\"exporter\": \"microsoft/perfview\", "); 
```

possibly with a version number.

swift-kim:
- Sort samples not only by time but also by depth to safely deal with recursive invocation of methods
- Prevent overlapping of adjacent samples caused by imprecise conversion of floating point numbers

Fixes #1178.

The result from the sample nettrace file in https://github.com/jlfwong/speedscope/issues/272:

| Format | Before | After |
|:-:|:-:|:-:|
| SpeedScope | image | image |
| Chromium | image | image |

Somehow I cannot run tests locally on my machine. Maybe the CI will do it?

@brianrob @adamsitnik Please take a look.

Thanks!

### DIFF
--- a/src/TraceEvent/Stacks/StackSourceWriterHelper.cs
+++ b/src/TraceEvent/Stacks/StackSourceWriterHelper.cs
@@ -44,7 +44,11 @@ namespace Microsoft.Diagnostics.Tracing.Stacks
                     if (!samplesPerThread.TryGetValue(threadInfo, out var samples))
                         samplesPerThread[threadInfo] = samples = new List<Sample>();
 
-                    samples.Add(new Sample(sample.StackIndex, -1, sample.TimeRelativeMSec, sample.Metric, -1));
+                    // The single-to-double floating point error may cause unexpected overlapping of samples.
+                    // At best we can add some minimum buffer between the samples to deal with it.
+                    var metric = sample.Metric - 1e-7;
+
+                    samples.Add(new Sample(sample.StackIndex, -1, sample.TimeRelativeMSec, metric, -1));
 
                     return;
                 }
@@ -143,7 +147,7 @@ namespace Microsoft.Diagnostics.Tracing.Stacks
                 var frameId = samplesInfo.Key;
                 var samples = samplesInfo.Value;
 
-                // this should not be required, but I prefer to be sure that the data is sorted
+                // make sure the samples are sorted by depth and then time
                 samples.Sort(CompareSamples);
 
                 Sample openSample = samples[0]; // samples are never empty
@@ -227,6 +231,11 @@ namespace Microsoft.Diagnostics.Tracing.Stacks
 
         private static int CompareSamples(Sample x, Sample y)
         {
+            int depthComparison = x.Depth.CompareTo(y.Depth);
+
+            if (depthComparison != 0)
+                return depthComparison;
+
             int timeComparison = x.RelativeTime.CompareTo(y.RelativeTime);
 
             if (timeComparison != 0)


adamsitnik:
Hi @swift-kim 

Thank you for a very elegant fix! The attached screenshots look great!

I am going to clone your fork right now and test few traces manually.

adamsitnik:
@swift-kim the latest speedscope version is giving me an error:

obraz

https://www.speedscope.app/#profileURL=https://gist.githubusercontent.com/adamsitnik/252468dff5c2b82cab913dce8f4b1d94/raw/fa46a2795965b727af7ca4eaa996b67e08980589/1184.speedscope.json

an older version of speedscope can open it:

obraz

I've analyzed the produced json file and it looks that it's my fault and the aggregation of samples is simply too aggressive or even invalid:

https://github.com/microsoft/perfview/blob/3aae482871d7c3c369e9733a31cdca975126c32a/src/TraceEvent/Stacks/StackSourceWriterHelper.cs#L186

If I remove the magic `1.2` multiplier:

```diff
- return left.RelativeTime + (left.Metric * 1.2) < right.RelativeTime;
+ return (left.RelativeTime + left.Metric) < right.RelativeTime
```

both speedscope and chromium open it.

The problem is that the size is growing (even up to +30%) and method names are not displayed (there is a lot of small boxes):

After:

obraz

Before:

obraz

I'll try to experiment with the aggregation and see if I can find a solution that would work fine for all scenarios.

adamsitnik:
Ok, I think I've found the source of the problem.

To repro, we need to set the time range of the provided trace file to `<1,243.000, 1,250.369>`

obraz

In the exported JSON file, frames from `0` to `14` are opened, and then `14`, `13` and **8** are closed. So we are not closing `12-9`

```json
"events": [
{ "type": "O", "frame": 0, "at": 1242.9994 },
{ "type": "O", "frame": 1, "at": 1242.9994 },
{ "type": "O", "frame": 2, "at": 1242.9994 },
{ "type": "O", "frame": 3, "at": 1242.9994 },
{ "type": "O", "frame": 4, "at": 1242.9994 },
{ "type": "O", "frame": 5, "at": 1242.9994 },
{ "type": "O", "frame": 6, "at": 1242.9994 },
{ "type": "O", "frame": 7, "at": 1242.9994 },
{ "type": "O", "frame": 8, "at": 1242.9994 },
{ "type": "O", "frame": 9, "at": 1242.9994 },
{ "type": "O", "frame": 10, "at": 1242.9994 },
{ "type": "O", "frame": 11, "at": 1242.9994 },
{ "type": "O", "frame": 12, "at": 1242.9994 },
{ "type": "O", "frame": 13, "at": 1242.9994 },
{ "type": "O", "frame": 14, "at": 1242.9994 },
// now we begin closing
{ "type": "C", "frame": 14, "at": 1248.8973839278624 },
{ "type": "C", "frame": 13, "at": 1248.8973839278624 },
// jump from 13 to 8, what about closing 12, 11, 10 and 9?
{ "type": "C", "frame": 8,  "at": 1248.8973839278624 }
```

This is because the following condition returns true for `System.Private.CoreLib!System.Reflection.RuntimeAssembly.GetCustomAttributes(class System.Type,bool)` 

https://github.com/microsoft/perfview/blob/3aae482871d7c3c369e9733a31cdca975126c32a/src/TraceEvent/Stacks/StackSourceWriterHelper.cs#L182-L183

Because it's called by two different methods:
 * `xamarin.forms.core!Xamarin.Forms.Internals.Registrar.RegisterStylesheets()`
*  `xamarin.forms.core!Xamarin.Forms.DependencyService.Initialize(class System.Reflection.Assembly[])`

but all other methods have the same caller and depthL

```log
A() => B() => X() => C() => D()
A() => B() => Y() => C() => D()
```

obraz

I'll try to come up with some way to recognize this situation (difference somewhere in the callstack)

swift-kim:
@adamsitnik Thank you for the detailed analysis!

Let's continue the discussion in the new PR.

PR #1185, opened 2020-06-01
 Fix aggregation of samples in the SpeedScope/Chromium exporter

adamsitnik:
I've cloned @swift-kim fork and started testing #1184 

I've found a few more things that needed to be changed (like using `CallerStackIndex` (not CallerFrameId)) which were old bugs of mine and decided to extend @swift-kim PR with some changes (see the commits)

 I was not sure whether I should send a PR to @swift-kim fork or to the upstream repo. I hope that @swift-kim doesn't mind that I am sending this as a new PR.

@brianrob I've tested this a LOT. PTAL

### DIFF
--- a/src/TraceEvent/Stacks/StackSourceWriterHelper.cs
+++ b/src/TraceEvent/Stacks/StackSourceWriterHelper.cs
@@ -44,7 +44,11 @@ namespace Microsoft.Diagnostics.Tracing.Stacks
                     if (!samplesPerThread.TryGetValue(threadInfo, out var samples))
                         samplesPerThread[threadInfo] = samples = new List<Sample>();
 
-                    samples.Add(new Sample(sample.StackIndex, -1, sample.TimeRelativeMSec, sample.Metric, -1));
+                    // The single-to-double floating point error may cause unexpected overlapping of samples.
+                    // At best we can add some minimum buffer between the samples to deal with it.
+                    var metric = sample.Metric - 1e-7;
+
+                    samples.Add(new Sample(sample.StackIndex, -1, sample.TimeRelativeMSec, metric, -1));
 
                     return;
                 }
@@ -143,7 +147,7 @@ namespace Microsoft.Diagnostics.Tracing.Stacks
                 var frameId = samplesInfo.Key;
                 var samples = samplesInfo.Value;
 
-                // this should not be required, but I prefer to be sure that the data is sorted
+                // make sure the samples are sorted by depth and then time
                 samples.Sort(CompareSamples);
 
                 Sample openSample = samples[0]; // samples are never empty
@@ -227,6 +231,11 @@ namespace Microsoft.Diagnostics.Tracing.Stacks
 
         private static int CompareSamples(Sample x, Sample y)
         {
+            int depthComparison = x.Depth.CompareTo(y.Depth);
+
+            if (depthComparison != 0)
+                return depthComparison;
+
             int timeComparison = x.RelativeTime.CompareTo(y.RelativeTime);
 
             if (timeComparison != 0)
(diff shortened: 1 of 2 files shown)


src/TraceEvent/Stacks/StackSourceWriterHelper.cs
```diff
@@ -44,7 +44,11 @@ internal static class StackSourceWriterHelper
                     if (!samplesPerThread.TryGetValue(threadInfo, out var samples))
                         samplesPerThread[threadInfo] = samples = new List<Sample>();
 
-                    samples.Add(new Sample(sample.StackIndex, -1, sample.TimeRelativeMSec, sample.Metric, -1));
+                    // The single-to-double floating point error may cause unexpected overlapping of samples.
+                    // At best we can add some minimum buffer between the samples to deal with it.
```
brianrob, line 48: Overall, this change looks good.  This is the one line that gives me pause, so I'd like to just get a bit more context on it.  Is the issue that it's possible that we have a case where a sample runs up right to the next sample, (e.g. sample starts at T=0 with metric = 5, and then next sample starts at T=5), or is it something else?  I'm wondering if there should be a more robust fix where we keep track of the last sample and it's metric so that we can reliably capture these cases, and only modify the results if we have to?  This would also capture cases where there might be larger inconsistencies in the data.  What do you think?

swift-kim: The real problem is with `StackSourceSample` which treats `TimeRelativeMSec` as double and `Metric` as float. The precision is already lost at

https://github.com/microsoft/perfview/blob/91a55f2e715fe1e094ad71304f413f3bf7e53ee8/src/TraceEvent/Computers/SampleProfilerThreadTimeComputer.cs#L533

when initializing the `Metric` value, and this later leads to inconsistent event sequencing as stated by https://github.com/microsoft/perfview/issues/1178. The first step to correctly manage this issue would be using a consistent data type (double) for all time values across the library. I thought it would be quite a heavy amount of work so I just made a little workaround.

If there's a better way to handle this problem or the comment can be updated with more context, I would be happy about it.

swift-kim:
[VoiceMemo.23092.zip](https://github.com/microsoft/perfview/files/4715923/VoiceMemo.23092.zip)

The above sample trace again resulted in a bad format error in SpeedScope.

```
Failed to load format Error: Tried to leave frame "ElmSharp!ElmSharp.Window.CreateHandle(class ElmSharp.EvasObject)" while frame "ElmSharp!ElmSharp.EvasObject.Realize(class ElmSharp.EvasObject)" was at the top at 279.1189620522523
```

It just turned out the value `1e-7` was not enough (increasing the value fixes the problem). As @brianrob pointed out however, it seems not a good idea to modify the metric value like this.

I'd rather suggest this change: https://github.com/swift-kim/perfview/commit/6655dc3f5ac5607c8f9e6b2ae15e17396559499a. In this way, we use double for other computation and convert to float when finally writing out ProfileEvents.

adamsitnik:
> I'd rather suggest this change

@swift-kim I've applied your suggestion and tested the provided `.nettrace` files and everything works fine. Thank you!

adamsitnik:
btw @brianrob that was a great question!

swift-kim:
@adamsitnik Thanks for all your effort!

src/TraceEvent/Stacks/StackSourceWriterHelper.cs
brianrob: You are correct that it would be a lot of work.  And in fact, I believe something along these lines has been attempted before and there is some subtle behavior that makes this difficult.

brianrob:
Nice job!  Thanks @swift-kim and @adamsitnik!

PR #1196, opened 2020-06-08
Tighten the sample aggregation threshold

swift-kim:
Improves the speedscope exporter for compatibility with speedscope 1.6.0 or above (https://github.com/jlfwong/speedscope/pull/273).

This is a workaround for https://github.com/microsoft/perfview/issues/1178#issuecomment-638664301. Since I don't know the reason for the original value (1.2), the new value is a simple compromise between 1.0 and 1.2.

@adamsitnik Could you test if you agree with this change?

### DIFF
--- a/src/TraceEvent/Stacks/StackSourceWriterHelper.cs
+++ b/src/TraceEvent/Stacks/StackSourceWriterHelper.cs
@@ -184,8 +184,7 @@ namespace Microsoft.Diagnostics.Tracing.Stacks
             if (left.CallerStackIndex != right.CallerStackIndex)
                 return true;
 
-            // 1.2 is a magic number based on some experiments ;)
-            return left.RelativeTime + (left.Metric * 1.2) < right.RelativeTime;
+            return left.RelativeTime + (left.Metric * 1.1) < right.RelativeTime;
         }
 
         /// <summary>


adamsitnik:
Hi @swift-kim 

Please excuse me for the delay.

I've started testing your fix today and I wanted to extend it with some validation logic that ensures that the output is correct (every close event is preceded by an open event with the same depth and frame id).

Code: https://github.com/adamsitnik/perfview/commit/f5415222a6400ccbb434579eab7549b96bdf61b1

Unfortunately, I've found some cases for very big trace files where even with aggregation threshold = 1.0 the exporter is producing invalid results. I need to wrap my head around it and figure out a proper solution. It might require rewriting the exporter from scratch.

swift-kim:
@adamsitnik Yes, I've also found such cases ([Weather.6625.zip](https://github.com/swift-kim/warehouse/blob/master/data/Weather.6625.zip)/[VoiceMemo.13755.zip](https://github.com/swift-kim/warehouse/blob/master/data/VoiceMemo.13755.zip) - I haven't yet investigated the files though) and been thinking about a proper way to validate outputs.

If you manage to do so, it will preclude the need for [this request](https://github.com/jlfwong/speedscope/issues/285) to the speedscope side. I really appreciate all your help!

PR #1203, opened 2020-06-12
Add processor stacks

vascoveigams:
Adds "Processor Stacks" advanced view if CPU stacks are available.
"Processor (x) Priority (y)" is added as the top frame to every stack, which allows per processor / priority analysis.
For convenience two grouping regexes are added to the view "Processor ({%}) Priority ({%})->Processor ($1)" and "Processor ({%}) Priority ({%})->Processor ($2)" that allow grouping by Processor and Priority respectively.
@Maoni0 , let me know if you have other thoughts around the view.

### DIFF
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -1,15 +1,3 @@
-using System;
-using System.Collections.Generic;
-using System.ComponentModel;
-using System.Diagnostics;
-using System.IO;
-using System.IO.Compression;
-using System.Linq;
-using System.Text.RegularExpressions;
-using System.Threading;
-using System.Windows;
-using System.Windows.Media;
-using System.Xml;
 using Diagnostics.Tracing.StackSources;
 using global::DiagnosticsHub.Packaging.Interop;
 using Graphs;
@@ -34,6 +22,19 @@ using Microsoft.DiagnosticsHub.Packaging.InteropEx;
 using PerfView.GuiUtilities;
 using PerfViewExtensibility;
 using PerfViewModel;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Windows;
+using System.Windows.Media;
+using System.Xml;
 using Utilities;
 using Address = System.UInt64;
 using EventSource = EventSources.EventSource;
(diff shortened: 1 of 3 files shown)


src/PerfView/PerfViewData.cs
```diff
@@ -5098,6 +5097,33 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                 eventSource.Kernel.AddCallbackForEvents<ObjectDuplicateHandleTraceData>(data => onHandleEvent(data.ObjectTypeName, data.Object, data.TargetHandle, data.TargetProcessID, data));
                 eventSource.Process();
             }
+            else if (streamName.StartsWith("Processor"))
+            {
+                eventSource.AllEvents += delegate (TraceEvent data)
```
brianrob, line 5102: Rather than hooking this up to `eventSource.AllEvents`, you should hook this up to `eventSource.Kernel.PerfInfoSample`.  Then, you can change the argument type on the delegate from `TraceEvent` to `SampledProfileTraceData` and you no longer have to process all events or do the cast.

src/PerfView/SupportFiles/UsersGuide.htm
```diff
@@ -2295,6 +2295,10 @@ <h2>
                     views that are more specialize investigations that are rarer than the common CPU,
                     wall clock time, and memory investigations.
                     <ul>
+                        <li>
+                            <strong><a id="ProcessorStacks">Processor Stacks</a></strong> - Shows what every
+                            processor is doing, including at what priority. Use the
```
brianrob, line 2300: Looks like you started this, but didn't finish :)

Maoni0:
this looks great to me; thanks so much for doing this!

src/PerfView/SupportFiles/UsersGuide.htm
vascoveigams: Yeah :) Added that GroupPats can be used to focus on processor or priority.

src/PerfView/PerfViewData.cs
vascoveigams: Is SampledProfileTraceData guaranteed to have a valid stack (data.CallStackIndex() != CallStackIndex.Invalid) ? If it does, then I can also remove the check for invalid stack.
I'll leave the check to be defensive for now if that's not true.

brianrob: It's good to leave it there - it is possible to be missing a stack.

src/PerfView/SupportFiles/UsersGuide.htm
brianrob: Awesome.  Thanks!

PR #1206, opened 2020-06-13
Fix EventPipeEventSource's event name generation logic for Start/Stop activities 

sywhang:
Originally reported in https://github.com/dotnet/diagnostics/issues/1189

In EventPipeEventSource, when we see an Event with Start/Stop suffix in their names, we associate that with the Start/Stop Opcode names, and append the "Start/Stop" word at the end when we display it on PerfView. (For example, RequestStart/Start).  

This causes the event names to not match what we see on ETW traces, which just says `Request/Start` instead of `RequestStart/Start`.

This PR fixes the Event name generation logic to match the ETW behavior. 

cc @brianrob @noahfalk @davidsh


src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -479,25 +479,54 @@ private void SetOpcode(DynamicTraceEventData template, int opcode)
         private DynamicTraceEventData CreateTemplate(EventPipeEventMetaDataHeader eventMetaDataHeader)
         {
             string opcodeName = ((TraceEventOpcode)eventMetaDataHeader.Opcode).ToString();
-
             int opcode = eventMetaDataHeader.Opcode;
             if (opcode == 0)
```
noahfalk, line 483: Seeing code snippets that set eventMetadataHeader.EventName and compute the substrings multiple times suggest this could probably be refactored to be less duplicative. For example:
````C#
            if (opcode == 0)
            {
                GetOpcodeFromEventName(eventMetaDataHeader.EventName, out opcode, out opcodeName);
            }
            FilterOpcodeNameFromEventName(eventMetaDataHeader, opcode);
````

File: src/TraceEvent/EventPipe/EventPipeEventSource.cs [hunk 2]
```diff
@@ -479,25 +479,54 @@ private void SetOpcode(DynamicTraceEventData template, int opcode)
...
+                {
+                    eventMetaDataHeader.EventName = eventNameWithoutOpcode;
+                }
+            }
+            else if (opcode == (int)TraceEventOpcode.Start || opcode == (int)TraceEventOpcode.Stop)
+            {
+                FilterOpcodeNameFromEventName(eventMetaDataHeader, opcode);
             }
 
             DynamicTraceEventData template = new DynamicTraceEventData(null, eventMetaDataHeader.EventId, 0, eventMetaDataHeader.EventName, Guid.Empty, opcode, null, eventMetaDataHeader.ProviderId, eventMetaDataHeader.ProviderName);
```
noahfalk, line 497: Is changing the name passed as parameter in this constructor sufficient to change the name shown in the trace? (I am guessing it is)
When possible it is usually simpler to reason about datastructures/variables that are only written once (a functional coding style) rather than mutating values in-place. It appears we are only assigning eventMetaDataHeader.EventName a 2nd time so that it can be read back at this line. Instead we could define a new variable to hold that value and leave the header immutable:
````C#
string eventName = FilterOpcodeNameFromEventName(eventMetaDataHeader.EventName, opcode);
DynamicTraceEventData template = new DynamicTraceEventData(null, eventMetaDataHeader.EventId, 0, eventName, Guid.Empty, opcode, null, eventMetaDataHeader.ProviderId, eventMetaDataHeader.ProviderName);
````

This prevents the updated name from leaking into CheckForWellKnownEventFields and eliminates the need to comment about the mutating field at 1188.

brianrob: Agreed with @noahfalk here.  If I understand what's happening here, in some cases, we're adding the opcode onto the end of the event name so that this gets seen later on and stripped off.  Did I get that right?  If so, it would be nice to see if there's a way to not do this.  We definitely need to discover cases where opcode wasn't specified because the format didn't specify it, but otherwise, if the opcode is specified, we should honor it and override the convention.

PR #1219, opened 2020-06-26
Expose CpuNumber in Stack Sample data

ivberg:
+Allow perfview to also open perf.data.txt directly

### DIFF
--- a/src/LinuxEvent.Tests/EventParseTests.cs
+++ b/src/LinuxEvent.Tests/EventParseTests.cs
@@ -17,7 +17,7 @@ namespace LinuxTracing.Tests
 
             for (int i = 0; i < stackSource.SampleIndexLimit; i++)
             {
-                var sample = stackSource.GetLinuxPerfScriptSampleByIndex((StackSourceSampleIndex)i);
+                var sample = stackSource.GetSampleByIndex((StackSourceSampleIndex)i);
 
                 var stackIndex = sample.StackIndex;
                 for (int j = 0; (int)stackIndex != -1; j++)
@@ -28,6 +28,10 @@ namespace LinuxTracing.Tests
                 }
 
                 Assert.Equal(-1, (int)stackIndex);
+
+                var linuxsample = stackSource.GetLinuxPerfScriptSampleByIndex((StackSourceSampleIndex)i);
+                Assert.NotNull(linuxsample);
+                Assert.NotEqual(StackSourceSampleIndex.Invalid, linuxsample.SampleIndex);
             }
         }
 
--- a/src/LinuxEvent.Tests/EventParseTests.cs
+++ b/src/LinuxEvent.Tests/EventParseTests.cs
@@ -17,7 +17,7 @@ namespace LinuxTracing.Tests
 
             for (int i = 0; i < stackSource.SampleIndexLimit; i++)
             {
-                var sample = stackSource.GetSampleByIndex((StackSourceSampleIndex)i);
+                var sample = stackSource.GetLinuxPerfScriptSampleByIndex((StackSourceSampleIndex)i);
 
                 var stackIndex = sample.StackIndex;
                 for (int j = 0; (int)stackIndex != -1; j++)
--- a/src/TraceEvent/Stacks/Linux/LinuxPerfScriptStackSource.cs
+++ b/src/TraceEvent/Stacks/Linux/LinuxPerfScriptStackSource.cs
@@ -302,6 +302,7 @@ namespace Microsoft.Diagnostics.Tracing.StackSources
             var sample = new StackSourceSample(this);
             sample.TimeRelativeMSec = linuxEvent.TimeMSec - StartTimeStampMSec;
             sample.Metric = (float)linuxEvent.Period;
+            sample.Count = linuxEvent.CpuNumber;
 
             stackIndex = InternFrames(frames.GetEnumerator(), stackIndex, linuxEvent.ProcessID, linuxEvent.ThreadID, doThreadTime ? blockedTimeAnalyzer : null);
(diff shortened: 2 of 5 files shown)


src/TraceEvent/Stacks/Stacks.cs
```diff
@@ -409,13 +409,15 @@ public class StackSourceSample
         /// </summary>
         public int Scenario { get; set; }
 
+        public int CpuNumber { get; set; }
```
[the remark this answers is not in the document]
ivberg: We are using this to expose the CPU # where a sample was taken in WPA. Need for graphing. In mind a sample stack would also have a CPU #, would it not - we just haven't exposed it in the past?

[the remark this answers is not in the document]
ivberg: Change from approach of modifying base StackSourceSample to using derived class LinuxPerfScriptStackSourceSample adding CpuNumber. This is because per brianrob not every StackSourceSample has a CPU #

src/TraceEvent/Stacks/Linux/LinuxPerfScriptStackSource.cs
```diff
@@ -291,17 +292,27 @@ public LinuxPerfScriptStackSource(string path, bool doThreadTime)
 
         public double TotalBlockedTime { get; set; }
 
+        #region private
+        internal /*protected*/ GrowableArray<LinuxPerfScriptStackSourceSample> m_LinuxPerfScriptSamples;
+        #endregion
+
+        public LinuxPerfScriptStackSourceSample GetLinuxPerfScriptSampleByIndex(StackSourceSampleIndex sampleIndex)
```
brianrob, line 299: Is this to avoid the cast from `StackSourceSample` to `LinuxPerfScriptStackSourceSample`?

src/LinuxEvent.Tests/EventParseTests.cs
```diff
@@ -17,7 +17,7 @@ private void DoStackTraceTest(string path, bool doBlockedTime, List<List<string>
 
             for (int i = 0; i < stackSource.SampleIndexLimit; i++)
             {
-                var sample = stackSource.GetSampleByIndex((StackSourceSampleIndex)i);
+                var sample = stackSource.GetLinuxPerfScriptSampleByIndex((StackSourceSampleIndex)i);
```
brianrob, line 20: If we are going to add a new API here, let's modify the test to run twice and test both of these APIs.  It's important that both work.

ivberg: Sure, makes sense

src/TraceEvent/Stacks/Linux/LinuxPerfScriptStackSource.cs
ivberg: yes, because the caller would just have to magically know that you could do that which isn't pretty. It's not super elegant API surface because the base .GetSampleByIndex() still shows up, but it seems it's required for this approach

brianrob: Ok, sounds good.

src/TraceEvent/Stacks/Linux/LinuxPerfScriptStackSource.cs
```diff
@@ -291,17 +291,27 @@ public LinuxPerfScriptStackSource(string path, bool doThreadTime)
...
+        {
+            return m_LinuxPerfScriptSamples[(int)sampleIndex];
```
brianrob, line 300: One other follow-up: I'd like to avoid this extra GrowableArray `m_LinuxPerfScriptSamples`.  What about implementing this API in terms of `GetSampleByIndex` and just cast the output?  Any concerns with this?

```suggestion
            return (LinuxPerfScriptStackSourceSample)GetSampleByIndex(sampleIndex);
```

src/TraceEvent/Stacks/Linux/LinuxPerfScriptStackSource.cs
```diff
@@ -333,6 +343,12 @@ public void AddSamples(IEnumerable<StackSourceSample> _samples)
             SampleEndTime = samples.Last().TimeRelativeMSec;
         }
 
+        public void AddSample(LinuxPerfScriptStackSourceSample sample)
```
brianrob, line 346: You'd be able to drop this method with the suggestion above.

src/TraceEvent/Stacks/Linux/LinuxPerfScriptStackSource.cs
ivberg: yes - I thought about this and tried this. It doesn't seem possible. You can downcast but not upcast. 

public StackSourceSample AddSample(StackSourceSample sample) is used to create a "real" sample. It was confusing to me as well how the existing code behaves

You cannot upcast StackSourceSample to LinuxPerfScriptStackSourceSample, and obviously StackSourceSample doesn't hold CpuNumber. So you need to add it and keep track of LinuxPerfScriptStackSourceSample somewhere because StackSourceSample is incapable of doing it

src/TraceEvent/Stacks/Linux/LinuxPerfScriptStackSource.cs
ivberg: see above

src/TraceEvent/Stacks/Linux/LinuxPerfScriptStackSource.cs
brianrob: You're right.  It's because `CopyStackSource.AddSample` makes a copy of the sample.

PR #1226, opened 2020-07-01
Fix the condemned reason table construction logic

cshung:
There are two bugs in the condemned reason table construction logic:

- To determine if we should render a column. We wanted to fill `columnHasContent` array. When my original code scans a row, it stops whenever it reaches a column with data in it. We should not stop there, the remaining column could also have data. The break was intended to stop trying other rows, not other columns, in order to do that, I need to swap the loop order.

- `FillCondemnedReason` is called twice, once for `PerHeapHistory` and again for `GlobalHeapHistory`. The intent was to capture both, but my implementation overwrote the data instead of accumulating them.


cshung:
Hi @brianrob, this is ready to be merged.

Issue #1222, opened 2020-06-30
SymbolReader's Age field rendering doesn't follow SSQP Key Conventions


danieljennings:
When building a path, SymbolReader renders the Age '10' as decimal '10' instead of 'a':

https://github.com/microsoft/perfview/blob/b4f39fcf5188d336415f8c6a643d6b7423010aa1/src/TraceEvent/Symbols/SymbolReader.cs#L229

From the [SSQP Key Conventions](https://github.com/dotnet/symstore/blob/master/docs/specs/SSQP_Key_Conventions.md):

> PDB-Signature-Age
> This applies to Microsoft C++ Symbol Format, commonly called PDB and using files with the .pdb file extension. The key is computed by extracting the Signature (guid) and Age (4 byte integer) values from the guid stream within MSF container. The final key is formatted:

and:

> Multi-byte integers: Convert to characters by first converting it to a big-endian byte sequence next convert the sequence as above and finally trim all leading '0' characters. Example 3,559,453,162 => 'd428f1ea', 114 => '72'

The behavior of WinDbg matches the spec, and disagrees with perfview:

> DBGHELP: C:\steam3\bld5946300\win64release\gc.exe - OK
> DBGENG:  C:\steam3\bld5946300\win64release\gc.exe - Mapped image memory
> Loading symbols for 00000001`40000000           gc.exe ->   SYMSRV:  BYINDEX: 0x1
>          c:\symbolcache*http://symbols/windows
>          gc_exe.pdb
>          A6DF2F6303FF414095B6ED112377FDD5a
> SYMSRV:  PATH: c:\symbolcache\gc_exe.pdb\A6DF2F6303FF414095B6ED112377FDD5a\gc_exe.pdb
> SYMSRV:  RESULT: 0x00000000
> gc.exe
> DBGHELP: gc - private symbols & lines 
>         c:\symbolcache\gc_exe.pdb\A6DF2F6303FF414095B6ED112377FDD5a\gc_exe.pdb

Where perfview would search instead for the nonexistent \A6DF2F6303FF414095B6ED112377FDD510\gc_exe.pdb

I imagine this is an easy fix to make, and I'm willing to write a PR but I would have to check with counsel about our current Microsoft CLA status. Let me know if that would be preferred and I can see if that's feasible from my end.

PR #1229, opened 2020-07-07
Fix for SymbolReader's Age field rendering not following SSQP Key Conventions

danieljennings:
Fix for issue I reported earlier: https://github.com/microsoft/perfview/issues/1222

This fixes lookup for PDBs with age greater than 9.

Ran the tests, all groups passed. Not sure if there's anything else to include for this bugfix since the details are in the issue thread.

Thanks for your consideration.


brianrob:
@noahfalk, can you please take a look at this one?

noahfalk (approved):
LGTM, thanks!

PR #1232, opened 2020-07-20
Runtime Loader advanced event view

davidwrighton:
Add support for viewing various when various runtime load operations occur during process execution. This is a view much like jit stats, but somewhat more flexible and less specialized to just the JIT.

Events visible are
- JIT
- R2R
- Assembly load
- Type load (When the runtime accepts pr dotnet/runtime#39656)

The view is presented linearly per thread, and can be shown nested or not (nested allows understanding of which operations occur as a result of other operations)

In addition, which events are shown is configurable via checkbox.


davidwrighton:
@billwer @brianrob Could one of you review this?

brianrob:
Yes, I will take a look @davidwrighton.

src/PerfViewCollect/PerfViewCollect.csproj
```diff
@@ -40,33 +40,26 @@
     <Compile Include="..\PerfView\ClrStats.cs" />
     <Compile Include="..\PerfView\GcStats.cs" />
     <Compile Include="..\PerfView\JitStats.cs" />
+    <Compile Include="..\PerfView\RuntimeLoaderStats.cs" />
```
brianrob, line 43: Are you wanting to enable this via the PerfViewCollect tool?

src/PerfViewCollect/PerfViewCollect.csproj
```diff
@@ -40,33 +40,26 @@
...
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles" Version="$(MicrosoftDiagnosticsTracingTraceEventSupportFilesVersion)" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All"/>
```
brianrob, line 47: It looks like there is a bunch of whitespace change here.  Can you revert these please?  Let me know if I missed an actual change here.

src/PerfView/SupportFiles/HtmlReportUsersGuide.htm
```diff
@@ -227,6 +227,29 @@ <h4>Understanding Tiered Compilation events</h4>
...
+        assembly load operations. This typically makes the data difficult to understand,
+        and less useful to most consumers, but for detailed investigation of runtime
+        behavior may be more useful. When a loader operation in this view occurs during
+        another operation, the nesting of the operation is represented in the view, and
+        the outer operation is broken up to show the time spent around the inner operation.
+
+        To enable data about all loader operations set the ".NET Loader" checkbox when collecting data.
+        Otherwise, the data captured will be restricted to JIT and assembly load operations. The /RuntimeLoading switch may also be used.
+
+        Also see the Runtime Loader stacks view, which ties this data into a view of CPU activity, to allow investigating CPU usage during specific loader operations.
```
brianrob, line 251: Thanks for adding help for this.  Can you also add a note that this only works on .NET Core (assuming we're not porting this to desktop)?

src/PerfView/Dialogs/RunCommandDialog.xaml
```diff
@@ -320,7 +320,11 @@
                     </TextBlock>                    
                     <CheckBox Grid.Row="0" Grid.Column="11" Name="CCWRefCountCheckBox" VerticalAlignment="Center"/>
 
-                    <CheckBox Grid.Row="0" Grid.Column="13" Name="Unused2" Visibility="Hidden" VerticalAlignment="Center"/>
+                    <TextBlock Grid.Row="0" Grid.Column="12" VerticalAlignment="Center" HorizontalAlignment="Right" Margin="8,2,5,2" 
+                               ToolTip="Log information about .NET Runtime loading." >
```
brianrob, line 324: It might be good to disambiguate this a bit in the tooltip to be clear that this is for more than just assembly loading, but also type loading, method loading, etc.

src/PerfView/RuntimeLoaderStats.cs
```diff
@@ -0,0 +1,219 @@
...
+            }
+        }
+
+        public static double TotalCPUMSec(Microsoft.Diagnostics.Tracing.Analysis.TraceProcess incompleteStatsProc, Microsoft.Diagnostics.Tracing.RuntimeLoaderStats runtimeOps)
+        {
+            TraceProcess process = null;
+
+            foreach (var proc in runtimeOps.EventSource.TraceLog.Processes)
+            {
+                if (proc.ProcessID == incompleteStatsProc.ProcessID)
```
brianrob, line 189: There are a couple of ways that you can match processes, while handing the fact that process IDs can be re-used.  One is to use TraceProcess.ProcessIndex, which is a unique value across the entire trace.  The second is to use ProcessID and timestamp, and find the closest match - TraceLog.Processes has an API for this.  If you leave this as is, it's possible that PID re-use will impact the results here.

src/PerfView/RuntimeLoaderStats.cs
```diff
@@ -0,0 +1,219 @@
...
+{
+    internal static class RuntimeLoaderStats
+    {
+        public static void ToHtml(TextWriter writer, Microsoft.Diagnostics.Tracing.Analysis.TraceProcess incompleteStatsProc, string fileName, Microsoft.Diagnostics.Tracing.RuntimeLoaderStats runtimeOps)
+        {
+            TraceProcess stats = null;
+
+            foreach (var proc in runtimeOps.EventSource.TraceLog.Processes)
+            {
+                if (proc.ProcessID == incompleteStatsProc.ProcessID)
```
brianrob, line 26: Same PID re-use issue here.

src/PerfView/PerfViewData.cs
```diff
@@ -7287,7 +7399,7 @@ protected override Action<Action> OpenImpl(Window parentWindow, StatusBar worker
             }
 
             advanced.Children.Add(new PerfViewJitStats(this));
-
+            advanced.Children.Add(new PerfViewRuntimeLoaderStats(this));
```
brianrob, line 7402: Can you condition this on having enough data to show something?

src/PerfView/RuntimeLoaderStats.cs
brianrob: In RuntimeLoaderStats.cs, you are using Microsoft.Diagnostics.Tracing.Etlx.TraceProcess, but here you're using Microsoft.Diagnostics.Tracing.Analysis.TraceProcess.  They are slightly different.  The first is layered on top of TraceLog, while this one is a level lower at TraceEventSource.  Since you already require Etlx.TraceProcess, I'd recommend just using that everywhere, and then you can use ProcessIndex rather than PID.

---
In reply to: [458422874](https://github.com/microsoft/perfview/pull/1232#discussion_r458422874) [](ancestors = 458422874)

src/TraceEvent/Computers/CLRRuntimeActivityComputer.cs
```diff
@@ -0,0 +1,343 @@
...
+            source.Clr.MethodJittingStarted += Clr_MethodJittingStarted;
+            source.Clr.MethodR2RGetEntryPoint += Clr_MethodR2RGetEntryPoint;
+            source.Clr.MethodLoadVerbose += Clr_MethodLoadVerbose;
+            source.Clr.MethodLoad += Clr_MethodLoad;
+            source.Clr.LoaderAssemblyLoad += Clr_LoaderAssemblyLoad;
+            source.Clr.MethodR2RGetEntryPointStart += Clr_R2RGetEntryPointStart;
+            source.Clr.LoaderTypeLoadStart += Clr_LoaderTypeLoadStart;
+            source.Clr.LoaderTypeLoadStop += Clr_LoaderTypeLoadStop;
+            source.Process();
+            source.Clr.MethodJittingStarted -= Clr_MethodJittingStarted;
```
brianrob, line 192: You should not need to unsubscribe these delegates.  The pattern used is to provide a fresh EventSource so that you are sure that other machinery isn't subscribed if you are concerned about this.  I don't think you'll find many, if any, places in where we attempt to unsubscribe in order to re-use the EventSource.

src/PerfView/SupportFiles/HtmlReportUsersGuide.htm
davidwrighton: It actually works to some extent on desktop. You can see JIT and assembly load events there.

src/PerfViewCollect/PerfViewCollect.csproj
davidwrighton: Eventually... yes. But for now, I'm settling for it just building cleanly. This change was required for that. The rest of these changes, are VS doing something I didn't want.

davidwrighton:
@brianrob I've addressed all of the feedback you've given me. As it turns out it was more difficult to consolidate on using the Etlx TraceProcess, so instead I consolidated on working with the Analysis TraceProcess, and some modestly dodgy logic involving pids and timestamps.

PR #1239, opened 2020-08-05
Ignore errors instantiating symbol checksum algorithm.

beteitle:
Ignore errors instantiating source checksum method during symbol source lookup.  For many use cases, source checksum is a "nice to have", but not a requirement.  Cryptography method instantiation may fail in certain cases.  For example, if FipsAlgorithmPolicy is in place on your host machine, you are not allowed to instantiate MD5.  Failing to checksum should not prevent source lookup workflow in my opinion (seems to work fine without it being performed for well behaving source server).

### DIFF
--- a/src/TraceEvent/Symbols/NativeSymbolModule.cs
+++ b/src/TraceEvent/Symbols/NativeSymbolModule.cs
@@ -626,15 +626,15 @@ namespace Microsoft.Diagnostics.Symbols
                 // 3 checksum generated with the SHA256 hashing algorithm.
                 if (sourceFile.checksumType == 1)
                 {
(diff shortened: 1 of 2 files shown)


brianrob:
@beteitle, thanks for your contribution.  In general, I want to keep these kinds of safeguards by default.  Not having them can result in invalid/incorrect data in some cases.  If you're running into cases where where the algorithms can't be executed, one option would be to have a flag, similiar to `/UnsafePDBMatch` where you can opt-out of these safety checks.  I would be supportive of this.

brianrob:

brianrob:

PR #1261, opened 2020-09-09
Fix msdia140 Dependencies in TraceEvent NuGet Package

brianrob:
Include msvcp140.dll and vcruntime140.dll in TraceEvent NuGet package.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -22,8 +22,8 @@
 
   <PropertyGroup>
     <!-- These are the versions of the things we are CREATING in this repository -->
-    <PerfViewVersion>2.0.59</PerfViewVersion>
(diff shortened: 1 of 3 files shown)


adamsitnik:
@brianrob I've recently used PerfView on Windows ARM64 (Surface Pro X) and it worked fine 👍 (it was using the `x86` emulation mode). I was wondering how this would work when there would be an `arm64` process that consumes `TraceEvent`.  Do arm-specific versions of the native libraries exist and we should add them as well? I would be very happy to help with testing.

brianrob:
@adamsitnik, I believe that there is an arm64 version of msdia140, but we don't package it today.  This is definitely something that we could do.  Do you want to file an issue for this?  We'd probably want to see about also including KernelTraceControl.dll so that collection works as well.

PR #1270, opened 2020-09-16
Don't Root TraceEventDispatcher in TraceProcessExtensions and TraceLoadedDotNetRuntimeExtensions

brianrob:
`TraceProcessExtensions` and `TraceLoadedDotNetRuntimeExtensions` both root the dispatcher to ensure non-concurrent access.  But once usage is complete, they hold the dispatcher until the process exits or the dispatcher is replaced.  Replace the strong reference with a weak reference so that memory associated with the dispatcher can be freed.

cc: @ivdiazsa


Maoni0 (approved):
LGTM!

brianrob:
Thanks @maoni0!

PR #1276, opened 2020-09-18
Add Process Metadata Events

### LINKED ISSUE #1265: Add abstract Process Start/Stop, Module Load/Unload and DbgId events
Today `TraceLog` depends on `ProcessStart`, `ProcessStop` (and their DCStart/DCStop equivalents), `ImageLoad`, `Image Unload` (and their DCStart/Stop equivalents) and `ImageID` events to be able to do useful event viewing and analysis of data.

As we add the ability to add multiple processes inside NetTrace files, it becomes important to represent these events. After discussion with @noahfalk and @brianrob it seems like having them not depend on the Windows Kernel events per-se seems useful and doesn't encumber the NetTrace format with these OS-specific concepts.

So, effectively these events will inform equivalent data but using different events. In fact, they will be like any other EventSource events, except their parsers will be part of the TraceEvent package and we will hook them up by default in the parser hook up for NetTrace and are "enlightened" to be acted up on when serializing TraceProcess/TraceModule, etc.

mjsabby:
Fixes #1265


mjsabby:
@brianrob Could you take a look at this one?

src/TraceEvent/TraceLog.cs
```diff
@@ -1786,6 +1788,95 @@ private unsafe void SetupCallbacks(TraceEventDispatcher rawEvents)
...
+            };
+
+            processMetadataParser.ProcessStop += delegate (ProcessExitArgsTraceData data)
+            {
+                Processes.GetOrCreateProcess((int)data.ProcessId, data.TimeStampQPC).ProcessEnd(data, data.CommandLine, data.Executable, (int)data.ParentProcessId, data.ExitCode);
+            };
+
+            processMetadataParser.ThreadStart += delegate (ThreadCreateArgsTraceData data)
+            {
+                TraceProcess process = processes.GetOrCreateProcess((int)data.ProcessId, data.TimeStampQPC);
```
brianrob, line 1806: Can we work these more complicated Threading and module load events into helper functions and then just call them from here and from the ETW equivalents?  There's enough here that it would be nice to only have one version of these.

src/TraceEvent/TraceLog.cs
```diff
@@ -30,6 +30,7 @@
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Diagnostics.Tracing.Parsers.ProcessMetadataEventSource;
```
brianrob, line 33: I know that in one of your other PRs, you had done some work to expose process name inside of the `EventPipe` source.  Do you think that would be worth plumbing in here as well?

brianrob (commented):
Thanks @mjsabby.  Some comments and questions.

brianrob:
@noahfalk  can you take a look at this as well?  Would like to get your thoughts, especially on the moduleload abstractions.

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        [Event(1, Opcode = EventOpcode.Start, Task = Tasks.Process)]
+        public void ProcessStart(long ProcessId, long ParentProcessId, string Executable, string CommandLine)
+        {
+            this.WriteEvent(1, ProcessId, ParentProcessId, Executable, CommandLine);
+        }
+
+        [Event(2, Opcode = EventOpcode.Stop, Task = Tasks.Process)]
+        public void ProcessExit(long ProcessId, long ParentProcessId, int ExitCode, string Executable, string CommandLine)
```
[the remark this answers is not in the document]
mjsabby: Done.

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.Diagnostics.Tracing
+{
+    [EventSource(Name = "ProcessMetadataEventSource")]
+    public sealed class ProcessMetadataEventSource : EventSource
+    {
+        public class Tasks
```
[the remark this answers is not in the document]
mjsabby: I intentionally steered away from them, because it would seem like process metadata is a sort of monolith. I'm happy to add them now or later if you see it differently.

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        [Event(4, Opcode = EventOpcode.Stop, Task = Tasks.Thread)]
+        public void ThreadDestroy(long ProcessId, long ThreadId, ulong StackBaseAddress, string ThreadName)
+        {
+            this.WriteEvent(4, ProcessId, ThreadId, StackBaseAddress, ThreadName);
+        }
+
+        [Event(5, Opcode = EventOpcode.Start, Task = Tasks.Module)]
+        public void ModuleLoad(long ProcessId, ulong LoadAddress, long ModuleSize, Guid DebugGuid, int DebugAge, string ModuleFilePath, string DebugModuleFileName)
```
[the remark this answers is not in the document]
mjsabby: As luck has it, ELF Debug IDs are 20-bytes. so I contemplated making this a byte[] but then thought about MachO which is also a GUID without any age, so I thought we could encode it as 0. And we could hack into the ELF IDs as guid + age to take the 20 bytes.

I just picked this and not a byte array because it was more convenient. @noahfalk thoughts?

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        [Event(2, Opcode = EventOpcode.Stop, Task = Tasks.Process)]
+        public void ProcessExit(long ProcessId, long ParentProcessId, int ExitCode, string Executable, string CommandLine)
+        {
+            this.WriteEvent(2, ProcessId, ParentProcessId, ExitCode, Executable, CommandLine);
+        }
+
+        [Event(3, Opcode = EventOpcode.Start, Task = Tasks.Thread)]
+        public void ThreadCreate(long ProcessId, long ThreadId, ulong StackBaseAddress, string ThreadName)
```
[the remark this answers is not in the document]
mjsabby: Rundown events or capture state events would have the same format, so I thought we could overload this, but I can also just add a specific thing for "Rundown", or maybe a more neutral term for catch up events.

src/TraceEvent/TraceLog.cs
mjsabby: Done.

noahfalk:
Its in my todo list now : )

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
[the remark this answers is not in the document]
brianrob: Gotcha - yes, interested in @noahfalk's thoughts here.  Do we have any other precedent?

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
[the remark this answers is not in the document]
brianrob: I do think it would be worth knowing if the thread has existed longer than the trace or not.  Perhaps something similar to Thread/DCStart?  Also, if you don't mind, let's change to Thread/Start and Thread/Stop top match existing precedent.

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            this.WriteEvent(3, ProcessId, ThreadId, StackBaseAddress, ThreadName);
```
noahfalk, line 31: Do we have a definition of what string to expect in ThreadName? For example .Net has a Thread.Name property and Linux has pthread_getname_np() but I don't know if the strings returned by those APIs are guaranteed to be equal.

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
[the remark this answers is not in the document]
noahfalk: I'd be tempted to encode it as a byte[] or string. Fixed size types would be a little more performant but given these shouldn't be frequent events I'd favor an encoding that feels straightforward. 

If the goal is to be able to look up the images and symbols on a symbol server then the data we need is:
format - PE/ELF/MachO
  - PE - filesize, timestamp, filename, debug signature, age, and the codeview debug directory major/minor version (major=0x100,minor=504D is a sentinel value that indicates portable pdb is supported - [PE spec](https://github.com/tmat/corefx/blob/master/src/System.Reflection.Metadata/specs/PE-COFF.md)). 
  - ELF - build-id and filename
  - MachO - uuid and name

[SSQP spec](https://github.com/dotnet/symstore/blob/master/docs/specs/SSQP_Key_Conventions.md)

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.Diagnostics.Tracing
+{
+    [EventSource(Name = "ProcessMetadataEventSource")]
+    public sealed class ProcessMetadataEventSource : EventSource
```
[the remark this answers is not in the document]
mjsabby: It is only for reference.

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
mjsabby: Are you suggesting a name change? I suppose it could be either, but I think adding two thread name events would be overkill.

src/TraceEvent/Parsers/ProcessMetadataTraceEventParser.cs
```diff
@@ -0,0 +1,683 @@
...
+#pragma warning disable 1591        // disable warnings on XML comments not being present
+
+// This code was automatically generated by the TraceParserGen tool, which converts
+// an ETW event manifest into strongly typed C# classes.
+namespace Microsoft.Diagnostics.Tracing.Parsers
+{
+    using Microsoft.Diagnostics.Tracing.Parsers.ProcessMetadataEventSource;
+
+    [System.CodeDom.Compiler.GeneratedCode("traceparsergen", "2.0")]
+    public sealed class ProcessMetadataEventSourceTraceEventParser : TraceEventParser
```
noahfalk, line 18: I realized above I said we didn't need to block on the name, but that probably also means we need to avoid putting the name in new public API surface. Can we switch to internal for now or should we just try to sort out the naming right away?

src/TraceEvent/TraceEvent.cs
```diff
@@ -194,6 +194,23 @@ public KernelTraceEventParser Kernel
             }
         }
 
+        /// <summary>
+        /// For convenience, we provide a property returns a ProcessMetadataTraceEventParser that knows 
+        /// how to parse all the Process Metadata events into callbacks.
+        /// </summary>
+        public ProcessMetadataEventSourceTraceEventParser ProcessMetadata
```
noahfalk, line 201: Similar to the comment above, another spot where the name is currently showing up in public API surface

src/TraceEvent/Parsers/ProcessMetadataTraceEventParser.cs
mjsabby: I'm fine but I don't have an opinion, I think it should be the .net name.

src/TraceEvent/EventSources/ProcessMetadataEventSource.cs
noahfalk: I am fine with the argument name, I am just trying to sort out what data do we expect to be provided? On Windows we have precedent from the kernel events but on all other platforms we have to define what is the equivalent data we are going to populate this with.

brianrob:

brianrob:

PR #1285, opened 2020-10-02
Match pdbcache folder in .diagsessions

pharring:
I noticed that the .etl.zip files that we capture via the Visual Studio Standard Collector Service use "PDBCache" for their symbols instead of symcache.

This causes errors when unzipping because the PDBCache folder isn't recognized. That further causes symbol resolution failures.


brianrob (approved):
LGTM.  Thanks @pharring!

PR #1288, opened 2020-10-14
Use the current culture's list separator as the separator for time ranges

paulomorgado:
When the user has regional settings that are not the expected English ones (like having a space as the digit grouping), the parsing of the start time will fail.

This PR uses the current culture information to parse and serialize these ranges.

/cc @brianrob


src/PerfView/RangeUtils.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>
+    /// Provides utility operations for <see cref="double"/> ranges.
+    /// </summary>
+    internal static class RangeUtils
+    {
+        /// <summary>
+        /// Array of list separators built from the current culture.
+        /// </summary>
+        /// <remarks>The user can set any character as a separator and that character can turn a regular expression invalid.</remarks>
+        private static readonly string[] listSeparators = new string[] { CultureInfo.CurrentCulture.TextInfo.ListSeparator };
```
brianrob, line 15: Is there a reason that you need to allocate a new array for this?  You should just be able to use the string `ListSeparator` as input to `string.Split` in `TryParse`.

brianrob (commented):
@paulomorgado, thanks for contributing this.  Overall looks good - one follow-up question.

src/PerfView/RangeUtils.cs
paulomorgado: That overload only exists for .NET Core. It doesn't exist for .NET Framework.

brianrob: Yes, you're right.

brianrob:
@paulomorgado, I did a bit of testing with your PR, and it seems that the Events view is partially broken with this change.  Here are some repro steps.  I looked at your code again, but the break wasn't immediately obvious to me.

1. Open a trace and pull up the Events view.
2. Multi-select some event types and hit enter to load the events.  They will load properly.
3. Multi-select another set of events and hit enter.

Instead of loading the events, the time range will be set to 0.00 - 0.00 and no events will appear.  You'll have to clear out the start and stop times to get the events to load again (and after another successful load, the next load will fail).

Can you take a look and see if you can debug it?

paulomorgado:
It doesn't repro on 2.0.61 but it doesn't run any code on this PR either.

brianrob:
Are you able to reproduce this on a local build?  I tested with and without your change with a local build, and was able to hit it with your change, but not without.

paulomorgado:
I was missing a breakpoint.

I think it's fixed now.

brianrob:
@paulomorgado, that issue does appear fixed.  Thanks.

I continued testing in the stack viewer (e.g. CPU Stacks), and things don't seem quite right there either.  When copying a time range on en-us, the value that gets written to the clipboard is <start>,<stop> instead of <start> <stop>.  This results in something like "0,34,000" instead of "0 34,000".  When pasting this into the start box, parsing fails.  This is because you're using the ListSeparator which on en-us is ','.  It's not clear to me what the right next step is here - perhaps looking into the way that numbers are being converted to text, and specifically what the separator characters are to see if there's a way to differentiate?

paulomorgado:
ARGH! We're back to where we started.

We could disallow pasting formatted values, but that would be a breaking change.

How about this to initialize the separator?

* Start with `" "`. (That's what it was before and there's no need to annoy users that don't have this issue.)
* If it's ambiguous, change to `CultureInfo.CurrentCulture.TextInfo.ListSeparator`.
* If it's still ambiguous, change to `"|"`. (I don't think this one will cause any issues.)

ilabutin:
There was an issue #235 with an attempt to fix parsing Start/End times exactly when space is a digit grouping symbol. And I tested it back then to verify it was fixed. Could you check the code in that old PR to better understand why it became broken (or was it incomplete)?
There was also a set of unit tests added for this functionality.

In my old issue, I was proposing some better "structured" representation in the clipboard like you suggest using `"|"`. I would say using `"|"` sounds like the best option. I doubt anybody uses time range from clipboard for anything but pasting back.

paulomorgado:
@ilabutin, do you mean this PR (#236)? It was [this comment in the code](https://github.com/microsoft/perfview/pull/236/files#diff-43365d915bf3956839783efe93cdaf607cba4b29954aea7f97cb9e6103bb00a4R2941):

    // TODO: This only works for cultures where a space is not the numeric group separator

I also use space as the digit grouping character and it's still broken for me.

ilabutin:
I see.
But I also know that I use `ru-RU` culture locally (where digit grouping symbol is a space). And here is my local test of number formatting in LinqPad:

image

At the same time, [there is a unit test](https://github.com/microsoft/perfview/pull/236/files#diff-2e51a772aa387d171dfec0cc80e19fa6a310ddaec05a0b9865832cbd0d98e88fR29) specifically for the `ru-RU` culture in the same PR. So, I wonder how the test passes then.

paulomorgado:
I haven't gone through that test, but this will fail:

```
var ru = CultureInfo.GetCultureInfo("ru-RU");
var text = $"{123456.789.ToString("n3", ru)} {234567.890.ToString("n3", ru)}".Dump();
Regex.Match(text, @"^\s*([\d\.,]+)\s+([\d\.,]+)\s*$").Dump();
```

Does it work for you in the released version of PerfView?

brianrob:
I agree that '|' may be the best choice.  If I understand correctly, as the code is written today it doesn't work for @ilabutin either.

paulomorgado:
I don't see how it would.

I have no problem implementing the 3-tier approach I [mentioned above](https://github.com/microsoft/perfview/pull/1288#issuecomment-712630575).

It would work as it works now for cultures that do not use space as a separator.
It would work with list separator if it's ambiguous.
It will work with pipe if the list separator is ambiguous.

I can submit a PR with this for you to test.

ilabutin:
I bet it worked for me during my tests because in 90% of cases I do trimming at the end, meaning that the start timestamp is 0,000 which is parsed OK (and the previous fix ensures that End time is parsed properly as well).

paulomorgado:
That's when I ran into this. The first time I needed to trim at the beginning.

paulomorgado:
@ilabutin, I just debugged your test and it never uses a range. That's why it always passes.

paulomorgado:
Introduced this logic for separator resolution:

```C#
private static string[] ResolveRangeSeparators()
{
    var listSeparator = "\u00A0";

    if (string.IsNullOrWhiteSpace(CultureInfo.CurrentCulture.NumberFormat.NumberGroupSeparator)
        || string.IsNullOrWhiteSpace(CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator))
    {
        listSeparator = CultureInfo.CurrentCulture.TextInfo.ListSeparator;

        if (string.IsNullOrWhiteSpace(listSeparator)
            || CultureInfo.CurrentCulture.NumberFormat.NumberGroupSeparator == listSeparator
            || CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator == listSeparator)
        {
            listSeparator = "|";
        }
    }

    return new string[] { listSeparator };
}
```

And added unit tests.

src/PerfView/StackViewer/StackWindow.xaml.cs
```diff
@@ -3615,44 +3615,32 @@ private bool ValidateStartAndEnd(StackSource newSource)
...
                 {
-                    if (double.TryParse(match.Groups[1].Value, out val1) &&
-                        double.TryParse(match.Groups[2].Value, out val2))
-                    {
-                        StartTextBox.Text = val1.ToString("n3");
-                        EndTextBox.Text = val2.ToString("n3");
-                    }
-                    else
-                    {
-                        StatusBar.LogError("Invalid number " + StartTextBox.Text);
```
brianrob, line 3652: We should preserve the behavior where we log the error here if something goes wrong.

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -1154,18 +1154,18 @@ public void Update()
...
             else
             {
-                var match = Regex.Match(StartTextBox.Text, @"^\s*([\d\.,]+)\s+([\d\.,]+)\s*$");
-                if (match.Success)
+                if (RangeUtilities.TryParse(StartTextBox.Text, out var startTimeRelativeMSec, out var endTimeRelativeMSec))
                 {
-                    m_source.StartTimeRelativeMSec = 0;
-                    double.TryParse(match.Groups[1].Value, out m_source.StartTimeRelativeMSec);
-
-                    m_source.EndTimeRelativeMSec = m_source.MaxEventTimeRelativeMsec;
```
brianrob, line 1163: The expected behavior here is that if we can't properly parse, then the range should be [0, MaxEventTimeRelativeMsec].  `RangeUtilities.TryParse` will set both of these to 0 if it can't properly parse an individual value.

src/PerfView/Utilities/RangeUtilities.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// Resolves the range separators for the current culture.
+        /// </summary>
+        /// <returns>The range separators for the current culture.</returns>
+        private static string[] ResolveRangeSeparators()
+        {
+            var listSeparator = " ";
+
+            if (string.IsNullOrWhiteSpace(CultureInfo.CurrentCulture.NumberFormat.NumberGroupSeparator)
```
brianrob, line 25: Can you describe what the behavior here should be (e.g. how would you tell someone to use this feature)?  I don't think I'm following this.

src/PerfView.Tests/Utilities/RangeUtilitiesTests.cs
```diff
@@ -0,0 +1,161 @@
...
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace PerfViewTests.Utilities
+{
+    public static class RangeUtilitiesTests
+    {
+        [Theory]
```
brianrob, line 16: Thanks for adding these!

src/PerfView/Utilities/RangeUtilities.cs
paulomorgado: 1. To determine the range separator, we start with ` `.
2. If the current culture's number group separator or number decimal separator is a white space, we change it to the list separator.
3. if the list separator is a white space or it matches the number group separator or number decimal separator, we change it to `|`.

I'll add this explanation to code and will gladly add to the docs, if you point me to where they are.

src/PerfView/StackViewer/StackWindow.xaml.cs
paulomorgado: What logging are we missing?

src/PerfView/EventViewer/EventWindow.xaml.cs
paulomorgado: Not exactly. We "if we can't match on the Regex". Because, after that, it's whatever `double.TryParse` outputs. Lines 1160 and 1163 are, effectively, no-op.

paulomorgado:
Added baseline tests for `StackWindow` as a separate PR: #1327 

I'd gladly add tests for `EventWindow`, but I couldn't figure out how to set up the tests.

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -1154,18 +1154,18 @@ public void Update()
...
-                    double.TryParse(match.Groups[2].Value, out m_source.EndTimeRelativeMSec);
+                    m_source.StartTimeRelativeMSec = startTimeRelativeMSec;
+                    m_source.EndTimeRelativeMSec = endTimeRelativeMSec;
                 }
                 else
                 {
-                    if (!double.TryParse(StartTextBox.Text, out m_source.StartTimeRelativeMSec))
+                    if (double.TryParse(StartTextBox.Text, out startTimeRelativeMSec))
+                    {
+                        m_source.StartTimeRelativeMSec = startTimeRelativeMSec;
```
brianrob, line 1166: I think you will want to keep the old behavior here in this else clause.  With your change here, if the value can't be parsed into a double, then the existing value of the field will be whatever the user just typed.  Without the change, it will revert back to 0, which is what we want.

src/PerfView/StackViewer/StackWindow.xaml.cs
brianrob: The call to StatusBar.LogError("Invalid number " + StartTextBox.Text);

---
In reply to: [524065367](https://github.com/microsoft/perfview/pull/1288#discussion_r524065367) [](ancestors = 524065367)

src/PerfView/Utilities/RangeUtilities.cs
brianrob: Is there a reason to have #2 at all?  I wonder if we can't just do #1 and if that fails, go straight to #3.  It would make things simpler.

---
In reply to: [524062988](https://github.com/microsoft/perfview/pull/1288#discussion_r524062988) [](ancestors = 524062988)

src/PerfView.Tests/Utilities/RangeUtilitiesTests.cs
paulomorgado: There's #1327 with just those tests

src/PerfView/Utilities/RangeUtilities.cs
paulomorgado: \#2 was my initial idea, but I found cases where it's still ambiguous. That's why I came up with \#3.

I'm OK with dropping \#2.

brianrob: Sounds good.  Go for it.

---
In reply to: [555350287](https://github.com/microsoft/perfview/pull/1288#discussion_r555350287) [](ancestors = 555350287)

brianrob (approved):

PR #1289, opened 2020-10-14
Display JIT Allocated Heap Size

kunalspathak:
Added a new event `MethodJitMemoryAllocatedForCode`  in https://github.com/dotnet/runtime/pull/44030 to log the JIT 's memory request for code. This PR displays that information in PerfView.

Here is the screenshot:

image

### DIFF
--- a/src/PerfView/JitStats.cs
+++ b/src/PerfView/JitStats.cs
@@ -160,7 +160,7 @@ namespace Stats
                     "<TH Title=\"The number of times the JIT was invoked for methods in this module\">Num Compilations</TH>" +
                     "<TH Title=\"The total amount of IL processed by the JIT for all methods in this module\">IL Size</TH>" +
                     "<TH Title=\"The total amount of native code produced by the JIT for all methods in this module\">Native Size</TH>" +
-                    "<TH Title=\"The total amount of heap memory allocated for the code produced by the JIT for all methods in this module\">LoaderHeap Allocation Size</TH>" +
+                    "<TH Title=\"The total amount of loader heap memory allocated for the code produced by the JIT for all methods in this module. Only available when 'LoaderHeap' on private CLR provider is enabled.  \">LoaderHeap Allocation Size</TH>" +
                     "<TH Title=\"Time spent jitting synchronously to produce code for methods that were just invoked. These compilations often consume time at startup.\">" + GetLongNameForThreadClassification(CompilationThreadKind.Foreground) + "<BR/>msec</TH>" +
                     "<TH Title=\"Time spent jitting asynchronously to produce code for methods the runtime speculates will be invoked in the future.\">" + GetLongNameForThreadClassification(CompilationThreadKind.MulticoreJitBackground) + "<BR/>msec</TH>" +
                     "<TH Title=\"Time spent jitting asynchronously to produce code for methods that is more optimized than their initial code.\">" + GetLongNameForThreadClassification(CompilationThreadKind.TieredCompilationBackground) + "<BR/>msec</TH>" +
@@ -183,7 +183,7 @@ namespace Stats
                     runtime.JIT.Stats().Count,
                     runtime.JIT.Stats().TotalILSize,
                     runtime.JIT.Stats().TotalNativeSize,
-                    runtime.JIT.Stats().TotalLoaderHeapAllocSize,
+                    runtime.JIT.Stats().TotalLoaderHeapAllocSizeIfAvailable,
                     runtime.JIT.Stats().TotalForegroundCpuTimeMSec,
                     runtime.JIT.Stats().TotalBackgroundMultiCoreJitCpuTimeMSec,
                     runtime.JIT.Stats().TotalBackgroundTieredCompilationCpuTimeMSec);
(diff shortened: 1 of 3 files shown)


src/PerfView/JitStats.cs
```diff
@@ -160,6 +160,7 @@ public static void ToHtml(TextWriter writer, TraceProcess stats, TraceLoadedDotN
                     "<TH Title=\"The number of times the JIT was invoked for methods in this module\">Num Compilations</TH>" +
                     "<TH Title=\"The total amount of IL processed by the JIT for all methods in this module\">IL Size</TH>" +
                     "<TH Title=\"The total amount of native code produced by the JIT for all methods in this module\">Native Size</TH>" +
+                    "<TH Title=\"The total amount of heap memory allocated for the code produced by the JIT for all methods in this module\">Allocated Heap Size</TH>" +
```
brianrob, line 163: For most people, these events aren't going to be collected, as they're too expensive to turn on by default.  I would recommend that you make the addition of this column conditional on having the data.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -3912,6 +3926,21 @@ internal static TraceJittedMethod MethodComplete(TraceLoadedDotNetRuntime stats,
             return _method;
         }
 
+        /// <summary>
+        /// Handles AllocRequest event for JIT
+        /// </summary>
+        internal static TraceJittedMethod JitAllocComplete(TraceLoadedDotNetRuntime stats, LoaderHeapAllocRequestTraceData data)
```
brianrob, line 3932: You can just return `void` here since you don't need to do anything with the method in the caller.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -1301,6 +1301,15 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
                     }
                 };
 
+                clrPrivate.LoaderHeapAllocRequest += delegate(LoaderHeapAllocRequestTraceData data)
+                {
+                    var process = data.Process();
+                    var stats = currentManagedProcess(data);
+                    var _method = JITStats.JitAllocComplete(stats, data);
+                    //TODO: What to do here of _method??
```
brianrob, line 1309: You should not need to do anything with `_method` here.  `JitAllocComplete` did all the work to capture the allocation.

src/TraceEvent/Computers/TraceManagedProcess.cs
brianrob, line 3932: I would recommend calling this method something like `LogJitAllocation` or something along those lines, since it's not actually a completion operation.  `MethodComplete`, for example, is responsible for marking the JIT operation complete, which is where the complete name comes from.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -4116,6 +4145,10 @@ public class TraceJittedMethod
         /// </summary>
         public int NativeSize;
         /// <summary>
+        /// Heap size allocated for JIT code of method
+        /// </summary>
+        public int AllocatedHeapSize;
```
brianrob, line 4150: Should we name this something like `LoaderHeapAllocSize` to be clear that these are loader heap allocations, instead of C++ heap allocations?  (If so, we should also update the various related fields.)

brianrob (commented):
@kunalspathak, this is looking good.  A few comments/questions.

src/PerfView/JitStats.cs
kunalspathak: Sure. What is the right way to check that? Just don't display all of this if `TotalAllocSize == 0` or have some flag somewhere in TraceManagerProcess.cs inside `LoaderHeapAllocRequest` delegate?

brianrob: Yes, using `TotalAllocSize==0` is a good way to do it.

kunalspathak: There is lot of conditional formatting code that needs to be turned off if we need to display this column conditionally. E.g. Adding the format for [extra column in moduleTableRow](https://github.com/microsoft/perfview/pull/1289/files#diff-02da9e6f69caee6ee8d847ae7dbceb6f9dde9c94428bf278c6cc056ab786fec7R175). I am thinking of displaying `N/A` or `-` if there is no data instead of `0`? In the help section, I will add a line saying how can someone view that data. Thoughts?

kunalspathak: Alternatively, I will need to just add this as a last column which will be easier to turn on and off.

brianrob: I think `N/A` would probably be OK.  The key is to make sure it's clear that the data isn't there (as opposed to 0).

kunalspathak:
@brianrob  - I think this should be ready for review.

brianrob (approved):
LGTM.  I will merge this once the corresponding runtime change is merged.

kunalspathak:
@brianrob - https://github.com/dotnet/runtime/pull/44030 is merged now.

PR #1294, opened 2020-10-28
Add API for reading logs from multiple ETW files

0b01:
This PR adds a constructor that takes a list of ETW files to ETWTraceEventSource and a MultiFile mode in TraceEventSourceType.


src/TraceEvent/TraceEvent.cs
```diff
@@ -3145,6 +3145,16 @@ public sealed class TraceEventDispatcherOptions
     /// </summary>
     public abstract unsafe class TraceEventDispatcher : TraceEventSource
     {
+        /// <summary>
+        /// Obtains the correct TraceEventDispatcher for multiple ETW files.
+        /// </summary>
+        /// <param name="fileNames">A list of paths to trace files.</param>
+        /// <returns>A TraceEventDispatcher for the given trace files.</returns>
+        public static TraceEventDispatcher GetDispatcherFromMultiEventTraceLogFiles(IEnumerable<string> fileNames, TraceEventDispatcherOptions options = null)
```
brianrob, line 3153: I don't think you should add this method, but instead just call `new ETWTraceEventSource` directly, and pass that to a new method in `TraceLog` that looks like this:

```
        public static string CreateFromEventTraceLogFile(TraceEventDispatcher source, string etlxFilePath, TraceLogOptions options = null)
        {
            if (etlxFilePath == null)
            {
                throw new ArgumentNullException(nameof(etlxFilePath));
            }

            if (source.EventsLost != 0 && options != null && options.OnLostEvents != null)
            {
                options.OnLostEvents(false, source.EventsLost, 0);
            }

            CreateFromTraceEventSource(source, etlxFilePath, options);

            return etlxFilePath;
        }

```

This ensures that if someone wants to do more custom things than what TraceEvent knows how to do, that they select the right `EventSource` class rather than using have to write a method that makes sure that all of the files are of the right type before creating the corresponding `EventSource`.  Also, most `EventSources` don't support multiple files.

src/TraceEvent/TraceLog.cs
```diff
@@ -51,27 +51,49 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
     /// </summary>
     public sealed class TraceLog : TraceEventSource, IDisposable, IFastSerializable, IFastSerializableVersion
     {
+        /// <summary>
+        /// Given an array of files, create an ETLX file for the data. 
+        /// <para>If etlxFilePath is null the output name is derived from etlFilePath by changing its file extension to .ETLX.</para>
+        /// <returns>The name of the ETLX file that was generated.</returns>
+        /// </summary>
+        public static string CreateFromEventTraceLogFiles(IEnumerable<string> fileNames, string etlxFilePath = null, TraceLogOptions options = null, TraceEventDispatcherOptions traceEventDispatcherOptions = null)
```
brianrob, line 59: I don't think you should create this method, but instead use the other one that you created that takes a dispatcher directly.

brianrob (commented):
@0b01, thanks for your contribution.  Some structural comments below.

src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -54,6 +54,18 @@ public ETWTraceEventSource(string fileOrSessionName, TraceEventSourceType type)
...
         }
+        /// <summary>
+        /// Open multiple ETL files for processing.
+        /// </summary>
+        /// <param name="fileNames">
+        /// If type == MultiFile, use only supplied files.
+        /// If type == MergeAll, for each file find related files in same directory.
+        /// <param name="type"></param>
+        // [SecuritySafeCritical]
+        public ETWTraceEventSource(IEnumerable<string> fileNames, TraceEventSourceType type)
```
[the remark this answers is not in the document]
0b01: `public ETWTraceEventSource(IEnumerable<string> fileNames)` constructor already exists? Is it okay if I add Initialize() to that?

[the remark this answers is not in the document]
brianrob: Ah, I didn't realize that.  Do you even need to call `Initialize`?  I suspect that the existing one already does what you want.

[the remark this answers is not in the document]
0b01: It won't work because the key is treating the files as the same trace. So when this.fileNames is set it'll process all the files instead of just 1.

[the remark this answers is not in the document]
brianrob: Gotcha.  That makes sense.  In that case, then yes, you should do what you were doing before, and keep this new `ETWTraceEventSource` constructor that takes a type.  Then if `type == MergeAll`, call `Initialize` to set things up.  I don't think you need to add a new value to `TraceEventSourceType`, as `FileOnly` is the type that the existing constructor represents.  Sorry for sending you back and forth.

src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -55,6 +56,21 @@ public ETWTraceEventSource(string fileOrSessionName, TraceEventSourceType type)
...
 
+        /// <summary>
+        /// Open multiple etl files as one trace for processing.
+        /// </summary>
+        /// <param name="fileOrSessionName">
+        /// If type == MergeAll, call Initialize
+        /// <param name="type"></param>
+        // [SecuritySafeCritical]
+        public ETWTraceEventSource(IEnumerable<string> fileNames, TraceEventSourceType type)
+        {
```
brianrob, line 67: Before checking the type, you should unconditionally set `this.fileNames = fileNames` like the other constructor does, since if `type != TraceEventSourceType.MergeAll` then the file names won't get saved.  Basically if `type == TraceEventSourceType.FileOnly` then we want to replicate the behavior of `ETWTraceEventSource..ctor(IEnumerable<string> fileNames)`.

src/TraceEvent/TraceLog.cs
```diff
@@ -57,21 +57,31 @@ public sealed class TraceLog : TraceEventSource, IDisposable, IFastSerializable,
         /// <returns>The name of the ETLX file that was generated.</returns>
         /// </summary>
         public static string CreateFromEventTraceLogFile(string filePath, string etlxFilePath = null, TraceLogOptions options = null, TraceEventDispatcherOptions traceEventDispatcherOptions = null)
+        {
+            using (TraceEventDispatcher source = TraceEventDispatcher.GetDispatcherFromFileName(filePath, traceEventDispatcherOptions))
```
brianrob, line 61: To preserve existing behavior, this method should make sure that `etlxFilePath != null`, and if it is, set it to `etlxFilePath = Path.ChangeExtension(filePath, ".etlx");`.

brianrob (commented):
@0b01, this is shaping up nicely.  A few more comments.

src/TraceEvent/ETWTraceEventSource.cs
0b01: Adding to the else clause when type != MergeAll

src/TraceEvent/TraceLog.cs
0b01: Good catch

brianrob:
@0b01, this looks good.  One last request - this seems like it wouldn't be hard to break in the future, so I would like to request that you create a unit test to protect the new `ETWTraceEventSource` constructor.  This involves creating a set of ETL files that get run through the new constructor, produce a `TraceLog` out of it like the tests in https://github.com/microsoft/perfview/blob/master/src/TraceEvent/TraceEvent.Tests/GeneralParsing.cs, and then converting it to a textual format so that it can be diffed with a baseline trace.  One test should be sufficient to provide coverage on the constructor, which is what I'm most concerned about.  The contents of the file should be sufficiently covered by the existing tests.

0b01:
@brianrob I added a simple test opening 4 etls as 1 trace. The files are from a DiagHub scenario that this PR will enable. Could you take a look and see if it's sufficient?

brianrob:
@0b01, can you please take a look at the test failure?  It looks to be from the new test.

0b01:
@brianrob I removed the size assertion which may be different for the CI environment. Now it's only using count assertion.

src/TraceEvent/TraceEvent.Tests/MultiFileMergeAll.cs
```diff
@@ -0,0 +1,43 @@
...
+            string etlFilePath = "diaghub-etls";
+            Output.WriteLine(string.Format("Processing the file {0}, Making ETLX and scanning.", etlFilePath));
+            string eltxFilePath = Path.ChangeExtension(etlFilePath, ".etlx");
+
+            TraceEventDispatcher source = new ETWTraceEventSource(fileNames, TraceEventSourceType.MergeAll);
+            TraceLog traceLog = new TraceLog(TraceLog.CreateFromEventTraceLogFile(source, eltxFilePath));
+
+            var traceSource = traceLog.Events.GetSource();
+
+            Assert.Equal(95506, traceLog.EventCount);
```
brianrob, line 40: I think that rather than using the count assert, it's important that you do what the other file validation tests do, and convert the contents of the events to a text file that can be diff'd.  This ensures that we don't break the most fragile part of the change, which makes sure that we process the files all as one, instead of sequentially.  I know it's a bit more work, but I think it's worth it.

src/TraceEvent/TraceEvent.Tests/MultiFileMergeAll.cs
```diff
@@ -0,0 +1,100 @@
...
+            TraceEventDispatcher source = new ETWTraceEventSource(fileNames, TraceEventSourceType.MergeAll);
+            TraceLog traceLog = new TraceLog(TraceLog.CreateFromEventTraceLogFile(source, eltxFilePath));
+
+            Assert.Equal(95506, traceLog.EventCount);
+            var stopEvents = traceLog.Events.Filter(e => e.EventName == "Activity2Stop/Stop");
+            Assert.Equal(55, stopEvents.Count());
+            Assert.Equal((uint)13205, (uint)stopEvents.Last().EventIndex);
+
+            using (var file = new StreamReader(UnZippedDataDir + "\\diaghub-etls\\diaghub-etls.txt"))
+            //using (var file = new StreamWriter(UnZippedDataDir + "\\diaghub-etls\\diaghub-etls.txt"))
```
brianrob, line 44: Nit: Can you remove this comment?

src/TraceEvent/TraceEvent.Tests/MultiFileMergeAll.cs
```diff
@@ -0,0 +1,100 @@
...
+                    // TODO FIX NOW, this is broken and should be fixed.  
+                    // We are hacking it here so we don't turn off the test completely.  
+                    if (eventName == "DotNet/CLR.SKUOrVersion")
+                    {
+                        return;
+                    }
+
+                    var evt = GeneralParsing.Parse(data);
+
+                    //file.WriteLine(evt);
```
brianrob, line 91: Nit: Can you please remove this comment?

src/TraceEvent/TraceEvent.Tests/MultiFileMergeAll.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            TraceEventDispatcher source = new ETWTraceEventSource(fileNames, TraceEventSourceType.MergeAll);
+            TraceLog traceLog = new TraceLog(TraceLog.CreateFromEventTraceLogFile(source, eltxFilePath));
+
+            Assert.Equal(95506, traceLog.EventCount);
+            var stopEvents = traceLog.Events.Filter(e => e.EventName == "Activity2Stop/Stop");
+            Assert.Equal(55, stopEvents.Count());
+            Assert.Equal((uint)13205, (uint)stopEvents.Last().EventIndex);
+
+            using (var file = new StreamReader(UnZippedDataDir + "\\diaghub-etls\\diaghub-etls.txt"))
```
brianrob, line 43: Can you rename `diaghub-etls.txt` to `diaghub-<version>-<bitness>-diagsession.baseline.txt` and move it out of the zip file?  Can you also rename the zip file to `diaghub-<version>-<bitness>-diagsession.zip`?  This gets things pretty close to the rest of the files, and will tell us in the future where the file came from.

brianrob (commented):
@0b01, this is looking really good.  A couple of minor things to bring this inline with the rest of the tests.

brianrob (approved):
LGTM.  Thanks!

PR #1295, opened 2020-10-28
Support parsing GCHeapStats_V2

cshung:
https://github.com/dotnet/runtime/pull/34549 introduced `GCHeapStats_V2`. The new event requires additional parsing logic to handle the extra fields.

The change is driven by adding the `GenerationSize4` and `TotalPromotedSize4` fields to `GCHeapStatsTraceData` class and add the appropriate parsing logic there, the rest is driven by consuming these fields in various places where we consumed `GenerationSize3` or `TotalPromotedSize3`.

Most of the changes are routine, except one place I marked with a `TODO` where I am not exactly sure what to do there.

I am sure the parsing logic is correct by taking a dump at the point when the event is emitted and comparing the parsed values with the dump. The correctness of the rest is less clear.


src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -2790,6 +2804,7 @@ internal static double GetGenSizeBeforeMB(List<TraceGC> GCs, TraceGC gc, Gens ge
                 return heapStats.GenerationSize2 / 1000000.0;
             }
 
+            // TODO, poh, what shall we do here?
```
cshung, line 2807: Note - this is important.

Maoni0 (approved):

src/TraceEvent/Computers/TraceManagedProcess.cs
brianrob: Good question. I would defer to @maoni0 on this.

brianrob (approved):
With the exception of the TODO, everything LGTM!

src/TraceEvent/Computers/TraceManagedProcess.cs
Maoni0: oops, I missed this comment. is there a problem doing the same as the other gens here, ie, return heapStats.GenerationSize4?

cshung: I am probably just being paranoid. Notice the `Math.Max` special case there for LOH just a few lines below. Wonder why we do it and does it apply to POH or not.

Maoni0: the Max just says that we want to the bigger one because we want the size before GC - either we do a gen2 GC, and this gc's heapstats could have a smaller number, in which if the previous one's is bigger we know that's closer to the actual value; or if it's not a gen2, and the currrent GC's is bigger then something was allocated and increased the gen size so we should take this gc's data. this is all an estimate. hopefully we never hit this code path because the perheap hist already tells us the accurate size_before so this is only there to handle the unusual cases. but the point is really there's no reason to handle LOH and POH differently.

tommcdon:
@karpinsn

src/TraceEvent/Samples/20_ObserveGCEvent.cs
```diff
@@ -63,12 +63,15 @@ public static void Run()
...
-                    Out.WriteLine("GC Collect:  Proc: {0,10} Gen0: {1,6:f1}M Gen1: {2,6:f1}M Gen2: {3,6:f1}M LargeObj: {4,6:f1}M",
+                    Out.WriteLine("GC Collect:  Proc: {0,10} Gen0: {1,6:f1}M Gen1: {2,6:f1}M Gen2: {3,6:f1}M LargeObj: {4,6:f1}M PinObj {4,6:f1}M",
                          GetProcessName(collectData.ProcessID),
                          collectData.GenerationSize0 / 1000000.0,
                          collectData.GenerationSize1 / 1000000.0,
                          collectData.GenerationSize2 / 1000000.0,
-                         collectData.GenerationSize3 / 1000000.0));
+                         collectData.GenerationSize3 / 1000000.0,
+                         0));
+                // TODO, enable POH display once TraceEvent package is updated
```
cshung, line 73: Note that this is pending on package update.

cshung:
@brianrob This is ready to be merged.

src/TraceEvent/Samples/20_ObserveGCEvent.cs
brianrob: @cshung, is this because the samples are dependent upon the package and not directly on the project?

cshung: Yes, it is.

brianrob (approved):

PR #1297, opened 2020-11-03
ArrayEquals fix

DamirAinullin:
ArrayEquals fix


brianrob (approved):
Thanks @DamirAinullin!

PR #1299, opened 2020-11-03
Fix infinite recursion for ElementName property

DamirAinullin:
Fix infinite recursion for ElementName property


DamirAinullin:
Do you know why CI was failed?

brianrob:
This is a transient failure.  Passed on re-run.

brianrob (approved):

PR #1302, opened 2020-11-04
Replace ContainsKey and [] with TryGetValue for Dictionary

DamirAinullin:
Replace ContainsKey and [] with TryGetValue for Dictionary, some additional code formatting

### DIFF
--- a/src/PerfView/HeapView/HeapAnalyzer.cs
+++ b/src/PerfView/HeapView/HeapAnalyzer.cs
@@ -932,11 +932,12 @@ namespace PerfView
             }
 
             // Get the process we want
-            if (!m_processLookup.TryGetValue(procID, out m_process))
+            if (!m_processLookup.TryGetValue(procID, out var process))
             {
                 return false;
             }
 
+            m_process = process;
             m_runtime = Microsoft.Diagnostics.Tracing.Analysis.TraceLoadedDotNetRuntimeExtensions.LoadedDotNetRuntime(m_process);
             return true;
         }
--- a/src/PerfView/HeapView/HeapAnalyzer.cs
+++ b/src/PerfView/HeapView/HeapAnalyzer.cs
@@ -716,9 +716,8 @@ namespace PerfView
         {
             if (data.ProcessID == m_procID)
             {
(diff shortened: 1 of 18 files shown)


brianrob:
@DamirAinullin, thank you for your contribution.  This change is reasonably large, and touches many different places, some of which I think might benefit from this, but others I suspect we wouldn't notice.  Do you have a scenario in mind where this makes a material performance difference?

DamirAinullin:
@brianrob, no, I just went through whole solution.

src/PerfView/HeapView/HeapAnalyzer.cs
```diff
@@ -933,12 +932,11 @@ public bool LoadEvents(int procID, int sampleInterval100ns)
             }
 
             // Get the process we want
-            if (!m_processLookup.ContainsKey(procID))
+            if (!m_processLookup.TryGetValue(procID, out m_process))
```
sharwell, line 935: 📝 This one appears to be a behavior change if `m_process` is not null when this line is reached.

src/PerfView/OverweightReport.cs
```diff
@@ -165,13 +165,12 @@ private class Result
             foreach (var key in d1.Keys)
```
sharwell, line 165:
```suggestion
            foreach (var (key, v1) in d1)
```

src/PerfView/OverweightReport.cs
```diff
@@ -165,13 +165,12 @@ private class Result
...
             {
                 // skip symbols that are not in both traces
-                if (!d2.ContainsKey(key))
+                if (!d2.TryGetValue(key, out var v2))
                 {
                     continue;
                 }
 
                 var v1 = d1[key];
```
sharwell, line 173:
```suggestion
```

sharwell (commented):
I've made it part way through. I agree with the primary change in principle, I'm seeing a whole bunch of unrelated formatting changes which is slowing down the review.

DamirAinullin:
@sharwell agree with yout point about formatting, sorry about that. It was just auto-formatting from VisualStudio.

PR #1306, opened 2020-11-06
Increase maximum size for GrowableArray<T>

sharwell:


brianrob:
@sharwell, this seems reasonable.  One question - in addition to increasing the maximum size, this also is a bit more aggressive at increasing the size of the array.  Is this for very large traces or heaps where you're seeing lots of calls to realloc?

sharwell:
> this also is a bit more aggressive at increasing the size of the array

This was not intentional. My only goal was to latch the size at `int.MaxValue` where the previous code would overflow somewhere around `int.MaxValue/3`.

src/FastSerialization/GrowableArray.cs
```diff
@@ -399,15 +399,20 @@ private void Realloc(int minSize)
             }
             else
             {
-                int expandSize = array.Length * 3 / 2 + 8;
-                if (minSize < expandSize)
+                long expandSize = (long)array.Length * 3 / 2 + 8;
+                if (minSize > expandSize)
```
brianrob, line 403: Would you mind reverting `>` back to `<`?  Otherwise, this change looks good to me.

sharwell: Update to closer match the original

brianrob (approved):

PR #1309, opened 2020-11-06
Make HistoryDictionary key generic

sharwell:


brianrob:
Can you share some context on this change, and what the ultimate goal is?  Thanks.

sharwell:
Some of the casts from thread ID (`int`) to address (`ulong`) were failing when I turned on overflow checking. Rather than use `unchecked(...)`, I fixed the data types to actually match the inputs.

brianrob (approved):

PR #1310, opened 2020-11-06
Set initial capacity for collections where it is predefined

DamirAinullin:
Set initial capacity for collections where it is predefined


brianrob (approved):
LGTM.  Thanks @DamirAinullin!

PR #1312, opened 2020-11-08
Added TraceEvent.props to nuspec also as buildTransitive

### LINKED ISSUE #1301: The "build" assets in the "Microsoft.Diagnostics.Tracing.TraceEvent" package should also be "buildTransitive"
## Problem

The `Microsoft.Diagnostics.Tracing.TraceEvent.props` file that's stored inside the `build` folder of the `Microsoft.Diagnostics.Tracing.TraceEvent` package is responsible for copying the native DLLs to the output directory. (Looking at the current version `2.0.61`)

However, due to the nature of how `build` assets work in NuGet, this will only be applied on the project that directly references this package. If there's another dependency step between, the native DLLs will not end up in the other projects. More info [here](https://github.com/NuGet/Home/wiki/Allow-package--authors-to-define-build-assets-transitive-behavior).

## Example

For example, `ProjectA` references the `Microsoft.Diagnostics.Tracing.TraceEvent` package, and `WebApplicationB` references `ProjectA`. Then the native DLLs won't end up in the final output of `WebApplicationB`, which is where they are needed, because `ProjectA` is just a class library, not an executable.

## Solution?

This can be solved by having the `Microsoft.Diagnostics.Tracing.TraceEvent.props` file twice in the NuGet package -- once in the `build` folder, and once in the `buildTransitive` folder.

tompazourek:
Fixes #1301, see that for more info.

### DIFF
--- a/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
+++ b/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
@@ -42,6 +42,7 @@
   <files>
     <!-- Support for native binaries -->
     <file src="Microsoft.Diagnostics.Tracing.TraceEvent.props" target="build\Microsoft.Diagnostics.Tracing.TraceEvent.props" />
+    <file src="Microsoft.Diagnostics.Tracing.TraceEvent.props" target="buildTransitive\Microsoft.Diagnostics.Tracing.TraceEvent.props" />
 
     <!-- Native binaries -->
     <file src="$OutDir$netstandard2.0\amd64\KernelTraceControl.dll" target="lib\native\amd64\KernelTraceControl.dll" />


sharwell:
❔ Are these dependencies required when loading an existing .etl file (as opposed to trying to collect a new one)?

sharwell:
💭 It seems like a better approach would be eliminating this dependency altogether.

tompazourek:
@sharwell I'm not actually sure if this dependency required (probably it isn't because if it was, more people would complain about its copying being broken in its current state). If the dependency on those native DLLs weren't there at all, it would be better, of course. But I don't think I can help with that part as I don't understand enough about that. If it's an optional dependency, maybe a better design would be to have a separate package for it.

All I notice is the current state with the `build` vs. `buildTransitive` discrepancy... The DLLs are copied, but only in the project that directly references the package, which might not be the executable project that actually needs to have them in the output. This PR simply makes this copying of DLLs consistent and working correctly as opposed to the current half-broken state. Whether those DLLs actually need to be copied is probably a separate issue. But for me, less dependencies is always better, so I'd be happy to see these gone 👍

brianrob:
There are a few dependencies here that get used for various things:

 - KernelTraceControl.dll: Used when capturing ETW traces.
 - msdia140.dll: Used to consume PDBs when resolving symbols.

There are also some other files as well that support these scenarios or are dependencies that these DLLs need in order to function (e.g. vcruntime140.dll).

I don't currently see a way where we get rid of these - the code is not open source, and I don't believe there are plans to open source it.

@sharwell, your point about by-design optional components is interesting.  Depending upon the goal of the app, I could believe that one or more DLLs might be optional, but the range of apps that would make all of them optional is likely low.  I'm inclined to take this change based on that.  Thoughts?

sharwell:
No objection here. I believe downstream projects could customize the build logic if they need to exclude some of the files.

brianrob (approved):

PR #1313, opened 2020-11-09
Enable loading symbols from long path names  ( > 250 characters) when…

Alois-xx:
… support for it is enabled on Windows.


Alois-xx:
I have created another with no merge conflicts

PR #1314, opened 2020-11-09
Issue1304 long path final

Alois-xx:
Enable Long Path Support on Windows Machines where it is enabled.


src/PerfView/app.config
```diff
@@ -7,4 +7,7 @@
   <startup>
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5"/>
   </startup>
+  <runtime>
+    <AppContextSwitchOverrides value="Switch.System.IO.UseLegacyPathHandling=false;Switch.System.IO.BlockLongPaths=false" />
```
sharwell, line 11: ❔ Any reason to not just update to .NET 4.7.2?

Alois-xx: It depends on which OS you can get data. See https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies
.NET 4.5 supports ✔️ 2012   | ✔️ 8

If you go to 4.7.2 you can only collect data on 

Windows versions | ✔️ 10 October 2018 Update (Version 1809)✔️ 10 April 2018 Update (Version 1803)➕ 10 Fall Creators Update (Version 1709)➕ 10 Creators Update (Version 1703)➕ 10 Anniversary Update (Version 1607)➕ 8.1➕7

✔️ Windows Server 2019✔️ Windows Server, version 1809✔️ Windows Server, version 1803➕ Windows Server, version 1709➕ 2016➕ 2012 R2➕ 2012➕ 2008 R2 SP1

Is this really what you want?

brianrob: Right, my preference would be to keep support for .NET 4.5.

src/PerfView/Properties/PerfView.manifest
```diff
@@ -0,0 +1,81 @@
...
+  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
+  <application xmlns="urn:schemas-microsoft-com:asm.v3">
+    <windowsSettings>
+      <longPathAware xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">true</longPathAware>
+    </windowsSettings>
+  </application>
+  <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
+    <security>
+      <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
+        <!-- UAC Manifest Options
```
sharwell, line 12: 💡 Can you remove the comments to simplify this file?

src/PerfView/Properties/PerfView.manifest
```diff
@@ -0,0 +1,81 @@
...
+            Specifying requestedExecutionLevel element will disable file and registry virtualization. 
+            Remove this element if your application requires this virtualization for backwards
+            compatibility.
+        -->
+        <requestedExecutionLevel level="asInvoker" uiAccess="false" />
+      </requestedPrivileges>
+    </security>
+  </trustInfo>
+
+  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
```
sharwell, line 29: 💡 All the elements from here to the end can be removed to simplify this file

src/PerfView/Properties/PerfView.manifest
```diff
@@ -0,0 +1,81 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
```
sharwell, line 3: ❔ Does this need to match the application name? If not, this entire file can be replaced by:
https://github.com/dotnet/roslyn/blob/e704ca635bd6de70a0250e34c4567c7a28fa9f6d/eng/config/app.manifest

sharwell: Ah, looks like support for Roslyn is implemented here:

https://github.com/dotnet/roslyn/blob/4758a7be7e9e59204bcee81eccd8309a5c04d393/eng/targets/Imports.targets#L322-L367

Alois-xx: yes but only for .NET 4.7.2 
Condition="'$(OutputType)' == 'Exe' And '$(TargetFramework)' == 'net472'">

Alois-xx: I do not worry about the assembly version in the manifest because no one would want to reference PerfView from unmanaged code via a manifest reference. I think we can keep it to 1.0.0.0 without any issues.

Alois-xx:
Is this ever beeing merged?

sharwell (approved):

AloisKraus:
can this now be merged?

raghavthind2005 (approved):

src/PerfView/app.config
brianrob, line 11: `UseLegacyPathHandling` gets set to `true` for container scenarios, so I'm suspicious that this will break long path support.  Does long path support require that `UseLegacyPathHandling==false`?

AloisKraus: UseLegacyPathHandling=true allows to use also the \\?\ notation which is for some paths an alternate way to reach long path support. But it breaks for UNC file share names. In general it should not be a problem in container scenarios if you are dealing with "normal" paths names. I added it to be more modern and faster because path normalization used to be more expensive when set to false.

AloisKraus:
Would I need to do something or who can do the update?

PR #1317, opened 2020-11-11
Remove Silverlight support

sharwell:
This change simplifies the transition to ClrMD 2.0, which does not support Silverlight.


src/HeapDump/ICorDebugGCHeap.cs
```diff
@@ -1,1407 +0,0 @@
-﻿using Microsoft.Diagnostics.Runtime;
```
brianrob, line 1: @sharwell, was this used by Silverlight, or more generally?  If I recall correctly, this is a fallback path after `CLRMD`?

sharwell: Currently the file is only used if `Silverlight.DebugActiveSilverlightProcess` returns non-null. Porting this file to ClrMD 2 will be extremely difficult, if possible.

brianrob: Gotcha - just wanted to be sure.  Thanks.

brianrob (approved):

PR #1318, opened 2020-11-11
Remove HeapDump argument /dumpSerializedException

sharwell:
This change simplifies the transition to ClrMD 2.0, which does not support .NET Native.


brianrob (approved):
LGTM

PR #1324, opened 2020-11-11
Fix build warnings

sharwell:


src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
```diff
@@ -7,13 +7,7 @@
     <authors>Microsoft</authors>
     <owners>Microsoft</owners>
     <projectUrl>https://github.com/Microsoft/perfview</projectUrl>
-    <licenseUrl>https://github.com/Microsoft/perfview/blob/master/LICENSE.TXT</licenseUrl>
```
brianrob, line 10: I know that licenseurl has been deprecated, but I think we still need to have the license in the package.  What's the right way to do that?

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
```diff
@@ -35,7 +29,15 @@
     <tags>TraceEvent EventSource Microsoft ETW Event Tracing for Windows</tags>
 
     <dependencies>
-      <dependency id="System.Runtime.CompilerServices.Unsafe" version="4.5.2" />
```
brianrob, line 38: Can you clarify what this change does?

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
```diff
@@ -44,17 +46,17 @@
...
-    <file src="$OutDir$netstandard2.0\amd64\msvcp140.dll" target="lib\native\amd64\msvcp140.dll" />
-    <file src="$OutDir$netstandard2.0\amd64\vcruntime140.dll" target="lib\native\amd64\vcruntime140.dll" />
-    <file src="$OutDir$netstandard2.0\amd64\vcruntime140_1.dll" target="lib\native\amd64\vcruntime140_1.dll" />
-
-    <file src="$OutDir$netstandard2.0\x86\KernelTraceControl.dll" target="lib\native\x86\KernelTraceControl.dll" />
-    <file src="$OutDir$netstandard2.0\x86\KernelTraceControl.Win61.dll" target="lib\native\x86\KernelTraceControl.Win61.dll" />
-    <file src="$OutDir$netstandard2.0\x86\msdia140.dll" target="lib\native\x86\msdia140.dll" />
-    <file src="$OutDir$netstandard2.0\x86\msvcp140.dll" target="lib\native\x86\msvcp140.dll" />
-    <file src="$OutDir$netstandard2.0\x86\vcruntime140.dll" target="lib\native\x86\vcruntime140.dll" />
+    <file src="$OutDir$netstandard2.0\amd64\KernelTraceControl.dll" target="build\native\amd64\KernelTraceControl.dll" />
```
brianrob, line 49: I would not be surprised to find that this breaks people who are copying files from the NuGet package directory.  Does this fix anything other than the warning?  I'm inclined to just suppress or ignore this warning.

brianrob (commented):
@sharwell, thanks for this.  A few comments and questions below.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
sharwell: It's covered here, in the steps for "Manually edited nuspec file":
https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5128#solution

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
sharwell: It's covered here, in the steps for "Manually edited nuspec file":
https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5128#solution

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
sharwell: The files are still in the package, but now they will be in the expected location in the package. The files are copied via the custom build logic which was updated to match.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
sharwell: According to https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-a-license-expression-or-a-license-file, it's either/or. This suggests it is not expected that a package would need both.

brianrob: Oh, I see.  Since it's MIT, license type should work.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
brianrob: Cool, thanks for the pointer.

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec
brianrob: So I definitely see that the TraceEvent/PerfView build should be fine.  I'm just wondering if other folks might be copying files out without using TraceEvent.props.  I suspect that we should be fine here because when folks ingest the new package they'll experience this and have the opportunity to fix it.

brianrob (approved):

PR #1327, opened 2020-11-16
Added range unit tests to StackWindow

paulomorgado:
Added unit tests for when setting `StartText` and `EndText` on `StackWindow`.

`TestSetTimeRangeInStartTextBoxRuRu03Async` is skipped because there's a bug in the implementation for cultures with white spaces as number group separator ou decimal number separator.


paulomorgado:
This tests could be simplified if XUnit was upgraded to allow the use of `WpfTheoryAttribute`.

PR #1329, opened 2020-11-21
Handle ThreadUninitialized case in SampleProfilerThreadTimeComputer

josalem:
fixes dotnet/diagnostics#1572

See dotnet/diagnostics#1572 for details on the issue.  Summary: There are some nettrace files that don't produce any stacks in speedscope or chromium trace conversions.

I'm pretty sure I've identified the issue.  I'm not 100% sure if this the "correct" solution, but I know it fixes the issue.  The logic that I believe is at fault is here:

https://github.com/microsoft/perfview/blob/e7a025a9ac9bec1348beeb118b8eb986fcd42703/src/TraceEvent/Computers/SampleProfilerThreadTimeComputer.cs#L475-L555

This section in particular is the one to blame:
```csharp
            public void LogThreadStack(double timeRelativeMSec, StackSourceCallStackIndex stackIndex, TraceThread thread, SampleProfilerThreadTimeComputer computer, bool onCPU)
            {
                if (onCPU)
                {
                    if (ThreadRunning) // continue running 
                    {
                        AddCPUSample(timeRelativeMSec, thread, computer);
                    }
                    else if (ThreadBlocked) // unblocked
                    {
                        AddBlockTimeSample(timeRelativeMSec, thread, computer);
                        LastBlockStackRelativeMSec = -timeRelativeMSec;
                    }

                    LastCPUStackRelativeMSec = timeRelativeMSec;
                    LastCPUCallStack = stackIndex;
                }
                else
                {
                    if (ThreadBlocked) // continue blocking
                    {
                        AddBlockTimeSample(timeRelativeMSec, thread, computer);
                    }
                    else if (ThreadRunning) // blocked
                    {
                        AddCPUSample(timeRelativeMSec, thread, computer);
                    }

                    LastBlockStackRelativeMSec = timeRelativeMSec;
                    LastBlockCallStack = stackIndex;
                }
            }
```

If all the samples in a trace are `onCPU`, the thread is never put into the blocked state, so `ThreadRunning` never returns true:

```csharp
public bool ThreadRunning { get { return LastBlockStackRelativeMSec < 0 && !ThreadDead; } }
```

If `ThreadRunning` is never true, no samples are put into the stack source because it never calls `AddCPUSample` or `AddBlockTimeSample`.  This results in no stacks in the output.

I fixed this by simply changing the first check to be `if (ThreadRunning || ThreadUninitialized)` so that the thread state starts correctly if all the samples are managed code.

CC @sywhang @adamsitnik


josalem:
I think there may be more to the problem than this PR is addressing.  This PR addresses the fact that a trace that contains samples **_only_** marked as managed will not be computed correctly.  There may be a lower level issue surrounding whether a stack is marked as managed or not.  Refer to the issue mentioned in the PR text for more details.

brianrob:
@josalem, I suspect you're right that this might not be the right solution, though it does work.  Can you tell from an individual event that the thread should be marked as OnCPU?  Usually, the right answer is when you see one of these, to update the data structure to make sure that the thread is considered OnCPU and running.  Is that do-able, or are we operating with not enough information here to do this?

josalem:
> Can you tell from an individual event that the thread should be marked as OnCPU?

Yes, you should be able to.  Every sample event has a payload that indicates the GC Mode when the sample was taken.  This is used to interpret whether a stack is "managed" (coop) or "external" (preemptive).  `onCPU`'s value is determined by whether the sample event is marked as "managed" (`true`) or "external" (`false`).

The main issue with the current logic is that `ThreadRunning` and `ThreadBlocked` are both determined using the value of `LastBlockStackRelativeMSec`.  Those two values are what cause `AddBlockTimeSample` and `AddCPUSample` to be called.  If all samples are marked as `onCPU` (managed samples), then `LastBlockStackRelativeMSec` will always be `0` (uninitialized) causing `ThreadRunning` and `ThreadBlocked` to be `false`.

https://github.com/microsoft/perfview/blob/e75e3856d53c123c7f7ba202579d44ba4d91a7b8/src/TraceEvent/Computers/SampleProfilerThreadTimeComputer.cs#L544-L547

This results in all samples _before_ an `onCPU == false` sample being thrown out since neither `Add*Sample` method will be called.

A more correct solution might be something like this:

```csharp
            public void LogThreadStack(double timeRelativeMSec, StackSourceCallStackIndex stackIndex, TraceThread thread, SampleProfilerThreadTimeComputer computer, bool onCPU)
            {
                if (onCPU)
                {
					if (ThreadUninitialized) // first sample is onCPU
					{
						AddCPUSample(timeRelativeMSec, thread, computer);
						LastBlockStackRelativeMSec = -1; // set ThreadRunning to true
					}
                    if (ThreadRunning) // continue running 
                    {
                        AddCPUSample(timeRelativeMSec, thread, computer);
                    }
                    else if (ThreadBlocked) // unblocked
                    {
                        AddBlockTimeSample(timeRelativeMSec, thread, computer);
                        LastBlockStackRelativeMSec = -timeRelativeMSec;
                    }

                    LastCPUStackRelativeMSec = timeRelativeMSec;
                    LastCPUCallStack = stackIndex;
                }
                else
                {
                    if (ThreadBlocked || Uninitialized) // continue blocking
                    {
                        AddBlockTimeSample(timeRelativeMSec, thread, computer);
                    }
                    else if (ThreadRunning) // blocked
                    {
                        AddCPUSample(timeRelativeMSec, thread, computer);
                    }

                    LastBlockStackRelativeMSec = timeRelativeMSec;
                    LastBlockCallStack = stackIndex;
                }
            }
```

to make sure the state machine covers the uninitialized state.  We simply assume that the starting state of the thread is "running" if the first event is `onCPU == true` and "blocked" if its `onCPU == false`.

josalem:
All of the samples in a trace collected on Windows will be marked as `external`, while they will be correctly split between `managed` and `external` on non-Windows platforms.  See dotnet/runtime#45179 for more details on the underlying runtime issue.

josalem:
Added two test cases using traces I generated locally that fail under the old logic.

brianrob (approved):
@josalem, thanks for continuing to look for the right fix.  This looks great!

adamsitnik:
@josalem thank you for fixing the issue and adding more tests!

PR #1330, opened 2020-11-23
TraceEvent symbol handler source file improvements

gregg-miskelly:
This PR exposes some additional information from the SourceFile classes. This also fixes some issues with source file handling

### New exposed APIs:
- In addition to the URL property, there is now a new 'GetSourceLinkInfo' method that exposes the URL and the relative file path
- The checksum value/algorithm is exposed
- NativeSymbolModule now implements IDisposable using IDiaSession3.dispose

### Additional fixes:
- The old code didn't calculate checksum's in Windows PDBs for managed code correctly - it only expected the 'C++ way' of encoding checksums, but that isn't how diasymreader.dll does it.
- The old code serched for sourcelink info in a native PDB via a very hacky method. This switches to doing things the right way. Along with this, I had to tweak the internal interface used to return source link data into something that could return multiple strings instead of just a single one.

### Testing:
Verified the source file info we get from:
1. A default C# .NET Framework console app (Windows PDB case)
2. A C# .NET Framework project with Source Link (Windows PDB case)
3. Newtonsoft Json's PDB (Portable PDB with Source Link info)
4. C++ project with Source Link
5. C++ project with srcsrv

### DIFF
--- a/src/TraceEvent/Symbols/NativeSymbolModule.cs
+++ b/src/TraceEvent/Symbols/NativeSymbolModule.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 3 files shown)


src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -1896,13 +1913,63 @@ public virtual string GetSourceFile(bool requireChecksumMatch = false)
...
+                {
+                    return "SHA1";
+                }
+                else if (_hashAlgorithm is MD5)
+                {
+                    return "MD5";
+                }
+                else
+                {
+                    Debug.Fail("Missing case in get_ChecskumAlgorithm");
```
brianrob, line 1942: Nit: Spelling --> get_ChecksumAlgorithm.

src/TraceEvent/Symbols/NativeSymbolModule.cs
```diff
@@ -620,7 +615,20 @@ internal unsafe MicrosoftPdbSourceFile(NativeSymbolModule module, IDiaSourceFile
             {
                 BuildTimeFilePath = sourceFile.fileName;
 
-                // 0 No checksum present.
+                if (sourceFile.checksumType == 0)
+                {
+                    // If the checksum type is zero, this means either this is a non-C++ PDB, or there is no checksum info
+                    TryInitializeManagedChecskum(module);
```
brianrob, line 621: Nit: Spelling --> TryInitializeManagedChecksum

brianrob (approved):

PR #1359, opened 2021-02-03
TraceParserGen: Remove internal keyword when not requested.

dorian-apanel-intel:
Partial revert of 46a2f3a73fd91fc2b0ae30440666589a9ca72212

I think that @billwert introduced issue here.
It looks like internalOpt should be used to insert "internal" keyword instead of hardcoding it in string.
If "Internal" would be set, the it would generate double "internal" in function header.

### DIFF
--- a/src/TraceParserGen/TraceParserGen.cs
+++ b/src/TraceParserGen/TraceParserGen.cs
@@ -289,7 +289,7 @@ internal class TraceParserGen
 
         output.WriteLine();
         output.WriteLine("        static private volatile TraceEvent[] s_templates;");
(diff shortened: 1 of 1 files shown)


dorian-apanel-intel:
@brianrob Hi,
Could you take a look? I have same issues failing on fresh main:
image
I guess this is not related to my PR

brianrob:
I have seen this intermittently in CI, but have not been able to repro it locally to debug.  If it's something that you are able to reproduce, I'd love it if you could attempt to debug.

Either way, I have re-triggered the run, and it passed.

brianrob (approved):

Issue #1364, opened 2021-02-09
Add a command-line generator from the collect/run GUI


paulomorgado:
As per @brianrob, this is already implemented here: https://github.com/microsoft/perfview/blob/3e09e663969d8679c6d29a28807d736fd1d2a0e7/src/PerfView/CommandProcessor.cs#L2756

Questions?

1. Should that include the `AcceptEULA` and `NoGUI` arguments?
2. Should that include the path to the currently executing PerfView?

brianrob:
To answer your questions: It is reasonable to include `NoGui`, but I would be hesitant to include `AcceptEULA` because this is a specific statement that each user should make.  I think including the path is really about what the use case is.  Are you going to script something locally, or is this really about handing out command lines to others?  If the former, than you'd want to include it.  For the latter, I think you'd not want to include it.  My gut tells me not to include it, but I'm persuadable.

PR #1374, opened 2021-02-16
Add a command-line option to RunCommandDialog.

paulomorgado:
Addresses #1364

### DIFF
--- a/src/PerfView/CommandProcessor.cs
+++ b/src/PerfView/CommandProcessor.cs
@@ -2753,7 +2753,7 @@ namespace PerfView
 #endif
         }
 
-        private static string ParsedArgsAsString(string command, CommandLineArgs parsedArgs)
+        public static string ParsedArgsAsString(string command, CommandLineArgs parsedArgs)
         {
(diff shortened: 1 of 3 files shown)


brianrob:
@paulomorgado, overall, I think this looks good.  One thing: Can you add "PerfView.exe" to the beginning of the command line.  Right now, the value that gets copied to the clipboard doesn't include the executable.

paulomorgado:
That's on purpose.

What should we do?
* Add the full path to the current executable (**PerfView** vs. **PerfView64**)?
* Add just the file name of the current executable (**PerfView** vs. **PerfView64**)?
* Add the **.exe** extension? Or not?
* If the user is in the same path as the **PerfView** executable on a **PowerShell** prompt, not prefixing with **.\** will fail to find **PerfView**.

brianrob:
I would say that it's fine to just stick PerfView.exe at the beginning and call it a day.

paulomorgado:
Done!

brianrob (approved):

PR #1379, opened 2021-03-04
Show POH PerHeapHistoryData in Raw XML view

cshung:
This change allows us to view the POH data in the raw XML view.

@brianrob
@Maoni0

### DIFF
--- a/src/PerfView/GcStats.cs
+++ b/src/PerfView/GcStats.cs
@@ -543,14 +543,14 @@ namespace Stats
                     writer.WriteLine(">");
 
                     var sb = new System.Text.StringBuilder();
-                    for (var gens = Gens.Gen0; gens <= Gens.GenLargeObj; gens++)
+                    for (int gens = 0; gens < perHeapHistory.GenData.Length; gens++)
                     {
                         sb.Clear();
                         sb.Append("        ");
-                        writer.Write(perHeapHistory.GenData[(int)gens].ToXml(gens, sb).AppendLine().ToString());
+                        writer.Write(perHeapHistory.GenData[(int)gens].ToXml((Gens)gens, sb).AppendLine().ToString());
                     }
 
-                    writer.Write("      </PerHeapHistory>");
+                    writer.WriteLine("      </PerHeapHistory>");
                     HeapNum++;
                 }
                 writer.WriteLine("      </PerHeapHistories>");
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -2458,7 +2458,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
         /// Per generation view of user allocated data
         /// </summary>
         [Obsolete("This is experimental, you should not use it yet for non-experimental purposes.")]
-        public double[] UserAllocated = new double[(int)Gens.Gen0After];
+        public double[] UserAllocated = new double[(int)Gens.GenCount];
         /// <summary>
         /// Heap size before gc (mb)
         /// </summary>
@@ -2468,7 +2468,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
         /// Per generation view of heap sizes before GC (mb)
         /// </summary>
         [Obsolete("This is experimental, you should not use it yet for non-experimental purposes.")]
-        public double[] GenSizeBeforeMB = new double[(int)Gens.Gen0After];
+        public double[] GenSizeBeforeMB = new double[(int)Gens.GenCount];
         /// <summary>
         /// This represents the percentage time spent paused for this GC since the last GC completed. 
         /// </summary>
@@ -4595,6 +4595,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
             TraceGC _event = GetLastGC(proc);
             if (_event != null)
             {
+                int numGenerations = data.HasCount ? data.Count : 4;
                 var hist = new GCPerHeapHistory()
                 {
                     FreeListAllocated = (data.HasFreeListAllocated) ? data.FreeListAllocated : -1,
@@ -4604,7 +4605,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
                     MemoryPressure = (data.HasMemoryPressure) ? data.MemoryPressure : -1,
                     HasMemoryPressure = data.HasMemoryPressure,
                     VersionRecognized = data.VersionRecognized,
-                    GenData = new GCPerHeapHistoryGenData[(int)Gens.GenLargeObj + 1],
+                    GenData = new GCPerHeapHistoryGenData[numGenerations],
                     CondemnReasons0 = data.CondemnReasons0,
                     CondemnReasons1 = (data.HasCondemnReasons1) ? data.CondemnReasons1 : -1,
                     HasCondemnReasons1 = data.HasCondemnReasons1,
@@ -4615,7 +4616,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
                     Version = data.Version
                 };
 
-                for (Gens GenIndex = Gens.Gen0; GenIndex <= Gens.GenLargeObj; GenIndex++)
+                for (Gens GenIndex = Gens.Gen0; GenIndex < (Gens)numGenerations; GenIndex++)
                 {
                     hist.GenData[(int)GenIndex] = data.GenData(GenIndex);
                 }
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -2458,7 +2458,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
         /// Per generation view of user allocated data
         /// </summary>
         [Obsolete("This is experimental, you should not use it yet for non-experimental purposes.")]
-        public double[] UserAllocated = new double[(int)Gens.GenCount];
+        public double[] UserAllocated = new double[(int)Gens.MaxGenCount];
         /// <summary>
         /// Heap size before gc (mb)
         /// </summary>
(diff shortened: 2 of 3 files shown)


src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -4941,8 +4941,8 @@ public enum Gens
         Gen1,
         Gen2,
         GenLargeObj,
-        Gen0After,
         GenPinObj,
+        GenCount,
```
Maoni0, line 4945: we do need to consider backward compatibility here - in 3.1 and priori there were 4 elements in this array instead of 5, we still need it to work with those older versions.

src/PerfView/GcStats.cs
```diff
@@ -543,14 +543,14 @@ public static void ToXmlAttribs(TextWriter writer, TraceProcess stats, TraceLoad
...
+                    for (int gens = 0; gens < perHeapHistory.GenData.Length; gens++)
                     {
                         sb.Clear();
                         sb.Append("        ");
-                        writer.Write(perHeapHistory.GenData[(int)gens].ToXml(gens, sb).AppendLine().ToString());
+                        writer.Write(perHeapHistory.GenData[(int)gens].ToXml((Gens)gens, sb).AppendLine().ToString());
                     }
 
-                    writer.Write("      </PerHeapHistory>");
+                    writer.WriteLine("      </PerHeapHistory>");
```
Maoni0, line 553: any reason to change this? it was more compact before

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -4595,6 +4595,7 @@ internal static void ProcessPerHeapHistory(TraceLoadedDotNetRuntime proc, GCPerH
             TraceGC _event = GetLastGC(proc);
             if (_event != null)
             {
+                int numGenerations = data.HasCount ? data.Count : 4;
```
Maoni0, line 4598: I would make this a property returned from the parser (ie in ClrTraceEventParser.cs) 'cause this looks quite internal to the event layout. in fact I don't think HasCount should even be a public property.

src/PerfView/GcStats.cs
cshung: I think it looks better that way, but I won't mind changing it back.

src/TraceEvent/Computers/TraceManagedProcess.cs
cshung: Sounds like a good idea.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
cshung: I am wondering about the backward compatibility requirements here. 

If we keep `Gen0After` to be 4, and somehow the value is used as the length to construct an array to put things in, and then the trace contains a POH, then the code will end up having an `ArrayOutOfBoundsException`. This change will probably help in the sense that when the user upgrades the TraceEvent package, they will get a compile error, and notice there is a need to change things for POH.

The ideal case is probably to avoid this problem in the first place by not having such an enum entry at all.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -5727,7 +5727,7 @@ private long[] GetIntPtrArray(int offset, int count)
             return arr;
         }
 
-        private const int maxGenData = (int)Gens.Gen0After + 1;
+        private const int maxGenData = (int)Gens.GenCount;
```
cshung, line 5730: I need to double-check this code path, something doesn't sound right here that we are unconditionally setting maxGenData to 5 but then use that to compute offset for parsing trace data.

cshung: Turn out it wasn't wrong. `Gen0After` was 4, and therefore the new `GenCount` is the old `Gen0After + 1`.
To clarify, I just hardcoded it to 5. This number is used only for parsing `GCHeapHistory` prior to version 3, so it has to be 5.

Maoni0: it actually does have something to do with max gen count, as in it was (max gen count + 1) since we had an extra one for gen0. could you please add a comment there to explain why this is 5? and make sure to mention that this is only used to parse older traces (as the newer versions no longer use it anymore.

Maoni0:
other than what I mentioned in [this comment](https://github.com/microsoft/perfview/pull/1379#discussion_r612801448) it LGTM!

Maoni0 (approved):
LGTM

brianrob (approved):

brandonh-msft:
@cshung & @brianrob 

This PR performed the following breaking change:
image

We have code in our codebase like this:

```csharp
        public TraceGCEx(TraceGC gc)
        {
            Number = gc.Number;
            RelativeTimestamp = gc.StartRelativeMSec;
            Generation = gc.Generation;
            Reason = gc.Reason.ToString();
            Type = gc.Type.ToString();
            PauseTime = gc.PauseDurationMSec;
            PinnedObjectsSize = (double)gc.GetPinnedObjectSizes() / (1024 * 1024);
            PinnedObjectPercent = gc.GetPinnedObjectPercentage() >= 0 ? (double)gc.GetPinnedObjectPercentage() : 0;
            SuspensionTime = gc.SuspendDurationMSec;
            Issues = new Dictionary<string, double>();

            // per heap calculations
            HeapSize = HeapFragmentation = 0;
            SurvivalPercent = new Dictionary<string, double>();
            
            for (int gen = (int)Gens.Gen0; gen < (int)Gens.Gen0After; gen++)
            {
                HeapSize += gc.HeapStats != null ? gc.GenSizeAfterMB((Gens)gen) : 0;
                HeapFragmentation += gc.GenFragmentationMB((Gens)gen);
                var surv = gc.SurvivalPercent((Gens)gen);
                if (!Double.IsNaN(surv))
                {
                    SurvivalPercent.Add(((Gens)gen).ToString(), surv);
                }
            }
```

What is the recommended way forward??

cshung:
> @cshung & @brianrob
> 
> This PR performed the following breaking change: image
> 
> We have code in our codebase like this:
> 
> ```cs
>         public TraceGCEx(TraceGC gc)
>         {
>             Number = gc.Number;
>             RelativeTimestamp = gc.StartRelativeMSec;
>             Generation = gc.Generation;
>             Reason = gc.Reason.ToString();
>             Type = gc.Type.ToString();
>             PauseTime = gc.PauseDurationMSec;
>             PinnedObjectsSize = (double)gc.GetPinnedObjectSizes() / (1024 * 1024);
>             PinnedObjectPercent = gc.GetPinnedObjectPercentage() >= 0 ? (double)gc.GetPinnedObjectPercentage() : 0;
>             SuspensionTime = gc.SuspendDurationMSec;
>             Issues = new Dictionary<string, double>();
> 
>             // per heap calculations
>             HeapSize = HeapFragmentation = 0;
>             SurvivalPercent = new Dictionary<string, double>();
>             
>             for (int gen = (int)Gens.Gen0; gen < (int)Gens.Gen0After; gen++)
>             {
>                 HeapSize += gc.HeapStats != null ? gc.GenSizeAfterMB((Gens)gen) : 0;
>                 HeapFragmentation += gc.GenFragmentationMB((Gens)gen);
>                 var surv = gc.SurvivalPercent((Gens)gen);
>                 if (!Double.IsNaN(surv))
>                 {
>                     SurvivalPercent.Add(((Gens)gen).ToString(), surv);
>                 }
>             }
> ```
> 
> What is the recommended way forward? If I change to using `.MaxGenCount` we get Index OOBs on our traces when processing and choosing `.GenPinObj` just because it happens to be the same int value as what we were using previously seems awfully dirty.

Fundamentally, the runtime introduced a new "generation" (technically, the pinned object heap is not a new generation, but it is convenient to represent it that way) in 5.0, so that bound needs to be dynamically adjusted depending on whether or not the trace was captured before 5.0 or not. 

To make this task easier, there is a new property named `NumGenerations` in this PR, can you use that property instead of hard coding?

brandonh-msft:
Thanks for that explanation, @cshung 

> To make this task easier, there is a new property named `NumGenerations` in this PR, can you use that property instead of hard coding?

I considered that, however even after updating to v3.0.4 the object we're using doesn't have that property :(
image

Any other ideas? I'm inheriting this code so not quite sure what the _right_ thing is ;)

cshung:
> Any other ideas? I'm inheriting this code so not quite sure what the right thing is ;)

I tried to look into this, and it appears to me there are some disconnects.

The new property `NumGenerations` is added on the `GCPerHeapHistoryTraceData`. That class seems to be mostly concerned with parsing trace data in binary format. The new property is used mostly because the trace format changed to have variable size content because we have a variable number of generations, and therefore the parsing need to deal with that.

But it appears to me all you have is a `TraceGC` object instead. I assumed `TraceGC` is being built as part of the parsing process, but I am unfamiliar with those structures, in particular, I am not sure how you can reach into the corresponding `GCPerHeapHistoryTraceData` object that was used during the parsing process to get access to that. This will need more study.

You mentioned that there is an index out of bound error when you were experimenting with the code. Do you happen to know which array are you accessing? Can you reach into that array and read the length of that array? Or surface the length of that array as a property of the `TraceGC` class?

brandonh-msft:
Yeah the blow-up happens in the loop I posted:
```csharp
            for (int gen = (int)Gens.Gen0; gen < (int)Gens.Gen0After; gen++)
            {
                HeapSize += gc.HeapStats != null ? gc.GenSizeAfterMB((Gens)gen) : 0;
                HeapFragmentation += gc.GenFragmentationMB((Gens)gen);
                var surv = gc.SurvivalPercent((Gens)gen);
```

Do you need me to execute it and get details on what `gc` looks like? It doesn't appear these objects/methods are proper arrays but rather methods which, in turn, index into an array.

cshung:
> Yeah the blow-up happens in the loop I posted:
> 
> ```cs
>             for (int gen = (int)Gens.Gen0; gen < (int)Gens.Gen0After; gen++)
>             {
>                 HeapSize += gc.HeapStats != null ? gc.GenSizeAfterMB((Gens)gen) : 0;
>                 HeapFragmentation += gc.GenFragmentationMB((Gens)gen);
>                 var surv = gc.SurvivalPercent((Gens)gen);
> ```
> 
> Do you need me to execute it and get details on what `gc` looks like? It doesn't appear these objects/methods are proper arrays but rather methods which, in turn, index into an array.

Yes, a stack trace with the TraceEvent bits included would be nice.

brandonh-msft:
image

does that shed any light on things?

cshung:
It looks like you have a GenSizeBeforeMB array that you could access, and the dimension is 4, which is probably the number you are looking for. What would happen if you replace `Gen.Gen0After` by the length of that array?

brandonh-msft:
Yeah, no good.

image

any other ideas?

cshung:
This is not going to be very productive to chat here once a day or two, would you mind finding me on teams?

brandonh-msft:
ThumbUpThumbUpEatingGIF

PR #1380, opened 2021-03-05
Explicitly send SIGINT to perf

brianrob:
Enables scripting against perfcollect, which is especially important in our performance labs.

cc: @sebastienros


sebastienros:
I can test this one before you merge.

brianrob:
Closing in favor of #1381.

PR #1389, opened 2021-03-17
Make ASP.NET Core Events Opt-In

brianrob:
- Remove collection of `Microsoft-AspNetCore-Hosting` and `Microsoft-ApplicationInsights-Core`.
 - Make ASP.NET Core DiagnosticSource events opt-in via `/AspNetCoreStartStop`.
 - Automatically enable ASP.NET Core DiagnosticSource events when `/ThreadTime` is specified unless `/AspNetCoreStartStop:false` is specified to explicitly opt-out.

cc: @stephentoub, @davidfowl

### DIFF
--- a/src/PerfView/CommandLineArgs.cs
+++ b/src/PerfView/CommandLineArgs.cs
@@ -125,6 +125,7 @@ namespace PerfView
         public bool DotNetCallsSampled;     // Sampling of .NET calls.  
         public bool DisableInlining;        // Force inlining to be disabled. (useful for DotNetCalls).  
         public bool JITInlining;            // Turn on logging of successful and failed JIT inlining
+        public bool? AspNetCoreStartStop;   // Turn on AspNetCore request start/stop events
         public int OSHeapProcess;           // Turn on OS Heap tracing for the process with the given process ID.
         public string OSHeapExe;            // Turn on OS heap tracing for any process with the given EXE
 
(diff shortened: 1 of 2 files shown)


src/PerfView/CommandLineArgs.cs
```diff
@@ -510,6 +511,7 @@ private void SetupCommandLine(CommandLineParser parser)
             parser.DefineOptionalQualifier("DotNetCallsSampled", ref DotNetCallsSampled, "Turns on per-call .NET profiling, sampling types in a smart way to keep overhead low.");
             parser.DefineOptionalQualifier("DisableInlining", ref DisableInlining, "Turns off inlining (but only affects processes that start after trace start.");
             parser.DefineOptionalQualifier("JITInlining", ref JITInlining, "Turns on logging of successful and failed JIT inlining attempts.");
+            parser.DefineOptionalQualifier("AspNetCoreStartStop", ref AspNetCoreStartStop, "Turns on logging of ASP.NET Core request start/stop events.");
```
davidfowl, line 514: nit: AspNetCoreRequests

davidfowl:
A couple of questions:
- Does this change turn on the ASP.NET Core diagnostic source events? I don't understand where that happens. I only see the HttpClient events.
- We changed the name of the ASP.NET Core event source in hosting (https://github.com/dotnet/aspnetcore/blob/dec7382dcf1bf1eab4863df27053525a769c9cb5/src/Hosting/Hosting/src/Internal/HostingEventSource.cs#L25). It's Microsoft.AspNetCore.Hosting
- We have a KestrelEventSource as well (https://github.com/dotnet/aspnetcore/blob/dec7382dcf1bf1eab4863df27053525a769c9cb5/src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelEventSource.cs#L15). That has much more details start stop information.

We have Start/Stop events that take advantage of ETW/EventSource activity ids for correlation. The diagnostic source events I assume are required for diagnostic source/open telemetry/activity like operations. Which ones do we care about more?

brianrob:
Per conversation with @davidfowl, I want to investigate these other EventSources.

PR #1391, opened 2021-03-18
Fix sample JSON and XML in release notes

pieter-venter:
Documentation only PR.

The samples provided in the release notes are invalid (and inconvenient due to `li`):

- The JSON is invalid - missing a `]`
- The XML is invalid - missing a closing `</Sample>` tag
- XML example text incorrectly names sample 2 as sample 1
- Changed the layout to `<pre>` so that you can copy/paste it

**For your convenience:**
JSON before:
json-before

JSON after:
json-after

XML before:
xml-before

XML after:
xml-after


brianrob (approved):
LGTM.  Thanks for fixing this @pieter-venter.

PR #1402, opened 2021-04-05
Enumerate objects to populate caches

sharwell:


brianrob:
@sharwell, I'm assuming that this is a performance improvement.  Can you share what scenario this benefits?

sharwell:
The first time objects in the heap are enumerated, ClrMD populates caches with type information. By populating the caches prior to allocating the large dictionaries, we reduce the likelihood of transient allocation failures that could occur from an OOM while attempting to access type information for a type not already in the cache. This exception is handled internally by silently ignoring the object, so it's easy to unintentionally review incomplete heap information.

brianrob (approved):

PR #1405, opened 2021-04-06
Disable ClrStack/Walk Events During Rundown

brianrob:
ClrStack/Walk events are unintentionally enabled during rundown if any of the rundown suppression arguments are used.  This can result in a large number of stacks that can't be matched to CLR events and overall slower conversion of an ETL file to ETLX.

An example configuration where this occurs is with `/GCCollectOnly` traces where the fully expanded commandline includes `/NoRundown`:
```
PerfView /RestartingToElevelate:collect /DataFile:PerfViewGCCollectOnly.etl /BufferSizeMB:256 /StackCompression /KernelEvents:Process,ImageLoad /ClrEventLevel:Informational /ClrEvents:GC /NoGui /SessionName:PerfViewGCSession /NoRundown /GCCollectOnly Collect
```

These events are not needed during rundown, but are enabled because rundown for V2.0 runtimes occurs via the public provider instead of the rundown provider.  The fix is to switch the non-existent starting rundown keywords to be a value that doesn't intersect with anything in the public provider or the rundown provider.

cc: @Maoni0

### DIFF
--- a/src/PerfView/CommandProcessor.cs
+++ b/src/PerfView/CommandProcessor.cs
@@ -3472,9 +3472,12 @@ namespace PerfView
                     if (parsedArgs.ClrEvents != ClrTraceEventParser.Keywords.None)
                     {
                         // Always enable minimal rundown, which ensures that we get the runtime start event.
-                        // We use the keyword 0x40000000 which does not match any valid keyword in the rundown provider.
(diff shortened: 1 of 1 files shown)


Maoni0:
thanks for fixing this!

Issue #1351, opened 2021-01-27
TraceParserGen: xxxTaskGuid not defined


dorian-apanel-intel:
I'm trying to compile cs code generated by TraceParserGen with custom manifest.

First problem I see is that guids passed to RegisterTemplate() are not defied.
Then I found this comment in code
```
// as of today, the generated code won't compile because the task GUID is not defined
// TODO: define the xxxTaskGuid based on eventGUID attribute of <task> elements of the .man file
```

Is it going to be fixed?
It happens both if I define eventGUID attribute or not.

brianrob:
@dorian-apanel-intel, this can be fixed, but wasn't an issue that I was aware of.  Would you be up for submitting a PR for this?

dorian-apanel-intel:
Hi, I fixed it for myself.
I'm going to propose a PR soon.

limulidae:
Is this fixed?

brianrob:
I don't believe that this has been fixed.  Fixes to TraceParserGen have been held up behind a goal to create a test-bed for the tool first, as each time we take a change, we tend to break something.

brianrob:
Fixed via #2308.

PR #1406, opened 2021-04-06
Fix TraceParserGen to generate meaningful parser for WPF

oatkins:
This attempts to resolve https://github.com/microsoft/perfview/issues/1351 and supports creating a parser for the Microsoft-Windows-WPF provider from the manifest at C:\Windows\Microsoft.NET\Framework\v4.0.30319\WPF\wpf-etw.man.

See also https://github.com/microsoft/perfview/issues/554.

Addresses various issues with the generated provider:

- Enum should not be normalized it if's already PascalCase, so the normalizing logic is skipped if no "_" is present in a field name.
- `RegisterTemplate()` seems not to exist; `source.RegisterEventTemplate` is apparently what was intended.
- Task GUID constants were previously referenced but not generated.
- Template class names were incorrectly referenced.

### DIFF
--- a/src/TraceParserGen/TraceParserGen.cs
+++ b/src/TraceParserGen/TraceParserGen.cs
@@ -134,6 +134,9 @@ internal class TraceParserGen
 
         // *********** GENERATE EVENTS ********** //
         GenerateEvents(output);
+
+        GenerateTaskGuids(output);
+
         output.WriteLine("");
         output.WriteLine("        #region private");
         output.WriteLine("        protected override string GetProviderName() { return ProviderName; }");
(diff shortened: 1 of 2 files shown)


oatkins:
See also #950 - need to make sure the problems resolved there are not reintroduced here.

src/TraceParserGen/TraceParserGen.cs
```diff
@@ -276,8 +304,10 @@ private void GenerateTemplateDefs(TextWriter output)
                 state = ", state";
             }
 
-            output.WriteLine("            return new {0}(action, {1}, {2}, \"{3}\", Guid.Empty, {4}, \"{5}\", ProviderGuid, ProviderName {6});",
-                templateClassName, evnt.Id, evnt.Task, TraceParserGen.ToCSharpName(evnt.TaskName), evnt.Opcode, TraceParserGen.ToCSharpName(evnt.OpcodeName), state);
+            var taskGuid = string.IsNullOrEmpty(evnt.TaskName) ? "Guid.Empty" : ToCSharpName(evnt.TaskName) + "TaskGuid";
```
dorian-apanel-intel, line 307: Shouldn't it check m_taskGuids?
If there is no "eventGUID" attributes in parsed manifest, I think it should return Guid.Empty.

brianrob:

PR #1408, opened 2021-04-13
Read Max ETW Loggers from Registry Key

pamcnult:
Reading Max ETW Loggers from Registry Key

### DIFF
--- a/src/TraceEvent/TraceEventSession.cs
+++ b/src/TraceEvent/TraceEventSession.cs
@@ -1389,15 +1389,19 @@ namespace Microsoft.Diagnostics.Tracing.Session
                 {
                     using (RegistryKey key = Registry.LocalMachine.OpenSubKey(MaxEtwRegistryKey))
                     {
-                        if (key != null && key.GetValue(MaxEtwPropertyName) != null)
+                        if (key != null)
                         {
-                            if (int.TryParse(key.GetValue(MaxEtwPropertyName).ToString(), out int propertyValue))
+                            var property = key.GetValue(MaxEtwPropertyName);
+                            if (property != null)
                             {
-                                // Ensure registry was set within permissable range as defined by
-                                // https://docs.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-starttracew
-                                if (propertyValue >= 32 && propertyValue <= 256)
+                                if (int.TryParse(property.ToString(), out int propertyValue))
                                 {
-                                    MaxEtwLoggers = propertyValue;
+                                    // Ensure registry was set within permissable range as defined by
+                                    // https://docs.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-starttracew
+                                    if (propertyValue >= 32 && propertyValue <= 256)
+                                    {
+                                        MaxEtwLoggers = propertyValue;
+                                    }
                                 }
                             }
                         }
--- a/src/TraceEvent/TraceEventSession.cs
+++ b/src/TraceEvent/TraceEventSession.cs
@@ -1338,7 +1338,7 @@ namespace Microsoft.Diagnostics.Tracing.Session
         /// <returns>A enumeration of strings, each of which is a name of a session</returns>
         public static unsafe List<string> GetActiveSessionNames()
         {
-            const int MAX_SESSIONS = 64;
+            int MAX_SESSIONS = GetETWMaxLoggers();
             int sizeOfProperties = sizeof(TraceEventNativeMethods.EVENT_TRACE_PROPERTIES) +
                                    sizeof(char) * TraceEventSession.MaxNameSize +     // For log moduleFile name 
                                    sizeof(char) * TraceEventSession.MaxNameSize;      // For session name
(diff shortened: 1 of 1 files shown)


src/TraceEvent/TraceEventSession.cs
```diff
@@ -1368,6 +1368,35 @@ public static unsafe List<string> GetActiveSessionNames()
             return activeTraceNames;
         }
 
+        /// <summary>
+        /// Get the maximum number of ETW loggers supported by the current machine
+        /// </summary>
+        /// <returns>The maximum number of supported ETW loggers</returns>
+        private static int GetETWMaxLoggers()
```
brianrob, line 1375: Can you cache the value in a static field?  When the registry key changes, the machine requires a reboot to apply the change, so it is safe to cache this for the life of the process.

src/TraceEvent/TraceEventSession.cs
```diff
@@ -1368,6 +1368,35 @@ public static unsafe List<string> GetActiveSessionNames()
...
+            const string MaxEtwPropertyName = "EtwMaxLoggers";
+            const int DefaultMaxETWLoggers = 64;
+
+            try
+            {
+                using (RegistryKey key = Registry.LocalMachine.OpenSubKey(MaxEtwRegistryKey))
+                {
+                    if (key != null)
+                    {
+                        if (int.TryParse(key.GetValue(MaxEtwPropertyName).ToString(), out int maxEtwLoggers))
```
brianrob, line 1387: The registry key is not present by default.  Can you check to see if the key and value are present first before trying to use them?  As is, I would expect this to throw in the common case when calling `GetValue().ToString()`, since the result will likely be `null`.

brianrob (changes requested):
@pamcnult thanks for fixing this.  A couple of requests below.

src/TraceEvent/TraceEventSession.cs
pamcnult: Good call. I have cached the value. 
I also added validation on permissible values defined https://docs.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-starttracew

src/TraceEvent/TraceEventSession.cs
pamcnult: I have added a check to see if the value is present alongside the key before using.

src/TraceEvent/TraceEventSession.cs
```diff
@@ -1368,6 +1368,50 @@ public static unsafe List<string> GetActiveSessionNames()
...
+                                    MaxEtwLoggers = propertyValue;
+                                }
+                            }
+                        }
+                    }
+                }
+                catch (Exception) { }
+
+                // If the value does not exist or cannot be read from the registry, return the default value
+                MaxEtwLoggers = MaxEtwLoggers ?? DefaultMaxETWLoggers;
```
brianrob, line 1409: Can we simplify this and just set `MaxEtwLoggers == 64` by default?  Then you don't need to worry about checking to see if `MaxEtwLoggers` has been set.  Also, you can change the type of `MaxEtwLoggers` to `int` so you don't have to cast it below.

src/TraceEvent/TraceEventSession.cs
```diff
@@ -1368,6 +1368,50 @@ public static unsafe List<string> GetActiveSessionNames()
             return activeTraceNames;
         }
 
+        /// <summary>
+        /// Maximum Number of MaxEtwLoggers the system supports
+        /// </summary>
+        private static int? MaxEtwLoggers = null;
```
brianrob, line 1374: Mentioned in another comment that we can simplify the code by setting `MaxEtwLoggers == 64` here.

brianrob (changes requested):
@pamcnult, thanks for the updates.  I'd like to simplify this just a bit, but I think we're getting close.

src/TraceEvent/TraceEventSession.cs
pamcnult: Don't we want to check if MaxEtwLoggers has been set?  This was done in order to achieve caching of MaxEtwLoggers.
If we have no value, check and set the value to the registry key or the default if the registry key does not exist.

If we set MaxEtwLoggers to 64 by default, in the instance the registry key does not exist, each time we call the function, we would have no way to know whether this was our first time checking, or if we have already checked and had to set the default value.

src/TraceEvent/TraceEventSession.cs
```diff
@@ -1368,6 +1368,50 @@ public static unsafe List<string> GetActiveSessionNames()
...
+            const string MaxEtwPropertyName = "EtwMaxLoggers";
+            const int DefaultMaxETWLoggers = 64;
+
+            if (MaxEtwLoggers == null)
+            {
+                try
+                {
+                    using (RegistryKey key = Registry.LocalMachine.OpenSubKey(MaxEtwRegistryKey))
+                    {
+                        if (key != null && key.GetValue(MaxEtwPropertyName) != null)
```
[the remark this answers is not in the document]
pamcnult: Sure. I believe the compiler would optimize and use the same value here? I left it for readability / simplicity. I'll break it out, and let me know what you think.

pamcnult:
> @pamcnult, thanks for the updates. I'd like to simplify this just a bit, but I think we're getting close.

No worries. Always open to feedback!  I have a couple comments above - let me know what you think.

src/TraceEvent/TraceEventSession.cs
brianrob: Good point.  What I was thinking would require another field to keep track of whether or not we've been through the initialization path.  I think what you have now is good.

brianrob (approved):

Issue #1415, opened 2021-04-21
Make BPerfEventSource can customize decompress algorithm


honghaic:
Today BPerfEventSource is using ULZ777 to decompress stream.  In some scenario we want to use FPGA to compress and decompress, the algorithm are different with ULZ777,  provide mechanism for end user to customize decompress algorithm can solve the issue.

PR #1416, opened 2021-04-21
Make BPerfEventSource support customize decompress algorithm

honghaic:
For fix issue https://github.com/microsoft/perfview/issues/1415


src/TraceEvent/BPerf/BPerfEventSource.cs
```diff
@@ -113,6 +147,18 @@ public BPerfEventSource(string btlFilePath, TraceEventDispatcherOptions options,
             this.skipQPC = 0;
             this.endFileOffset = long.MaxValue;
             this.canResolveSymbols = !skipReadingUnreachableEvents;
+            this.DecompressDelegate = decompressDelegate;
+            if(this.DecompressDelegate == null)
+            {
+                foreach(var key in BPerfEventSource.BPerfDecompressDelegateDictionary.Keys)
```
brianrob, line 153: Each instance of `BPerfEventSource` is responsible for opening exactly one file.  Thus, it feels to me like the registration mechanism and dictionary are probably more than we need.  What do you think about allowing callers of the `BPerfEventSource` to specify a decompression delegate, and if it's not present, just default to the existing implementation?

honghaic: fixed, thanks for your time.

src/TraceEvent/BPerf/BPerfEventSource.cs
```diff
@@ -521,7 +542,7 @@ private unsafe int ProcessBTLInner(int eof)
                 fixed (byte* uncompressedBufferPtr = &this.uncompressedBuffer[0])
```
brianrob, line 542: With this change, do you still need `uncompressedBufferPtr` and `compressedBufferPtr`?  It looks like you pass the objects into the delegate and then let the delegate pin if it needs to.

src/TraceEvent/BPerf/BPerfEventSource.cs
```diff
@@ -610,6 +631,19 @@ private static long GetOffset(string indexFile, DateTime requestTimestamp, out l
             return 0;
         }
 
+        /// <summary>
+        /// The delegate function to decompress by using ULZ777 algorithm
+        /// </summary>
+        /// <returns></returns>
+        public unsafe static int ULZ777Decompress(byte[] input, int inputOffset, int inputLength, byte[] output, int outputLength)
```
brianrob, line 638: Do you need this to be public?  I'd recommend marking this private and just let it be an implementation detail.

brianrob (commented):
Thanks @honghaic.  A couple more questions.

src/TraceEvent/BPerf/BPerfEventSource.cs
honghaic: removed compressedBufferPtr.   but uncompressedBufferPtr  still needed because in next five lines use it for DeserializeEventRecord

src/TraceEvent/BPerf/BPerfEventSource.cs
honghaic: changed to private, thanks

brianrob (approved):

PR #1421, opened 2021-05-04
Enable more provider levels to be enabled up the ETW max of 255 (current max is 5)

ivberg:
Enable more provider levels to be enabled. The current max is TraceEventLevel.Verbose(5) but ETW can go up to 255. For example a TraceLogging provider I am working with uses a special level of 6 and cannot be enabled with the current TraceEvent code.

Existing TraceEventLevel are preserved and now API users can optionally provider a raw byte level as well.


ivberg:
This is not really necessary. Workaround is to do this instead, "(TraceEventLevel) 6" to set a level above TraceEventLevel.Verbose(5)

PR #1427, opened 2021-05-13
Use numbers for startValue and endValue

russcam:
This commit updates SpeedScopeStackSourceWriter to emit startValue and endValue as numbers, in line with the [JSON schema](https://www.speedscope.app/file-format-schema.json).


adamsitnik (approved):
LGTM, thank you @russcam !

PR #1442, opened 2021-06-07
Add public method converting from QPC timestamp into TraceEvent timestamp.

lateralusX:
New events like ExecutionCheckpoint carries an EventPipe QPC timestamp that can be used to calculate time spans between interesting checkpoints and regular EventPipe events. In order to be able to compare times, EventPipeEventSource needs to expose method that can convert a QPC timestamp into a timestamp on same format as used by TraceEvent open up the ability to compare TraceEvent.TimeStamp with timestamps returned from EventPipeEventSource.QPCTimeToTimeStamp.


src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -83,6 +83,11 @@ private EventPipeEventSource(PinnedStreamReader streamReader, string name)
             StackCache = new StackCache();
         }
 
+        public DateTime QPCTimeToTimeStamp(long QPCTime)
```
brianrob, line 86: Question: Do we explicitly want a local time as opposed to a UTC time?

Also, either way, can you rename the API to specify the type of `DateTime` being returned (essentially using the name of the base API or if local time, perhaps `QPCTimeToLocalDateTime`)?

lateralusX: I picked local since that is what you get from TraceEvent.TimeStamp and that is what you will compare with, I picked UTC first but when starting to compare with the TimeStamp of events you get a timezone diff you don't wan't. Didn't put it in the name to match TraceEvent.TimeStamp, its also a local DateTime.

brianrob: Gotcha.  Makes sense.  Can you add a test to ensure that we protect this code path?  It would be reasonable to check in a small nettrace next to the other files and use it to ensure that the new scenario that your dotnet/runtime PR enables works properly here.

lateralusX: @brianrob Added test with minimal nettrace file only including 3 ExecutionCheckpoint events. Test uses timestamps in those events, convert to TimeStamp using new API and make sure they align with a set of predefined deltas.

lateralusX:
Test failure seems unrelated and appears on other PR's as well:

https://dev.azure.com/ms/perfview/_build/results?buildId=177519&view=logs&jobId=b10d7ded-e7f8-50a1-ede8-c44d21cb8292&j=b10d7ded-e7f8-50a1-ede8-c44d21cb8292&t=b0d9302a-4207-5610-a2f8-f14d39461b90

src/TraceEvent/TraceEvent.Tests/EventPipeParsing.cs
```diff
@@ -513,6 +513,60 @@ public void WellKnownDiagnosticSourceEventsHavePayloads()
...
+
+            UnicodeEncoding unicode = new UnicodeEncoding();
+            using (var source = new EventPipeEventSource(eventPipeFilePath))
+            {
+                source.Dynamic.All += (TraceEvent obj) =>
+                {
+                    ushort eventID = (ushort)obj.ID;
+
+                    // ExecutionCheckpoint.
+                    if (eventID == 300)
```
brianrob, line 538: Question on this: Do we want to update `ClrTraceEventParser` to allow this new event to be parsed using TraceEvent?  I am assuming the answer is yes, since this is the way to consume events from EventPipe.  If so, I would recommend including that update in this PR, and then you can simplify this test to just use the parser to parse the event.  Sorry, it didn't occur to me to mention this ahead of time.

lateralusX: Yeah, I was about to ask that question when implementing the test, so whats the process in updating ClrTraceEventParser to include this new event currently in dotnet/runtime main branch?

brianrob: The best way to do this is to use run `TraceParserGen` against the manifest in the runtime repo.  `TraceParserGen` lives here: https://github.com/microsoft/perfview/tree/main/src/TraceParserGen.  You'll get a C# file out of this, and I would recommend just copying out the parts of the file for your event into https://github.com/microsoft/perfview/blob/main/src/TraceEvent/Parsers/ClrTraceEventParser.cs.  Theoretically, you're supposed to be able to run `TraceParserGen` and just overwrite the file, but there are customizations, and other issues that prevent this.

Once this is done, then you should just be able to consume the event by calling `source.Clr.YourEvent += delegate(YourEventData data { // do stuff here };`

lateralusX: @brianrob Added parsing into ClrTraceEventParser handling both ExecutionCheckpoint emitted from clr provider (currently not emitted) and ExecutionCheckpointDCEnd emitted by clr rundown provider. Updated test to parse the clr rundown ExecutionCheckpoint events.

brianrob (approved):
LGTM.  Thanks @lateralusX!

PR #1444, opened 2021-06-08
Expose the generational aware views

cshung:
As discussed in https://github.com/dotnet/runtime/issues/48937, I removed the check to `AppLog.InternalUser` for the views and expose them directly.


brianrob (approved):
LGTM.  Thanks!

PR #1450, opened 2021-06-25
Add ..\build to paths to native binaries in build/.props

### LINKED ISSUE #1449: Native binaries (msdia140.dll vcruntime etc.) are not copied in transitive builds
Repro:
1. Create a solution with two projects: One an EXE targeting net5.0-windows and one a class library targeting netstandard2.0.
2. Have the EXE reference the class lib project
3. In the class lib, add a package reference to Microsoft.Diagnostics.Tracing.TraceEvent
4. Build the whole solution and examine the output folder
Expect:
You should see native DLLs (msdia140.dll etc.) in the x86/amd64 folders of the output.

Result:
Missing native DLLs.

I finally tracked down this bug to an authoring problem with the .props file in buildTransitive.
The buildTransitive folder was introduced by #1312 to fix #1301
However, the .props file there refers to native DLLs using paths relative to `$(MSBuildThisFileDirectory)`
Unfortunately, the native DLLs don't exist there. They are in the "build" folder.

Manually patching the .props file in the buildTransitive folder to refer to `$(MSBuildThisFileDirectory)..\build\native` instead fixes this issue.

pharring:
Microsoft.Diagnostics.Tracing.TraceEvent.props

Fixes #1449

### DIFF
--- a/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.props
+++ b/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.props
@@ -1,52 +1,52 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
-    <None Condition="Exists('$(MSBuildThisFileDirectory)native\x86\KernelTraceControl.dll')" Include="$(MSBuildThisFileDirectory)native\x86\KernelTraceControl.dll">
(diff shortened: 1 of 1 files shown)


brianrob:
Thanks @pharring.  To confirm my understanding, this continues to work for the build directory because it just drops down a directory and back into the build directory.  Is that right?

pharring:
@brianrob yes, that's right. If that's too subtle, we would have to add a 2nd .props files with the `..\build` pieces just for buildTransitive.

brianrob:
Thanks @pharring.  Not too subtle, just wanted to make sure I understood.

brianrob (approved):

PR #1451, opened 2021-06-26
Add Advanced tabs for File Queries and Directory Enumeration

bwadswor:
Adding advanced tabs for File Queries and Directory Enumeration operations.

These are both a simple copy of the File I/O tab.  Both use hit count as the metric where File I/O uses total I/O size.


brianrob (approved):
:shipit:

PR #1452, opened 2021-06-29
Removed duplicated text from UsersGuide.htm

wilka:
Removed the text "under the assumption that the file" for the help of `/noView run` because it was duplicated in the sentence.


brianrob (approved):
LGTM.  Thanks!

PR #1458, opened 2021-07-08
Fix failure to identify types defined in mscorlib

sharwell:
Fixes the GC Heap Alloc Ignore Free view.

Before:

(Note the presence of `?[]` and `<Unknown>`)

image

After:

image

### DIFF
--- a/src/EtwClrProfiler/CorProfilerTracer.cpp
+++ b/src/EtwClrProfiler/CorProfilerTracer.cpp
@@ -72,12 +72,14 @@ public:
 class ModuleInfo
 {
 public:
-	ModuleInfo() { ID = 0; AssemblyID = 0; MetaDataImport = NULL; Path = NULL; }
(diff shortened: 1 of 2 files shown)


brianrob:
Thanks @sharwell.  Just to make sure I understand correctly, is the issue here that mscorlib is a multi-module assembly, and the existing code doesn't handle this?

sharwell:
No, the issue here is the code assumes `m_info->GetModuleInfo` will provide the AssemblyID. Failure to return an AssemblyID causes GetModuleInfo to return NULL from that point on. I rewrote the code to remove the reliance on obtaining the AssemblyID from GetModuleInfo, because for mscorlib it is called prior to the AttachModuleToAssembly event.

brianrob:
Gotcha.  Thanks @sharwell.

brianrob (approved):

PR #1466, opened 2021-07-20
Update docs to mention .NET 4.6.2

kant2002:
This is reflects fact that minimum requirements for PervView is .NET 4.6.2
And as such minimum OS supported is OS which able to install .NET 4.6.2
When made changes I made some assumptions on supported OS,
which maybe too strict, and actual support wider.

### DIFF
--- a/src/PerfView/SupportFiles/UsersGuide.htm
+++ b/src/PerfView/SupportFiles/UsersGuide.htm
@@ -57,15 +57,15 @@
         way, right clicking allows you to discover what PerfView's can do for you.
     </p>
     <p>
-        PerfView is a V4.5 .NET application.&nbsp;&nbsp; Thus you need to have installed
-        a V4.5 .NET Runtime on the machine which you actually run PerfView.&nbsp;&nbsp;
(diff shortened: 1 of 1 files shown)


brianrob:
@kant2002, thank you for this contribution.  Looking at the .NET 4.5 requirements, it is possible to run on Windows 7 (see: https://www.microsoft.com/en-us/download/details.aspx?id=30653).

Also, I think there is a word missing here:
```
On machines that don't have V4.5 or later of the .NET runtime installed, it is also to collect ETL data
```

I think it should be "it is also possible to collect ETL data".

kant2002:
@brianrob can you take a look at this?

kant2002:
@brianrob friendly ping.

brianrob:
@kant2002, thanks for reminding me of this.  I hate to do this, but with the release of PerfView 3.0 just a couple of days ago, the baseline requirement is now .NET 4.6.2  Would you mind updating your PR to account for this?

kant2002:
@brianrob that's fine. I update text, so if you have other suggestions let me know.

kant2002:
@brianrob friendly ping again.

brianrob:
LGTM.  Thanks!

PR #1468, opened 2021-07-23
Fix opening NativeAOT images

kant2002:
After running ImageSize command on the NativeAOT generated image
It fails beause `SampleWindowsForms__Module___StartupCodeMain` symbol has `src="<stdin>"` after importing image. I simply workaround this issue, assuming that no other cases like that exists.


brianrob (approved):
LGTM.  Thanks @kant2002!

PR #1469, opened 2021-07-25
Update Source Server URL

brianrob:


src/TraceEvent/Symbols/NativeSymbolModule.cs
```diff
@@ -592,6 +595,12 @@ protected override string GetSourceFromSrcServer()
                             _log.WriteLine("Source Server command {0} is not recognized as safe (sd.exe or tf.exe), failing.", fetchCmdStr);
                             return null;
                         }
+
+                        if (fetchCmdStr.Contains(OldSourceServerUrl))
```
pharring, line 599: Nit: The .Contains check isn't really necessary. Replace returns the input string unmodified if there are no replacements.

brianrob: Good call @pharring.  Thanks.

PR #1478, opened 2021-08-16
Fix Registry Tracking using Rundown Events

brianrob:
Not all registry events contain the key name, so we do extra work to share the key name across events by using the KeyHandle as the common identifier.  The existing code does not include a couple of the rundown events, and also was saving many instances of string.Empty for the KeyName.

This change identifies a couple more rundown events, and also only saves the key name if it is non-null and non-empty.


brianrob:
This fix is incomplete, and will require more work.  I'm going to close this and will re-open when a new code delta is ready.

Issue #1445, opened 2021-06-18
TraceLog.IsKernelAddress makes wrong assumptions for none Windows platforms.


lateralusX:
Runing EventPipe on Android Arm32 will produce trace files that can have code address in high address space. Trying to convert that nettrace file to SpeedScope or trying to load it in PerfView will give incomplete stack traces (without any method information). After digging through the parser code I located the issue to be and assumption done in TraceLog.IsKernelAddress where it assumes the following:

```
if (pointerSize == 4)
{
    return ip >= 0x80000000;
}

return ip >= 0xFFFF000000000000;        // TODO I don't know what the true cutoff is.
```

and that could fail on all none Windows platforms. Eliminating this check makes it possible to load and analyse the file as expected.

This is a blocking issue for .net6 since we will roll up a number of new platforms into EventPipe and nettrace (Android/iOS/x64/x86/Arm64/Arm32) that could have code located in high address ranges. The check is also not relevant for nettrace files since they currently only includes managed methods and no native frames.

lateralusX:
//CC @brianrob @josalem

lateralusX:
Fixed by #1480.

PR #1480, opened 2021-08-30
Fix issue with IsKernelAddress on none Windows platforms.

lateralusX:
https://github.com/microsoft/perfview/issues/1445 shows issues with current schema to detect if a address is in kernel address space. On none Windows platforms the implemented rules are not correct causing issues to display stack traces on nettrace files captured on some configurations of none Windows platforms (Android is one example).

EventPipe nettrace files only include stack traces with managed frames and only user mode code addresses, so addresses should never be in kernel address space and since the heuristics miss fire on none Windows platforms, this commit adds logic to always detect addresses using a process hooked up to an EventPipeEventSource to not be in kernel address mode. None EventPipeEventSource event sources will continue to work as before.

### DIFF
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -2513,7 +2513,7 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
             // we get a def, we have to know which one to look in (we do this based on the address
             // in the def (a bit kludgey in my opinion).
             IncompleteStack stackInfo;
-            if (IsKernelAddress(data.InstructionPointer(data.FrameCount - 1), data.PointerSize))
+            if (IsWindowsKernelAddress(data.InstructionPointer(data.FrameCount - 1), data.PointerSize))
             {
                 // We have a kernel mode definition, look up in the kernel mode table.
                 if (kernelStackKeyToInfo.TryGetValue(data.StackKey, out stackInfo))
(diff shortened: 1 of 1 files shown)


lateralusX:
//CC @brianrob, @josalem

josalem (approved):
I'll defer to @brianrob for comments on the impact to PerfView. In terms of the nettrace ecosystem, this seems appropriate and gives us a spot to change the behavior if we ever do need this to return `true`.

src/TraceEvent/TraceLog.cs
```diff
@@ -5800,6 +5800,15 @@ internal int ParentDepth()
 #endif
         #endregion
 
+        internal bool IsKernelAddress(Address ip, int pointerSize)
+        {
+            // EventPipe doesn't generate kernel address events and current heauristics are not deterministic on none Windows platforms.
+            if (log?.rawEventSourceToConvert is EventPipeEventSource)
```
brianrob, line 5806: Overall this strategy is a fine way to do this.  I would like to not put this method on `TraceProcess` since it isn't really a per-process concept - it is a per-trace concept.  Can you move this method to `TraceLog`?  It can still be accessed from `process.Log`.  You will probably need to rename it, perhaps to something like `IsWindowsKernelAddress`.

lateralusX: @brianrob Renamed TraceLog.IsKernelAddress to TraceLog.IsWindowsKernelAddress.

brianrob (approved):

PR #1484, opened 2021-09-07
Add support to parse SYSTEMTIME type

xleza:
Currently the DynamicTraceEventParser parses SYSTEMTIME in a wrong way, because it treats SYSTEMTIME as a FILETIME that causes this problem. I added support to parse the SYSTEMTIME in a correct way.


xleza:
Currently the DynamicTraceEventParser parses SYSTEMTIME in a wrong way, because it treats SYSTEMTIME as a FILETIME that causes this problem. I added support to parse the SYSTEMTIME in a correct way.

brianrob:
@xleza this change seems pretty reasonable to me.  Would you be able to capture a small trace that contains a SYSTEMTIME argument that you can add to the test bed here so that we don't break this in the future?

gelasultana:
@brianrob Added .etl file which contains primitives in order to check how does the DynamicTraceEventParser parses them. So SYSTEMTIME parsing is fully covered.

brianrob:
Thanks @gelasultana!

brianrob (approved):

PR #1486, opened 2021-09-08
Let SegmentedList Handle more than 2^31-1 elements

mpeyrotc:
Since we want to support collections that are larger than 2^31 - 1 elements, this PR changes all our indexing logic to use longs and makes sure that in interface methods from ICollection and IReadOnlyList that require us to continue using ints we throw when we pass their max limit.


mpeyrotc:
@brianrob @sharwell this is a prerequisite to changing segmented list to support more elements. It should not have any functionality change. Could I bother you with a review?

sharwell:
> Since we want to support collections that are larger than 2<sup>31</sup> - 1

To clarify, you are referring to the number of elements in the collection or just the number of bytes it takes to represent those elements?

mpeyrotc:
I fixed the message, elements.

sharwell:
What about just making `ToArray()` and `CopyTo()` throw an exception if the collection contains more than the supported number of elements?

brianrob:
Agreed with @sharwell, I would like to keep the implementation, but let's throw if the operation can't be run because the collection is too large.

src/FastSerialization/SegmentedList.cs
```diff
@@ -328,13 +339,14 @@ public void AppendFrom(SegmentedList<T> from, int index, int count)
...
+                    int sourceOffset = (int)(index % from.segmentSize);
                     int sourceLength = from.segmentSize - sourceOffset;
-                    int targetSegment = this.count >> this.segmentShift;
-                    int targetOffset = this.count & this.offsetMask;
+                    int targetSegment = (int)(this.count >> this.segmentShift);
+                    int targetOffset = (int)(this.count & this.offsetMask);
                     int targetLength = this.segmentSize - targetOffset;
-                    int countToCopy = Math.Min(count, Math.Min(sourceLength, targetLength));
+                    // We can safely cast to int since source and target lengths will never surpass int.MaxValue
+                    int countToCopy = (int)Math.Min(count, Math.Min(sourceLength, targetLength));
```
brianrob, line 349: Is this safe to cast, as count is now a long?

src/FastSerialization/SegmentedList.cs
```diff
@@ -403,10 +421,28 @@ public T[] ToArray()
         /// <param name="arrayIndex">Destination array starting index.</param>
         public void CopyTo(T[] array, int arrayIndex)
         {
-            int remain = this.count;
+            if (array == null)
+            {
+                throw new ArgumentNullException(nameof(array));
+            }
+
+            if (arrayIndex < 0 || arrayIndex > array.Length)
```
brianrob, line 429: I believe this should be `arrayIndex >= array.Length`, and the argument text updated.

src/FastSerialization/SegmentedList.cs
```diff
@@ -424,9 +460,38 @@ public void CopyTo(T[] array, int arrayIndex)
...
-        public void CopyRangeTo(T[] array, int arrayIndex, int startIndex, int endIndex)
+        public void CopyRangeTo(T[] array, int arrayIndex, long startIndex, long endIndex)
         {
-            int remain = Math.Min(this.count, endIndex - startIndex + 1);
+            if (array == null)
+            {
+                throw new ArgumentNullException(nameof(array));
+            }
+
+            if (arrayIndex < 0 || arrayIndex > array.Length)
```
brianrob, line 470: Same here.  I think this should be `arrayIndex >= array.Length`.

src/FastSerialization/SegmentedList.cs
mpeyrotc: Yes, it is safe. Because this logic copies stuff over by segments, which are always 8_192 items long.

brianrob (approved):
Merging this into the feature branch to unblock progress.  I will do further review there.

PR #1488, opened 2021-09-14
SegmentedDictionary Initial Implementation and Tests

mpeyrotc:
When a `MemoryGraph` has many addresses, the internal `Dictionary` instance it uses to keep track of them might run out of space due to its internal array's limitations. Hence, this PR adds a new `SegmentedDictionary` class that will not run into this issue. The original implementation of the class is located in the [dotnet/roslyn](https://github.com/dotnet/roslyn) repo and can be found [here](https://github.com/dotnet/roslyn/blob/release/dev17.0/src/Dependencies/Collections/SegmentedDictionary%602.cs).

The new class should have a similar performance to the `Dictionary` class it is replacing. However, we will continue to use the old class unless we detect that we might run into the memory issue.

This PR adds also the Generic `Dictionary` tests from the [dotnet/runtime](https://github.com/dotnet/runtime) repo to guarantee that the new implementation is behaving as expected.


brianrob:
Overall, looks good.  A couple of requests:

1. Can you move the new files into a new SegmentedDictionary subdirectory of FastSerialization?  This will help to keep track of what files were copied.
2. Can you move the tests from `PerfView.Tests` to `TraceEvent.Tests`?  This is where the other FastSerialization tests live.  They should also live in a SegmentedDictionary subdirectory as well.

src/FastSerialization/SegmentedList.cs
```diff
@@ -93,6 +93,8 @@ public int Count
             }
         }
 
+        public int Capacity => this.capacity;
```
brianrob, line 96: Would like to understand the need for these two new public APIs.  Are these concerned internally by SegmentedDictionary?

src/MemoryGraph/MemoryGraph.cs
```diff
@@ -10,7 +10,17 @@ public class MemoryGraph : Graph, IFastSerializable
         public MemoryGraph(int expectedSize)
             : base(expectedSize)
         {
-            m_addressToNodeIndex = new Dictionary<Address, NodeIndex>(expectedSize);
+            // If we have too many addresses we will reach the Dictionary's internal array's size limit and throw.
+            // Therefore use a new implementation of it that is similar in performance but that can handle the extra load.
+            if (expectedSize > 200_000)
```
brianrob, line 15: Can you share where this number comes from?  I realize that we have to pick something if we want to only impact "large" dumps, but just want to understand the math behind this.

brianrob (approved):
A couple of questions here, but I'll go ahead and merge this into the feature branch to unblock further progress.

PR #1498, opened 2021-09-23
Use new SegmentedList and SegmentedDictionary to process large heap dumps

mpeyrotc:
These changes are a continuation of the following PRs:

* [1486](https://github.com/microsoft/perfview/pull/1486) Let SegmentedList Handle more than 2^31-1 elements
* [1488](https://github.com/microsoft/perfview/pull/1488) SegmentedDictionary Initial Implementation and Tests
* [1491](https://github.com/microsoft/perfview/pull/1491) Opt-In 64-bit StreamLabel Support

In summary we are hooking everything together, and able to process large memory heap dumps. The actual changes are:

1. SegmentedDictionary.cs and SegmentedList.cs have minor improvements and make sure we don't delete public methods and that new methods that were added for this scenario remain internal.
2. SegmentedStreamReader.cs takes a `SerializationConfiguration` instance specifying if it should treat `StreamLabel` instances as 32 or 64bit objects; as well as making the necessary adjustments to take advantage of this fact when processing a memory graph.
3. SegmentedMemoryStreamWriter.cs is the same as SegmentedStreamReader.cs
4. GCHeapDump.cs did not require making the `XmlGcHeapDump` class public for it to be used in the PerfView.Tests project.
5. MemoryGraph.cs should not use the expected size to determine when to use a `SegmentedDictionary` instance since it is not that accurate and the range is subject to change if the internal implementation of a `Dictionary` changes.
6. Graph.cs makes sure that the `SegmentedMemoryStreamWriter` instance it creates is passed the correct `StreamLabel` size information.

### DIFF
--- a/src/FastSerialization/SegmentedDictionary/SegmentedDictionary.cs
+++ b/src/FastSerialization/SegmentedDictionary/SegmentedDictionary.cs
@@ -781,7 +781,8 @@ namespace System.Collections.Generic
                 throw new ArgumentOutOfRangeException(nameof(capacity));
             }
 
-            var currentCapacity = _entries.Capacity;
+            // Normal usage of a dictionary should never ask for a capacity that exceeds int32.MaxValue.
+            var currentCapacity = (int)_entries.Capacity;
             if (currentCapacity >= capacity)
             {
                 return currentCapacity;
--- a/src/FastSerialization/SegmentedMemoryStreamReader.cs
+++ b/src/FastSerialization/SegmentedMemoryStreamReader.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.IO;
 using System.Text;
 
 namespace FastSerialization
@@ -13,12 +12,31 @@ namespace FastSerialization
         /// <summary>
         /// Create a IStreamReader (reads binary data) from a given byte buffer
         /// </summary>
-        public SegmentedMemoryStreamReader(SegmentedList<byte> data) : this(data, 0, (int)data.Count) { }
+        public SegmentedMemoryStreamReader(SegmentedList<byte> data, SerializationConfiguration config = null) : this(data, 0, data.Count, config) { }
         /// <summary>
         /// Create a IStreamReader (reads binary data) from a given subregion of a byte buffer 
         /// </summary>
-        public SegmentedMemoryStreamReader(SegmentedList<byte> data, int start, int length)
+        public SegmentedMemoryStreamReader(SegmentedList<byte> data, long start, long length, SerializationConfiguration config = null)
         {
+            SerializationConfiguration = config ?? new SerializationConfiguration();
+
+            if (SerializationConfiguration.StreamLabelWidth == StreamLabelWidth.FourBytes)
+            {
+                readLabel = () =>
+                {
+                    return (StreamLabel)(uint)ReadInt32();
+                };
+                sizeOfSerializedStreamLabel = 4;
+            }
+            else
+            {
+                readLabel = () =>
+                {
+                    return (StreamLabel)(ulong)ReadInt64();
+                };
+                sizeOfSerializedStreamLabel = 8;
+            }
+
             bytes = new SegmentedList<byte>(65_536, length);
             bytes.AppendFrom(data, start, length);
             position = start;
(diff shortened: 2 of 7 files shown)


src/FastSerialization/SegmentedMemoryStreamReader.cs
```diff
@@ -174,6 +204,11 @@ public void Dispose()
         /// Dispose pattern
         /// </summary>
         protected virtual void Dispose(bool disposing) { }
+
+        /// <summary>
+        /// Returns the SerializationConfiguration for this stream reader.
+        /// </summary>
+        public SerializationConfiguration SerializationConfiguration { get; private set; }
```
brianrob, line 211: If not needed, let's keep this internal.  I'm not sure how much of this is going to live forever vs. disappear relatively quickly once we have backwards compat in-place.

src/FastSerialization/SegmentedMemoryStreamWriter.cs
```diff
@@ -1,16 +1,31 @@
...
-        public SegmentedMemoryStreamWriter(int initialSize)
+        public SegmentedMemoryStreamWriter(SerializationConfiguration config = null) : this(64, config) { }
+        public SegmentedMemoryStreamWriter(long initialSize, SerializationConfiguration config = null)
         {
+            SerializationConfiguration = config ?? new SerializationConfiguration();
+
+            if (SerializationConfiguration.StreamLabelWidth == StreamLabelWidth.FourBytes)
+            {
+                writeLabel = (value) =>
+                {
```
brianrob, line 17: There's a good pattern of asserting here that (int)value <= int.MaxValue.  Can you add that here as well?

src/FastSerialization/SegmentedMemoryStreamWriter.cs
```diff
@@ -85,10 +98,15 @@ public void WriteToStream(Stream outputStream)
...
             var readerBytes = bytes;
-            return new SegmentedMemoryStreamReader(readerBytes, 0, (int)readerBytes.Count);
+            return new SegmentedMemoryStreamReader(readerBytes, 0, readerBytes.Count, SerializationConfiguration);
         }
         public void Dispose() { }
 
+        /// <summary>
+        /// Returns the SerializationConfiguration for this stream writer.
+        /// </summary>
+        public SerializationConfiguration SerializationConfiguration { get; private set; }
```
brianrob, line 108: Let's keep this one internal as well.

src/FastSerialization/SegmentedDictionary/SegmentedDictionary.cs
```diff
@@ -781,7 +781,8 @@ public int EnsureCapacity(int capacity)
                 throw new ArgumentOutOfRangeException(nameof(capacity));
             }
 
-            var currentCapacity = _entries.Capacity;
+            // Normal usage of a dictionary should never ask for a capacity that exceeds int32.MaxValue.
```
brianrob, line 784: Is the cast here necessary?

mpeyrotc: I believe so, since the underlying data structure supports long indices, but the dictionary does not.

brianrob: Wondering if the cast should actually be on capacity and a cast to (long) so that this doesn't truncate the value if it's greater than int.MaxValue?

src/FastSerialization/SegmentedMemoryStreamWriter.cs
brianrob: I think this should be `Debug.Assert((long)value <= int.MaxValue)`.  Sorry, I said (int) earlier.

src/FastSerialization/SegmentedDictionary/SegmentedDictionary.cs
mpeyrotc: the issue is the return value, more than that the value of the capacity arg?

brianrob: Oh, I see.  In that case, should not matter because the return value has to be an int anyway.  Got it.

mpeyrotc: correct. Are you ok in resolving it then?

PR #1506, opened 2021-10-12
Fix Tutorial.cs links

eduherminio:
Changes can be verified [here](https://htmlpreview.github.io/?https://github.com/eduherminio/perfview/blob/patch-1/src/PerfView/SupportFiles/UsersGuide.htm#Tutorial) against my branch.


eduherminio:
Oh, I see the confusion now, my bad.
It's from GitHub where it fails to find those files, I just double checked it works from PerfView.

So maybe this is not the right fix after all 😅

PR #1520, opened 2021-11-11
Set PerfView as the default project

davkean:
VS picks the first project in the solution as the default project, set this to PerfView so the first time you open the solution you can immediately F5.


brianrob:
@davkean, thanks for contributing this.  I think it would be good to revert the README change so that if folks happen to have changed the startup project inadvertently, this information is available.  On first run through these instructions, this will be a no-op, but folks will know about it for the future.

davkean:
@brianrob Done.

brianrob (approved):

PR #1522, opened 2021-11-11
Remember the last opened directory

davkean:


brianrob:
@davkean thanks for contributing this.  I am torn on this one.  I know that for myself, I prefer to have PerfView open the current directory unless I specify otherwise.  Do you have a workflow in-mind that this enables?

davkean:
Two cases which result in a non-ideal behavior today:

1. I pin PerfView to the taskbar. If I launch it from there, it defaults to the perfview location which I've put in program files.
2. If I launch PerfView without elevation and then choose Collect -> Collect, it will relaunch PerfView defaulting to Documents.

PR #1527, opened 2021-11-14
Enforce naming conventions

davkean:
- Add, what I believe, to be the naming conventions of this project in editorconfig so that it gets used during refactorings and enforced by VS
- Baseline existing violations to prevent the bleeding

I plan to go back and fix the violations, however, there's some non-trivial changes needed. This is due to the common pattern of having a public property and internal field that would otherwise result in the same name, I will make these as separate changes so they don't get lost in the trivial renames.


brianrob:
This is a change that I want to think about for a bit.  There are unfortunately several areas that don't necessarily share coding conventions, and some of these also take updates from code generators such as TraceParserGen.  I'm not sure I want to enforce naming conventions.  Let me think on this and get back to you.

davkean:
@brianrob We can exclude specific directories, and turn these into refactoring hints so that the compiler doesn't warn about it, but refactoring uses the conventions.

PR #1529, opened 2021-11-14
Add Open Folder/Open Containing Folder commands

davkean:
This adds two new commands mimicing Visual Studio's behavior.

**Open Folder:** Opens a folder in Windows Explorer
image

**Open Containing Folder:** Opens a file, selecting it, in Windows Explorer
image

This also has a behavior change for Menu Items, which is to hide them when they are not enabled.


brianrob (approved):
Thanks for contributing this.  Seems like this could be handy.

PR #1531, opened 2021-11-15
Add icon for non-expanded folders/groups 

davkean:
This includes https://github.com/microsoft/perfview/pull/1530, so will be merged after it has been merged. Just review https://github.com/microsoft/perfview/commit/5f979d2df9941c109c9e571ec40d8a93f094dac6.

This adds a icon for non-expanded folders/groups.

ExpandedIcon


brianrob:
I think I'm missing something obvious here, but I don't see a difference between the behavior of #1530 and #1531 with respect to folders and groups.  I'm watching the animation in this PR and then looking at a baseline version of PerfView, and I don't see a difference.  What did I miss?

davkean:
@brianrob There is no "folder closed" icon. They are always expanded in the baseline.

PR #1536, opened 2021-11-23
Remove Default Event Limit

brianrob:
Remove the 20M event limit now that ETLX files can support file sizes larger than 4GB.


MokoSan:
@brianrob was this change incorporated in a new release?  If not, would this be released anytime soon? Thanks for adding this! Our tests diagnosis would greatly benefit from this.

brianrob:
It is not.  In fact, I reverted the change from main, as I am working on a set of changes for a PerfView 3.0, and want to include this in a major release.  I just created a PR to the feature branch here: https://github.com/microsoft/perfview/pull/1549

PR #1542, opened 2021-12-03
Add support for VSDStubCreated event to TraceEvent

jakobbotsch:
Add support for new event that is being added in https://github.com/dotnet/runtime/pull/62255

### DIFF
--- a/src/TraceEvent/Parsers/ClrTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/ClrTraceEventParser.cs
@@ -191,6 +191,11 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
             TypeDiagnostic = 0x8000000000,
 
             /// <summary>
+            /// Events about stubs created by the runtime.
+            /// </summary>
+            Stubs = 0x40000000000,
+
+            /// <summary>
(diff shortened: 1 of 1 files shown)


src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -3614,8 +3639,8 @@ protected internal override void Dispatch()
         }
         protected internal override void Validate()
         {
-            Debug.Assert(!(Version == 0 && EventDataLength != 0 + (TypeParameterCount * 8) + 36));
-            Debug.Assert(!(Version > 0 && EventDataLength < 0 + (TypeParameterCount * 8) + 36));
+            Debug.Assert(!(Version == 0 && EventDataLength != 0 + (TypeParameterCount * 8) + 32));
+            Debug.Assert(!(Version > 0 && EventDataLength < 0 + (TypeParameterCount * 8) + 32));
```
jakobbotsch, line 3643: Is this change right? I hit an assert here in debug.

brianrob: To be honest, I'm not sure.  The size of the event is fixed, so I'm surprised that the existing code isn't right.

PR #1544, opened 2021-12-14
Fix download link for perfview.exe

### LINKED ISSUE #1543: Download link is not working as described in Downloading.md
In the file `Downloading.md`, section "Shortcut to Download the Latest PerfView.exe" (linked on the GitHub main page) it states:

> In the common case, you only need one file, PerfView.exe, to use the tool. The most recent copy of this file can be downloaded here:
> 
> GitHub release (latest SemVer)
> 
> Once you click the above link in your browser it will start downloading, the details of which vary from browser to browser. In some cases it will prompt for more information (IE) and in others (Chrome) it may not be obvious that you clicked on anything (look at the bottom of the pane for changes). The result, however will be a PerfView.exe on your local machine.

However, when I click on the link, neither with Firefox, nor with Chrome nor Edge the download starts. Instead, I see this page (see screenshot below), but there is nothing pointing me to `PerfView.exe`:

----
image
 
Can someone fix the wording on the documentation _or/and_ add the missing exe file into the download section? Or am I just missing something else?

mregen:
The old link confusingly only lists the TraceEvent release, which is tagged as latest. The updated link shows the list of releases, so you can easily find the perfview release underneath TraceEvent.
fixes #1543

### DIFF
--- a/documentation/Downloading.md
+++ b/documentation/Downloading.md
@@ -16,7 +16,7 @@ care about specific bug fixes or features go there.
 In the common case, you only need one file, PerfView.exe, to use the tool.  The most recent copy of
 this file can be downloaded here:
 
-[![GitHub release (latest SemVer)](https://img.shields.io/github/v/release/microsoft/perfview)](https://github.com/microsoft/perfview/releases/download/latest/PerfView.exe)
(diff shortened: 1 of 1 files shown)


brianrob:
Thanks @mregen for this.  I just made a very similar fix this morning without realizing you had already done the work.

PR #1554, opened 2022-01-17
Fix JSON escaping

adamsitnik:
Today I've tried to convert a trace provided by the customer in https://github.com/dotnet/runtime/issues/63621#issuecomment-1010030287 to Chromium trace file format. I was able to save the file, but unable to load it in Chromium Developer Tools (F12).

It turned out that the Chromium/Speedscope exporters were not escaping some of the strings correctly.


adamsitnik:
@brianrob PTAL

brianrob (approved):
LGTM!

PR #1555, opened 2022-01-18
Upgrading to .NET framework 4.6.2

cshung:
As [this](https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/) blog post indicates, .NET 4.5 has gone unsupported for a while, and even .NET 4.6.1 will reach the end of support by this April. 

This change retargets PerfView to run on top of .NET 4.6.2 instead. This will also make it possible to build with Visual Studio 2022. (Without this change, it is impossible to build with Visual Studio 2022, and there is no way you can install .NET 4.5 support for it)


brianrob:
Closing this PR as we'll take this change as part of 3.0 in #1579.

PR #1557, opened 2022-01-25
Removed the Obsolete Attribute for GC properties

MokoSan:
There are a number of properties in the ``TraceGC`` and its related classes that contain the ``Obsolete`` attribute. In reality these __were __experimental and can now that decoration can be removed as they are used across the board. 

The presence of this attribute is causing noise with warnings in all the clients that make use of this API.  

CC: @Maoni0


Maoni0 (approved):
LGTM!

brianrob (approved):

PR #1560, opened 2022-01-27
Add configurable alignment to FastSerializer

josalem:
IOStreamStreamReader had a hard coded 8 byte alignment. EventPipe streams do not have any alignment requirements (save for some 4 byte alignments inside blocks). This means that blocks can have sizes that are not aligned. EventPipeEventSource will not dispatch events read from the stream until reads complete. In live cases where these blocks are unaligned, but no new data is coming down the pipe, the reader can attempt to enforce the alignment and end up blocking until new data arrives. This results in latency for event dispatch. For infrequent events, this can be a long period of time causing users to think they haven't received events.

For example:

EventBlock 1, EventBlock 2, and EventBlock 3 all arrive with aligned lengths. EventBlock 4 arrives with an unaligned length. Until more data arrives on the stream, none of the events in EventBlock 4 will be dispatched to the user. 

This change adds a configuration switch for changing the alignment of the reader. The default is the old behavior (8 byte alignment) and the live reading for EventPipe was changed to 1 byte alignment.

Changing to an alignment of 1 uncovered a bug where calling `Goto(StreamLabel)` with a label greater than the alignment length would cause the subsequent read to return bytes from the wrong position in the stream. Up until now, this wasn't an issue for EventPipe because all `Goto` calls were for jumps < 4 due to the internal alignment, which is always less than 8.

This patch also fixes that bug.

TODO: tag all the issues this will close. I believe there are issues in dotnet/runtime, dotnet/diagnostics, dotnet/dotnet-monitor, and microsoft/perfview that are tracking this issue.

CC @tommcdon @jander-msft @wiktork

### DIFF
--- a/src/FastSerialization/FastSerialization.cs
+++ b/src/FastSerialization/FastSerialization.cs
@@ -84,6 +84,20 @@ namespace FastSerialization
     };
 
     /// <summary>
+    /// Allows users of serialization and de-serialization mechanism to specify the alignment required by the
+    /// reader.
+    /// </summary>
+#if FASTSERIALIZATION_PUBLIC
+    public
+#endif
+    enum StreamReaderAlignment
+    {
+        OneByte     = 1,
+        FourBytes   = 4,
+        EightBytes  = 8
+    };
+
+    /// <summary>
     /// These settings apply to use of Serializer and Deserializer specifically.
     /// </summary>
 #if FASTSERIALIZATION_PUBLIC
@@ -92,6 +106,7 @@ namespace FastSerialization
     sealed class SerializationConfiguration
     {
         public StreamLabelWidth StreamLabelWidth { get; set; }
+        public StreamReaderAlignment StreamReaderAlignment { get; set; } = StreamReaderAlignment.EightBytes;
     }
 
     /// <summary>
--- a/src/FastSerialization/FastSerialization.cs
+++ b/src/FastSerialization/FastSerialization.cs
@@ -90,7 +90,7 @@ namespace FastSerialization
 #if FASTSERIALIZATION_PUBLIC
     public
 #endif
-    enum StreamReaderAlignment
+    enum StreamReaderAlignment : int
     {
         OneByte     = 1,
         FourBytes   = 4,
@@ -106,7 +106,6 @@ namespace FastSerialization
     sealed class SerializationConfiguration
     {
         public StreamLabelWidth StreamLabelWidth { get; set; }
-        public StreamReaderAlignment StreamReaderAlignment { get; set; } = StreamReaderAlignment.EightBytes;
     }
 
     /// <summary>
--- a/src/FastSerialization/StreamReaderWriter.cs
+++ b/src/FastSerialization/StreamReaderWriter.cs
@@ -718,8 +718,12 @@ namespace FastSerialization
                     while (amountToThrowOut > 0)
                     {
                         int read = inputStream.Read(bytes, 0, (int)Math.Min(amountToThrowOut, (long)bytes.Length));
-                        amountToThrowOut -= read;
                         inputStreamBytesRead += read;
+                        if (read == 0)
+                        {
+                            throw new Exception("Read past end of stream.");
+                        }
+                        amountToThrowOut -= read;
                     }
                 }
 
--- a/src/FastSerialization/StreamReaderWriter.cs
+++ b/src/FastSerialization/StreamReaderWriter.cs
@@ -646,10 +646,10 @@ namespace FastSerialization
         /// Create a new IOStreamStreamReader from the given System.IO.Stream.   Optionally you can specify the size of the read buffer
         /// The stream will be closed by the IOStreamStreamReader when it is closed.  
         /// </summary>
-        public IOStreamStreamReader(Stream inputStream, int bufferSize = defaultBufferSize, bool leaveOpen = false, SerializationConfiguration config = null)
-            : base(new byte[bufferSize + (((int?)config?.StreamReaderAlignment) ?? 8)], 0, 0, config)
+        public IOStreamStreamReader(Stream inputStream, int bufferSize = defaultBufferSize, bool leaveOpen = false, SerializationConfiguration config = null, StreamReaderAlignment alignment = StreamReaderAlignment.EightBytes)
+            : base(new byte[bufferSize + (int)alignment], 0, 0, config)
         {
-            align = (int?)config?.StreamReaderAlignment ?? 8;
+            align = (int)alignment;
             Debug.Assert(bufferSize % align == 0);
             this.inputStream = inputStream;
             this.leaveOpen = leaveOpen;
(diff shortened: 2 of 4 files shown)


src/TraceEvent/TraceEvent.Tests/EventPipeParsing.cs
```diff
@@ -254,6 +255,21 @@ public void V4EventPipeFileHasProcNumbers()
...
         }
 
+        [Fact]
+        public void GotoWorksForPositionsGreaterThanAlignment()
+        {
+            using (var reader = new PinnedStreamReader(new MockHugeStream((long)uint.MaxValue + 5_000_000), config: new SerializationConfiguration { StreamReaderAlignment = StreamReaderAlignment.EightBytes }))
+            {
+                reader.Goto((StreamLabel)0x148);
+                var buf = new byte[100_000];
+                // Specifically doing a read larger than the default buffer size (16KB) to avoid caching path.
```
jander-msft, line 265: nit: Do you want to fail the test if someone changes the default buffer size to be larger than the buffer length here so your assumption is not accidentally invalidated?

josalem: Probably easier to specifically set the buffer size of the stream to be smaller/an explicit size. I'll update the test.

src/FastSerialization/StreamReaderWriter.cs
```diff
@@ -710,6 +711,18 @@ public override void Read(byte[] data, int offset, int length)
             // remainder from the stream
             if (length > (bytes.Length - align))
             {
+                // check if our inputstreambytesread < position in stream then throw out delta
+                if (inputStreamBytesRead < positionInStream)
+                {
+                    long amountToThrowOut = positionInStream - inputStreamBytesRead;
+                    while (amountToThrowOut > 0)
+                    {
+                        int read = inputStream.Read(bytes, 0, (int)Math.Min(amountToThrowOut, (long)bytes.Length));
```
wiktork, line 720: nit: Did you need to handle Read() returning 0?

josalem: I think we do. That would signify we've read past the end of the stream. In the live scenario, it would block until the stream closed in which case, we would also need to error if we read 0 bytes. I think there's an exception defined for that I can add here.

hoyosjs: Doesn't read returning 0 signify end of stream? The error is a bit unclear to me as it stands.

src/FastSerialization/FastSerialization.cs
```diff
@@ -92,6 +106,7 @@ enum StreamLabelWidth
     sealed class SerializationConfiguration
     {
         public StreamLabelWidth StreamLabelWidth { get; set; }
+        public StreamReaderAlignment StreamReaderAlignment { get; set; } = StreamReaderAlignment.EightBytes;
```
noahfalk, line 109: I interpreted SerializationConfiguration as describing what is supported in the file format. Other settings that solely affect implementation behavior/performance like the cache buffer size and whether the Stream should be left open are passed as parameters to the IOStreamStreamReader constructor. If we were following that convention I'd expect this StreamReaderAlignment to also be a constructor parameter rather than a property here.

I figured I'd mention it in case anyone cares, but I don't care much myself. If you and @brianrob are happy with this as a property it is fine for me.

src/FastSerialization/StreamReaderWriter.cs
```diff
@@ -647,8 +647,9 @@ public IOStreamStreamReader(string fileName, SerializationConfiguration config =
         /// The stream will be closed by the IOStreamStreamReader when it is closed.  
         /// </summary>
         public IOStreamStreamReader(Stream inputStream, int bufferSize = defaultBufferSize, bool leaveOpen = false, SerializationConfiguration config = null)
-            : base(new byte[bufferSize + align], 0, 0, config)
+            : base(new byte[bufferSize + (((int?)config?.StreamReaderAlignment) ?? 8)], 0, 0, config)
```
brianrob, line 650: I think I might be missing something here, but why does this need to be casted to int?.  I would expect this to just be an int, and not need the cast.

src/TraceEvent/TraceEvent.Tests/EventPipeParsing.cs
```diff
@@ -254,6 +255,21 @@ public void V4EventPipeFileHasProcNumbers()
...
+        [Fact]
+        public void GotoWorksForPositionsGreaterThanAlignment()
+        {
+            using (var reader = new PinnedStreamReader(new MockHugeStream((long)uint.MaxValue + 5_000_000), bufferSize: 0x4000 /* 16KB */, config: new SerializationConfiguration { StreamReaderAlignment = StreamReaderAlignment.EightBytes }))
+            {
+                reader.Goto((StreamLabel)0x148);
+                var buf = new byte[100_000];
+                // Specifically doing a read larger than the default buffer size (16KB) to avoid caching path.
+                reader.Read(buf, 0, buf.Length);
+                // 0x14 is the 148th byte of the MockHugeStream (it is deterministic).
```
noahfalk, line 267:
```suggestion
                // 0x14 is the 0x148th byte of the MockHugeStream (it is deterministic).
```

noahfalk (approved):
LGTM

src/FastSerialization/StreamReaderWriter.cs
josalem: `StreamReaderAlignment` is an enum where `EightBytes` is 8, etc.

src/FastSerialization/FastSerialization.cs
brianrob: It's a good point.  The thing that I actually am thinking is that `SerializationConfiguration` gets used on both serialization and deserialization, and so I'm wondering if a deserialization-only property makes sense here.  I think I would prefer this as a parameter with a specified default value.  @josalem what do you think about just putting this as an optional parameter on the second constructor here?

src/FastSerialization/StreamReaderWriter.cs
```diff
@@ -766,7 +783,7 @@ protected override void Dispose(bool disposing)
             base.Dispose(disposing);
         }
 
-        internal /*protected*/  const int align = 8;        // Needs to be a power of 2
+        internal /*protected*/  readonly int align = 8;        // Needs to be a power of 2
```
brianrob, line 786: nit: Can you move this down with the other non-const fields?

brianrob (commented):
Overall, LTGM.  A couple of small things.

src/FastSerialization/FastSerialization.cs
josalem: I can certainly move it. The first version I wrote up did it as a parameter. Thoughts on leaving it as an enum to restrict values?

brianrob (approved):

Issue #1558, opened 2022-01-26
[AutomatedAnalysis] Expand the mechanisms of discovery by AnalyzerResolver


WilliamXieMSFT:
Currently the AnalyzerResolver looks in an "Analyzer" folder relative to the TraceEvent.dll for analyzer dlls and configuration xmls.

Potentially read from environment variable, MEF, or other mechanisms.

brianrob:
Thanks for filing this @WilliamXieMSFT.  I'm not 100% sure what the mechanism should look like, but off the top of my head, I think it should be something like an interface that can be implemented by anyone, and then maybe we have some pre-build ones.  One thing in my mind that I haven't determined is how the resolver is related to the analyzers themselves.  For example, is there an attribute that must exist at the assembly level for all Analyzer DLLs so that anyone that writes a resolver can find them?  Or do we need to support multiple resolvers running for a given instance of `AutomatedAnalysisManager` because the analyzers are written to be discovered by a particular resolver.  My gut says the former, but I haven't formalized the design in my head yet.

WilliamXieMSFT:
Perhaps we don't need to have a resolver interface. The current resolver's design which searches by assembly-level `AnalyzerProviderAttribute` can be extended to include capabilities such as CPU threshold, memory limits, etc. The `AutomatedAnalysisManager` can then be updated so clients can configure things such as where to probe for analyzers and what capabilities they're targeting. 

This allows analyzer authors to not worry about writing an additional resolver. Clients that run analysis are responsible for picking which analyzers to run.

Thoughts?

brianrob:
Fixed via #1589.

PR #1561, opened 2022-01-28
[Automated Analysis] Add support for custom probe path

WilliamXieMSFT:
Current behavior was probing relative to the TraceEvent.dll. This change allows whoever is instantiating the analysis manager to pass in their own path to search for analyzers and configuration.


WilliamXieMSFT:
#1558

clguiman (approved):

WilliamXieMSFT:
The CI build is weirdly failing on `Debug.Assert/Fail` & `Trace.Assert/Fail` not throwing exceptions. I don't repro this locally.

PR #1563, opened 2022-02-05
POC for supporting downloading symbols from ADO

zivcaspi:


src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -1043,6 +1043,19 @@ internal bool GetPhysicalFileFromServer(string serverPath, string pdbIndexPath,
                             // So set it (this is what the symsrv code on Windows sets).   On NetStandard1.6 we give up since we dont' have this API available.  
                             req.UserAgent = "Microsoft-Symbol-Server/6.13.0009.1140";
 #endif
+
+#if NET45
+                            // Ugly support for ADO sources
+                            if (fullUri.StartsWith("https://") && fullUri.Contains("artifacts.visualstudio.com/"))
```
brianrob, line 1049: Can you share what the URL structure looks like?  If I recall correctly its something like "https://<something>.artifacts.visualstudio.com/..."  If so, I'm thinking it would be good to change this to `fullUri.Host.EndsWith("artifacts.visualstudio.com", StringComparison.OrdinalIgnoreCase)`

src/TraceEvent/Symbols/AdalAuth.cs
```diff
@@ -0,0 +1,59 @@
...
+                    return null;
+                }
+                s_lastAttempt = now;
+            }
+
+            var tenantId = "microsoft.com";
+            var authority = $"https://login.microsoftonline.com/{tenantId}";
+            var resourceId = "499b84ac-1321-427f-aa17-267ca6975798"; // ADO
+
+            var clientId = "1950a258-227b-4e31-a9cf-717495945fc2"; // Azure PowerShell
```
brianrob, line 38: Is there something that we need to do to register a clientId for PerfView, and then plumb that through.  I'm a bit weary of merging this as is, since it means we look like Azure PowerShell.

src/TraceEvent/Symbols/SymbolReader.cs
zivcaspi: Yes, but we still need to check the scheme (so as to never send a security token to an unverified target or through an unencrypted channel)

src/TraceEvent/Symbols/AdalAuth.cs
zivcaspi: You can certainly register PerfView as a 1st-party client application (because you're in Microsoft). Note that the client itself is _not_ secret ("public client" in AAD terms), so that 1st-"partiness" is really about having the app available to all AAD tenants, without needing registration to a particular tenant.

(I don't really know why they built their system like that -- anybody can copy the Azure PowerShell app ID, or any other app ID for that matter, and it really provides 0 security. Oh well.)

src/TraceEvent/Symbols/AdalAuth.cs
```diff
@@ -0,0 +1,59 @@
...
+            var redirectUri = new Uri("urn:ietf:wg:oauth:2.0:oob");
+
+            var authContext = new Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContext(authority);
+
+            try
+            {
+                var token = authContext.AcquireTokenAsync(resourceId, clientId, redirectUri, new PlatformParameters(PromptBehavior.Auto)).Result;
+                lock (s_lock)
+                {
+                    s_accessToken = token.AccessToken;
```
pharring, line 48: You also need to record the expiration time so that you know when to refresh the token.

src/TraceEvent/Symbols/AdalAuth.cs
```diff
@@ -0,0 +1,59 @@
...
+            var resourceId = "499b84ac-1321-427f-aa17-267ca6975798"; // ADO
+
+            var clientId = "1950a258-227b-4e31-a9cf-717495945fc2"; // Azure PowerShell
+            var redirectUri = new Uri("urn:ietf:wg:oauth:2.0:oob");
+
+            var authContext = new Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationContext(authority);
+
+            try
+            {
+                var token = authContext.AcquireTokenAsync(resourceId, clientId, redirectUri, new PlatformParameters(PromptBehavior.Auto)).Result;
```
pharring, line 45: My suggestion would be to make _this_ method async (i.e. `var token = await authContext.AcquireTokenAsync(...)`) and put the onus on the caller to make the "sync over async" trade-off. In this case, it looks like the caller is already doing "sync over async", but _this_ code doesn't have to be part of the problem.

zivcaspi: Done

src/TraceEvent/Symbols/AdalAuth.cs
zivcaspi: Done

src/TraceEvent/Symbols/SymbolReader.cs
brianrob: Agreed.

src/TraceEvent/Symbols/AdalAuth.cs
brianrob: I would like to have a TraceEvent ID.  If you can point me to documentation internally, I will set this up.

src/TraceEvent/Parsers/KernelTraceEventParser.cs
```diff
@@ -7597,7 +7597,7 @@ protected internal override Delegate Target
         }
         protected internal override void Dispatch()
         {
-            Debug.Assert(!(Version >= 1 && EventDataLength < HostOffset(52, 1)));
+            // Debug.Assert(!(Version >= 1 && EventDataLength < HostOffset(52, 1)));
```
brianrob, line 7600: Once you're ready to merge, please undo the changes to the asserts.  I am trying to work on this separately.

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -1043,6 +1043,19 @@ internal bool GetPhysicalFileFromServer(string serverPath, string pdbIndexPath,
                             // So set it (this is what the symsrv code on Windows sets).   On NetStandard1.6 we give up since we dont' have this API available.  
                             req.UserAgent = "Microsoft-Symbol-Server/6.13.0009.1140";
 #endif
+
+#if NET462
```
brianrob, line 1047: NET462 makes this work for PerfView, but other TraceEvent consumers that run against .NET Core can't use it.  The identity packages appear to support various netstandard versions.  Any reason we can't relax this?

src/TraceEvent/TraceEvent.csproj
```diff
@@ -63,6 +63,7 @@
     <Reference Include="System.Net.Http" />
     <Reference Include="System.Web" />
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
+    <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="5.2.9" />
```
brianrob, line 66: Per https://www.nuget.org/packages/Microsoft.IdentityModel.Clients.ActiveDirectory/, this package is no longer maintained, and https://www.nuget.org/packages/Microsoft.Identity.Client/ is the recommended alternative.  Would you be able to convert to using the new recommended package?

brianrob:
@zivcaspi thanks for contributing this.  This is great, and in fact, I've recently received a request for such functionality, so it is timely as well.

src/TraceEvent/Symbols/AdalAuth.cs
zivcaspi: You can contact the 1st Party Engagement Team at firpet@microsoft.com

The process is to go to [https://firstpartyportal.msidentity.com](https://firstpartyportal.msidentity.com) (from SAW), and start “Create Application” flow.

src/TraceEvent/Symbols/SymbolReader.cs
buckleyGI: There is also a second url that is equivalent
https://<orgName>.artifacts.visualstudio.com/defaultcollection/_apis/symbol/symsrv/
https://artifacts.dev.azure.com/<orgName>/_apis/symbol/symsrv/

I remember noting it down from the DevOps documentation. Both work for me.

pharring:
@zivcaspi, thank you for the contribution. Your POC inspired the work in #1673. Unless you see a reason to keep this around, please can you close this?

zivcaspi:
Closed as the work was done in another branch in a better way :)

PR #1570, opened 2022-02-16
Reduce the number of events and stack walks required for /GCCollectOnly

brianrob:
- Reduce the number of events collected for /GCCollectOnly.
 - Add /GCTriggeredStacks to enable stack collection for each GC.


Maoni0 (approved):
awesome, thanks!!

PR #1579, opened 2022-03-03
Upgrading to .NET framework 4.6.2

cshung:
This is the same as https://github.com/microsoft/perfview/pull/1555 targetting feature/perfview-3.0


brianrob (approved):

MagicAndre1981:
why don't you target .net 6.0 to get more perf fixes from the runtime? Is Perfview using .net framework only API?

brianrob:
This is certainly something that could be considered, but would need to be evaluated to determine how much work it is.

MagicAndre1981:
> This is certainly something that could be considered, but would need to be evaluated to determine how much work it is.

ok, I created an issue to track this #1594

PR #1580, opened 2022-03-07
Added a GUI warning popup when closing PerfView during rundown/zipping

### LINKED ISSUE #1293: Provide GUI warning popup when closing PerfView during rundown/zipping
Hi all,

first of all: Thanks for PerfView, it has recently become a very valuable tool in our company! 👍 

We deploy a very large .NET application to our customers and hence we (developers) don't run PerfView but instead our customers do. This is an issue since they very often don't wait for the rundown/zipping. Of course we try to give as detailed instructions as possible and tell them to await the rundown (while it blinks in the bottom right corner) but it's often in vain. About 30% of the ETL files are prematurely aborted :/

Hence: Would it be possible to have a warning window if the user wants to quit PerfView while the rundown/zipping is going on?

Cheers from Germany

nicnike:
We prompt the user to ensure that he's aware of the cancelation.

fixes #1293

### DIFF
--- a/src/PerfView/CommandProcessor.cs
+++ b/src/PerfView/CommandProcessor.cs
@@ -52,6 +52,7 @@ namespace PerfView
 
         public bool CommandLineCommand;
         public bool CollectingData;
+        public bool RundownActive;
         public TextWriter LogFile
         {
             get { return m_logFile; }
@@ -1162,6 +1163,7 @@ namespace PerfView
 
         public void Stop(CommandLineArgs parsedArgs)
         {
+            RundownActive = true;
             if (parsedArgs.DataFile == null)
             {
                 parsedArgs.DataFile = "PerfViewData.etl";
@@ -1344,6 +1346,7 @@ namespace PerfView
 
             DateTime stopComplete = DateTime.Now;
             LogFile.WriteLine("Stop Completed at {0}", stopComplete);
+            RundownActive = false;
         }
         public void Mark(CommandLineArgs parsedArgs)
         {
--- a/src/PerfView/CommandProcessor.cs
+++ b/src/PerfView/CommandProcessor.cs
@@ -52,7 +52,7 @@ namespace PerfView
 
         public bool CommandLineCommand;
         public bool CollectingData;
-        public bool RundownActive;
+        public bool StopInProgress;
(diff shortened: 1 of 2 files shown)


src/PerfView/CommandProcessor.cs
```diff
@@ -1344,6 +1346,7 @@ public void Stop(CommandLineArgs parsedArgs)
 
             DateTime stopComplete = DateTime.Now;
             LogFile.WriteLine("Stop Completed at {0}", stopComplete);
+            RundownActive = false;
```
brianrob, line 1349: I would recommend doing this in a finally block, and put the rest of the method in a try block. This ensures that the value of `RundownActive` doesn't become stale if an exception is thrown, which can happen in this path.

src/PerfView/MainWindow.xaml.cs
```diff
@@ -1097,12 +1097,18 @@ public MainWindow()
...
                 if (StatusBar.IsWorking)
                 {
-                    var result = MessageBox.Show(this, "Rundown is still in progress. If you abort now the collected trace will be unusable. Are you sure you want to abort?", "Collecting data in progress", MessageBoxButton.YesNo);
-                    if (result == MessageBoxResult.No)
+                    if (App.CommandProcessor.RundownActive)
                     {
-                        e.Cancel = true;
-                        return;
+                        var result = MessageBox.Show(this,
+                            "Rundown is still in progress. If you abort now the collected trace will be unusable. Are you sure you want to abort?",
```
brianrob, line 1103: This is more than just rundown, so I'd recommend going with something like "Closing PerfView while the trace is being processed will result in a trace that is unusable if copied off of this machine.  Would you still like to close PerfView?"  Will that be a reasonable prompt for the problem that you're trying to solve?

brianrob: Also recommend changing "RundownActive" to "StopInProgress".

brianrob (commented):
Overall, looking good.  A few small requests.

nicnike:
Thanks for the review. 
I have implemented the changes and added another commit.

brianrob (approved):
LGTM.  Thanks!

PR #1584, opened 2022-03-09
Add support for OSR jitted code

AndyAyersMS:
Parser support for OSR.

OSR methods jit on foreground threads.


AndyAyersMS:
We have been setting the proper info in the ETW records for quite a while now. But OSR has always been an experimental feature, so I never bothered to update any tooling.

Now we're getting close to enabling it by default for some architectures. Methods jitted this way should start showing up in .NET 7 Preview 3 or 4.

Without this change, OSR methods show up as "ReadyToRun" in tiering annotated views, and mess up the jit stats report.

AndyAyersMS:
@kouvel can you PTAL -- I think you added the tiering aware views.
cc @brianrob

kouvel (approved):

brianrob (approved):
:shipit:

PR #1585, opened 2022-03-09
Added Attribute UseCulture("en-US") to prevent another time format from being taken

### LINKED ISSUE #1581: Test GenerateBaselineXmlFiles fails when running it with a german os
When running the tests with a German operating system, both GenerateBaselineXmlFiles tests fail because the two files have different time formats. One has 12h format and the other the 24h format.

A possible solution might be attribute [UseCulture("en-US")]

2022-03-07_14h20_47

nicnike:
To prevent the test GenerateBaselineXmlFiles from using two different time formats, the attribute UseCulture("en-US") is added.
fixes #1581


brianrob (approved):
LGTM.  Thanks!

PR #1588, opened 2022-03-14
Non-blocking EventPipeEventSource Ctor

### LINKED ISSUE #1172: EventPipeEventSource constructor blocks until the first event is read
Moving https://github.com/dotnet/diagnostics/issues/1060 to here because this repo is a better place to keep track of this issue:

From @davmason:

If you follow the sample code on how to use EventPipeEventSource but replace the provider with a provider that doesn't produce any events, it will block forever in the constructor to EventPipeEventSource.
```csharp
    var providers = new List<EventPipeProvider>()
    {
        new EventPipeProvider("Microsoft-Windows-DotNETRuntime",
            EventLevel.Informational, (long)ClrTraceEventParser.Keywords.GC)
    };

    var client = new DiagnosticsClient(processId);
    using (var session = client.StartEventPipeSession(providers, false))
    {
        var source = new EventPipeEventSource(session.EventStream);

        source.Clr.All += (TraceEvent obj) => {
            Console.WriteLine(obj.EventName);
        }
        try
        {
            source.Process();
        }
        // NOTE: This exception does not currently exist. It is something that needs to be added to TraceEvent.
        catch (EventStreamException e)
        {
            Console.WriteLine("Error encountered while processing events");
            Console.WriteLine(e.ToString());
        }
    }
```
It will block on this line: https://github.com/microsoft/perfview/blob/cb1943b8ded584e136e140746e5771ca63668c5a/src/TraceEvent/EventPipe/EventPipeEventSource.cs#L68

This seems like a pitfall, and it's not clear what is happening. I've used that code pattern lots before and it's never failed, then all of a sudden it hangs my app and I have to do a lot of digging to understand why. I imagine customers will run in to it as well.

josalem:
This PR makes the constructor for `EventPipeEventSource` non-blocking. Previously, the constructor would block until data was sent on the stream. This brings `EventPipeEventSource` closer in behavior to other similar APIs in the TraceEvent library.

~One issue that isn't solved by this PR (yet) is that we need to handle the case where users try to get metadata about the source before data arrives. Before this PR, users were guaranteed to have that information when the ctor returned. Now, this is only true for opening a completed file, hence the special casing for files. Currently, this PR makes no attempt to stop users from reading metadata properties. _I am hoping for some feedback on whether this should throw an exception or now. Until this is decided, I'll leave this in draft mode._~ I'm going to leave this alone for now, and address it if users are hitting this behavior. The only use case for these Properties seems to be covered fine by differentiating streaming vs non-streaming.

fixes #1172

### DIFF
--- a/src/TraceEvent/EventPipe/EventPipeEventSource.cs
+++ b/src/TraceEvent/EventPipe/EventPipeEventSource.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Diagnostics.Tracing
         {
         }
 
-        private EventPipeEventSource(PinnedStreamReader streamReader, string name, bool isLive)
+        private EventPipeEventSource(PinnedStreamReader streamReader, string name, bool isStreaming)
         {
             _deserializerIntializer = () =>
             {
@@ -83,7 +83,7 @@ namespace Microsoft.Diagnostics.Tracing
             osVersion = new Version("0.0.0.0");
             cpuSpeedMHz = 10;
 
-            if (!isLive)
+            if (!isStreaming)
                 _deserializer = _deserializerIntializer();
 
             EventCache = new EventCache();
--- a/src/TraceEvent/EventPipe/EventPipeEventSource.cs
+++ b/src/TraceEvent/EventPipe/EventPipeEventSource.cs
@@ -848,8 +848,6 @@ namespace Microsoft.Diagnostics.Tracing
         private StreamLabel _endOfEventStream;
 #endif
         private Dictionary<int, EventPipeEventMetaDataHeader> _eventMetadataDictionary = new Dictionary<int, EventPipeEventMetaDataHeader>();
-        private PinnedStreamReader _streamReader;
-        private string _name;
         private Deserializer _deserializer;
         private Dictionary<TraceEvent, DynamicTraceEventData> _metadataTemplates =
             new Dictionary<TraceEvent, DynamicTraceEventData>(new ExternalTraceEventParserState.TraceEventComparer());
--- a/src/TraceEvent/EventPipe/EventPipeEventSource.cs
+++ b/src/TraceEvent/EventPipe/EventPipeEventSource.cs
@@ -28,55 +28,63 @@ namespace Microsoft.Diagnostics.Tracing
     /// </summary>
     public unsafe class EventPipeEventSource : TraceEventDispatcher, IFastSerializable, IFastSerializableVersion
     {
-        public EventPipeEventSource(string fileName) : this(new PinnedStreamReader(fileName, 0x20000), fileName)
+        public EventPipeEventSource(string fileName) : this(new PinnedStreamReader(fileName, 0x20000), fileName, false)
         {
         }
 
         public EventPipeEventSource(Stream stream)
-            : this(new PinnedStreamReader(stream, alignment: StreamReaderAlignment.OneByte), "stream")
+            : this(new PinnedStreamReader(stream, alignment: StreamReaderAlignment.OneByte), "stream", true)
         {
         }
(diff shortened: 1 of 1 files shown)


noahfalk:
> Currently, this PR makes no attempt to stop users from reading metadata properties

I'd say throwing is desirable if the values aren't initialized yet, but if it was a significant hassle to implement I don't think it would be awful to return some default 0/null/empty string values instead. What do you think @brianrob?

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -28,55 +28,63 @@ namespace Microsoft.Diagnostics.Tracing
...
         }
 
         public EventPipeEventSource(Stream stream)
-            : this(new PinnedStreamReader(stream, alignment: StreamReaderAlignment.OneByte), "stream")
+            : this(new PinnedStreamReader(stream, alignment: StreamReaderAlignment.OneByte), "stream", true)
         {
         }
 
-        private EventPipeEventSource(PinnedStreamReader streamReader, string name)
+        private EventPipeEventSource(PinnedStreamReader streamReader, string name, bool isLive)
```
noahfalk, line 40: Naming suggestion:
```suggestion
        private EventPipeEventSource(PinnedStreamReader streamReader, string name, bool isStreaming)
```

josalem:
> I'd say throwing is desirable if the values aren't initialized yet

Agreed.

There are `Debug.Assert`s scattered around that check whether this metadata is set before certain operations. These are mostly buried in the `TraceLog` APIs from what I saw. These caused the file-based tests to fail until I special cased the file-based ctor.

Depending on @brianrob's thoughts, I'll look at making these properties throw if `Process` hasn't been called yet.

brianrob:
I would tend to agree that if there is metadata that can't be filled before returning from the constructor, then throwing is a good idea.  If you look at how we handle realtime ETW sessions, the session is created and setup before you get the source from it, which means that the source can make calls to the session to get metadata.  Would such a pattern help here?  If not, I would say throwing is the next best thing.

josalem:
> Would such a pattern help here?

I think yes, however, the API for configuring, starting, and stopping an EventPipe session is in the diagnostics client library and _not_ in TraceEvent. That makes it a little more complicated to make this pattern happen.

Right now, the flow for "trace on startup" is this:

1. [Diag Client Lib] Create configuration (list of providers, buffersize, etc.)
2. [Diag Client Lib] Start EventPipe session (get `Stream` object)
3. [TraceEvent lib] Create `EventPipeEventSource` using `EventPipeEventSource(Stream stream)` ctor
4. [TraceEvent lib] Call `EventPipeEventSource.Process()`
5. [Diag Client lib] Resume target
6. [Diag Client lib] Stop EventPipe session

In the "trace on startup" scenario, we need to wait for the target process to resume before the session starts. Without this change, the ctor call in step 3 would block until we resume the target. After this change, the ctor would complete, but the metadata wouldn't be populated yet. Any usage of that `EventPipeEventSource` before resuming the target could cause errors in other parts of TraceEvent regardless of whether `EventPipeEventSource.Process` has been called. The metadata should only be present after step 5.

brianrob:
Gotcha.  So this becomes a bit more complicated.

I think it is OK to simplify and just throw in APIs that don't have data yet.  Just wanted to make sure that there wasn't another pattern that might help.

brianrob:
@josalem, I see that you just pushed another commit here.  Please let me know when this is ready for another review, or is ready for merge.  Thanks.

josalem:
I think this is ready for final review and merge at your discretion.

brianrob (approved):

PR #1589, opened 2022-03-15
Refactor Analyzer Resolution

brianrob:
- Resolvers provide one or more assemblies that contain Analyzers by implementing the `AnalyzerResolver.Resolve` method and calling `ConsumeAssembly` for each assembly.
 - The infrastructure will look for the registered `IAnalyzerProvider` in each assembly, instantiate it, and ask it to provide a list of `Analyzer` instances.
 - The infrastructure will call `AnalyzerResolver.OnAnalyzerLoaded` for each `Analyzer`, giving the resolver an opportunity to decide if the `Analyzer` should be run.


WilliamXieMSFT (approved):
:shipit:

PR #1609, opened 2022-04-07
Ignore the Microsoft-Windows-DotNETRuntime Dynamic Manifest

### LINKED ISSUE #1591: StopOnBGCFinalPauseOverMsec incurs an exception
with this command 

`PerfView /nogui /accepteula /StopOnBGCFinalPauseOverMsec:30 /Process:corerun /DelayAfterTriggerSec:0 /CollectMultiple:1 /KernelEvents=ThreadTime+Memory+VirtualAlloc /ClrEvents:GC+Stack /BufferSize:3000 /CircularMB:3000 /Merge:TRUE /ZIP:True collect`

I'm seeing this exception

`Only allowing process corerun with ID 61308 stop the trace.
Collecting   10 sec: Size= 94.5 MB. Error: Exception thrown during monitoring: Unable to cast object of type 'Microsoft.Diagnostics.Tracing.Parsers.DynamicTraceEventData' to type 'Microsoft.Diagnostics.Tracing.Parsers.Clr.GCSuspendEETraceData'.`

with the latest ver of perfview.

brianrob:
Starting with .NET 6, the runtime will emit the Microsoft-Windows-DotNETRuntime manifest into the event stream.  This results in two instances of the manifest:

 - The pre-compiled copy of the manifest: `ClrTraceEventParser`
 - The dynamic copy of the manifest which enables `DynamicTraceEventParser` to parse the CLR events

Normally, this would not be a problem, except that there is lots of code out there that knows that Microsoft-Windows-DotNETRuntime is not a dynamic provider, and so now that this provider manifest is written into the event stream, that code breaks.

This fix will cause `DynamicTraceEventParser` to ignore manifest events for the `Microsoft-Windows-DotNETRuntime` provider so that the previous behavior is restored.

Fixes #1591.

cc: @noahfalk, @josalem

### DIFF
--- a/src/TraceEvent/DynamicTraceEventParser.cs
+++ b/src/TraceEvent/DynamicTraceEventParser.cs
@@ -198,6 +198,14 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
                 return false;
             }
 
+            // Starting with .NET 6, the manifest is written into the event stream, which results in
+            // duplicate event dispatch.  To avoid this, filter out the Microsoft-Windows-DotNETRuntime provider
+            // if it is present in the event stream.
+            if (data.ProviderGuid == ClrTraceEventParser.ProviderGuid)
+            {
+                return false;
+            }
+
             // Look up our information. 
             List<PartialManifestInfo> partialManifestsForGuid;
             if (!partialManifests.TryGetValue(data.ProviderGuid, out partialManifestsForGuid))


noahfalk:
I think we had run into other issues relating to this in the past where the hard-coded provider information differed from the dynamic one and you get different behavior depending on which parser examines it. You might find some other issues in your backlog that can be duped to this one and marked fixed. No worries on my end either way and glad it is fixed!

PR #1610, opened 2022-04-09
Kernel based event can not be disabled on UI

iaalm:
When KernelBasedCheckBox is false, the default value is `KernelTraceEventParser.Keywords.Default`, which still has kernel events enabled.


iaalm:
@brianrob Could you help reviewing?

brianrob (approved):
LGTM.  Thanks!

PR #1613, opened 2022-04-18
Add docs on UNMANAGED_CODE_TIME and CPU_TIME

josalem:
fixes dotnet/diagnostics#1166


documentation/TraceEvent/TraceEventProgrammersGuide.md
```diff
@@ -936,3 +936,41 @@ Unfortunately when a provider registers with the operating system (the Win32 `Ev
...
+
+During a typical CPU usage investigation, you may want to ask about on-CPU vs off-CPU time to determine when your code is blocked waiting or actively doing work. Due to the limitation of only knowing about managed frames, CPU samples collected via EventPipe can't give exact on/off-CPU information. Instead, the TraceEvent library uses a heuristic to add pseudo-frames to the trave that indicate whether there are additional native frames on the stack or no. It uses relative time deltas between events and the state of the Garbage Collector at the time the sample was take to insert either `UNMANAGED_CODE_TIME` or `CPU_TIME` onto the stack in the "Thread Time" view or when exporting to SpeedScope or other formats. `UNMANAGED_CODE_TIME` represents stacks where there are one or more native frames after the last managed frame and that these frames may be blocked waiting _or_ actively on the CPU. `CPU_TIME` represents stacks where the last managed frame is the function currently on the CPU and doing work.
```
noahfalk, line 972: The last two sentences seem to contradict the 3rd from last.

> `UNMANAGED_CODE_TIME` represents stacks where there are one or more native frames after the last managed frame and that these frames may be blocked waiting _or_ actively on the CPU. `CPU_TIME` represents stacks where the last managed frame is the function currently on the CPU and doing work.

I read these sentences saying that what matters is whether there are native frames at the leaf of the stack.

> It uses relative time deltas between events and the state of the Garbage Collector at the time the sample was take to insert either `UNMANAGED_CODE_TIME` or `CPU_TIME` onto the stack in the "Thread Time" view or when exporting to SpeedScope or other formats

I read this sentence saying that what matters is the GC state and time deltas.

documentation/TraceEvent/TraceEventProgrammersGuide.md
```diff
@@ -936,3 +936,41 @@ Unfortunately when a provider registers with the operating system (the Win32 `Ev
 ## On the fly filtering AND Logging using `ETWReloggerTraceEventSource`
 
 TODO
+
+## `UNMANAGED_CODE_TIME` and `CPU_TIME` when analyzing CPU traces collected with EventPipe
```
noahfalk, line 940: This is useful for people targeting TraceEvent programmatically. For users using PerfView I'd expect them to look here instead: https://github.com/microsoft/perfview/blob/main/src/PerfView/SupportFiles/UsersGuide.htm. That page is the guide that is included in the PerfView tool where all the in-app help links resolve to.

mikelle-rogers:
@josalem how many people look at this repo? In the meeting we were talking about how perfview is not the tool we really try to advertise. I think it is important to have this documented somewhere people can see it because UNMANAGED_CODE_TIME and CPU_TIME show up on any traces (not just when the traces are analyzed in perfview).

documentation/TraceEvent/TraceEventProgrammersGuide.md
josalem: The phrasing is definitely confusing. The GC state is what informs the computer whether there were native leaf frames, and the relative time deltas tell us how long that was the case.

For example, if I have a function that is 50% managed work and 50% P/Invoked work, I'd end up with two sets of thread time stacks. One that is `MyFunction->CPU_TIME` corresponding to the aggregate of samples where the GC was in COOP mode and another that is `MyFunction->UNMANAGED_CODE_TIME` for when the GC was in PRE mode.

The duration of the `CPU_TIME` and `UNMANAGED_CODE_TIME` frames come from the relative deltas and which frame gets used comes from the GC mode during the sample.

Let me try to reword this so it's more obvious that all these things are describing the same set of information.

documentation/TraceEvent/TraceEventProgrammersGuide.md
josalem: Yes, I think that's a better spot for this.

josalem:
> I think it is important to have this documented somewhere people can see it because UNMANAGED_CODE_TIME and CPU_TIME show up on any traces (not just when the traces are analyzed in perfview).

These pseudo frames show up in 2 places and only if the trace is a nettrace file collected via EventPipe:

1. PerfView Thread Time view
2. Exported speedscope and chromium trace files from PerfView or `dotnet-trace`

While `dotnet-trace` can export in those formats and generate these pseudo-frames, all the code/API for doing so lives in TraceEvent.

We could link to this blurb in the user guide (after I move it there) from the `dotnet-trace` docs, but I'm not sure it makes sense to put it directly on `dotnet-trace`.

For example: "dotnet-trace uses the TraceEvent library for converting nettrace files to other formats. For details on the conversion, see: <link>".

mikelle-rogers:
@josalem 
> For example: "dotnet-trace uses the TraceEvent library for converting nettrace files to other formats. For details on the conversion, see: ".

That sounds great to me!

documentation/TraceEvent/TraceEventProgrammersGuide.md
noahfalk: I'd suggest deleting the part about GC state and time deltas in that case. I think the critical information for the reader is to understand what callstack produces what descriptive term. The techniques we use to determine the callstack are implementation details that appear unnecessary as long as they are accomplishing the job reliably.

brianrob:
> For example: "dotnet-trace uses the TraceEvent library for converting nettrace files to other formats. For details on the conversion, see: ".

Agreed, sounds great.

noahfalk (approved):
LGTM!

Issue #410, opened 2017-10-19
Group by field value


sharwell:
It would be nice if the **Refs-To** tab support a **Group by Field** command.

In order to group by a field, the children of the selected node are replaced by the set of values observed for that field in the selected instances. For example, suppose a **Refs-To** node indicates 100,000 instances (exclusive). If a user groups by a Boolean field, the children of this node would be **True** and **False**, and the sum of the exclusive counts for True and False would be 100,000.

This feature is the first step towards improving the ability to reason about the objects in a graph, as opposed to just looking at the shape of the graph.

adamsitnik:
I spoke today with @sharwell  about more details about this issue.

The issue is not only about implementing the group by value in PerfView UI.

The current problems are:
* current memory dump does not contain information about field values
* if we add the field values to dump, we are going to make it grow
* if the dump is too big we are going to hit the 32bit limit issue

So to implement this feature we would first of all need to solve the problem of 32 bit limit for the dump files. (#401)

An idea that Sam proposed: instead of adding the group by field value feature we could implement an interactive C# scripting window (similar to the one from VS), where we could expose the data from dump to the users. Users could write any script they want, which would make it more flexible solution
/cc @jorive

brianrob:
This is an interesting idea.  As @adamsitnik points out, this is not only a UI issue - the data to make this possible is not currently in the gcdump file.  There is also a question as to whether or not we want to include object contents in the file as there seems to be a fine line as to what is OK to include in a gcdump that is copied out of data centers or production deployments.

If there is a desire to implement such a thing, we should have a design discussion and work through the details.

PR #1621, opened 2022-04-20
Update ClrMD to 2.0

leculver:
Update ClrMD to 2.0.

- Updated to ClrMD 2.0.
- Removed special casing of dependent handles.  These are tracked by ClrMD now and will show up as normal edges in the graph.  We can do more here if this is confusing.
- Removed a lot of the heap walking retry code.  The benefit of not retrying is it's a LOT faster, and the old code was not entirely correct.  The downside is that when there is a legitimate heap walking failure then we can miss part of a segment.

### DIFF
--- a/src/PerfView/PerfView.csproj
+++ b/src/PerfView/PerfView.csproj
@@ -333,7 +333,7 @@
       <Link>Microsoft.Diagnostics.Tracing.TraceEvent.xml</Link>
       <Visible>False</Visible>
     </EmbeddedResource>
-    <EmbeddedResource Include="$(NuGetPackageRoot)\Microsoft.Diagnostics.Runtime\$(MicrosoftDiagnosticsRuntimeVersion)\lib\net45\Microsoft.Diagnostics.Runtime.dll">
+    <EmbeddedResource Include="$(NuGetPackageRoot)\Microsoft.Diagnostics.Runtime\$(MicrosoftDiagnosticsRuntimeVersion)\lib\net461\Microsoft.Diagnostics.Runtime.dll">
       <Type>Non-Resx</Type>
       <WithCulture>false</WithCulture>
       <LogicalName>.\x86\Microsoft.Diagnostics.Runtime.dll</LogicalName>
--- a/src/PerfView/PerfView.csproj
+++ b/src/PerfView/PerfView.csproj
@@ -8,6 +8,7 @@
     <StartupObject>PerfView.App</StartupObject>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <ExtrasEnableWpfProjectSetup>true</ExtrasEnableWpfProjectSetup>
+    <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
 
     <Description>PerfView</Description>
     <Company>Microsoft</Company>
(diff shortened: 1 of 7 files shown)


sharwell:
oh, super exciting! I worked on this in the past and ran into some cases where behavior different more than desired. Will take a look.

sharwell:
One of the main issues I ran into was the fact that ClrMD 2.0 doesn't include information about the names of static variables which form GC roots. I'm guessing based on the commit titles that this was resolved specifically in the second commit. Can you confirm this?

leculver:
> oh, super exciting! I worked on this in the past and ran into some cases where behavior different more than desired. Will take a look.

There's definitely some different behavior here.  The end result looks correct in the new version but the produced .gcdump uis not _exactly_ 1 to 1 with the previous version.  Still, the resulting gcdump seems correct and it's an order of magnitude faster and the resulting .gcdump analysis.

> One of the main issues I ran into was the fact that ClrMD 2.0 doesn't include information about the names of static variables which form GC roots. I'm guessing based on the commit titles that this was resolved specifically in the second commit. Can you confirm this?

Right, we now (correctly) don't report static variables as roots because that's not how the GC sees them.  To mimic the behavior of the previous version of perfview, we simply walk static variables and treat them as their own root kind in PerfView.  You can see that here:

https://github.com/microsoft/perfview/pull/1621/files#diff-1a87bf3e000fe3a9085c111836cb22024f272498e508552bbd5567a9956db099R1117-R1130

I've been chatting with Brian.  Overall this seems to give pretty good results.  The only major change between the old version and the new version is that I've stripped out the heap walking "retry" code.  That code was actually incorrect and resulted in actually dead (already collected) objects being reported.  Especially around allocation contexts.

In the future, if we are worried about missing segments of the GC heap when it's not walkable, we should put that enumeration code in ClrMD, something like `ClrSegment.EnumerateObjectsCarefully` or something similar, which would be slower but could "recover" when we have a region of unwalkable memory because we paused the process at a bad time (or there's just heap corruption).

leculver:
@brianrob This is ready for code review.  I think I'm going to leave it up to you to figure out the right way to embed ClrMD's dependencies.  I know it's just editing the .csproj file, but clearly I'm doing something wrong here.  If you don't mind, I will hand this over to you to finish that portion of the change...I got this 95% of the way there.  =)

brianrob:
@leculver, no problem at all.  I will take care of the dependencies.  Thank you!

sharwell:
I'll update my branch to use this as well and see how it fares

src/PerfView/PerfView.csproj
```diff
@@ -8,6 +8,7 @@
     <StartupObject>PerfView.App</StartupObject>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <ExtrasEnableWpfProjectSetup>true</ExtrasEnableWpfProjectSetup>
+    <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
```
sharwell, line 11: We probably shouldn't add this. The custom assembly resolver knows how to load AnyCPU assemblies from SupportFileDir, so there's no need to copy them into the amd64 and x86 directories. In other words, you want this:

https://github.com/microsoft/perfview/blob/65a4b1b706fea223e22a470813f2b94766a99762/src/PerfView/Utilities/SupportFiles.cs#L83-L87

Instead of what you are currently relying on:

https://github.com/microsoft/perfview/blob/65a4b1b706fea223e22a470813f2b94766a99762/src/PerfView/Utilities/SupportFiles.cs#L89-L94

src/PerfView/PerfView.csproj
```diff
@@ -255,13 +251,97 @@
       <Link>amd64\HeapDump.exe.config</Link>
       <Visible>False</Visible>
     </EmbeddedResource>
+    <EmbeddedResource Include="$(NuGetPackageRoot)\System.Buffers\4.5.1\lib\net461\System.Buffers.dll">
```
sharwell, line 254: ⚠️ You shouldn't assume this package is in this location. Add an explicit PackageReference for the package, and use the resulting generated property `$(PkgSystem_Buffers)`.

```suggestion
    <EmbeddedResource Include="$(PkgSystem_Buffers)\lib\net461\System.Buffers.dll">
```

sharwell:
Here's the first logs for before and after this change:

Before:

```
649.1s: Done Dumping all the segments.
649.1s: Done collecting data.
Number of bad objects during trace 0
649.1s: Finished heap dump 4/22/2022 12:02:29 PM
Object count 206415K less than 0K, Dumped all objects.
Dump Created from a .DMP file, no live statistics
The process being dumped is 64 Bit
Actual number of objects dumped = 206,415,277
Actual number of types = 14,618
Size of dumped objects 15,803.4 MB
```

After:

```
781.9s: Done Dumping all the segments.
781.9s: Dump RCW/CCW information781.9s: Done collecting data.
Number of bad objects during trace 0
781.9s: Finished heap dump 4/22/2022 1:55:31 PM
Object count 206332K less than 0K, Dumped all objects.
Dump Created from a .DMP file, no live statistics
The process being dumped is 64 Bit
Actual number of objects dumped = 206,332,404
Actual number of types = 13,051
Size of dumped objects 15,802.1 MB
```

It shows a slight regressing in extraction speed, but I haven't run the profiler to see if it's just something unintentional.

leculver:
Was the build retail?  It was universally faster on my machine running 2.0.  If it was retail it would be a good idea to put it under a profiler.

Also making sure that one run didn't hit the symbol server on a dump whereas the other run already pre-filled the symbol cache.

sharwell:
@leculver Can you double check the extraction of weak references?

image

sharwell:
Ah, found the issue here. The default Priority box is this:

> `v4.0.30319\%!->-1;v2.0.50727\%!->-1;[*local vars]->-1000;mscorlib!Runtime.CompilerServices.ConditionalWeakTable->-10000;[COM/WinRT Objects]->-1000000;[*handles]->-2000000;[other roots]->-2000000`

The item of interest here is:

> `[*handles]->-2000000`

This adjustment stopped behaving as expected because a node name changed:

image

sharwell:
> Was the build retail? It was universally faster on my machine running 2.0. If it was retail it would be a good idea to put it under a profiler.
> 
> Also making sure that one run didn't hit the symbol server on a dump whereas the other run already pre-filled the symbol cache.

Yes it's a release build. Do you have SeLockMemoryPrivilege set? The time here seems to be in the ArrayPool, and even when I tried to set AWE it overrides that setting by having a bad check here:

https://github.com/microsoft/clrmd/blob/5444e45c321f681bdd6b5fbfb4d2115de19c39a2/src/Microsoft.Diagnostics.Runtime/src/Windows/CachedMemoryReader.cs#L53-L57

This heap dump has more than 80% of it's ~7200 segments smaller than 64K, but the remaining ones are 200MB. The algorithm should check wasted space as a percentage of total size instead of checking segment counts.

leculver:
Ah interesting.  Yes I have SeLockMemoryPrivilege set, but I should probably turn it off for testing ClrMD things.  =)

> This heap dump has more than 80% of it's ~7200 segments smaller than 64K, but the remaining ones are 200MB. The algorithm should check wasted space as a percentage of total size instead of checking segment counts.

Feel free to suggest a better algorithm for choosing AWE or not, but we need to optimize for non-AWE use.  I'd be interested to take a look at the dump you are using to test this with if it's possible to send it to me offline?  I'll put it under a profiler and see if there's something we should be doing to make the fallback (non-AWE) caching better.

In any case, we shouldn't block this PR on perf unless it's really, really bad.  Updating to 2.0 is for correctness and bug fixing.  I can chase down perf fixes and push new versions once we get the baseline working.

> Can you double check the extraction of weak references dependent handles?

(This is something Brian and I chatted about via call, here's the context.)

As mentioned in the checkin notes, Dependent Handles are treated completely differently in this checkin.  Instead of explicitly enumerating them, they are now simply an edge in the graph.  IE if there is a dependent handle from A -> B then when we walk A's references, B will be enumerated as if A had a field pointing to B.  So while the `[Dependent Handles]` tag is gone, the references are still there and handled appropriately.

That comes from this line:  https://github.com/microsoft/perfview/pull/1621/files#diff-1a87bf3e000fe3a9085c111836cb22024f272498e508552bbd5567a9956db099R1490.  `foreach (var childObj in obj.EnumerateReferences(carefully: true, considerDependantHandles: true))`.

If we want better handling of dependent handles, I suggested to Brian that in the future we use `EnumerateReferencesWithFields` which will tell you that something came from a dependent handle (or the actual `m_foo` field name) and we could add that to the graph if it was interesting.  We agreed to leave that for a later date.

sharwell:
The heavy churn appears to come from this line (or similar):
https://github.com/microsoft/clrmd/blob/4318a412226509f9deb27d02a3e9bc701413938b/src/Microsoft.Diagnostics.Runtime/src/Implementation/ClrmdHeap.cs#L264

Since the implementation is eagerly loading the object type for the target of a reference (even though PerfView does not use this information), it prevents the efficient loading of pages during the heap walk.

sharwell:
The heap dump I'm testing is 15.6GiB with 206M objects. I can find a similar one in a feedback ticket if you don't have a large example already.

leculver:
We could add a simplified `EnumerateReferenceAddresses` that returns only object addresses and not the fully baked ClrObject.  That would be a simple fix in a future checkin.

sharwell:
I think I like the new behavior for dependent handles more than the old.

Eventually I think a new application would be better suited for heap dump investigation (PerfView's stack viewer is the wrong approach in a lot of ways), but inside the bounds of what it supports the current behavior is usable.

sharwell:
If we wanted PerfView to handle dependent handles, we could always use something like #410, or otherwise make field names the immediate children of a type node.

leculver:
We can simply disable the AWE reader for PerfView.  There's a flag to do that when setting caching options, iirc.  Would be good to track that OOM down and do something about it, but this should at least make perfview more predictable.

src/PerfView/PerfView.csproj
brianrob: @sharwell I just tried, this but it doesn't work.  It looks like there are some $pkg variables being created, but only for a subset of packages.

image

Do you know if there's something that I need to do other than add a `<PackageReference/>`?

src/PerfView/PerfView.csproj
brianrob: This is the PerfView resolver, but we're talking about HeapDump.exe here.  That said, your point is valid.  I have created a HeapDump directory for IL dependencies and updated the resolver for HeapDump.exe to look there.

brianrob:
It took me a while to get back to this, but I think we're getting close here.  @leculver do you want to make any other changes such as disabling the AWE reader prior to merge?

leculver:
If possible I'd like to merge this as-is and submit a new PR within a day.  I'm releasing a new copy of ClrMD (now 2.1!) which will have a fix for Sam's reported `EnumerateReferenceAddresses` issue and switching off AWE.  If you would prefer I do that in this checkin I can, but it may take longer.  =)

brianrob:
We can do this in two separate PRs - that is fine.  Assuming that 2.1 will be an easier transition than 2.0?

leculver:
Yes, it's just a version bump with some minor fixes to use the new `EnumerateReferenceAddresses` + turn off AWE.

brianrob:
Sounds good.  I'll get this merged shortly.

Issue #1629, opened 2022-05-07
Feature Request - a new Filtering capability for the Events view


Maoni0:
The Events view often comes in handy for in-depth perf analysis. However, it can be cumbersome to use when you want to look at events in detail - often you want to do this based on some criteria. For example, if I want to look at CSwitch events in depth along with some GC events that happen on a certain processor, it would be very useful to only show only those events with that ProcessorNumber. Or if I'm only interested in a certain thread and how it gets switched in and out. Of course there are tools that would show you an already processed view of the CSwitch/ReadyThread events. But that processed view may not be exactly what you want and it would be difficult to change it. So being able to filter on the raw events (to be perfectly accurate, the Events view is of course also a processed view but with very little processing) definitely has its own advantages when it comes to looking at events in greater detail. This proposal is to add such a filtering capability to the Events view. This will mostly be used in the "Columns to Display" text box but can also be used in the "Find" text box. I will give a mockup, scenario description and example usage below. 

1) CSwitch analysis –
![filter-Events-view](https://user-images.githubusercontent.com/10837357/167238596-da8ad5bc-ecb6-4989-b5db-166ffcb5f33a.jpg)

I put the filtering expression in []. C++ expressions seem fine for this purpose. Really all we need is some standard ==, >, >=, <, <=, and || and &&. We could consider supporting the “contains” operator. 

If I want to make sure the thread id 1000 is from ABC.exe and not some other process I could do

[((OldThreadID == 1000) && (OldProcessName=”ABC”)) || ((NewThreadID == 1000) && (OldProcessName=”ABC”))] OldThreadID OldProcessName NewThreadID NewThreadPriority OldThreadWaitReason

There may be fields named the same in multiple events so if you want to distinguish the same filed between different you could do 

[(CSwitch::ThreadID == 1000) || …]

So qualify the field with the name of the event (our event names are in general pretty long so we might want to just use the last part which is sufficient for 99% of the cases. 

This gives us a view of how a thread is executing. You could also for example filter based on NewThreadWaitTime >= x to see when it was ready but took a long time to actually got to run. 

Another typical usage for this scenario is to filter based on the processor number so you know the detailed view of what’s running on a specific processor (or multiple specific processors). The Processor view is good for aggregation but sometimes I’d like to see the detail.

UI wise, I think I’d be fine with it if it’s easier to just add another textbox as the Filter for events to display so we don’t have a lot of text in this “Columns To Display” textbox.

[...]

PR #1632, opened 2022-05-18
Added the ability to filter events based on boolean expression(s) in the Events Window.

MokoSan:
## Summary

Added the ability to filter events in the event view using boolean expression(s). This PR Addresses a large part of https://github.com/microsoft/perfview/issues/1629 (CC: @Maoni0). In a nutshell, this PR allows the ability to specify a filter query (currently only) in the Columns To Display textbox to apply to all events so that we can easily filter the data to retrieve pertinent events.

## Design Details

In a nutshell, here's what's happening under the hood:

1. If the user specifies a filter query within ``[]`` in the "Columns To Display" textbox, we extract the intended query and create a ``FilterQueryExpressionTree`` out of it based on the string provided by the user.
    * A ``FilterQueryExpressionTree`` encapsulates the collection of ``FilterQueryExpression``s and the matching of the trace event on the collection of these expressions. 
    * ``FilterQueryExpression``s are of the following format: ``LeftOperand Operator RightOperand`` where:
        - ``LeftOperand`` can either be a name of a property to search on in any event e.g. ``ThreadID`` or be specified specifically for a particular event name (we are simply checking if the event name passed in by the user is contained in the full ``EventName`` property of the TraceEvent) by using ``::``. For example: ``CSwitch::OldThreadID``.
        - ``Operator`` can be any of the following: ``!=, =, <, <=, >=, >, Contains``.
        - ``RightOperand`` can either be a ``string`` or a numeric that's used as a ``double``.
    * Multiple ``FilterQuerieExpression``s can be combined using either ``||`` or ``&&``. 
2. The algorithm used to deduce the logical expression is a modification of the [Shunting Yard Algorithm](https://en.wikipedia.org/wiki/Shunting_yard_algorithm) that takes in an infix expression, converts it into it's postfix (aka Reverse Polish Notation (RPN)) equivalent.
    * The design decision to convert to the RPN equivalent is based on making the boolean expression reduction more performant when we iterate through all TraceEvents to discern which ones to display.
3. Once the RPN expression is obtained, and the user invokes a UI update, the entire RPN expression is applied to the TraceEvents and a single boolean is returned on the basis of which, we decide to display or discard the particular event. 

While working on this PR, the solution was prototyped [here](https://github.com/MokoSan/BooleanExpressionParser). A number of unit tests have also been added with this PR.

## Current Implementation Limitations

1. The way I have designed this, I have imposed a restriction of 26 Filter Query Expressions. This is implementation specific and I'll be happy to change this, if need be. The trade-off here was to keep the construction of final RPN expression quick and easy to reduce in the more perf demanding part of the code.
2. To construct an expression, you _have_ to add a space between the Left Operand, Operator and Right Operand. e.g. ``Depth<1`` will not work - instead you'll need to enter: ``Depth < 1``. This is an implementation detail, as well; to construct ``FilterQueryExpression``s, I am splitting on empty spaces to distinguish between the different parts of the expression.
4. Right now, only _one_ query will be used to filter the event i.e. you can't specify multiple ``[..]``s in the Columns To Display textbox.

### Demo of Greater + Less Than or Equal to Operators:
https://user-images.githubusercontent.com/4951960/168762353-416a619c-0c2f-4b59-887c-213a7fe2f77f.mp4

### Demo of More Complex Operations With A Bunch of Nesting:
https://user-images.githubusercontent.com/4951960/168765726-f6644215-77e3-4f5a-a70f-a1359a82cdc0.mp4

Some of the queries I demo'd based on the ``GC/Start`` and ``GC/Stop`` events are:
1. ``[(Count > 10 && (Depth >= -1) && (Count <= 30) && (Count <= 30 || ProcessorNumber = 2))] Count ProcessorNumber Depth``
2. ``[(Count > 10  && (Count <= 30) && (Count <= 30 || ProcessorNumber = 2))] Count ProcessorNumber Depth``
5. ``[(Count > 10  && (Count <= 30) && (Count <= 30 || ProcessorNumber = 2))] Count ProcessorNumber``
6. ``[(Count > 10  && (Count <= 30)  && (Count <= 30 || ProcessorNumber = 2))] Count ProcessorNumber``

Happy to address any feedback!

### DIFF
--- a/src/CSVReader/EventSource.cs
+++ b/src/CSVReader/EventSource.cs
@@ -114,7 +114,8 @@ namespace EventSources
 
             if (matched.Success)
             {
-                filterQueryExpressionTree = new FilterQueryExpressionTree(matched.Value.Replace("[", "").Replace("]", ""));
+                var cleanedExpressionString = matched.Value.Replace("[", string.Empty).Replace("]", string.Empty);
+                filterQueryExpressionTree = new FilterQueryExpressionTree(cleanedExpressionString);
             }
 
             var regex = new Regex(@"\s*(\S+)\s*");
--- a/src/CSVReader/EventSource.cs
+++ b/src/CSVReader/EventSource.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using Microsoft.Diagnostics.Tracing.TraceUtilities.FilterQueryExpression;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text.RegularExpressions;
@@ -66,6 +67,10 @@ namespace EventSources
         /// Optionally you can provide a list of columns available.  
         /// </summary>
         public virtual ICollection<string> AllColumnNames(List<string> eventNames) { return null; }
+        /// <summary>
+        /// An optional filter query expression tree that'll be used to filter different events. 
+        /// </summary>
+        public FilterQueryExpressionTree FilterQueryExpressionTree { get; set; } = null;
 
         /// <summary>
         /// The number of fields that are NOT put in the 'rest' column.  Defaults to 4
@@ -92,14 +97,26 @@ namespace EventSources
         /// Utility function that takes a specification for columns (which can include *) and the raw list of
         /// all possible columns and returns the list that match the specification. 
         /// </summary>
-        public static List<string> ParseColumns(string columnSpec, ICollection<string> columnNames)
+        public static List<string> ParseColumns(string columnSpec, ICollection<string> columnNames, out FilterQueryExpressionTree filterQueryExpressionTree)
         {
+            filterQueryExpressionTree = null;
+
             if (string.IsNullOrWhiteSpace(columnSpec))
             {
                 return null;
             }
 
             var ret = new List<string>();
+
+            var escapedMatch = new Regex(@"\[[^\]]*\]");
+            var matched = escapedMatch.Match(columnSpec);
+            columnSpec = escapedMatch.Replace(columnSpec, string.Empty);
+
+            if (matched.Success)
+            {
+                filterQueryExpressionTree = new FilterQueryExpressionTree(matched.Value.Replace("[", "").Replace("]", ""));
+            }
+
             var regex = new Regex(@"\s*(\S+)\s*");
             var index = 0;
             for (; ; )
--- a/src/CSVReader/EventSource.cs
+++ b/src/CSVReader/EventSource.cs
@@ -95,7 +95,7 @@ namespace EventSources
 
         /// <summary>
         /// Utility function that takes a specification for columns (which can include *) and the raw list of
-        /// all possible columns and returns the list that match the specification. 
+        /// all possible columns and returns the list that match the specification. This function also extracts the event filter query and creates a FilterQueryExpressionTree.
         /// </summary>
         public static List<string> ParseColumns(string columnSpec, ICollection<string> columnNames, out FilterQueryExpressionTree filterQueryExpressionTree)
         {
--- a/src/PerfView/EtwEventSource.cs
+++ b/src/PerfView/EtwEventSource.cs
@@ -327,6 +327,15 @@ namespace PerfView
                         }
                     }
 
+                    if (FilterQueryExpressionTree != null)
+                    {
+                        var match = FilterQueryExpressionTree.Match(data);
+                        if (!match)
+                        {
+                            return;
+                        }
+                    }
+
                     cnt++;
                     if (MaxRet < cnt)
                     {
--- a/src/PerfView/EventViewer/EventWindow.xaml.cs
+++ b/src/PerfView/EventViewer/EventWindow.xaml.cs
@@ -1,6 +1,7 @@
 ﻿using EventSources;
 using Microsoft.Diagnostics.Symbols;
 using Microsoft.Diagnostics.Tracing.Etlx;
+using Microsoft.Diagnostics.Tracing.TraceUtilities.FilterQueryExpression;
 using Microsoft.Diagnostics.Utilities;
 using System;
 using System.Collections.Generic;
@@ -1205,8 +1206,23 @@ namespace PerfView
             }
 
             m_source.SetEventFilter(eventFilter);
-            m_source.ColumnsToDisplay = EventSource.ParseColumns(ColumnsToDisplayTextBox.Text, m_source.AllColumnNames(eventFilter), out var filterQueryExpresionTree);
-            m_source.FilterQueryExpressionTree = filterQueryExpresionTree;
+            try
+            {
+                m_source.ColumnsToDisplay = EventSource.ParseColumns(ColumnsToDisplayTextBox.Text, m_source.AllColumnNames(eventFilter), out var filterQueryExpresionTree);
+                m_source.FilterQueryExpressionTree = filterQueryExpresionTree;
+            }
+
+            // Appropriately log any filter query expression parsing issue to give as much info to the user.
+            catch (FilterQueryExpressionTreeParsingException fqpEx)
+            {
+                StatusBar.Log(fqpEx.Message);
+                m_source.FilterQueryExpressionTree = null; 
+            }
+            catch (FilterQueryExpressionParsingException fqepEx)
+            {
+                StatusBar.Log(fqepEx.Message);
+                m_source.FilterQueryExpressionTree = null; 
+            }
 
             for (int i = 0; i < m_userDefinedColumns.Count; i++)
             {
--- a/src/PerfView/EventViewer/EventWindow.xaml.cs
+++ b/src/PerfView/EventViewer/EventWindow.xaml.cs
@@ -1205,7 +1205,9 @@ namespace PerfView
             }
 
             m_source.SetEventFilter(eventFilter);
-            m_source.ColumnsToDisplay = EventSource.ParseColumns(ColumnsToDisplayTextBox.Text, m_source.AllColumnNames(eventFilter));
+            m_source.ColumnsToDisplay = EventSource.ParseColumns(ColumnsToDisplayTextBox.Text, m_source.AllColumnNames(eventFilter), out var filterQueryExpresionTree);
+            m_source.FilterQueryExpressionTree = filterQueryExpresionTree;
+
             for (int i = 0; i < m_userDefinedColumns.Count; i++)
             {
                 if (m_source.ColumnsToDisplay != null && i < m_source.ColumnsToDisplay.Count)
--- a/src/TraceEvent/TraceUtilities/FilterQueryExpression/FilterQueryExpressionTree.cs
+++ b/src/TraceEvent/TraceUtilities/FilterQueryExpression/FilterQueryExpressionTree.cs
@@ -5,7 +5,6 @@ namespace Microsoft.Diagnostics.Tracing.TraceUtilities.FilterQueryExpression
 {
     public sealed class FilterQueryExpressionTree
     {
-        private readonly string _originalExpression;
         private readonly FilterQueryExpression _simpleFilterQueryExpression;
         private readonly Dictionary<char, FilterQueryExpression> _expressionMap;
         private readonly string _postFixExpression;
@@ -26,14 +25,15 @@ namespace Microsoft.Diagnostics.Tracing.TraceUtilities.FilterQueryExpression
 
             else
             {
-                _originalExpression = expression;
+                OriginalExpression = expression;
 
                 // Prime Expression -> Post Fix
-                string primedExpression = ShuntingYard.PrimeExpression(_originalExpression, out var expressionMap);
-                _expressionMap = expressionMap;
+                string primedExpression = ShuntingYard.PrimeExpression(OriginalExpression, out var _expressionMap);
                 _postFixExpression = ShuntingYard.ToPostFix(primedExpression);
             }
         }
+
+        public string OriginalExpression { get; }
 
         public bool Match(TraceEvent @event)
         {
--- a/src/TraceEvent/TraceUtilities/FilterQueryExpression/FilterQueryExpressionTree.cs
+++ b/src/TraceEvent/TraceUtilities/FilterQueryExpression/FilterQueryExpressionTree.cs
@@ -3,6 +3,9 @@ using System.Collections.Generic;
 
 namespace Microsoft.Diagnostics.Tracing.TraceUtilities.FilterQueryExpression
 {
+    /// <summary>
+    /// Class responsible for encapsulating the collection of FilterQueryExpressions and the logic to combine them to deduce if we match on a particular trace event.
+    /// </summary>
     public sealed class FilterQueryExpressionTree
     {
         private readonly FilterQueryExpression _simpleFilterQueryExpression;
@@ -13,41 +16,52 @@ namespace Microsoft.Diagnostics.Tracing.TraceUtilities.FilterQueryExpression
         {
             if (string.IsNullOrEmpty(expression))
             {
-                throw new ArgumentException($"{nameof(expression)} is null. Invalid Filter Expression Tree.");
+                throw new FilterQueryExpressionTreeParsingException($"{nameof(expression)} is null.", expression); 
             }
 
-            // Base Case: Simple Expression Without Parentheses.
+            // Faster computation if a single Expression Without Parentheses i.e. one simple expression is provided.
             if (FilterQueryExpression.IsValidExpression(expression) &&
                 (!expression.Contains("(") || expression.Contains(")")))
             {
                 _simpleFilterQueryExpression = new FilterQueryExpression(expression);
             }
 
+            // Compute the PostFix (RPN) representation of the expression for easier logical deduction where perf matters.
             else
             {
                 OriginalExpression = expression;
-
-                // Prime Expression -> Post Fix
-                string primedExpression = ShuntingYard.PrimeExpression(OriginalExpression, out var _expressionMap);
+                string primedExpression = ShuntingYard.PrimeExpression(OriginalExpression, out var expressionMap);
+                _expressionMap = expressionMap; 
                 _postFixExpression = ShuntingYard.ToPostFix(primedExpression);
             }
         }
 
+        /// <summary>
+        /// Keep track of the original expression the user provided.
+        /// </summary>
         public string OriginalExpression { get; }
 
+        /// <summary>
+        /// This method is responsible for matching the entire user specified expression on an event.  
+        /// </summary>
+        /// <param name="event"></param>
+        /// <returns></returns>
         public bool Match(TraceEvent @event)
         {
+            // Handle the simple case where a single simple filter query expression is provided.
             if (_simpleFilterQueryExpression != null)
             {
                 return _simpleFilterQueryExpression.Match(@event);
             }
 
+            // Map each of the expressions to lower-case alphabet and the match result of the expression i.e. a boolean.
             Dictionary<string, bool> convertedExpressionMap = new Dictionary<string, bool>();
             foreach (var kvp in _expressionMap)
             {
                 convertedExpressionMap[kvp.Key.ToString()] = kvp.Value.Match(@event);
             }
 
+            // Conduct a Shunting Yard Match using the modified postfix expression.
             return ShuntingYard.Match(_postFixExpression, convertedExpressionMap);
         }
     }
--- a/src/TraceEvent/TraceUtilities/FilterQueryExpression/FilterQueryExpressionTree.cs
+++ b/src/TraceEvent/TraceUtilities/FilterQueryExpression/FilterQueryExpressionTree.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Diagnostics.Tracing.TraceUtilities.FilterQueryExpression
+{
+    public sealed class FilterQueryExpressionTree
+    {
+        private readonly string _originalExpression;
+        private readonly FilterQueryExpression _simpleFilterQueryExpression;
+        private readonly Dictionary<char, FilterQueryExpression> _expressionMap;
+        private readonly string _postFixExpression;
+
+        public FilterQueryExpressionTree(string expression)
+        {
+            if (string.IsNullOrEmpty(expression))
+            {
+                throw new ArgumentException($"{nameof(expression)} is null. Invalid Filter Expression Tree.");
+            }
+
+            // Base Case: Simple Expression Without Parentheses.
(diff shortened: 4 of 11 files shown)


brianrob:
@MokoSan, this is super cool.  I am playing around with this locally and starting to look at the code.

MokoSan:
> @MokoSan, this is super cool. I am playing around with this locally and starting to look at the code.

Thanks for taking a look! Please let me know if you have any feedback. I am also working on getting this functionality working from the Find text box; might submit that as a separate PR.

MokoSan:
Hi @brianrob, I think this PR is ready to be merged sans the seemingly unrelated build failures; have added the ability to do the query search for the Find text box and added a plethora of unit tests. I was unable to figure out how to rerun the CI checks. How can I do so? Thanks for your help!

src/CSVReader/EventSource.cs
```diff
@@ -90,16 +95,22 @@ public abstract class EventSource
 
         /// <summary>
         /// Utility function that takes a specification for columns (which can include *) and the raw list of
-        /// all possible columns and returns the list that match the specification. 
+        /// all possible columns and returns the list that match the specification. This function also extracts the event filter query and creates a FilterQueryExpressionTree.
         /// </summary>
-        public static List<string> ParseColumns(string columnSpec, ICollection<string> columnNames)
+        public static List<string> ParseColumns(string columnSpec, ICollection<string> columnNames, out FilterQueryExpressionTree filterQueryExpressionTree)
```
brianrob, line 100: This looks like a breaking change for any extension that might use the events view to show non-standard data.  Do you think that it is possible to remove the new parameter and depend on the `FilterQueryExpressionTree` property?  From my review so far, I'm not sure how to think about the property vs. this out parameter.

src/PerfView/EtwEventSource.cs
```diff
@@ -327,6 +327,15 @@ public override void ForEach(Func<EventRecord, bool> callback)
                         }
                     }
 
+                    if (FilterQueryExpressionTree != null)
```
brianrob, line 330: I like that this is completely opt-in and so if a user doesn't specify a filter, then none of this runs.

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -1108,20 +1112,64 @@ public bool Find(string pat)
...
+
+                    if (m_source.ColumnsToDisplay != null)
+                    {
+                        for(int displayFieldIdx = 0; displayFieldIdx < item.DisplayFields.Length; displayFieldIdx++)
+                        {
+                            data[m_source.ColumnsToDisplay[displayFieldIdx]] = item.DisplayFields[displayFieldIdx];
+                        }
+                    }
+
+                    foundItem = tree.Match(data, item.EventName);
```
brianrob, line 1148: I'm wondering if it makes sense to pass the entire `EventRecord` in here and look for ways to optimize here, such as not parsing `Rest` until we have to.

src/TraceEvent/TraceUtilities/FilterQueryExpression/FilterQueryExpressionTree.cs
```diff
@@ -0,0 +1,87 @@
...
+        public bool Match(TraceEvent @event)
+        {
+            // Handle the simple case where a single simple filter query expression is provided.
+            if (_simpleFilterQueryExpression != null)
+            {
+                return _simpleFilterQueryExpression.Match(@event);
+            }
+
+            // Map each of the expressions to lower-case alphabet and the match result of the expression i.e. a boolean.
+            Dictionary<string, bool> convertedExpressionMap = new Dictionary<string, bool>();
```
brianrob, line 58: Same here - wondering if we can cache this or otherwise re-use it.

brianrob (commented):
@MokoSan, thank you very much for building this.  It's great!  For the most part, I am pretty happy with how this is architected.  A few thoughts that I have as I reviewed the code and tried it out:

1. I like that this is fairly self-contained and doesn't run at all if you don't specify a filter.  That's great from a "if we mess something up we don't affect existing functionality" perspective.
2. I tried this out on my machine and it's super cool.  One thing I noticed is that the code is very whitespace particular.  For example, `[(ProcessorNumber = 2)]` works but `[(ProcessorNumber =2)]` throws an exception.  I also noticed that if you don't specify any fields then you get an empty `Rest` column.  Is that intentional?
3. I am a bit leary of perf.  I see lots of allocations of dictionaries, calls to `string.Split`, etc. and some N-squared algorithms (e.g. looking up each field in a set of field names, which results in a O(N) lookup per-field).  I am wondering how this behaves over a reasonably large trace.  I'm thinking that there is likely room for optimization, but I'm wondering if you've profiled any of this.  I want to be pragmatic and not force perfection here - just want to make sure that we aren't egregious.  :)

Happy to chat about these and work with you on plans going forward.  Thank you again for this - it's going to be a great addition.

MokoSan:
Hi @brianrob, thank you very much for taking the time to review the PR! 

Here are my responses:

> 2. I tried this out on my machine and it's super cool.  One thing I noticed is that the code is very whitespace particular.  For example, `[(ProcessorNumber = 2)]` works but `[(ProcessorNumber =2)]` throws an exception.  I also noticed that if you don't specify any fields then you get an empty `Rest` column.  Is that intentional?

Yes, this is intentional - I mentioned this in the first comment in the PR but let me know if I should be adding this anywhere else to better broadcast the usage. 

**Details**: The way I have this currently implemented is by splitting on empty spaces to deduce the operands and the operator. Do let me know if this should be changed. I have a couple of ideas but will require some more work.  

> 3. I am a bit leary of perf.  I see lots of allocations of dictionaries, calls to `string.Split`, etc. and some N-squared algorithms (e.g. looking up each field in a set of field names, which results in a O(N) lookup per-field).  I am wondering how this behaves over a reasonably large trace.  I'm thinking that there is likely room for optimization, but I'm wondering if you've profiled any of this.  I want to be pragmatic and not force perfection here - just want to make sure that we aren't egregious.  :)

Definitely will be happy to iterate on this further. I have addressed a majority of your comments with the latest update.

## Microbenchmarking 

I used Benchmark.NET to get details about the wall-clock time and allocations based on creating and matching simple and complex queries. 

NOTE: By clicking on the ▸in any of the sections below, you should have access to all the Benchmark.NET results and code.

### For Simple Queries 

For simple queries, the average wall-clock time it takes to create a filter query tree and to match on it by a simple query is ~1.5μs allocating ~1.2KB per call.

<details>

#### Results

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
11th Gen Intel Core i7-1165G7 2.80GHz, 1 CPU, 8 logical and 4 physical cores
  [Host]                                    : .NET Framework 4.8 (4.8.4510.0), X86 LegacyJIT  [AttachedDebugger]
  FilterQueryTreePerf_CreateAndMatch_Simple : .NET Framework 4.8 (4.8.4510.0), X86 LegacyJIT

Job=FilterQueryTreePerf_CreateAndMatch_Simple  InvocationCount=100  IterationCount=100  
UnrollFactor=1  

```
|                          Method |    SimpleExpression |       Mean |     Error |    StdDev |     Median |        Min |         Q1 |         Q3 |        Max | Allocated |
|-------------------------------- |-------------------- |-----------:|----------:|----------:|-----------:|-----------:|-----------:|-----------:|-----------:|----------:|
| **CreateAndMatch_SimpleExpression** |    **ThreadID = 1,001** | **1,245.3 ns** | **105.53 ns** | **311.17 ns** | **1,136.0 ns** |   **830.0 ns** | **1,019.5 ns** | **1,464.0 ns** | **1,961.0 ns** |     **901 B** |
| **CreateAndMatch_SimpleExpression** |    **ThreadID &gt; 1,000** | **1,599.4 ns** | **128.19 ns** | **373.93 ns** | **1,547.5 ns** |   **974.0 ns** | **1,304.0 ns** | **1,889.8 ns** | **2,407.0 ns** |     **901 B** |
| **CreateAndMatch_SimpleExpression** |   **ThreadID &gt;= 1,001** | **1,428.0 ns** |  **82.64 ns** | **239.75 ns** | **1,447.5 ns** | **1,033.5 ns** | **1,207.5 ns** | **1,641.5 ns** | **1,921.5 ns** |     **983 B** |
| **CreateAndMatch_SimpleExpression** | **ThreadID Contains 1** |   **962.4 ns** |  **27.08 ns** |  **76.82 ns** |   **945.0 ns** |   **831.0 ns** |   **903.0 ns** | **1,013.0 ns** | **1,200.0 ns** |     **983 B** |

#### Code

```csharp
    [SimpleJob(targetCount: 100, invocationCount: 100, id: "FilterQueryTreePerf_CreateAndMatch_Simple")]
    [MinColumn, Q1Column, MedianColumn, Q3Column, MaxColumn]
    [MemoryDiagnoser]
    [MarkdownExporter]
    public class FilterQueryTreePerf_CreateAndMatch_Simple
    {
        [Params("ThreadID = 1,001", "ThreadID >= 1,001", "ThreadID > 1,000", "ThreadID Contains 1")]
        public string SimpleExpression { get; set; }

        [Benchmark]
        public void CreateAndMatch_SimpleExpression()
        {
            FilterQueryExpressionTree tree = new FilterQueryExpressionTree(SimpleExpression);
            var data = new Dictionary<string, string>()
            {
                { "ThreadID", "1,001" },
            };
            tree.Match(data, "GC/Start");
        }
    }
```

</details>

### For Complex Queries

For more complex queries, the average wall-clock time varies fairly significantly depending on the query. E.g. for a fairly nested complex query such as: ``((GC/Start::Depth = 10 && (GC/Start::Depth <= 11 && (GC/Start::Depth <= 12 && (Depth <= 13 && (GC/Start::Depth <= 14))))))`` , we average around 12.5μs of wall-clock time and allocate 9KB of memory as opposed to a query like: ```"((Depth <= 10) && (Depth <= 20)) || ((Depth != 15) && (Depth > 5))"`` where we average around 7.22μs and allocate 6KB. 

<details>

#### Results

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
11th Gen Intel Core i7-1165G7 2.80GHz, 1 CPU, 8 logical and 4 physical cores
  [Host]                            : .NET Framework 4.8 (4.8.4510.0), X86 LegacyJIT  [AttachedDebugger]
  FilterQueryTreePerf_Match_Complex : .NET Framework 4.8 (4.8.4510.0), X86 LegacyJIT

Job=FilterQueryTreePerf_Match_Complex  InvocationCount=100  IterationCount=100  
UnrollFactor=1  

```
|                           Method |    ComplexExpression |      Mean |     Error |    StdDev |    Median |      Min |        Q1 |        Q3 |       Max | Allocated |
|--------------------------------- |--------------------- |----------:|----------:|----------:|----------:|---------:|----------:|----------:|----------:|----------:|
| **CreateAndMatch_ComplexExpression** | **((Dep(...)&gt; 5)) [66]** |  **7.224 μs** | **0.1879 μs** | **0.5269 μs** |  **7.043 μs** | **6.484 μs** |  **6.893 μs** |  **7.424 μs** |  **8.817 μs** |      **6 KB** |
| **CreateAndMatch_ComplexExpression** |  **((GC(...))))) [122]** | **12.576 μs** | **0.8231 μs** | **2.3350 μs** | **11.783 μs** | **9.009 μs** | **10.781 μs** | **13.876 μs** | **20.172 μs** |      **9 KB** |
| **CreateAndMatch_ComplexExpression** | **((GC/(...) 10)) [92]** |  **8.980 μs** | **0.4311 μs** | **1.2299 μs** |  **8.584 μs** | **7.514 μs** |  **8.077 μs** |  **9.632 μs** | **12.423 μs** |      **7 KB** |
| **CreateAndMatch_ComplexExpression** | **((Thr(...)1001) [68]** |  **5.686 μs** | **0.2163 μs** | **0.6171 μs** |  **5.372 μs** | **5.008 μs** |  **5.232 μs** |  **5.922 μs** |  **7.278 μs** |      **5 KB** |
| **CreateAndMatch_ComplexExpression** | **((Thr(...)1001) [45]** |  **4.266 μs** | **0.3320 μs** | **0.9631 μs** |  **3.907 μs** | **3.209 μs** |  **3.526 μs** |  **4.858 μs** |  **7.092 μs** |      **3 KB** |
| **CreateAndMatch_ComplexExpression** | **(Dept(...)= 20) [30]** |  **3.746 μs** | **0.1663 μs** | **0.4744 μs** |  **3.582 μs** | **3.216 μs** |  **3.379 μs** |  **3.997 μs** |  **5.059 μs** |      **3 KB** |
| **CreateAndMatch_ComplexExpression** | **(GC/S(...)s 10) [50]** |  **4.714 μs** | **0.2925 μs** | **0.8202 μs** |  **4.503 μs** | **3.763 μs** |  **4.039 μs** |  **5.058 μs** |  **7.478 μs** |      **4 KB** |
| **CreateAndMatch_ComplexExpression** | **(GC/S(...)= 20) [48]** |  **5.564 μs** | **0.4654 μs** | **1.3577 μs** |  **5.129 μs** | **3.916 μs** |  **4.441 μs** |  **6.520 μs** |  **9.515 μs** |      **4 KB** |
| **CreateAndMatch_ComplexExpression** | **(OldP(...)est2) [77]** |  **6.068 μs** | **0.2337 μs** | **0.6704 μs** |  **5.792 μs** | **5.370 μs** |  **5.574 μs** |  **6.674 μs** |  **8.434 μs** |      **6 KB** |
| **CreateAndMatch_ComplexExpression** | **Depth(...)!= 10 [41]** |  **4.936 μs** | **0.3329 μs** | **0.9552 μs** |  **4.509 μs** | **3.970 μs** |  **4.222 μs** |  **5.545 μs** |  **7.585 μs** |      **4 KB** |
| CreateAndMatch_ComplexExpression | Depth(...)!= 10 [41] |  5.365 μs | 0.3979 μs | 1.1289 μs |  4.941 μs | 3.942 μs |  4.538 μs |  5.749 μs |  8.388 μs |      4 KB |
| **CreateAndMatch_ComplexExpression** | **Depth(...)&lt;= 20 [26]** |  **3.672 μs** | **0.2279 μs** | **0.6576 μs** |  **3.416 μs** | **2.957 μs** |  **3.165 μs** |  **4.032 μs** |  **5.578 μs** |      **3 KB** |

#### Code

```csharp 
    [SimpleJob(targetCount: 100, invocationCount: 100, id: "FilterQueryTreePerf_Match_Complex")]
    [MinColumn, Q1Column, MedianColumn, Q3Column, MaxColumn]
    [MemoryDiagnoser]
    [MarkdownExporter]
    public class FilterQueryTreePerf_CreateAndMatch_Complex
    {
        [Params("(Depth >= 10) && (Depth <= 20)",
        "Depth <= 10 || Depth <= 20",
        "((Depth <= 10) && (Depth <= 20)) || ((Depth != 15) && (Depth > 5))",
        "Depth != 10 && Depth != 10 && Depth != 10",
        "Depth <= 10 && Depth <= 10 && Depth != 10",
        "((ThreadData != 1,0033) && ThreadData = 1001)",
        "((ThreadData != 1,0033 && ThreadData != 1,002) && ThreadData = 1001)",
        "(OldProcessName = test && OldProcessName != test1 || OldProcessName != test2)",
        "(GC/Start::Depth >= 10 && GC/Start::Depth <= 20)",
        "((GC/Start::Depth = 10 && (GC/Start::Depth <= 11 && (GC/Start::Depth <= 12 && (Depth <= 13 && (GC/Start::Depth <= 14))))))",
        "((GC/Start::Depth = 10) && (Depth <= 10)) || ((GC/Start::ThreadData != 10) || (Depth >= 10))",
        "(GC/Start::Depth = 10) && (ThreadData Contains 10)")]
        public string ComplexExpression { get; set; }

        [Benchmark]
        public void CreateAndMatch_ComplexExpression()
        {
            FilterQueryExpressionTree tree = new FilterQueryExpressionTree(ComplexExpression);
            var data = new Dictionary<string, string>
                {
                    { "Depth", "10" },
                    { "ThreadData", "10" }
                };
            tree.Match(data, "GC/Start");
        }
    }
```

</details>

## CPU Analysis Based On Tracing PerfView

I collected a trace that examined another trace with the events window open with ~1M+ CSwitch events with the following Columns to display textbox filled:

Cols: ``OldProcessName ProcessorNumber [(ProcessName = devenv) && (ProcessorNumber = 5)]``

image
image

The results indicated that to traverse 1M events and to filter 10k out of them for a 2 expression query, it takes ~1 second of CPU time to Match (this was much lesser than what I expected). Additionally, for the Shunting Yard Algorithm, the lifeblood of the boolean expression deduction, ``string[] rawTokens = postfixNotation.Split(' ');`` and `` if (_operandRegex.IsMatch(rawToken))`` were the two most expensive calls. For Matching individual queries, Splitting the query expression and looking in the payload of the Trace Events took the most time.

I wouldn't call these egregious with these latest changes but do let me know if you'd want me to optimize further.

Further details:

Name | Inc % | Inc | Exc % | Exc
-- | -- | -- | -- | --
microsoft.diagnostics.tracing.traceevent!FilterQueryExpressionTree.Match | 28.3 | 1,817 | 0.3 | 18
| +   microsoft.diagnostics.tracing.traceevent!ShuntingYard.Match | 14.3 | 919 | 0.2 | 10
|+   microsoft.diagnostics.tracing.traceevent!FilterQueryExpression.Match | 11.2 | 721 | 0.3 | 17
|+   mscorlib.ni!System.String.CtorCharCount(Char, Int32) | 1.1 | 70 | 0.0 | 0
|+   mscorlib.ni!System.Collections.Generic.Dictionary`2[System.__Canon,System.Boolean].Insert(System.__Canon,   Boolean, Boolean) | 0.6 | 39 | 0.5 | 31
|+   mscorlib.ni!System.Collections.Generic.Dictionary`2[System.__Canon,System.Boolean].Clear() | 0.5 | 34 | 0.0 | 3
|+   mscorlib.ni!System.Collections.Generic.Dictionary`2+Enumerator[System.Char,System.__Canon].MoveNext() | 0.2 | 13 | 0.2 | 12

Zooming in on the Shunting Yard algorithm, the lifeblood of the matching side of things:

Name | Inc % | Inc | Exc % | Exc
-- | -- | -- | -- | --
| +   microsoft.diagnostics.tracing.traceevent!ShuntingYard.Match | 14.3 | 919 | 0.2 | 10
|+   system.ni!Regex.IsMatch | 6.6 | 422 | 0.0 | 0
|+   mscorlib.ni!String.SplitInternal | 3.8 | 247 | 0.0 | 2

Zooming in on the individual query Match side of things:

Name | Inc % | Inc | Exc % | Exc
-- | -- | -- | -- | --
|+   microsoft.diagnostics.tracing.traceevent!FilterQueryExpression.Match | 11.2 | 721 | 0.3 | 17
|+ mscorlib.ni!String.Split | 5.5 | 351 | 0.1 | 9
|+   microsoft.diagnostics.tracing.traceevent!FilterQueryUtilities.ExtractPayloadByName | 5.4 | 349 | 0.1 | 9
|+ mscorlib.ni!String.EqualsHelper | 0.0 | 2 | 0.0 | 2

The command I used to collect the trace: ``.\PerfView.exe /nogui /accepteula collect withperf.etl``.

src/PerfView/EventViewer/EventWindow.xaml.cs
MokoSan: >> I'm wondering if it makes sense to pass the entire EventRecord in here

Definitely thought about that. I held back so far because: I figured I leave the input data type to be PerfView construct agnostic; specifically, the ``EventRecord`` type exists in the PerfView project but since I am building this functionality in the TraceEvent project I would either have to move the entire implementation from TraceEvent to PerfView or move the ``EventRecord`` definition to the TraceEvent library. 

My thinking here was that if there could be other uses of this filtering mechanism, might as well have it exposed as behavior in the TraceEvent side of things such as for cases where we'd possibly want to programmatically filter events outside Perfview. 

I'd be curious to know your opinion here and could definitely move things around. 

>>  such as not parsing Rest until we have to.

Definitely, the latest implementation takes care of this where we are only parsing the Rest column if the rest of the checks fail.

MokoSan:
Hi @brianrob,

I believe I have addressed the following issues in the latest commit:

- Added documentation in the ``Columns To Display`` and ``Find`` sections.
- Removed the requirement of providing spaces amongst the operands and the operator per expression i.e. we can now write expressions such as ``[(ProcessName=devenv) && (GC/Start::Depth >=2)]``. 
- Added more robustness to all the Filter Query Expression calls by wrapping around try-catch blocks and writing to the log file in case of an exception rather than bubbling it up to the user.

Do let me know if you have any further feedback! Thanks for your help!

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -1075,7 +1075,21 @@ public bool Find(string pat)
             int curPos = SelectionStartIndex();
             var startingNewSearch = false;
 
-            string modifiedPat = FilterQueryUtilities.TryExtractFilterQueryExpression(pat, out FilterQueryExpressionTree tree);
+            FilterQueryExpressionTree tree;
+            try
+            {
+                string modifiedPat = FilterQueryUtilities.TryExtractFilterQueryExpression(pat, out tree);
```
brianrob, line 1081: I did some additional testing and found an example that didn't work.  If you filter the view to just AllocationTick events, and then use the following expression `[TypeName = System.Char[]]`, I get no results even if they do exist.  I can confirm this with `[TypeName contains System.Char]`.  I'm assuming this is because the type is an array.

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -1286,6 +1300,11 @@ public void Update()
                 StatusBar.LogError(fqepEx.Message);
                 m_source.FilterQueryExpressionTree = null; 
             }
+            catch(Exception ex)
+            {
+                StatusBar.LogError(ex.Message);
```
brianrob, line 1305: I don't think this quite did what I wanted - the error seems to be overwritten by the number of records found, but it does at least write something to the log, so that's good.  I think this will be sufficient for now, and we can see if we need to something here in the future.

brianrob (commented):
@MokoSan, this is looking really good.  A couple of comments - I think most of these should be reasonably easy, but let me know if you want to defer any of these to post merge.

src/PerfView/EventViewer/EventWindow.xaml.cs
MokoSan: Yes, I think this is expected behavior with the current implementation. Unfortunately, I have developed this in a way where we are matching only on the first [] and ignoring the rest. I have a couple of ideas as to how this can be fixed (I'll take this on post-merge of this PR if that's fine with you).

brianrob (approved):
LGTM!

PR #1633, opened 2022-05-18
Upgrade to net48 for projects 

ivberg:
so VS2022 works out of the box without installing targeting pack


ivberg:
No worries @brianrob - feel free to reject it and I will keep it around for posterity while 4.6.2 is still in support. I may be wrong, but it seems 4.8 is as widely distributed as 4.6.2 on major Windows OSes (even more so on dev machines?) but I can understand if there are some things that consume these libs that are stuck on 4.6.2

brianrob:
Sounds good.  Yes, there is still pretty solid penetration in the 4.6.2 and 4.7.2 versions.

PR #1636, opened 2022-05-20
Explicitly Version FastSerialization

brianrob:
FastSerialization does not currently have an explicit version set, so it is always 1.0.0.0.  Breaking changes to FastSerialization can break when there are apps that have conflicting copies of FastSerialization with different versions.

This change allows us to explicitly version FastSerialization and sets the version number to 3.0.0.


brianrob:
cc: @karpinsn

PR #1641, opened 2022-06-02
Update to ClrMD 2.1

leculver:
Version bump ClrMD to 2.1.

- Disable AWE reader.
- Use ClrObject.EnumerateReferenceAddresses to avoid looking up the type when we do not need it.


brianrob:
@leculver thanks for doing this.

Are you able to build and run PerfView locally with this change?  It looks like ClrMD 2.1 doesn't target net461, so I would expect the PerfView build to fail.  This will require you to update PerfView.csproj to pull in the netstandard2.0 version of Microsoft.Diagnostics.Runtime.dll.  See https://github.com/microsoft/perfview/blob/main/src/PerfView/PerfView.csproj#L367.

It also looks like the CI is failing due to a missing dependency.  I would have expected this to work, but perhaps we were getting lucky.  Can you fix this one up with an explicit package reference in HeapDump.csproj?

Let me know if you'd prefer me to make these changes - I will just want your help validating the change since we'll be pulling in a netstandard copy of Microsoft.Diagnostics.Runtime instead of net461.

leculver:
Sorry I'm still working on this one.  Forgot to mark it as draft and let you know, I got distracted after creating this PR.

brianrob (approved):

PR #1665, opened 2022-07-18
Target netcoreapp3.1 in PerfViewCollect since 2.1 is out of support

pharring:
This also eliminates a warning from the IDE when you load PerfView.sln on a machine without the .NET Core 2.1 runtime


pharring:
@brianrob PTAL

brianrob (approved):

PR #1667, opened 2022-07-24
Allow alternative line endings when matching the source checksums

pharring:
When downloading source code via SourceLink directives, particularly if the source code provider is based on git, the line endings in the downloaded file may be different from the line-endings used to compile the binary and PDB. For example, if the assembly is compiled on Windows, the source code typically has CR+LF line endings. When committed to git, those line endings may be modified so that, when downloaded via a REST API, the same source file has Unix-style LF line endings.

When doing "go to source", PerfView checks the checksum (hash) of the downloaded file against the checksum recorded in the PDB -- originally written by the compiler. If the line endings don't match, then the checksum won't match either.

To fix that, this PR follows the same algorithm as the Visual Studio debugger and computes a second hash with the "other" line ending. If the checksum in the PDB matches this second hash, then we know we have matching source code for the PDB.


pharring:
@brianrob can we get this into the next release?

brianrob:
Yes, thanks for pinging me on this.  I had looked at this a few days ago but never got to merging.

brianrob (approved):

PR #1670, opened 2022-07-26
[WIP] Add authentication handlers for Azure DevOps and GitHub

pharring:
This change adds handlers to SymbolReader that authenticate requests for Azure DevOps and GitHub. This allows PerfView to download symbols from Azure DevOps symbol servers and to "go to source" for private Azure DevOps and GitHub repositories.

For Azure DevOps, it replaces #1563 with a more robust solution that uses DefaultAzureCredential from Azure.Identity
The code discovers the target tenant by parsing WWW-Authenticate challenge responses. If the user has Visual Studio or VSCode installed, and they have signed in, then PerfView will use that account to access Azure DevOps. If not, then a browser window is shown to let the user sign in.

For GitHub, we use Device Flow authentication to grant access to an OAuth application registered with GitHub.com. The user is prompted to visit github.com, enter an 8-digit code and authorize the application.

Both of these are implemented as custom handlers for a DelegatingHandler that is passed to the SymbolReader.

TODO:
- [x] Register an official OAuth application in GitHub for PerfView in the Microsoft organization.
- [x] Update the PerfViewClient ID in GitHubHandler with the new client ID
- [ ] Research if there's a narrower scope than "repo" for GitHub, since we don't need full write-access to repositories.
- [ ] Add Git Credential Manager Support
- [ ] Investigate using a GitHub Application instead of an OAuth application.
- [ ] Add a UI (checkable menu items?) to configure authentication


brianrob:
@pharring thanks for doing this!  I have started reviewing this PR and will add some comments as I go.  Do let me know once this is ready to go - I see you have one more TODO item.

Also, I will need to investigate whether or not we can carry the GitHub image inside of PerfView or not.

pharring:
@brianrob Don't spend too long on it. I have an improvement coming that adds Git Credential Manager support. It uses the same basic idea (chains of "symbol reader handlers"), but I refactored and shuffled the code quite a bit.

I also plan to add a UI so users can select which auth mechanism to use. I'll set defaults that work for me, but some users might need finer control.

Finally, the maintainers of the Microsoft org asked me to try and use a "GitHub Application" instead of an OAuth app for the device flow. Apparently, legacy OAuth apps are harder to maintain. Still haven't looked at that.

pharring:
On the GitHub image, it was downloaded from here: https://github.com/logos
From that page:
> Use the Octocat or GitHub logo to advertise that your product has built-in GitHub integration

I'm confident that this is appropriate usage of the logo.

brianrob:
> On the GitHub image, it was downloaded from here: https://github.com/logos From that page:
> 
> > Use the Octocat or GitHub logo to advertise that your product has built-in GitHub integration
> 
> I'm confident that this is appropriate usage of the logo.

I suspect you're right.

brianrob:
> @brianrob Don't spend too long on it.

Sounds good.  I'll wait for now.

pharring:
I'm going to abandon this and open a new PR that includes GCM support.

Issue #1064, opened 2019-11-27
Question: How to use an Azure DevOps Symbol Server inside of PerfView?


JustinKaffenberger:
It seems for some time now Azure DevOps has supported [publishing symbols](https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/symbols?view=azure-devops), and, as far as I can tell, each Azure DevOps organization has a dedicated symbol server.

My question is, how can I, if possible, use this Symbol Server inside of PerfView (for example, using the *Set Symbol Path* action)?

brianrob:
From a quick read of the article, it appears that AzDO supports various types of symbol drops, from file shares to indexed symbols.  At a minimum you should be able to point at a UNC file path or a symbol server using Set Symbol Path, and that should work.  If it doesn't, then the PerfView log should contain information about what went wrong.

JustinKaffenberger:
That makes sense, I think I'll need to open a separate issue with Microsoft (not sure where yet) to clarify what the actual URI is to the symbol server. My other concern is that its a private symbol server, so I'm not sure how PerfView would behave in that scenario.

mjsabby:
@JustinKaffenberger At this very moment you would need a proxy server that uses your PAT Token and issues the request on your behalf and you configure that proxy server in PerfView. I'm looking to see if there is a way we could make this setup easier by having PerfView host a proxy server and excavating your AAD credentials from your login, but that is a ways out for me.

In the meantime, however, you could basically spin up a new ASP.NET Core app every time you'd like to debug or what I've done is setup a Windows service that starts it automatically on my dev machine.

I've pasted below all the code you'd need to open up in VS and press F5, then put in `http://127.0.0.1:5000/api/symbols/download` as your symbol server in PerfView, or you could set `_NT_SYMBOL_PATH` so it is even more seamless.

The code below expects your PAT Token in the environment variable "PATToken". Of course follow all the guidelines of creating scoped PAT tokens, etc. and note that saving them in environment variables registry is not considered secure, yada yada ...

Here is AADSymbolProxy.csproj

```xml
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.1</TargetFramework>
  </PropertyGroup>

</Project>
```

Here is `Program.cs`

```csharp
namespace AADSymbolProxy
{
    using System;
    using System.Collections.Generic;
    using System.Net;
[...]

PR #1673, opened 2022-07-30
Add authentication handlers for Azure DevOps and GitHub

pharring:
This replaces #1670 

This change adds handlers to SymbolReader that authenticate requests for Azure DevOps and GitHub. This allows PerfView to download symbols from Azure DevOps symbol servers and to "go to source" for private Azure DevOps and GitHub repositories.

The preferred authentication mechanism is Git Credential Manager, which is installed with Git for Windows. It works for Azure DevOps (symbols and source) and GitHub. GitLab support has been requested in #1064 so we should try adding that next.

If Git Credential Manager is not installed, or the user decides not to use it, there are two alternatives:

For Azure DevOps, there's a replacement for #1563 that uses DefaultAzureCredential from Azure.Identity
The code discovers the target tenant by parsing WWW-Authenticate challenge responses. If the user has Visual Studio or VSCode installed, and they have signed in, then PerfView will use that account to access Azure DevOps. If not, then a browser window is shown to let the user sign in.

For GitHub, we use Device Flow authentication to grant access to an OAuth application registered with GitHub.com. The user is prompted to visit github.com, enter an 8-digit code and authorize the application.


pharring:
@brianrob, this is ready to review now.

src/PerfView/SymbolReaderHttpHandler.cs
```diff
@@ -0,0 +1,2010 @@
...
+        {
+            HttpRequestMessage request = context.Request;
+            if (request.RequestUri.Scheme == Uri.UriSchemeHttps &&  // Require a secure connection
+                request.Headers.Authorization is null &&            // No existing auth header
+                TryGetAuthority(request.RequestUri, out Uri authority) // Is a recognized authority
+                )
+            {
+                if (!TryGetCachedToken(authority, logExpiration: true, out AuthToken cachedToken))
+                {
+                    return GetAuthTokenAsyncCoreAsync(context, next, authority, cancellationToken);
```
brianrob, line 531: Nit: Two asyncs?

src/PerfView/Authentication.cs
```diff
@@ -0,0 +1,222 @@
...
+            {
+                throw new ArgumentNullException(nameof(mainWindow));
+            }
+
+            handler.ClearHandlers();
+
+            // The order isn't critical, but we chose to put GCM last
+            // because the user might want to use GCM for GitHub and
+            // Developer identity for Azure DevOps. If GCM were first,
+            // it would always pre-empt the developer identity handler.
```
brianrob, line 205: Per our discussion, is this the right ordering given that the delegate chain order is in reverse?

src/PerfView/SymbolReaderHttpHandler.cs
```diff
@@ -0,0 +1,2010 @@
...
+        }
+
+        /// <summary>
+        /// Try to locate the GCM executable in various places on disk.
+        /// </summary>
+        /// <returns>An enumeration of places to search. The values may not necessarily exist.</returns>
+        private static IEnumerable<string> EnumeratePossibleGcmExecutableLocations()
+        {
+            // The GCM_CORE_PATH environment variable.
+            yield return Environment.GetEnvironmentVariable("GCM_CORE_PATH");
```
brianrob, line 1066: Can we wrap this in a call to App.IsElevated, or otherwise make this conditional, so that someone can't set this and then cause arbitrary code to run elevated?

src/PerfView/PerfView.csproj
```diff
@@ -55,9 +55,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="1.6.0" />
```
brianrob, line 58: The DLLs for these packages are going to need to be packaged into PerfView as <EmbeddedResource/> entries so that they are present on client machines.

pharring: Can you elaborate? Doesn't it happen automatically?

brianrob: The unpacking of reosurces happens automatically, but what gets packed as resources does not.  Here's an example of how files get packaged as resources: https://github.com/microsoft/perfview/blob/main/src/PerfView/PerfView.csproj#L261.  I think you'll want to do this for the DLLs in these packages that are required to make these flows work.

If you look at the bin directory, the DLLs will be present there, but if you copy PerfView to a separate directory and then run it, it will use the unpacked version in %appdata%, which will mean that the new dependencies won't be present, and I'd expect these new flows to fail.

pharring: Ah, OK.

pharring: I fixed this by using a variant of the .targets file we use in the Snapshot Debugger to create a self-contained EXE. It automatically embeds all the resolved reference assemblies.

src/PerfView/Authentication.cs
pharring: I fixed it in SymbolReaderHttpHandler by composing the chain in reverse order.

brianrob:
@pharring, these changes look good.  One thing - as nice as it might be to automate the inclusion of resource files, this is including many more files than are actually needed, and is increasing the size of the resulting exe.  Would you mind reverting this and just including the files from the new NuGet packages explicitly?  Thanks.

pharring:
I can't really avoid increasing the size of the EXE, but I know what you mean: You want to include bare minimum (necessary and sufficient) set of assemblies needed to get the new scenarios working.
... while leaving behind a continually-growing list of manually curated assemblies for someone to maintain :stuck_out_tongue_winking_eye:

brianrob (approved):
LGTM.  Thanks @pharring!

PR #1676, opened 2022-08-01
Added the Number of Heaps in the Summary of the GCStats window

mrsharm:
The GCStats summary for a process should have the Number of Heaps:

image

CC: @Maoni0


brianrob (approved):
LGTM.  Thanks!

PR #1681, opened 2022-08-09
Update System.Reflection.Metadata to 1.8.1

inthemedium:
`System.Reflection.Metadata` 1.8.1 is the first package to include an explicit
`net461` which allows the build to target inbox `System.*` dlls.
Specifically, this PR will cause the  `System.Net.Http` reference to drop down from `4.2.0.0` to
the correct version of `4.0.0.0`.

From the `GetDependsOnNETStandard` task in the binlog (previous output was `true`):

```
GetDependsOnNETStandard
    Assembly = C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\Microsoft.NET.Build.Extensions\\tools\net472\Microsoft.NET.Build.Extensions.Tasks.dll
    Parameters
        References
            System.IO.Compression
            System.Net.Http
...
    OutputProperties
        DependsOnNETStandard = False
```

### DIFF
--- a/src/TraceEvent/TraceEvent.csproj
+++ b/src/TraceEvent/TraceEvent.csproj
@@ -70,7 +70,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles" Version="$(MicrosoftDiagnosticsTracingTraceEventSupportFilesVersion)" />
-    <PackageReference Include="System.Reflection.Metadata" Version="1.5.0" />
+    <PackageReference Include="System.Reflection.Metadata" Version="1.8.1" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="5.0.0" />
     <PackageReference Include="System.Security.Cryptography.Algorithms" Version="4.3.0" />
 


inthemedium:
Let me know if there is a different version of `System.Reflection.Metadata` I should update to. Only requirement is should have a `net461` or `net462` target

brianrob:
@inthemedium, can you tell me more about the goal of this PR?  Thanks.

inthemedium:
> @inthemedium, can you tell me more about the goal of this PR? Thanks.

The goal is to eliminate the need for the following assembly binding redirect when this is used in a .NET framework project:
```
    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
      <dependentAssembly>
        <assemblyIdentity name="System.Net.Http" Culture="neutral" PublicKeyToken="b03f5f7f11d50a3a" />
        <bindingRedirect oldVersion="0.0.0.0-4.2.0.0" newVersion="4.0.0.0" />
      </dependentAssembly>
    </assemblyBinding>
```
If an assembly has a reference to `System.Net.Http Version=4.2.0.0` it implies a reference to .NET Standard which is not actually needed for something that is just .NET framework and adds a lot of build and runtime complexity.

Specifically, `System.Reflection.Metadata` version 1.8.1 is the first version to add a `net461` target as [per this guidance](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting)

> 
> ✔️ CONSIDER adding a target for net461 when you're offering a netstandard2.0 target.
> 
> Using .NET Standard 2.0 from .NET Framework has some issues that were addressed in .NET Framework 4.7.2. You can improve the experience for developers that are still on .NET Framework 4.6.1 - 4.7.1 by offering them a binary that is built for .NET Framework 4.6.1.

brianrob (approved):

PR #1684, opened 2022-08-12
Add UserCommand to save the CPUStacks view as a CSV

EmosewaMC:
This PR adds a UserCommand to get PerfView data in a CSV format that is easy to understand since it directly correlates to the CPUStacks view and reuses the math PerfView already does when loading in the trace and outputs the data to a CSV.
This PR also adds another method of getting the process which is by the most amount of CPU msec used in the trace.

cc: @slevenstein who helped me understand how to get this data from within the PerfView application and wrote the basis which I iterated on and improved

### DIFF
--- a/src/PerfView/Extensibility.cs
+++ b/src/PerfView/Extensibility.cs
@@ -1000,6 +1000,11 @@ namespace PerfViewExtensibility
         /// <param name="outputFileName"> The file name the data will be written to </param>
         public void SaveAsCsvByName(string outputFileName)
         {
+            if (string.IsNullOrEmpty(outputFileName))
+            {
+                throw new ArgumentException($"{nameof(outputFileName)} is null or empty.");
+            }
+
             if (File.Exists(outputFileName))
             {
                 File.Delete(outputFileName);
@@ -1011,15 +1016,14 @@ namespace PerfViewExtensibility
                 foreach (var callTreeNode in callTree)
                 {
                     var frameUpdated = callTreeNode.Name.Replace(",", ";");
-                    csvFile.Write("{0},{1},{2},{3},{4},{5},{6},{7}\r\n",
-                        frameUpdated,
-                        callTreeNode.ExclusiveMetric,
-                        callTreeNode.ExclusiveMetricPercent,
-                        callTreeNode.InclusiveCount,
-                        callTreeNode.InclusiveMetricPercent,
-                        callTreeNode.ExclusiveFoldedMetric,
-                        callTreeNode.FirstTimeRelativeMSec,
-                        callTreeNode.LastTimeRelativeMSec);
+                    csvFile.WriteLine($"{frameUpdated}," +
+                        $"{callTreeNode.ExclusiveMetric}," +
+                        $"{callTreeNode.ExclusiveMetricPercent}," +
+                        $"{callTreeNode.InclusiveCount}," +
+                        $"{callTreeNode.InclusiveMetricPercent}," +
+                        $"{callTreeNode.ExclusiveFoldedMetric}," +
+                        $"{callTreeNode.FirstTimeRelativeMSec}," +
+                        $"{callTreeNode.LastTimeRelativeMSec}");
                 }
             }
         }
--- a/src/PerfView/Extensibility.cs
+++ b/src/PerfView/Extensibility.cs
@@ -995,6 +995,36 @@ namespace PerfViewExtensibility
         }
 
         /// <summary>
+        /// Saves the stacks as a CSV in the same format as it would appear in the CPUStacks GUI
+        /// </summary>
+        /// <param name="outputFileName"> The file name the data will be written to </param>
+        public void SaveAsCsvByName(string outputFileName)
+        {
+            if (File.Exists(outputFileName))
+            {
+                File.Delete(outputFileName);
+            }
+            using (var csvFile = File.CreateText(outputFileName))
+            {
+                csvFile.Write("Name,Exc,Exc%,Inc,Inc%,Fold,First,Last\r\n");
+                var callTree = ByName;
+                foreach (var callTreeNode in callTree)
+                {
+                    var frameUpdated = callTreeNode.Name.Replace(",", ";");
+                    csvFile.Write("{0},{1},{2},{3},{4},{5},{6},{7}\r\n",
+                        frameUpdated,
+                        callTreeNode.ExclusiveMetric,
+                        callTreeNode.ExclusiveMetricPercent,
+                        callTreeNode.InclusiveCount,
+                        callTreeNode.InclusiveMetricPercent,
+                        callTreeNode.ExclusiveFoldedMetric,
+                        callTreeNode.FirstTimeRelativeMSec,
+                        callTreeNode.LastTimeRelativeMSec);
+                }
+            }
+        }
+
+        /// <summary>
         /// Saves the stacks as a XML file (or a ZIPed XML file).  Only samples that pass the filter are saved.
         /// Also all interesting symbolic names should be resolved first because it is impossible to resolve them 
         /// later.   The saved samples CAN be regrouped later, however.  
--- a/src/PerfView/UserCommands.cs
+++ b/src/PerfView/UserCommands.cs
@@ -231,17 +231,63 @@ namespace PerfViewExtensibility
         }
 
         /// <summary>
-        /// Save the entire CPU stacks from 'etlFileName' as a csv.  If the /process qualifier is present use it to narrow what
-        /// is put into the file to a single process.
+        /// Returns the process with the most amount of CpuMsec time.  Should this time be equal, the oldest ID is prioritized
         /// </summary>
-        public void SaveCPUStacksAsCsv(string etlFileName, string processName = null, string warmSymbolLookupMinimumValue = "10")
+        /// <param name="processName"> The process name to look for in the stacks </param>
+        /// <returns> The process with the greatest cpuMSec, or the oldest process if there was a tie. </returns>
+        TraceProcess ProcessWithGreatestCpuMSec(ETLDataFile etlDataFile, string processName)
+        {
+            if (etlDataFile == null || string.IsNullOrWhiteSpace(processName))
+            {
+                return null;
+            }
+            float greatestMSec = 0.0f;
+            TraceProcess ret = null;
+            for (int i = 0; i < etlDataFile.Processes.Count; i++)
+            {
+                TraceProcess process = etlDataFile.Processes[(ProcessIndex)i];
+                if (string.Compare(process.Name, processName, StringComparison.OrdinalIgnoreCase) == 0)
+                {
+                    if (greatestMSec == process.CPUMSec && (ret == null || process.ProcessID < ret.ProcessID))
+                    {
+                        ret = process;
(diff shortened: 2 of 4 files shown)


src/PerfView/UserCommands.cs
```diff
@@ -1512,6 +1533,37 @@ private TraceEvents GetTraceEventsWithProcessFilter(ETLDataFile etlFile)
...
         }
 
+        /// <summary>
+        /// Save the CPU stacks for an ETL file into a perfView.xml.zip file.
+        /// </summary>
+        /// <param name="etlFile">The ETL file to save.</param>
+        /// <param name="process">The process to save. If null, save all processes.</param>
+        /// <param name="filter">The filter to apply to the stacks. If null, apply no filter.</param>
+        /// <param name="outputName">The name of the file to output data to. If null, use the default.</param>
+        private static void SaveCPUStacksForProcessAsCsv(ETLDataFile etlFile, TraceProcess process = null, FilterParams filter = null, string outputName = null)
```
slevenstein, line 1543: the output name is always null, so you can probably just get rid of that parameter

src/PerfView/UserCommands.cs
```diff
@@ -1512,6 +1533,37 @@ private TraceEvents GetTraceEventsWithProcessFilter(ETLDataFile etlFile)
...
+            }
+
+            if (filter == null)
+            {
+                filter = new FilterParams();
+            }
+
+            var stacks = etlFile.CPUStacks();
+            stacks.Filter = filter;
+            stacks.LookupWarmSymbols(10); // Look up symbols (even on the symbol server) for modules with more than 50 inclusive samples
```
slevenstein, line 1558: nit pick but you can replace "with more than 50 inclusive samples" in the comment with "with at least 10 inclusive samples"

src/PerfView/UserCommands.cs
slevenstein: I guess it's

slevenstein: same thing with filter, but I guess the SaveCPUStacksForProcess does the same thing below so you could keep both, get rid of both, or make SaveCPUStacks be able to take in a filter and outputName

EmosewaMC: Fixed

src/PerfView/UserCommands.cs
EmosewaMC: Fixed

src/PerfView/UserCommands.cs
EmosewaMC: Just going to deprecate both since they arent used at the moment

src/PerfView/UserCommands.cs
```diff
@@ -230,6 +230,27 @@ public void SaveCPUStacks(string etlFileName, string processName = null)
             }
         }
 
+        /// <summary>
+        /// Save the entire CPU stacks from 'etlFileName' as a csv.  If the /process qualifier is present use it to narrow what
+        /// is put into the file to a single process.
+        /// </summary>
+        public void SaveCPUStacksAsCsv(string etlFileName, string processName = null)
```
MokoSan, line 237: Might be worth adding an Async version of this method making use of `File.WriteAllTextAsync` under the hood to prevent blocking on IO.

src/PerfView/UserCommands.cs
```diff
@@ -1512,6 +1533,36 @@ private TraceEvents GetTraceEventsWithProcessFilter(ETLDataFile etlFile)
...
+            }
+
+            if (filter == null)
+            {
+                filter = new FilterParams();
+            }
+
+            var stacks = etlFile.CPUStacks();
+            stacks.Filter = filter;
+            stacks.LookupWarmSymbols(10); // Look up symbols (even on the symbol server) for modules with more than 10 inclusive samples
```
MokoSan, line 1551: Is it worth adding the inclusive sample threshold as a parameter with a default value of 10?

src/PerfView/UserCommands.cs
EmosewaMC: Good shout.  Should this be a separate method from the current one or be the default?

src/PerfView/UserCommands.cs
EmosewaMC: Probably!  This is what the previous one was doing but who says we cant iterate on the changes and make it better and more modular :)

src/PerfView/UserCommands.cs
MokoSan: I would add two different methods like System.IO's ``File.Read`` / ``File.Write`` methods but will leave it up to you. :)

src/PerfView/Extensibility.cs
```diff
@@ -994,6 +994,36 @@ public void LookupSymbolsForModule(string simpleModuleName, SymbolReaderOptions
             }
         }
 
+        /// <summary>
+        /// Saves the stacks as a CSV in the same format as it would appear in the CPUStacks GUI
+        /// </summary>
+        /// <param name="outputFileName"> The file name the data will be written to </param>
+        public void SaveAsCsvByName(string outputFileName)
+        {
```
MokoSan, line 1002: NIT: Might be worth adding a precondition check on the outputFileName being null / empty.

```csharp
  if (string.IsNullOrEmpty(fileName))
  {
      throw new ArgumentException($"{nameof(fileName)} is null or empty.");
  }

File: src/PerfView/Extensibility.cs [hunk 6]
```diff
@@ -994,6 +994,36 @@ public void LookupSymbolsForModule(string simpleModuleName, SymbolReaderOptions
...
+                File.Delete(outputFileName);
+            }
+            using (var csvFile = File.CreateText(outputFileName))
+            {
+                csvFile.Write("Name,Exc,Exc%,Inc,Inc%,Fold,First,Last\r\n");
+                var callTree = ByName;
+                foreach (var callTreeNode in callTree)
+                {
+                    var frameUpdated = callTreeNode.Name.Replace(",", ";");
+                    csvFile.Write("{0},{1},{2},{3},{4},{5},{6},{7}\r\n",
```
MokoSan, line 1014: NIT: I have been bit by this before so, thought would mention that string interpolation reduces the risk of misordering the columns especially if you add / change them. 

```csharp
csvFile.Write($"{frameUpdated},{callTreeNode.ExclusiveMetric},{callTreeNode.ExclusiveMetricPercent},
{callTreeNode.InclusiveCount},{callTreeNode.InclusiveMetricPercent},{callTreeNode.ExclusiveFoldedMetric},{callTreeNode.FirstTimeRelativeMSec},{callTreeNode.LastTimeRelativeMSec}\r\n");

**Comment by EmosewaMC** [comment]
hmm, the debug build is failing on here but locally it succeeds just fine?  Is there something I am missing?

File: src/PerfView/Extensibility.cs (hunk 6 above)
Comment (EmosewaMC) [reply src/PerfView/Extensibility.cs:1014]: I think this is a good call.  I realized how hard it was to view the columns if they are formatted like this.  Will fix this so its easier to read and maintain in the future

**Comment by MokoSan** [comment]
> hmm, the debug build is failing on here but locally it succeeds just fine? Is there something I am missing?

Test failures indicate that it could be because of a seemingly flakey test: https://github.com/microsoft/perfview/pull/1684/checks?check_run_id=7815017436. Specifics: https://dev.azure.com/ms/perfview/_build/results?buildId=348014&view=logs&j=dc31e2ab-11d5-59fa-122f-b0830955303b&t=ae395ee7-e160-5d25-b658-cc4a5f2ce7d9&l=1263

File: src/PerfView/UserCommands.cs (hunk 3 above)
Comment (EmosewaMC) [reply src/PerfView/UserCommands.cs]: It doesn't look like the current version of dotnet PerfView uses supports File.WriteAllTextAsync?  Is this correct?  If not what would I be missing to enable this feature?  None of the other File.Write...Async are available either so this may be an issue on my end but I'm not sure.

Comment (MokoSan) [reply src/PerfView/UserCommands.cs]: Sorry, you are right. Looks like the version of .NET PerfView is using is not high enough :(. Thank you very much for checking and all your effort.

Comment (EmosewaMC) [reply src/PerfView/UserCommands.cs]: My pleasure!  Anything to make a tool better and faster :D

File: src/PerfView/UserCommands.cs [hunk 7]
```diff
@@ -1512,6 +1533,33 @@ private TraceEvents GetTraceEventsWithProcessFilter(ETLDataFile etlFile)
...
+        private static void SaveCPUStacksForProcessAsCsv(ETLDataFile etlFile, TraceProcess process = null, int warmSymbolLookupMinimumValue = 10)
+        {
+            // Focus on a particular process if the user asked for it via command line args.
+            if (process != null)
+            {
+                etlFile.SetFilterProcess(process);
+            }
+
+            var stacks = etlFile.CPUStacks();
+            stacks.Filter = new FilterParams();
```
brianrob, line 1551: I don't believe that you should need to assign a `FilterParams` object if you're not setting any parameters.

src/TraceEvent/TraceLog.cs
```diff
@@ -5301,6 +5301,36 @@ public TraceProcess FirstProcessWithName(string processName, double afterTimeRel
             }
             return null;
         }
+
+        /// <summary>
+        /// Returns the process with the most amount of CpuMsec time.  Should this time be equal, the oldest ID is prioritized
+        /// </summary>
+        /// <param name="processName"> The process name to look for in the stacks </param>
+        /// <returns> The process with the greatest cpuMSec, or the oldest process if there was a tie. </returns>
+        public TraceProcess ProcessWithGreatestCpuMSec(string processName)
```
brianrob, line 5310: The precedent in other user commands of this type is to use the last process in the trace with the specified name.  I can see that this might not be the right process.  That said, if you want to select based on CPU msec, I would recommend that you move this method to UserCommands.cs, so that it is private to the user command.  This isn't something that I think we should expose publicly in TraceEvent.

brianrob (commented):
Thanks @MokoSan and @slevenstein for reviewing this.  A couple of additional comments.

src/TraceEvent/TraceLog.cs
EmosewaMC: That's fair.  My assumption here is if you want this data via the command line, you are likely investigating the process with the greatest cpu msec, should there be more than one of the same process.  My reasoning for this assumption is that the most costly process is likely the one you are investigating.  I'll add another parameter so you can choose to either do greatest msec or last process of that's fine (and also move this method)

brianrob: Sounds great.  Thanks @EmosewaMC!

src/PerfView/UserCommands.cs
EmosewaMC: Addresed

src/TraceEvent/TraceLog.cs
EmosewaMC: Resolved

EmosewaMC:
Is it required that the tests pass for this to be merged?  I'm not sure what could be failing here with my changes since they succeed locally in my release build and debug build

brianrob:
@EmosewaMC, this was just a transient failure that succeeded on re-run.

brianrob (approved):
LGTM.  Thanks!

PR #1685, opened 2022-08-16
Add null check for Disposing EventPipeEventSource deserializer

EmosewaMC:
When running the UserCommand `NetperfToSpeedScope` with an invalid file name, PerfView crashes due to _deserializer being null.  When applied, this PR will do a null check before attempting to dispose of the object.


brianrob (approved):
Thanks for contributing this!

PR #1690, opened 2022-08-19
Add EtlToSpeedScope UserCommand

### LINKED ISSUE #1688: Adding a UserCommand to convert etl files to speedscope.json or updating NetperfToSpeedScope
I've noticed that there is a UserCommand called `NetperfToSpeedScope` however giving that method a provided etl file, as it is stated in its comments results in an error due to the etlx file not being found.  I am unsure if there is a way to give the UserCommand a .netperf file since as far as I can tell from documentation the .netperf format is deprecated but this would likely cause the program to also look for a .etlx file which may not exist since I cant find where these are generated.  

My main question is, is it worth the time to put in a PR to add a UserCommand to convert an etl file to SpeedScope, or does the current method work but need some more specific files that I am not looking for?

If NetPerf is not deprecated, how would I obtain this data using PerfView so I can quickly convert traces to the SpeedScope format?

EmosewaMC:
Resolves #1688 

Adds a UserCommand to convert etl files to .speedscope.json for data parsing.  Added some parameters to filter the data down as well that are also used for other commands like process to filter on and warm symbol lookups.

### DIFF
--- a/src/PerfView/UserCommands.cs
+++ b/src/PerfView/UserCommands.cs
@@ -70,6 +70,42 @@ namespace PerfViewExtensibility
                 LogFile.WriteLine("[Converted {0} to {1}  Use https://www.speedscope.app/ to view.]", netPerfFileName, outputName);
             }
         }
+        
+        /// <summary>
+        /// Converts an etl file to the SpeedScope json format.  If processToFilterOn is specified,
+        /// that process will be the only process that will be saved to the SpeedScope file.
+        /// </summary>
+        /// <param name="etlFileName"> The etl file to be converted </param>
+        /// <param name="lookupWarmSymbolsMinimumValue"> The minimum number of trace events needed to lookup symbols before processing </param>
+        /// <param name="processToFilterOn"> The process to filter the trace data on </param>
+        public void EtlToSpeedScope(string etlFileName, string processToFilterOn = null, string lookupWarmSymbolsMinimumValue = "10")
+        {
+            string outputName = Path.ChangeExtension(etlFileName, ".speedscope.json");
+            using (var etlFile = new ETLDataFile(etlFileName))
+            {
+                if (processToFilterOn != null)
+                {
+                    var process = etlFile.Processes.LastProcessWithName(processToFilterOn);
+                    if (process == null)
+                    {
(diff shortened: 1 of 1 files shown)


EmosewaMC:
its been a couple of months since this was looked at - just wanted to throw a quick bump here

src/PerfView/UserCommands.cs
```diff
@@ -70,6 +70,42 @@ public void NetperfToSpeedScope(string netPerfFileName)
...
+                    if (lookupWarmSymbolsMinimumValue == null || string.IsNullOrWhiteSpace(lookupWarmSymbolsMinimumValue))
+                    {
+                        throw new ApplicationException("Invalid number to lookup with.");
+                    }
+                    int warmSymbolLookupThreshold = int.Parse(lookupWarmSymbolsMinimumValue);
+                    if (warmSymbolLookupThreshold <= 0)
+                    {
+                        throw new ApplicationException("Invalid number to lookup with.");
+                    }
+                    etlFile.CPUStacks().LookupWarmSymbols(warmSymbolLookupThreshold);
```
brianrob, line 104: The call to `CPUStacks()` creates a new instance of `Stacks` unconditionally, meaning that the last line of this method will override this line if the caller specifies a process filter.  You will want to save the `StackSource` in a local variable and then use it to generate the file.

src/PerfView/UserCommands.cs
```diff
@@ -70,6 +70,42 @@ public void NetperfToSpeedScope(string netPerfFileName)
...
+                }
+                SpeedScopeStackSourceWriter.WriteStackViewAsJson(etlFile.CPUStacks().StackSource, outputName);
```
brianrob, line 106: `lookupWarmSymbolsMinimumValue` only has an impact if a process filter is specified.  Presumably this should be used unconditionally?

brianrob (commented):
Thanks for pinging this.  A couple of comments below.

src/PerfView/UserCommands.cs
brianrob: Correcting myself - the filter does stick.  However, the stacks get calculated twice.

brianrob:

PR #1697, opened 2022-08-25
Add more checks for managed processes

WilliamXieMSFT:
A .nettrace file collected for an Asp.Net 6 app running on Azure App Service for Linux does not have coreclr as part of its loaded modules list.

### DIFF
--- a/src/TraceEvent/AutomatedAnalysis/Extensions.cs
+++ b/src/TraceEvent/AutomatedAnalysis/Extensions.cs
@@ -28,12 +28,12 @@ namespace Microsoft.Diagnostics.Tracing.AutomatedAnalysis
 
         public static bool ManagedProcess(this TraceProcess process)
         {
-            return process.LoadedModules.Where(module =>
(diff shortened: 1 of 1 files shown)


src/TraceEvent/AutomatedAnalysis/Extensions.cs
```diff
@@ -29,9 +29,11 @@ public static StackSource CPUStacks(this TraceLog eventLog, TraceProcess process
         public static bool ManagedProcess(this TraceProcess process)
         {
             return process.LoadedModules.Where(module =>
```
pharring, line 31: Use `.Any(predicate)` instead of `.Where(predicate).Count() > 0`

brianrob:
@WilliamXieMSFT, this makes sense.  Some of the existing analyzers won't work, since they look for mscorlib, but you can fix this in any of the analyzers you own directly.  I have this as a follow-up to address in the analyzers.

PR #1699, opened 2022-08-27
Support embedded source files

### LINKED ISSUE #1487: Question: Does Perfview have support for source files  included via EmbedAllSources
Hi,

I logged this [here](https://github.com/microsoft/perfview/discussions/1483), which might be the wrong place., adding here as well. 

In addition to source link support, does perfview support getting source code from a dll where the source code has been embedded in the dll via EmbedAllSources? If not what might it take to include such support? I would like to take a look at that, if it does not support it.
In addition can it also find embedded pdb's?

pharring:
Fixes #1487

Goto Source now works with source that's embedded in a portable PDB.


Xhanti:
This is the lords work 🙏🏾

brianrob:
@pharring, this is great!  It would be awesome to also have a regression test along with this change so that we don't break it in the future.

pharring:
I'll see what I can do.

brianrob:
Cool, thanks @pharring.

pharring:
@brianrob I added a test for this. PTAL.

brianrob (approved):
LGTM.  Thanks @pharring.  Really appreciate all your work in this space!

PR #1712, opened 2022-09-16
Espace invalid URL characters when constructing Source Link URL

### LINKED ISSUE #1711: TraceEvent does not escape invalid URL characters in Source Link.
While investigating [this issue](https://github.com/OmniSharp/omnisharp-vscode/issues/5290) with the Microsoft C# Extension, we inspected other implementations of Source Link to see if they were subject to the same issue. By inspection we can see that TraceEvent does not escape invalid characters while constructing its Source Link URL's.

chuckries:
Fixes #1711
This brings the URL escaping in line with [this implementation](https://github.com/dotnet/sourcelink/blob/16ba25a5b808ab4794d8eeb945a836b32f2efc15/src/SourceLink.Tools/SourceLinkMap.cs#L214) from dotnet.


gregg-miskelly (approved):

brianrob (approved):
LGTM.  Thanks!

PR #1717, opened 2022-09-26
Add dark mode

### LINKED ISSUE #1116: Request: Dark Mode support of Perfview
I set Windows 10 2004 (19041) to dark mode for Shell+Apps, but Perfview is still white. 

Can we get Dark mode support in Perfview?

rzikm:
Closes #1116.

This PR adds dark mode to PerfView based on https://github.com/AngryCarrot789/WPFDarkTheme.

The theme can be switched via menu in Options on main window. Changes require restarting the application to take effect.

Besides minor color changes (the gray background in dialogs is slightly lighter now), the light theme is unaffected.

Dark theme may bring some visual changes (I had to redo the control templates in order to apply color changes), but there are no functional changes AFAICT.

The theme was tested on Windows 11, it would be nice if someone tried this on Windows 7 to make sure I did not break some compatibility.


MagicAndre1981:
ok, the menu items are unreadable and the embedded help is still white. Is this possible to fix?

perfview_dark_menu_items

rzikm:
Must have broken it when fixing tests, should be good now.

The embedded help is a browser pointed to embedded html files. It sure can be themed, but at the moment I am not sure how to condition it on the current theme. I have at least changed the help on main window which is just TextArea

MagicAndre1981:
looks better now. Any idea how to theme the "about" messagebox? 

Also having an option "follow system design settings" would be nice.

Moving mouse over the "Get Started" shows a blue border which is a bit annoying after some time. 

NExt thing, opening the TraceInfo, Processes, Memory->GCStats, adv->JITStats, Runtime Loader, EventStats and Module Versioninfo of an ETL still show white. Also the Events view shows the same gray colors which also makes this view unreadable.

rzikm:
> looks better now. Any idea how to theme the "about" messagebox?

Probably creating custom MessageBox window, I might look into it later.

> Also having an option "follow system design settings" would be nice.

Done

> Moving mouse over the "Get Started" shows a blue border which is a bit annoying after some time.

Fixed

> NExt thing, opening the TraceInfo, Processes, Memory->GCStats, adv->JITStats, Runtime Loader, EventStats and Module Versioninfo of an ETL still show white

These all are opening a browser window to a generated HTML file. I would consider these for phase 2 and style them together with the help dialogs.

> Also the Events view shows the same gray colors which also makes this view unreadable.

Good catch, I missed this one, should be good now.

brianrob:
@rzikm, thanks for doing this work.  I have not had a chance to review this yet, but wanted you to know that I've seen this and will review it as soon as I am able.

MagicAndre1981:
> These all are opening a browser window to a generated HTML file. I would consider these for phase 2 and style them together with the help dialogs.
> 

ok, it looks good so far for me. 1 bonus would be that the alternate color in the grid (currently yellow) would be the [accent color of Windows ](https://support.microsoft.com/en-us/windows/change-colors-in-windows-d26ef4d6-819a-581c-1581-493cfcc005fe)

perfview_darkmode_accentcolor

If you can later fix the other mentioned views I'm happy

rzikm:
> ok, it looks good so far for me. 1 bonus would be that the alternate color in the grid (currently yellow) would be the [accent color of Windows](https://support.microsoft.com/en-us/windows/change-colors-in-windows-d26ef4d6-819a-581c-1581-493cfcc005fe)

I am not sure about this, since you can't use the color as is. E.g. the red one you have selected on the screen is too bright for dark mode (and for light mode as well perhaps). So you would have to adapt it slightly (lower the saturation/darken it).

It is not terribly hard, but would need quite some experimentation with parameters so I would rather do it separately in another PR.

brianrob:
@rzikm, I've started trying this out on my local machine.  Overall, it's pretty cool.  A couple of things that I wanted to point out:

1. In both the Stack and Events views, if you try to multi-select a set of cells to copy and paste, it works, but in dark mode, you can't see which cells you've selected.
2. I think that the OSS resources that you've used to implement this will be fine, but we'll know for sure once we review and merge here and the internal backend jobs get to run against it.  Just want you to be aware in case we end up having to revert it.

I'll keep playing with it and start reviewing the code as well - what I've glanced at so far looks reasonable to me.

rzikm:
@brianrob How is it going? I have fixed the multiselection and made the colors more consistent. Is there anything more that should change?

brianrob:
@rzikm, thank you for your patience.  I just did another round of review - overall, this looks really good.  One comment above on the Registry.  Also, I see that you fixed the selection highlighting.  It looks like there is still a bug there, but a little more subtle - if you click and hold on a node in the stack viewer, and then drag down, I would expect to highlight everything, but it doesn't.  It only highlights the top and bottom nodes.  Can you take a look at this?

brianrob (approved):
LGTM.  Thanks for implementing this @rzikm!  I will go ahead and merge this, and will reach out if there are any follow-up concerns around the OSS included in this PR.

MagicAndre1981:
> > Next thing, opening the TraceInfo, Processes, Memory->GCStats, adv->JITStats, Runtime Loader, EventStats and Module Versioninfo of an ETL still show white
> 
> These all are opening a browser window to a generated HTML file. I would consider these for phase 2 and style them together with the help dialogs.

if this can be later fixed I'm fully happy. Current implementation works so far fine for me.

PR #1720, opened 2022-09-29
Expose the GenerationCount property on TraceGC for usability

### LINKED ISSUE #1719: GC Traces should expose # of Generations
From my comments on #1379 regarding the enum change

We currently have code in our codebase that loops through a GC trace and attempts to assess things like HeapSize, HeapFragmentation, SurvivalPercent, etc.

However, in the current SDK it blows up with an Index OOB:

image

After digging into the underlying collection that we're exceeding, we modified the code to be:

```csharp
if (gc.PerHeapHistories?.Count > 0)
{
    for (var genInt = (int)Gens.Gen0; genInt < gc.PerHeapHistories[0]GenData.Length; genInt++)
    {
        var gen = (Gens)genInt;
        this.HeapSize += gc.HeapStats != null ? gc.GenSizeAfterMB(gen) : 0;
        this.HeapFragmentation += gc.GenFragmentationMB(gen);
        var surv = gc.SurvivalPercent(gen);
        if (!double.IsNaN(surv))
        {
            this.SurvivalPercent.Add(gen.ToString(), surv);
        }
    }
}
```

but I don't like that we're limiting our history for HeapSize and SurvivalPct based on the array underpinning HeapFragmentation. Instead, it would be great if a GC trace simply told us how many generations there were and we could use that as the upper bound on this loop.

in v2.0.66 (before enum change) we were able to do this:

```csharp
for (var genInt = (int)Gens.Gen0; genInt < (int)Gens.Gen0After; genInt++)
{
    var gen = (Gens)genInt;
    this.HeapSize += gc.HeapStats != null ? gc.GenSizeAfterMB(gen) : 0;
    this.HeapFragmentation += gc.GenFragmentationMB(gen);
    var surv = gc.SurvivalPercent(gen);
    if (!double.IsNaN(surv))
    {
        this.SurvivalPercent.Add(gen.ToString(), surv);
    }
}
```

w/o hitting OOBs

cc @cshung

cshung:
Fixes #1719
@brandonh-msft

### DIFF
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -1877,6 +1877,10 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
         /// </summary>
         public int HeapCount { get; private set; } = -1;
         /// <summary>
+        /// Number of Generations (0,1,2,LOH,[POH]), which is either 4 or 5 depending on version
+        /// </summary>
+        public int GenerationCount { get { return this.PerHeapHistories[0].GenData.Length; } }
+        /// <summary>
         /// Calculate the size of all pinned objects
         /// </summary>
         /// <returns></returns>
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -1879,7 +1879,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
         /// <summary>
         /// Number of Generations (0,1,2,LOH,[POH]), which is either 4 or 5 depending on version
         /// </summary>
-        public int GenerationCount { get { return this.PerHeapHistories[0].GenData.Length; } }
+        public int GenerationCount { get { return this.PerHeapHistories?.FirstOrDefault()?.GenData.Length ?? 0; } }
         /// <summary>
         /// Calculate the size of all pinned objects
         /// </summary>
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -1879,7 +1879,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
         /// <summary>
         /// Number of Generations (0,1,2,LOH,[POH]), which is either 4 or 5 depending on version
         /// </summary>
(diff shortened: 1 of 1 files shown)


cshung:
@brianrob, can you please take a look? 

Something seems wrong with CLRMD TFM in the CI, which is probably not related to my change.

brianrob:
@cshung this occurs in CI builds with no changes, but does not occur in local builds.  I'm trying to figure out the next investigative step.

brianrob:
Looks like the version of NuGet that was being downloaded was causing this.  Upgrading to the latest fixed this.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -1877,6 +1877,10 @@ public void EnsureBGCRevisitInfoAlloc()
         /// </summary>
         public int HeapCount { get; private set; } = -1;
         /// <summary>
+        /// Number of Generations (0,1,2,LOH,[POH]), which is either 4 or 5 depending on version
+        /// </summary>
+        public int GenerationCount { get { return this.PerHeapHistories[0].GenData.Length; } }
```
brianrob, line 1882: From a review of the existing code in this file, other accesses of `PerHeapHistories` is guarded with a `null` check and validation that the length of the array is non-zero.  Can you add this here as well please?

brandonh-msft:
```suggestion
        public int GenerationCount { get { return this.PerHeapHistories?.FirstOrDefault()?.GenData.Length ?? 0; } }
```

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -1877,6 +1877,10 @@ public void EnsureBGCRevisitInfoAlloc()
         /// </summary>
         public int HeapCount { get; private set; } = -1;
         /// <summary>
+        /// Number of Generations (0,1,2,LOH,[POH]), which is either 4 or 5 depending on version
+        /// </summary>
+        public int GenerationCount { get { return this.PerHeapHistories?.FirstOrDefault()?.GenData.Length ?? 0; } }
```
brandonh-msft, line 1882: Do we need to check `.GenData` for `null` as well?
```suggestion
        public int GenerationCount { get { return this.PerHeapHistories?.FirstOrDefault()?.GenData?.Length ?? 0; } }
```

brianrob: Yes, I think this would be a good change.

brianrob (approved):

PR #1721, opened 2022-09-29
Use Encoded String in Regression Report

marshall-lerner:
When I used the regression report tool on two GC Dumps from an x64 Mac, it would try to make tags out of frames such as "Lib<</Applications/Example.app/Lib!Task<Example>>>" and turn it into things such as "LIB<>>". This change encodes reserved HTML characters, which should prevent this (In my example, it properly shows the original frame)


marshall-lerner:
The PR before this one is also failing with the same CI problems, so it doesn't look to be an issue in this PR

brianrob (approved):
LGTM.  Thanks!

PR #1727, opened 2022-10-12
Support Arbitrarily Large ETW/nettrace Heap Snapshots

brianrob:
- Set `isVeryLargeGraph=true` for the in-memory graph representation.  NOTE: This graph doesn't get written to disk.
 - Remove the 10M node limit.

cc: @Maoni0, @cshung


Maoni0:
awesome, thanks @brianrob!!

Maoni0 (approved):

PR #1734, opened 2022-10-25
Avoid integer overflow in symbol reader log

### LINKED ISSUE #1732: Symbol download shows negative sizes for symbol files > 2 GiB
```
.................................................................................................... -1837,0 Meg
.................................................................................................... -1836,0 Meg
.................................................................................................... -1835,0 Meg
```

Presumably somewhere there's an int32 used for the size during download.

(Chrome for example has a symbol file that's close to 4 GiB.)

pharring:
Avoid integer overflow by using long instead of int to count bytes in symbol reader.
Also, increase the buffer size from 8K to 80K to improve download speed.

Fixes #1732


brianrob (approved):
Thanks @pharring!

PR #1740, opened 2022-10-31
TraceEvent: expose timestamps in UTC format

vitalych:
Under heavy system load (10K+ events per second), timezone conversion accounts for the vast majority of time when reading the timestamp. In case the client does not need local time, the new properties saves a considerable amount of time.

image

### DIFF
--- a/src/TraceEvent/TraceEvent.cs
+++ b/src/TraceEvent/TraceEvent.cs
@@ -240,14 +240,13 @@ namespace Microsoft.Diagnostics.Tracing
         /// <summary>
         /// The time when session started logging. 
         /// </summary>
-        public DateTime SessionStartTime
-        {
-            get
-            {
-                var ret = QPCTimeToDateTimeUTC(sessionStartTimeQPC);
-                return ret.ToLocalTime();
-            }
-        }
+        public DateTime SessionStartTime => SessionStartTimeUTC.ToLocalTime();
+
+        /// <summary>
+        /// The time when session started logging in UTC format.
+        /// </summary>
+        public DateTime SessionStartTimeUTC => QPCTimeToDateTimeUTC(sessionStartTimeQPC);
+
 
         /// <summary>
         /// The time that the session stopped logging.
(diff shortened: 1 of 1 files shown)


vitalych:
That's all there is to this PR. This property is to be used by clients that link with the library. The screenshot would show our own code above the get_Timestamp method.

For the context, we have an application that gets a stream of TraceEvent objects and needs to get timestamps quickly. The workaround we have  is to get TimeStampQPC and convert it to UTC ourselves using the same logic that the library uses internally.

vitalych:
@microsoft-github-policy-service agree [company="Cyberhaven, Inc"]

vitalych:
@microsoft-github-policy-service agree company="Cyberhaven, Inc"

brianrob:
This seems reasonable to me.  Having some tests in https://github.com/microsoft/perfview/tree/main/src/TraceEvent/TraceEvent.Tests would be a good thing.  Probably just create a new `TimeStamps` directory.

@pharring, do you have concerns with this?  From my perspective, this just seems like goodness, and I don't see any concerns around not being able to implement all of these new UTC properties for all current and future formats.

joanna-lee-0:
> This seems reasonable to me. Having some tests in https://github.com/microsoft/perfview/tree/main/src/TraceEvent/TraceEvent.Tests would be a good thing. Probably just create a new `TimeStamps` directory.
> 
> @pharring, do you have concerns with this? From my perspective, this just seems like goodness, and I don't see any concerns around not being able to implement all of these new UTC properties for all current and future formats.

Fuck you!

pharring:
I've no concerns.

Once this PR is in, you could consider:
1. Deprecating the non-UTC properties via [Obsolete] and steering people to use the UTC versions.
2. Auditing all the callers in PerfView and see if they could be modified to use the UTC versions. Perhaps PerfView should have an option to show times in UTC or the local time zone.

brianrob:
Thanks @pharring.  Much appreciated.  Yes, I think these are reasonable things to do.

@vitalych, would you mind adding some tests?

brianrob:

brianrob:

PR #1744, opened 2022-11-04
Build the HeapDumpDLL project for netstandard20

brianrob:
Build the HeapDumpDLL project against the netstandard20 target framework.

Does not include the `GCHeapDumper` and associated code because some of its dependencies are not available in the netstandard20 set of APIs.


asundheim (approved):

mpeyrotc (approved):
LGTM

PR #1748, opened 2022-11-10
Make it easy to build PerfView with custom ClrMD

cshung:
This change introduces a new build mode so that we can easily build against a locally custom-built ClrMD. This can be very useful for diagnosing issues related to `HeapDump.exe`.

It does not change the default build, but we can change `MicrosoftDiagnosticsRuntimeVersion` in `Directory.Build.props` to local so that it will build against the local, custom-built ClrMD.

### DIFF
--- a/src/EtwHeapDump/EtwHeapDump.csproj
+++ b/src/EtwHeapDump/EtwHeapDump.csproj
@@ -28,7 +28,16 @@
     </FilesToSign>
 
     <PackageReference Include="MicroBuild.Core" Version="0.2.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(MicrosoftDiagnosticsRuntimeVersion)' != 'local'">
     <PackageReference Include="Microsoft.Diagnostics.Runtime" Version="$(MicrosoftDiagnosticsRuntimeVersion)" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(MicrosoftDiagnosticsRuntimeVersion)' == 'local'">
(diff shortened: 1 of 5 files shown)


src/EtwHeapDump/EtwHeapDump.csproj
```diff
@@ -28,8 +28,18 @@
     </FilesToSign>
 
     <PackageReference Include="MicroBuild.Core" Version="0.2.0" />
+    <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
```
brianrob, line 31: Is this only needed when using the local copy of ClrMD?  I would expect this to be pulled in via NuGet.  If so, maybe we put this under the local condition?

brianrob (commented):
Overall, looks good.  One question below.

src/EtwHeapDump/EtwHeapDump.csproj
cshung: Yes, this is needed only for local ClrMD.

if I don't add this and built with local flipped on, the build fails with some version mismatch warnings.

It feels like Nuget pulled the wrong version and then complain the pull was wrong, so I added it to make sure Nuget pull the right one.

brianrob: Gotcha.  In that case, what do you think about putting this under the local condition, so that the this doesn't apply when we use NuGet?

cshung: Sure

brianrob (approved):

PR #1754, opened 2022-11-17
Add support for last branch recording

jakobbotsch:
Add /LbrSources and /LbrFilters command line options. When the former is specified, last branch recording is enabled from the specified sources (comma-separated string of numeric source IDs or "PmcInterrupt"). /LbrFilters can optionally be specified to filter which branches are recorded (comma-separated list using same syntax as xperf).

No special parsing of these events yet, but this at least allows recording them.

### DIFF
--- a/src/PerfView/CommandLineArgs.cs
+++ b/src/PerfView/CommandLineArgs.cs
@@ -745,7 +745,7 @@ namespace PerfView
                 switch (entry.ToLowerInvariant())
                 {
                     case "stackmode": flags |= LbrFilterFlags.CallstackEnable; break;
-                    case "conditionalbranches": flags |= LbrFilterFlags.FilterNearRelCall; break;
+                    case "conditionalbranches": flags |= LbrFilterFlags.FilterJcc; break;
                     case "nearrelativecalls": flags |= LbrFilterFlags.FilterNearRelCall; break;
                     case "nearindirectcalls": flags |= LbrFilterFlags.FilterNearIndCall; break;
                     case "nearreturns": flags |= LbrFilterFlags.FilterNearRet; break;
--- a/src/PerfView/CommandLineArgs.cs
+++ b/src/PerfView/CommandLineArgs.cs
@@ -1,7 +1,9 @@
 ﻿using Microsoft.Diagnostics.Tracing;
 using Microsoft.Diagnostics.Tracing.Parsers;
+using Microsoft.Diagnostics.Tracing.Session;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using Utilities;
 
 namespace PerfView
@@ -108,6 +110,8 @@ namespace PerfView
 
         // Start options.
         public bool StackCompression = true;    // Use compresses stacks when collecting traces. 
+        public CommandLineLbrSources LastBranchRecordingSources;
+        public CommandLineLbrFilters LastBranchRecordingFilters;
         public int BufferSizeMB = 256;
         public int CircularMB;
         public bool InMemoryCircularBuffer;         // Uses EVENT_TRACE_BUFFERING_MODE for an in-memory circular buffer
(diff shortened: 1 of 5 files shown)


jakobbotsch:
PTAL @brianrob 

(Also the test failure does not look related to me -- but please let me know if you think it is and I can look closer.)

JulieLeeMSFT:
Ping @brianrob for code review.

brianrob:
Thank you for the reminder.  I am catching up from extended time out of office.  I plan to review this soon.

brianrob:
@jakobbotsch, thanks for your patience on this.  I've reviewed this PR, and would like to step back and understand more about the ultimate goal here before moving forward with this PR.  Once this is done, PerfView and TraceEvent will be able to capture these events, but won't be able to view or consume them.  Can you share details about what these events look like, and what you anticipate wanting to do to consume them?

jakobbotsch:
Spoke with @brianrob offline and we decided that the parser support should be added as part of this PR to make this more useful when consuming programmatically.

src/PerfView/CommandLineArgs.cs
```diff
@@ -679,5 +690,91 @@ private void SetupCommandLine(CommandLineParser parser)
...
+                    default: throw new CommandLineParserException($"Could not parse '{entry}' as an LBR filter");
+                }
+            }
+
+            const LbrFilterFlags callstackAllowedFlags =
+                LbrFilterFlags.CallstackEnable | LbrFilterFlags.FilterKernel | LbrFilterFlags.FilterUser;
+            if ((flags & LbrFilterFlags.CallstackEnable) != 0 &&
+                (flags & ~callstackAllowedFlags) != 0)
+            {
+                throw new CommandLineParserException("/LbrFilters: Only 'Kernel' or 'User' can be specified alongside 'StackMode'");
```
brianrob, line 766: Should this be "stackmode" instead of "StackMode"?

src/TraceEvent/TraceEventSession.cs
```diff
@@ -738,10 +738,30 @@ public unsafe bool EnableKernelProvider(KernelTraceEventParser.Keywords flags, K
...
             }
         }
 
+        private unsafe void EnableLastBranchRecordingIfConfigured()
+        {
+            uint[] sources = m_LastBranchRecordingProfileSources;
+            if (sources == null || sources.Length == 0)
+                return;
+
+            uint filters = (uint)m_LastBranchRecordingFilters;
```
brianrob, line 753: Can you please add a check here to make sure that this code only runs on supported operating systems?  There is an example of how to do this at [https://github.com/microsoft/perfview/blob/main/src/TraceEvent/TraceEventSession.cs#L1441](https://github.com/microsoft/perfview/blob/main/src/TraceEvent/TraceEventSession.cs#L1441).  In this case, I think throwing an exception that states which operating systems are supported will be a good way to avoid having to investigate issues where LBR support isn't present in the OS.

src/TraceEvent/TraceEventSession.cs
```diff
@@ -2391,6 +2453,15 @@ private static unsafe void CopyStringToPtr(char* toPtr, string str)
             }
         }
 
+        /// <summary>
+        /// Get the max number of last branch recording sources that can be specified at the same time.
+        /// </summary>
+        public static int GetMaxLastBranchRecordingSources()
+        {
+            const int ETW_MAX_LBR_EVENTS = 4;
```
brianrob, line 2461: Is there a way to determine this from the OS that we should be using instead of a constant?

brianrob:
@jakobbotsch this looks great!  Just a couple of small requests please.

src/TraceEvent/TraceEventSession.cs
jakobbotsch: This is defined in the Windows SDK headers, I don't think there is an OS API to get it. But perhaps perfview has a native component I expose this via to avoid having to hardcode it like this?

src/TraceEvent/TraceEventSession.cs
brianrob: Thanks for adding this @jakobbotsch.  For now, this is fine - if it turns out that we're seeing lots of failures in this area because its being run on old versions, we may need to tighten this up.

src/TraceEvent/TraceEventSession.cs
brianrob: Not easily - this is OK.  Thank you.

brianrob (approved):

PR #1758, opened 2022-12-02
Parsing of new GC events

cshung:
This change performs the parsing of the new events introduced in https://github.com/dotnet/runtime/pull/55888.

At a high level, here are the code changes:

- Make the `GenAware` trace parsing more consistent with the manifest, in particular, I collapsed the `GenAwareBeginTraceData` and `GenAwareEndTraceData` into a single class `GenAwareTemplateTraceData`. These two classes are identical anyway. I also renamed the event `GCGenAwareStart` to `GCGenAwareBegin` to be consistent with the manifest as well.

> My motivation for doing this change is to experiment with my idea at https://github.com/cshung/perfview/tree/private/trace-event-matcher. It is a prototype that I built to compare `ClrTraceEventParser.cs` with `ClrEtwAll.man` so that we can find holes in our event parsing logic. I have no plans to productize yet.

- Introduce the new `GCLOHCompact`, `GCFitBucketInfo`, and `GCSettingsRundown` events. These events are newly introduced in .NET 6.

- Parse the new V4 `GCAllocationTick` event that includes a new field named `ObjectSize`

- Parse the new V4 `GCGlobalHeapHistory` event that includes an array of timestamps.

- Add the missed `PinnedObjectHeap` enum value for the `CreateSegment` event.

- Add the GC keyword to the default rundown keywords.

In addition to the events exposed on `ClrTraceEventParser`, we also expose the data through `TraceGarbageCollector` as public property and `TraceGC` as public fields (to be consistent with other fields). 

This info is also reflected in the raw XML for the `GCStat` page.


src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -700,6 +700,7 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
                     }
                 };
 
+                // TODO, register the new event handlers to capture new data here so that we can stuff them in TraceGC
```
Maoni0, line 703: >// TODO, register the new event handlers to capture new data here so that we can stuff them in TraceGC [](http://example.com/codeflow?start=16&length=102)

what are these new event handlers? also are you planning to do this soon? if so you probably don't need to add a comment here.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2590,16 +2629,99 @@ public override object PayloadValue(int index)
...
-        public int Count { get { return GetInt32At(0); } }
-        public int ClrInstanceID { get { return GetInt16At(4); } }
+        /// <summary>
+        /// TODO
+        /// </summary>
+        public int TimePlan
+        {
+            get
+            {
+                return BitConverter.ToInt32(m_dataArray, 0);
```
Maoni0, line 2645: I'm wondering why you are using `BitConverter` instead of the standard `GetInt16At`, `GetInt32At` like all the other places in this file?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2641,16 +2771,198 @@ public override object PayloadValue(int index)
...
+                    Debug.Assert(false, "Bad field index");
+                    return null;
+            }
+        }
+
+        private int SizeOfGCLOHCompactInfo
+        {
+            get
+            {
+                return PointerSize == 8 ? 32 : 20;
```
brianrob, line 2787: Can you please add a comment that describes the field layout that is used to generate these values?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2641,16 +2771,198 @@ public override object PayloadValue(int index)
...
                     return null;
             }
         }
 
-        private event Action<GenAwareEndTraceData> m_target;
+        private int SizeOfGCFitBucket
+        {
+            get
+            {
+                return PointerSize == 8 ? 16 : 12;
```
brianrob, line 2961: Same here.  Can you please add a comment that describes the field order and size?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2641,16 +2771,198 @@ public override object PayloadValue(int index)
...
+            Debug.Assert(!(Version == 0 && EventDataLength != size0));
+            Debug.Assert(!(Version > 0 && EventDataLength < size0));
+        }
+        public override StringBuilder ToXml(StringBuilder sb)
+        {
+            Prefix(sb);
+            XmlAttrib(sb, "ClrInstanceID", ClrInstanceID);
+            XmlAttrib(sb, "BucketKind", BucketKind);
+            XmlAttrib(sb, "TotalSize", TotalSize);
+            XmlAttrib(sb, "Count", Count);
```
brianrob, line 2920: I would expect buckets to be written out here as well.

brianrob:
>             sb.Append("/>");

I would expect the Count and Times to be written out here.

---
Refers to: src/TraceEvent/Parsers/ClrTraceEventParser.cs:6977 in 6c0019c. [](commit_id = 6c0019c8beef70614c52fa44c36df7d734d2cb52, deletion_comment = False)

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2590,16 +2629,107 @@ public override object PayloadValue(int index)
...
+
+        private int m_timePlan;
+        private int m_timeCompact;
+        private int m_timeRelocate;
+        private long m_totalRefs;
+        private long m_zeroRefs;
+        #endregion
+    }
+
+    public class GCLOHCompactTraceData : TraceEvent
```
brianrob, line 2713: Can you please add a new test trace that includes these new events so that they are covered by tests?

brianrob (commented):
@cshung, a few comments, but overall, looks good.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
Maoni0, line 2713: shouldn't this be a sealed class like all the other *TraceData classes?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2590,16 +2629,107 @@ public override object PayloadValue(int index)
...
+        /// The time spent on planing the LOH for compaction
+        /// </summary>
+        public int TimePlan
+        {
+            get
+            {
+                return m_timePlan;
+
+            }
+        }
```
Maoni0, line 2648: nit formating, other places with a one liner getter all put them on one line like this -

`public int TimePlan { get { return m_timePlan; } }

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2641,16 +2771,198 @@ public override object PayloadValue(int index)
...
+        protected internal override Delegate Target
+        {
+            get { return Action; }
+            set { Action = (Action<GCFitBucketInfoTraceData>)value; }
+        }
+        protected internal override void Validate()
+        {
+            int size0 = 14 + Count * SizeOfGCFitBucket;
+            Debug.Assert(!(Version == 0 && EventDataLength != size0));
+            Debug.Assert(!(Version > 0 && EventDataLength < size0));
```
Maoni0, line 2912: there's only version 0 so I don't understand what this assert is supposed to achieve?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2641,16 +2771,198 @@ public override object PayloadValue(int index)
...
+            Action(this);
+        }
+        protected internal override Delegate Target
+        {
+            get { return Action; }
+            set { Action = (Action<GCFitBucketInfoTraceData>)value; }
+        }
+        protected internal override void Validate()
+        {
+            int size0 = 14 + Count * SizeOfGCFitBucket;
```
Maoni0, line 2910: nit - not sure why this var is named size0... instead of just size..

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -9585,6 +9918,96 @@ public override object PayloadValue(int index)
...
 
+    public sealed class GCSettingsRundownTraceData : TraceEvent
+    {
+        public long HardLimit { get { return GetInt64At(0); } }
+        public long LOHThreshold { get { return GetInt64At(8); } }
+        public long PhysicalMemoryConfig { get { return GetInt64At(16); } }
+        public long Gen0MinBudgetConfig { get { return GetInt64At(24); } }
+        public long Gen0MaxBudgetConfig { get { return GetInt64At(32); } }
+        public int HighMemPercentConfig { get { return GetInt32At(40); } }
+        public int BitSettings { get { return GetInt32At(44); } }
```
Maoni0, line 9929: we should include the definition of bits in this BitSettings field since this is a contract (we don't change the meaning of the bits), like MarkRootType

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -13437,8 +13881,8 @@ protected internal override void EnumerateTemplates(Func<string, string, EventFi
 
                 // New style
                 templates[22] = TieredCompilationSettingsDCStartTemplate(null);
-
                 templates[23] = ExecutionCheckpointDCEndTemplate(null);
+                templates[24] = GCSettingsRundownTemplate(null);
```
Maoni0, line 13885: just a general question - so the new style doesn't require the event id and all that? can all event data be specified this way? @brianrob

src/PerfView/GcStats.cs
```diff
@@ -226,6 +226,16 @@ public static void ToXml(TextWriter writer, TraceProcess stats, TraceLoadedDotNe
...
             writer.Write(" ProcessID={0}", StringUtilities.QuotePadLeft(stats.ProcessID.ToString(), 5));
+            if (runtime.GC.GCSettings != null)
+            {
+                writer.Write(" HardLimit=\"{0}\"", runtime.GC.GCSettings.HardLimit);
+                writer.Write(" LOHThreshold=\"{0}\"", runtime.GC.GCSettings.LOHThreshold);
+                writer.Write(" PhysicalMemoryConfig=\"{0}\"", runtime.GC.GCSettings.PhysicalMemoryConfig);
+                writer.Write(" Gen0MinBudgetConfig=\"{0}\"", runtime.GC.GCSettings.Gen0MinBudgetConfig);
+                writer.Write(" Gen0MaxBudgetConfig=\"{0}\"", runtime.GC.GCSettings.Gen0MaxBudgetConfig);
+                writer.Write(" HighMemPercentConfig=\"{0}\"", runtime.GC.GCSettings.HighMemPercentConfig);
+                writer.Write(" BitSettings=\"{0}\"", runtime.GC.GCSettings.BitSettings);
```
Maoni0, line 237: these bits should be expanded to the actual meaning

src/PerfView/GcStats.cs
```diff
@@ -536,6 +546,15 @@ public static void ToXmlAttribs(TextWriter writer, TraceProcess stats, TraceLoad
                                 }
                             }
                         }
+                        if (gc.LOHCompactInfos.Count > 0)
+                        {
+                            GCLOHCompactInfo lohCompactInfo = gc.LOHCompactInfos[HeapNum];
+                            writer.Write(" LohTimePlan =\"{0:n3}\" ", lohCompactInfo.TimePlan);
```
Maoni0, line 552: > LohTimePlan [](http://example.com/codeflow?start=42&length=12)

nit - could we please use LOH instead of Loh here? seems a bit better...

src/TraceEvent/Parsers/ClrTraceEventParser.cs
brianrob: All of the information still has to be specified - it's just specified in a different spot.  For this event it's specified here: https://github.com/microsoft/perfview/pull/1758/files#diff-5a4cce33f428c644a7af2b6ce4e0f6eff4f05f9e81ace26b0f145357bd61775fR13850

src/TraceEvent/Parsers/ClrTraceEventParser.cs
brianrob: Yes, please do this.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
cshung: It seems to be the way how other `Validate` method works. My interpretation is that whenever we have a newer version, the newer version should have the earlier layout keep unchanged, that's why from a size perspective it should be at least as large as the earlier version.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
cshung: @brianrob, but `Bucket` is an array of struct, how exactly will we do that?

brianrob: My thought is that you'd need to write some code to cast each of these entries to the struct, and then write the struct out in XML.  This stuff tends to be useful when you don't trust what's in the events, or want to dump their values into XML in the PerfView log.

cshung:
>             sb.Append("/>");

Similarly, Times is an array, how exactly do we do that?

---
In reply to: [1370367894](https://github.com/microsoft/perfview/pull/1758#issuecomment-1370367894) [](http://example.com/codeflow?ancestors=1370367894,1370367894)

---
Refers to: src/TraceEvent/Parsers/ClrTraceEventParser.cs:6977 in 6c0019c. [](commit_id = 6c0019c8beef70614c52fa44c36df7d734d2cb52, deletion_comment = False)

src/TraceEvent/Parsers/ClrTraceEventParser.cs
cshung: Got it, I think we can do something along the line of `GCBulkTypeTraceData.ToXml()`

brianrob: Yes, that sounds great.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
Maoni0: that just takes me to Andrew's commit 😀 which lines in which file should I be looking at?

brianrob: Weird.  The link does actually point to the line in question, but it didn't expand `ClrTraceEventParser.cs`.  Check out line13850.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
cshung: @brianrob, I needed some help with this. I have a program that can generate all these three types of events, but I am not sure how to add them to the tests.

The program is simple enough, it is just these few lines:

```c#
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.

using System;
using System.Reflection;
using System.Runtime;
using System.Diagnostics;

namespace CoreLab
{
    internal static class Program
    {
        private static void Main()
        {
            Console.ReadLine(); // So that I have them to attach dotnet-trace
            // Here is how I can trigger the FitBucket event, the event is enable for verbose level only.
            GC.Collect(
                /* generation = */ 1,
                /* mode       = */ GCCollectionMode.Forced,
                /* blocking   = */true);

            // Here is how I can trigger the GCLOHCompact event
            GCSettings.LargeObjectHeapCompactionMode = GCLargeObjectHeapCompactionMode.CompactOnce;
            GC.Collect();
        }
    }
}
```

The `dotnet-trace` command-line needs to be tailored to contain verbose trace to get it. Here is the command line:
```sh
dotnet-trace collect --clrevents gc --clreventlevel verbose -p <pid>
```

Because of [this](https://github.com/dotnet/runtime/pull/79671) bug, the new `GCSettingRundown` event can be present only in the latest runtime after that PR is merged. For ETW traces, it should work on .net 6+, but for the trace collection side, it requires the new GC keyword I introduced in the `ClrRundownTraceEventParser.Keywords.Default` in this PR to work because otherwise, the keyword is still missing.

Maoni0 (approved):
LGTM!

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2590,16 +2629,93 @@ public override object PayloadValue(int index)
...
+        public int Count { get { return GetInt16At(2); } }
+
+        public GCLOHCompactInfo Info(int heapIndex)
+        {
+            Debug.Assert(0 <= heapIndex && heapIndex < Count);
+            int baseAddress = 4 + SizeOfGCLOHCompactInfo * heapIndex;
+            int timePlan = GetInt32At(baseAddress);
+            int timeCompact = GetInt32At(baseAddress + 4);
+            int timeRelocate = GetInt32At(baseAddress + 8);
+            long totalRefs = (PointerSize == 8) ? GetInt64At(baseAddress + 16) : (long)GetInt32At(12);
```
brianrob, line 2712: Something doesn't seem right to me here.  Does pointer size affect the size of totalRefs and zeroRefs?  If so, I would expect totalRefs to be at `baseAddress+12` and then zeroRefs to be at `baseAddress+16` or `baseAddress+20`.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2641,16 +2762,213 @@ public override object PayloadValue(int index)
...
+        // <struct name="Values"   count="Count"  >
+        //     <data name="Index" inType="win:UInt16" />
+        //     <data name="Count" inType="win:UInt32" />
+        //     <data name="Size" inType="win:Pointer" outType="win:HexInt64" />
+        // </struct>
+        private int SizeOfGCFitBucket
+        {
+            get
+            {
+                return PointerSize == 8 ? 16 : 12;
```
brianrob, line 2967: My calculations are two bytes lower than what you have here.  Are you including two bytes for count?  From what I see here, count would not be included.

cshung: The size is correct, the 2 bytes comes from alignment

Here is how the struct look like in 32 bits
```
0:000> dt coreclr!wks::etw_bucket_info
   +0x000 index            : Uint2B
   +0x004 count            : Uint4B
   +0x008 size             : Uint4B
```
and here is how the struct look like in 32 bits
```
0:000> dt coreclr!wks::etw_bucket_info
   +0x000 index            : Uint2B
   +0x004 count            : Uint4B
   +0x008 size             : Uint8B
```

brianrob (approved):

PR #1764, opened 2022-12-10
Fix CtfChannel Improper Read Behavior

cincuranet:
Various places ignored when `Stream.Read` would read less on first try, and would read more on next attempt (as standard `Stream` behavior).


cincuranet:
@brianrob With this fix, all the "missing events" issues we discussed are gone.

LoopedBard3 (approved):
LGTM!

cincuranet:
@brianrob Any ETA when version with this will be published on NuGet (package `Microsoft.Diagnostics.Tracing.TraceEvent`)?

brianrob:
I am in the process of going through the backlog of PRs from last month while I was out.  Once that's done, I'll plan to create a new release.  Expect this to be in ~1 week.

cincuranet:
Awesome. Thanks.

PR #1769, opened 2022-12-26
Fixed handles events not firing issue

LordKiRon:
Due to this bug the TraceEvent fails to issue event calls on all Handle events (nothing registered)
So, this fixes it


LordKiRon:
No idea why test fails on Azure, works fine locally and unrelated top my changes

PR #1781, opened 2023-01-10
Ensure .new files are cleaned up

pharring:
I found a scenario where temporary files created during conversion to ETLX were being locked ("in use by another process") on disk. I think this happens when there's an exception during serialization (e.g. out of disk space).

Here are the changes:
1. Add a finalizer to MemoryStreamWriter. It's a based class that implements IDisposable with a `protected virtual void Dispose(bool disposing)` member. Without the finalizer, leaked instances will not get cleaned up. In particular, the derived `IOStreamStreamWriter` class won't close its stream which, in most cases is a `FileStream`.
2. Added a FileShare mode option to the `Serializer` constructor that takes a file path. This allows us to specify `FileShare.Delete` when we're doing conversion to ETLX. This means that the `File.Delete` on the error recovery path can succeed, even if the file isn't closed.
3. Removed unnecessary File.Exists checks in conversion code. File.Delete doesn't throw if the file isn't there. This is partly a correctness fix (File.Exists is always racy), partly to reduce I/O.
4. Formatting in TraceLog.cs

### DIFF
--- a/src/FastSerialization/FastSerialization.cs
+++ b/src/FastSerialization/FastSerialization.cs
@@ -511,9 +511,12 @@ namespace FastSerialization
     sealed class Serializer : IDisposable
     {
         /// <summary>
-        /// Create a serializer writes 'entryObject' to a file.  
+        /// Create a serializer that writes 'entryObject' to a file.
         /// </summary>
-        public Serializer(string filePath, IFastSerializable entryObject) : this(new IOStreamStreamWriter(filePath), entryObject) { }
+        /// <param name="filePath">The destination file.</param>
+        /// <param name="entryObject">The object to serialize.</param>
+        /// <param name="share">Optional sharing mode for the destination file. Defaults to <see cref="FileShare.Read"/>.</param>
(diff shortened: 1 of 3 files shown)


pharring:
The thing that's fragile here is trying to do all the conversion work inside Serializer's constructor. If an exception is thrown during serialization, then it'll be thrown out of the constructor and calling code has no way to clean up the intermediate stream (call Dispose).

brianrob:
@pharring this seems like a good change.  Question - is there a reason that we should require callers to specify `FileShare.Delete` explicitly, vs. just specifying it unconditionally?  I'm wondering if we can avoid the surface changes, and just change this internally.

pharring:
> is there a reason that we should require callers to specify FileShare.Delete explicitly, vs. just specifying it unconditionally?

I didn't want to change the defaults for a public class. Possibly other consumers of FastSerialization might expect certain sharing mode. But if you think it's a safe change, I can make it unconditional. You tell me.

brianrob:
Gotcha.  I think that sounds good - let's leave it as is.

brianrob (approved):

PR #1782, opened 2023-01-10
Use SafeTraceHandle to wrap Win32 TRACEHANDLE

pharring:
This makes the closing of Win32 ETW sessions more robust.
Use SafeTraceHandle instead of ulong.
Like SafeHandle, SafeTraceHandle has a finalizer that will try to close the session.

I mainly need this for robust closing of the TRACEHANDLE when reading ETW files (ETWTraceEventSource).
However, I saw the "TODO" in TraceEventSession and decided to tackle it there too.

There is a Debug.Assert in the SafeTraceHandle's finalizer. During testing, it fired because the ClrRundown session was not being closed. I fixed it, but it's in such a heavily trodden code path that I'd appreciate extra diligence when reviewing that part (CommandProcessor.cs)


src/TraceEvent/TraceEventNativeMethods.cs
```diff
@@ -499,33 +588,93 @@ internal unsafe struct EVENT_FILTER_EVENT_ID
...
+            [In] TRACEHANDLE TraceHandle,
             [In] int IsEnabled,
             [In] byte Level,
             [In] ulong MatchAnyKeyword,
             [In] ulong MatchAllKeyword,
             [In] uint EnableProperty,
             [In] EVENT_FILTER_DESCRIPTOR* filterData);
 
+        internal static int EnableTraceEx(
+            Guid ProviderId,
```
brianrob, line 652: Should this be `ref Guid ProviderId`?

src/TraceEvent/TraceEventNativeMethods.cs
```diff
@@ -534,6 +683,26 @@ internal unsafe struct EVENT_FILTER_EVENT_ID
...
+            uint ControlCode,          // See EVENT_CONTROL_CODE_*
+            TraceEventLevel Level,
+            ulong MatchAnyKeyword,
+            ulong MatchAllKeyword,
+            int Timeout,
+            ref ENABLE_TRACE_PARAMETERS EnableParameters)
+        {
+            return EnableTraceEx2(
+                TraceHandle.DangerousGetHandle(),
+                ref ProviderId, ControlCode,
```
brianrob, line 698: Nit: ControlCode on separate line?

src/TraceEvent/TraceEventNativeMethods.cs
```diff
@@ -499,33 +588,93 @@ internal unsafe struct EVENT_FILTER_EVENT_ID
 
         [DllImport("advapi32.dll", CharSet = CharSet.Unicode)]
         internal static extern int StartTraceW(
```
brianrob, line 590: Any reason that all the pinvokes that have helpers here can't be private?

src/TraceEvent/TraceEventSession.cs
```diff
@@ -2250,24 +2251,22 @@ private void EnsureStarted(TraceEventNativeMethods.EVENT_TRACE_PROPERTIES* prope
...
                 Stop();
                 m_Stopped = false;
-                Thread.Sleep(100);  // Give it some time to stop. 
-                retCode = TraceEventNativeMethods.StartTraceW(out m_SessionHandle, m_SessionName, properties);
+                Thread.Sleep(100);  // Give it some time to stop.
+                retCode = TraceEventNativeMethods.StartTrace(out m_SessionHandle, m_SessionName, properties);
             }
             if (retCode == 5 && OperatingSystemVersion.AtLeast(51))     // On Vista and we get a 'Accessed Denied' message
             {
-                m_SessionHandle = TraceEventNativeMethods.INVALID_HANDLE_VALUE; // StartTrace sets to 0 on failure.  We use INVALID_HANDLE_VALUE to represent failure
```
brianrob, line 2264: If `m_sessionHandle` currently gets set to 0 on failure, and then we update it to `INVALID_HANDLE_VALUE`, the behavior here will change because `TraceEventNativeMethods.StartTrace` will determine that the return code is non-zero, and that the `sessionHandle` is non-zero, and will call `CloseTrace` on it.  We should avoid the call to `CloseTrace` in these failure cases.

brianrob:
Appreciate you submitting this change @pharring.  A few questions/comments.

src/TraceEvent/TraceEventNativeMethods.cs
pharring: Since it's not a DllImport, I took the opportunity to remove the ref. It's a wrinkle that the GUIDs are passed byref in the P/Invoke definition, meaning you have to use "ref providerId" at the call-site. I ironed out the wrinkle for when you're using the SafeHandle overload.
I wonder if `in` would work.

src/TraceEvent/TraceEventNativeMethods.cs
pharring: Will fix (where possible).

src/TraceEvent/TraceEventNativeMethods.cs
pharring: `in` works as a replacement for `ref` where it's an input-only value. Will use that.

src/TraceEvent/TraceEventNativeMethods.cs
pharring: ok

src/TraceEvent/TraceEventSession.cs
pharring: This is a case I considered and I believe the new `StartTrace` overload handles this case carefully and correctly.
It returns a new `SafeTraceHandle` only if retCode is zero (success).
Otherwise (the P/Invoke StartTraceW returned a non-zero value), we set the `out` param to null.
There's an edge case, also handled, where we have a non-zero retCode but a valid TRACEHANDLE; the StartTrace overload will call CloseTrace prior to returning (with a null out param).

pharring: StartTraceW docs say "Do not use this handle if the function fails. Do not compare the session handle to INVALID_HANDLE_VALUE. The session handle is 0 if the handle is not valid."
I believe that means we shouldn't call CloseHandle if retCode is non-zero. I will fix that in the StartTrace overload.

pharring: Fixed in latest iteration (removed the CloseHandle from the StartTrace overload)

src/TraceEvent/TraceEventNativeMethods.cs
pharring: Use `in` instead of `ref`

src/TraceEvent/TraceEventNativeMethods.cs
pharring: Fixed

brianrob (approved):
@pharring, thank you for your patience.  LGTM!

PR #1798, opened 2023-02-01
Log FileVersions for In-Use .NET Runtimes During Collection

brianrob:
Newer versions of .NET Core log unique version numbers, but most versions of .NET Framework and .NET Core in the world don't log unique enough version information.  When they are present, the `KernelTraceControl/FileVersion` events provide this.  There are cases when these events are not present, and capturing them is prohibitively expensive.

This change causes PerfView to capture the `FileVersion` from each .NET runtime DLL that is identified during collection.  This is done with a real-time listener that listens to `RuntimeInformation/Start` events and then uses their `runtimeDllPath` to lookup the `FileVersion` and log it into the trace.

Example events:
image


brandonh-msft:
Will the ProcessName map back to the _real_ process which loaded the runtime DLL or will it always show `PerfView`?

brianrob:
> Will the ProcessName map back to the _real_ process which loaded the runtime DLL or will it always show `PerfView`?

The process will always be PerfView because PerfView is the process that is logging the event.  The expectation for this event is that you can match the version information to a process by using the DLL path in the `PerfView/RuntimeVersion` event and the `Runtime/Start` event.  Expect to only see one `PerfView/RuntimeVersion` event per runtime DLL because the code keeps track of which DLLs we've seen so that we don't emit duplicate data.

brandonh-msft:
Gotchya - just thought I'd ask as it'll affect what the readers have to do to correlate if not alongside the process itself. 🚀

brandonh-msft:
@brianrob  do you think this'll make it into the next release?

brianrob:
Possibly.  I am planning to do a bit more investigation to make sure that the design of this event is what we want.  One I confirm this, I will update the events necessary, and then merge.

PR #1799, opened 2023-02-06
Remove not needed dependencies

kant2002:


brianrob:
This is a great improvement, as it removes the System.Web dependency.  Thanks!

brianrob (approved):

PR #1801, opened 2023-02-06
Avoid logging code addresses for LBR event

jakobbotsch:
TraceEvent does not expose these addresses symbolized anyway and the data can easily be too voluminous to handle with the existing mechanism.


brianrob (approved):
LGTM. Thanks!

PR #1802, opened 2023-02-07
TraceEvent - don't use undocumented APIs

idigdoug:
Stop using ETWControl and ETWParsing from OSExtensions.dll. Replace them with implementations that use Microsoft-supported APIs.

- Use TraceSetInformation for configuring profile interval (no change from existing behavior other than using supported API).
- Use TraceSetInformation for configuring stack caching (same behavior, except that the new API is only supported on Windows 10 19582+).
- Use TdhEnumerateManifestProviderEvents for reading manifest data (works perfectly for manifests, but doesn't work for MOF).

Still using OSExtensions.dll for KernelTraceControl stuff, but that's supported by Microsoft. So perhaps the KernelTraceControl stuff could move into TraceEvent and then OSExtensions.dll would no longer be needed at all.

### DIFF
--- a/src/TraceEvent/RegisteredTraceEventParser.cs
+++ b/src/TraceEvent/RegisteredTraceEventParser.cs
@@ -1,7 +1,6 @@
 ﻿//     Copyright (c) Microsoft Corporation.  All rights reserved.
 using FastSerialization;
 using Microsoft.Diagnostics.Tracing.Compatibility;
-using Microsoft.Diagnostics.Tracing.Extensions;
 using Microsoft.Diagnostics.Tracing.Session;
 using System;
 using System.Collections.Generic;
@@ -88,7 +87,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
         public static string GetManifestForRegisteredProvider(Guid providerGuid)
         {
             int buffSize = 84000;       // Still in the small object heap.  
-            byte* buffer = (byte*)System.Runtime.InteropServices.Marshal.AllocHGlobal(buffSize);
+            var buffer = new byte[buffSize]; // Still in the small object heap.
             byte* enumBuffer = null;
 
             TraceEventNativeMethods.EVENT_RECORD eventRecord = new TraceEventNativeMethods.EVENT_RECORD();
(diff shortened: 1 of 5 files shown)


src/TraceEvent/TraceEventSession.cs
```diff
@@ -733,9 +740,14 @@ public unsafe bool EnableKernelProvider(KernelTraceEventParser.Keywords flags, K
                 Marshal.ThrowExceptionForHR(TraceEventNativeMethods.GetHRFromWin32(dwErr));
                 m_IsActive = true;
 
-                if (OperatingSystemVersion.AtLeast(62) && StackCompression)
+                if (StackCompression && OperatingSystemVersion.AtLeast(OperatingSystemVersion.Win10))
                 {
-                    ETWControl.EnableStackCaching(m_SessionHandle.DangerousGetHandle());
+                    var info = new TraceEventNativeMethods.TRACE_STACK_CACHING_INFO{ Enabled = 1 };
+                    TraceEventNativeMethods.TraceSetInformation(
+                        m_SessionHandle.DangerousGetHandle(),
```
pharring, line 747: We should add an overload of TraceSetInformation that takes a safe handle to "hide" the call to DangerousGetHandle

idigdoug: Looks like that already exists.

pharring: Actually, that overload (taking a SafeTraceHandle) already exists. Can we use it instead?
The call to TraceSetInformation a little higher up that passes zero for the TRACEHANDLE, we can allow `null` there? I think the "null" case is why the "dangerous" version of TraceSetInformation was not made private.

src/TraceEvent/RegisteredTraceEventParser.cs
```diff
@@ -979,6 +996,21 @@ private DynamicTraceEventData.PayloadFetchClassInfo ParseFields(int startField,
             ref int infoSize
         );
 
+        [DllImport("tdh.dll")]
+        internal static extern int TdhEnumerateManifestProviderEvents(
+            [In] ref Guid Guid,
```
pharring, line 1001: We can actually use `in` instead of `ref` now with P/Invokes so that the caller doesn't have to use the ref keyword.

pharring: See, for example, what we do with the provider Guid in `EnableTrace` in TraceEventNativeMethods.cs.

src/TraceEvent/RegisteredTraceEventParser.cs
```diff
@@ -979,6 +996,21 @@ private DynamicTraceEventData.PayloadFetchClassInfo ParseFields(int startField,
...
+            [Out] byte[] pBuffer, // Actually PROVIDER_EVENT_INFO*
+            ref int pBufferSize
```
pharring, line 1003: Missing`[In, Out]` attributes?

idigdoug: `ref` defaults to `[In, Out]`. I suppose I could make it explicit, but that just makes the binary larger.

pharring:
@brianrob code LGTM, but over to you to decide if this is moving the right direction vis-a-vis OSExtensions.dll (and possible support for downlevel OSes)

brianrob:
Thanks for reviewing @pharring.  I am in the process of testing this to see how it impacts existing scenarios.

pharring:
The PerfCollect unit tests are failing due to a leaking SafeTraceHandle. Implying that we forgot to Dispose a SafeTraceHandle.

idigdoug:
This is puzzling to me because my change does not create or dispose any handles.

pharring:
Are you able to run unit tests locally in VS?

I'm taking more than a passing interest in this because I introduced the SafeTraceHandle quite recently. There could certainly be a bug there or there could be legitimate cases where leaking a SafeTraceHandle is necessary or unavoidable.
If you click through the "Details" of the failing build and then through to the build log (assuming you have access), you can get more details, but personally, I can't determine exactly which test failed. The leak is reported as a debug assert, but it looks like that doesn't result in a specific test failure. Of course, finalizers can run any time, and most likely (long) after the culpable test has completed. I think XUnit calls GC.Collect and WaitForPendingFinalizers, but I'm not certain it's after every test case. All you can know is that it's at least one of the tests prior to the Debug.Assert.

pharring:
Good job fixing the leak, @idigdoug Thank you!

brianrob (approved):
Thanks @idigdoug!  I did some testing with this and all looks good.

Issue #1448, opened 2021-06-23
Hide Utilities namespace


andkamashev:
Lots of devs has Utilities static class(es) in project and after installing this nuget all calls of methods like Utilities.Sum(228,1488) throw error on compilation. It is quite easy to fix, but not if you have hundreds of projects.

brianrob:
From a quick look at the Utilities namespace, it looks like this would not change any public surface area.  Would you be interested in submitting a PR to merge the Utilities namespace with Microsoft.Diagnostics.Utilities?

andkamashev:
Sure, it will be nice, anything except global Utilities namespace will be perfect

JpWebster:
Any chance of getting these namespaces cleaned up, we are hitting this same issue as https://github.com/microsoft/perfview/issues/1049

brianrob:
Fixed via #1805.

Issue #1049, opened 2019-11-14
Types in Utilities namespace


alexandre-mckinnon:
In the Microsoft.Diagnostics.Tracing.TraceEvent assembly, some types are in the Utilities namespace.

Those types should probably be in the Microsoft.Diagnostics.Tracing.Utilities namespace:

Utilities.Cache<K,T>
Utilities.Command
Utilities.CommandOptions
Utilities.IndexMap
Utilities.OperatingSystemVersion

alexandre-mckinnon:
PR #1062 doing this.

twwilliams:
This causes a big problem for us since we have internal Utilities classes already.

While our classes are in a proper company-named namespace hierarchy, when we end up using OurCompany.Internal in a file (which is most places), when we add a reference to the Microsoft.Diagnostics.Tracing.TraceEvent assembly to our project, none of our Utilities.Whatever() calls can be resolved because this top-level Utilities namespace takes precedence.

JpWebster:
Any chance we could get this change made? We are running into the same issue as reported.

brianrob:
Fixed via #1805.

PR #1805, opened 2023-02-07
hided all utilities namespaces

andkamashev:
Hided public utilities namespaces.
Solves:
#1448 
#1049


brianrob:
@andkamashev it looks like the CI is failing with this:

```
src\PerfView\Triggers.cs(333,27): Error CS0234: The type or namespace name 'Command' does not exist in the namespace 'Utilities' (are you missing an assembly reference?)
```

brianrob (approved):

PR #1806, opened 2023-02-08
[Auto-Analyzers] Add an issue id to AnalzyerResult

xiaomi7732:
So that the analyzers could decide to put a durable id onto any AnalyzerIssue instance.

### DIFF
--- a/src/TraceEvent/AutomatedAnalysis/AnalyzerIssue.cs
+++ b/src/TraceEvent/AutomatedAnalysis/AnalyzerIssue.cs
@@ -15,6 +15,7 @@ namespace Microsoft.Diagnostics.Tracing.AutomatedAnalysis
         /// <param name="url">A URL pointing to further documentation.</param>
         /// <exception cref="InvalidOperationException">Instances of AnalyzerIssue can only be created during Analyzer execution.</exception>
         public AnalyzerIssue(
+            Guid id,
             string title,
             string description,
             string url)
@@ -30,6 +31,7 @@ namespace Microsoft.Diagnostics.Tracing.AutomatedAnalysis
             Title = title;
             Description = description;
             URL = url;
+            Id = id;
         }
 
         /// <summary>
@@ -37,6 +39,12 @@ namespace Microsoft.Diagnostics.Tracing.AutomatedAnalysis
         /// </summary>
         public Analyzer Analyzer { get; private set; }
 
+        /// <summary>
+        /// Gets the id of the issue.
+        /// </summary>
+        /// <value></value>
+        public Guid Id { get; protected set; }
+        
         /// <summary>
         /// The title of the issue.
         /// </summary>
--- a/src/TraceEvent/AutomatedAnalysis/AnalyzerIssue.cs
+++ b/src/TraceEvent/AutomatedAnalysis/AnalyzerIssue.cs
@@ -10,7 +10,8 @@ namespace Microsoft.Diagnostics.Tracing.AutomatedAnalysis
         /// <summary>
         /// Create a new instance of AnalyzerIssue.
         /// </summary>
-        /// <param name="id">A durable id, in form of GUID, to uniquely identify a performance issue.</param>
+        /// <param name="id">A durable id, in form of GUID, to uniquely identify a performance issue. Note: This id is intended to be stable for a given issue.
+        /// Once the analyzer decides the id for a specific issue, it does not change.</param>
         /// <param name="title">A string title.</param>
         /// <param name="description">A string description.</param>
         /// <param name="url">A URL pointing to further documentation.</param>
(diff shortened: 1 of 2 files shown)


src/TraceEvent/AutomatedAnalysis/AnalyzerIssue.cs
```diff
@@ -30,13 +31,20 @@ public class AnalyzerIssue
...
 
         /// <summary>
         /// The Analyzer that created the issue.
         /// </summary>
         public Analyzer Analyzer { get; private set; }
 
+        /// <summary>
+        /// Gets the id of the issue.
+        /// </summary>
+        /// <value></value>
```
pharring, line 45: What is `<value></value>` for?

src/TraceEvent/AutomatedAnalysis/AnalyzerIssue.cs
```diff
@@ -15,6 +15,7 @@ public class AnalyzerIssue
         /// <param name="url">A URL pointing to further documentation.</param>
```
pharring, line 15: Add new `<param name="id">...`

src/TraceEvent/AutomatedAnalysis/AnalyzerIssue.cs
xiaomi7732: A mistake. Fixed.

src/TraceEvent/AutomatedAnalysis/AnalyzerIssue.cs
xiaomi7732: Appended.

src/TraceEvent/TraceEvent.Tests/AutomatedAnalysis/AnalyzerExecutionTests.cs
```diff
@@ -79,7 +79,7 @@ protected override AnalyzerExecutionResult Execute(AnalyzerExecutionContext exec
             if (processContext.Process.DisplayID == PID && ProcessDescription.Equals(processContext.Process.Description))
             {
                 Assert.Null(Issue);
-                Issue = new AnalyzerIssue("Test Title", "Test Description", "http://test-url");
+                Issue = new AnalyzerIssue(Guid.NewGuid(), "Test Title", "Test Description", "http://test-url");
```
brianrob, line 82: Can you please create constant Guid and pass it in here for testing purposes?

xiaomi7732: Done.

src/TraceEvent/AutomatedAnalysis/AnalyzerIssue.cs
```diff
@@ -10,11 +10,14 @@ public class AnalyzerIssue
...
         /// Create a new instance of AnalyzerIssue.
         /// </summary>
+        /// <param name="id">A durable id, in form of GUID, to uniquely identify a performance issue. Note: This id is intended to be stable for a given issue.
+        /// Once the analyzer decides the id for a specific issue, it does not change.</param>
         /// <param name="title">A string title.</param>
         /// <param name="description">A string description.</param>
         /// <param name="url">A URL pointing to further documentation.</param>
         /// <exception cref="InvalidOperationException">Instances of AnalyzerIssue can only be created during Analyzer execution.</exception>
         public AnalyzerIssue(
+            Guid id,
```
brianrob, line 20: Follow-up question @xiaomi7732 - should we force a value for the ID (e.g. throw if `id` is `Guid.Empty`)?  My gut reaction is yes we should force this.

brianrob: Also, this change is going to break the `Automated CPU Analysis` view in PerfView.  We should address this before merge.

xiaomi7732: Hi Brian,
I can add a Guid.Empty check easily.
Could you please point me to the code for `Automated CPU Analysis`? And where do I invoke it on PerfView UI? I can't seem to locate it in this repo.

brianrob: It's located under the `Experimental` group of views.  This only shows up on corpnet today, but I am going to change that to be configurable in the PerfView UI.  If you're building PerfView, you can comment out the `InternalUser` check here:

https://github.com/microsoft/perfview/blob/main/src/PerfView/PerfViewData.cs#L7545

brianrob: Also, here is the exception that fires with your change, which is expected since the signature changed:
image

xiaomi7732: @brianrob Things starts to make sense a bit. The exception is expected since the signature of the ctor changed. We will need a newer package for the analyzers to make it work.

But what is the flow?

How can I sync source between this repo and the repo with the analyzers? The analyzers wouldn't compile until it knows the new signature of the constructor of the analyzer issue. 

Or do you want me to directly work on the other repo, so that we can release a newer package for the analyzers, and then bump up the reference to the analyzer package here?

An alternative could be, instead of changing the ctor, there could be an overload, but I want to avoid it since we control both package versions.

brianrob: I have been doing some work recently here and have made some other changes that will break in the same way.  I'll go ahead and fix the rules along with my changes.

PR #1808, opened 2023-02-08
Drop netstandart1.6

kant2002:
Seems to be only practical reason to support netstandard1.6 is for .NET Core 1.1, Mono 4.6, Xamarin.Android 7.0, Xamarin.iOS 10.0 which probably not needed. If so, that's probably dead code.

The only potential place to look closer is ETWTraceEventSource.GetOSPointerSize()

### DIFF
--- a/src/TraceEvent/AutomatedAnalysis/DirectoryAnalyzerResolver.cs
+++ b/src/TraceEvent/AutomatedAnalysis/DirectoryAnalyzerResolver.cs
@@ -1,5 +1,4 @@
-﻿#if NET462 || NETSTANDARD2_0
-using System.IO;
+﻿using System.IO;
 using System.Reflection;
 
 namespace Microsoft.Diagnostics.Tracing.AutomatedAnalysis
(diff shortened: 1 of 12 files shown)


brianrob:
@kant2002, this seems like a reasonable thing to do - I don't think we need netstandard1.6 anymore.  The CI is failing right now because the NuGet package build expects to see netstandard1.6 binaries and it can't find them.  You will need to modify the nuspec file to fix this: 

https://github.com/microsoft/perfview/blob/main/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec#L35
https://github.com/microsoft/perfview/blob/main/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.nuspec#L87

You'll also need to update the props file that gets pulled into the NuGet package:

https://github.com/microsoft/perfview/blob/main/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.props#L85

brianrob:
Superceded by #1822.

PR #1817, opened 2023-02-23
Fix heap count again

cshung:
We witnessed in some customer traces that for .NET Framework, the `ThreadInfo` is not reliable. It is possible that we have only one thread with the `.NET Server GC` thread info and therefore leading PerfView to believe there is only one heap.

Actually, PerfView already has provision to do the right thing during `ProcessGlobalHistory`, but it is not doing its work just because we set the flag and heap counts to the wrong value based on the unreliable data, so this change just gets rid of it.


brianrob (approved):
This looks like a good change.  Thanks @cshung.

PR #1838, opened 2023-03-21
Add TaskWaitSend.Behavior to AnyStacks

ezsilmar:
With the help of TPL TaskWaitSend event we can find all the places that use sync-over-async pattern
image


brianrob (approved):

brianrob:
@ezsilmar thanks for contributing this.

PR #1840, opened 2023-03-22
Dispose cached PDBs properly

pharring:
The main fix here is in Utlities\Cache.cs to call Dispose when evicting an entry from the cache (see the new call to ClearEntry from within GetFreeEntry)

In addition, add a finalizer to NativeSymbolModule so that we don't leak native IDiaSource COM objects which hold PDB files open on disk.


pharring:
In case it isn't clear, you can repro this in PerfView by loading up a trace, going to the CPU view and doing a warm symbol load. The SymbolReader has a small (size 10) cache of in-memory modules. When the 11th is loaded, we evict an entry (but don't call Dispose), thereby leaking the NativeSymbolModule. Since no-one called Dispose, the IDiaSource leaks which locks the PDB on disk.

PR #1841, opened 2023-03-22
Enable pdb file name extractions for Windows file names when running on Linux.

clguiman:
If a .NET app is built on Windows, then deployed on Linux, the pdb path will have Windows path separators. On Unix, `Path.GetFileName` will not consider '\\' as a path separator and will fail to extract the file name.
The change adds a new utility method: `PathUtil.GetPlatformIndependentFileName` that, when running on non-Windows platforms, will check to see if the path contains any '\\' separators and return only the string after the last separator. `SymbolReader.FindSymbolFilePath` will use this new method to get the pdb file name.

### DIFF
--- a/src/TraceEvent/Symbols/SymbolReader.cs
+++ b/src/TraceEvent/Symbols/SymbolReader.cs
@@ -182,13 +182,13 @@ namespace Microsoft.Diagnostics.Symbols
             }
 
             string pdbIndexPath = null;
-            string pdbSimpleName = Path.GetFileName(pdbFileName);        // Make sure the simple name is really a simple name
+            string pdbSimpleName = PathUtil.GetPlatformIndependentFileName(pdbFileName);        // Make sure the simple name is really a simple name
 
             // If we have a dllPath, look right beside it, or in a directory symbols.pri\retail\dll
             if (pdbPath == null && dllFilePath != null)        // Check next to the file. 
             {
                 m_log.WriteLine("FindSymbolFilePath: Checking relative to DLL path {0}", dllFilePath);
-                string pdbPathCandidate = Path.Combine(Path.GetDirectoryName(dllFilePath), Path.GetFileName(pdbFileName));
+                string pdbPathCandidate = Path.Combine(Path.GetDirectoryName(dllFilePath), PathUtil.GetPlatformIndependentFileName(pdbFileName));
                 if (PdbMatches(pdbPathCandidate, pdbIndexGuid, pdbIndexAge))
                 {
                     pdbPath = pdbPathCandidate;
--- a/src/Utilities/FileUtilities.cs
+++ b/src/Utilities/FileUtilities.cs
@@ -10,6 +10,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Runtime.InteropServices;
 
 namespace Microsoft.Diagnostics.Utilities
 {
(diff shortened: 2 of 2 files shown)


WilliamXieMSFT (approved):
:shipit:

src/Utilities/FileUtilities.cs
```diff
@@ -310,5 +311,28 @@ public static string PathRelativeTo(string path, string relativeToDirectory)
...
+        /// Returns the file name and extension of the specified path string.
+        /// Path.GetFileName will not trim the file name from a Windows path 
+        /// when running on Unix because `\` is a legal file char on Unix.
+        /// See the Remarks section in the docs: https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getfilename
+        /// This method ignores that logic and provides the string after the last '\' when running on non-Windows platform and Path.GetFileName returns the same string.
+        /// </summary>
+        public static string GetPlatformIndependentFileName(string path)
+        {
+            var fileName = Path.GetFileName(path);
+#if !NETFRAMEWORK
```
brianrob, line 325: TraceEvent is only compiled for `netstandard20` now, so I don't think that this #if is necessary anymore.

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -182,13 +182,13 @@ public string FindSymbolFilePath(string pdbFileName, Guid pdbIndexGuid, int pdbI
...
             string pdbIndexPath = null;
-            string pdbSimpleName = Path.GetFileName(pdbFileName);        // Make sure the simple name is really a simple name
+            string pdbSimpleName = PathUtil.GetPlatformIndependentFileName(pdbFileName);        // Make sure the simple name is really a simple name
 
             // If we have a dllPath, look right beside it, or in a directory symbols.pri\retail\dll
             if (pdbPath == null && dllFilePath != null)        // Check next to the file. 
             {
                 m_log.WriteLine("FindSymbolFilePath: Checking relative to DLL path {0}", dllFilePath);
-                string pdbPathCandidate = Path.Combine(Path.GetDirectoryName(dllFilePath), Path.GetFileName(pdbFileName));
+                string pdbPathCandidate = Path.Combine(Path.GetDirectoryName(dllFilePath), PathUtil.GetPlatformIndependentFileName(pdbFileName));
```
brianrob, line 191: Just doing a quick search for `Path.GetFileName`, there are many more of these.  I support wanting to run more workloads using TraceEvent on Linux.  At the same time, I'd like to understand a bit more about what you're looking to do.  We probably also need to start thinking more broadly about how the TraceEvent codebase is going to deal with these sorts of platform-specific things.  The pattern you've set forth here might be the right one, but I've not thought about it much.

src/Utilities/FileUtilities.cs
```diff
@@ -310,5 +311,28 @@ public static string PathRelativeTo(string path, string relativeToDirectory)
...
         }
+
+        /// <summary>
+        /// Returns the file name and extension of the specified path string.
+        /// Path.GetFileName will not trim the file name from a Windows path 
+        /// when running on Unix because `\` is a legal file char on Unix.
+        /// See the Remarks section in the docs: https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getfilename
+        /// This method ignores that logic and provides the string after the last '\' when running on non-Windows platform and Path.GetFileName returns the same string.
+        /// </summary>
+        public static string GetPlatformIndependentFileName(string path)
```
brianrob, line 322: We should mark this as internal so that we minimize the API surface that we expose.  One could argue that this belongs in the core libraries.

brianrob (commented):
Some comments and questions below.

src/Utilities/FileUtilities.cs
pharring: I believe that's handled by the `#if UTILITIES_PUBLIC` around the whole FileUtilities class. It's not publicly exposed from TraceEvent because TraceEvent.csproj doesn't define UTILITIES_PUBLIC.

src/Utilities/FileUtilities.cs
pharring: Yes, but this is a shared source file that is included in several other projects (e.g. PerfView.csproj) that compile for .NET Framework.

brianrob:
@clguiman, @WilliamXieMSFT thank you for clarifying the scenario offline.  This change looks good to me.

brianrob (approved):

brianrob:
@pharring thanks for reviewing.

PR #1843, opened 2023-03-24
TraceEvent: produce a portable pdb during build

clguiman:
Since TraceEvent is a netstandard2.0 library that could also be used on Linux, it would be nice if debugging with symbols also works so I'm proposing changing the pdb format to portable.


brianrob:
@clguiman looks good.  Can I ask you to also do this for `FastSerialization` as well?  This will also be pulled in by `TraceEvent`.

PR #1845, opened 2023-03-28
Only Include Arch-Specific DLLs for RID-Specific Builds That Reference TraceEvent

brianrob:
Today when a RID-specific build of TraceEvent is done, all of the native DLLs for all architectures are included in the output.  This means that there are several DLLs that are never needed.  Worse, in the case of single-file builds, those DLLs are included in the final image, which results in a larger binary with no benefit.

Change the behavior so that when a RID-specific build is done, only the needed native DLLs are included.

### DIFF
--- a/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.props
+++ b/src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.props
@@ -1,77 +1,77 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
-    <None Condition="Exists('$(MSBuildThisFileDirectory)..\build\native\x86\KernelTraceControl.dll')" Include="$(MSBuildThisFileDirectory)..\build\native\x86\KernelTraceControl.dll">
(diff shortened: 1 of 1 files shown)


src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.props
```diff
@@ -1,77 +1,77 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
-    <None Condition="Exists('$(MSBuildThisFileDirectory)..\build\native\x86\KernelTraceControl.dll')" Include="$(MSBuildThisFileDirectory)..\build\native\x86\KernelTraceControl.dll">
+    <None Condition="Exists('$(MSBuildThisFileDirectory)..\build\native\x86\KernelTraceControl.dll') And ('$(ProcessorArchitecture)' == 'x86' Or '$(ProcessorArchitecture)' == 'msil')" Include="$(MSBuildThisFileDirectory)..\build\native\x86\KernelTraceControl.dll">
```
rainersigwald, line 4: I was a bit worried because the transformations from RID to `ProcessorArchitecture` happen in `.targets`, but this is in an item, so it's processed later despite being imported earlier. These conditions wouldn't work in a `PropertyGroup` but they're fine here.

rainersigwald (approved):

src/TraceEvent/Microsoft.Diagnostics.Tracing.TraceEvent.props
brianrob: Thanks.  That's not something that I was aware of.

PR #1847, opened 2023-03-29
Fix PDBCache Merge From DiagSession to Local SymCache

brianrob:
@pharring pointed out that there was already code in PerfView to merge the DiagSession's PDBCache into the local SymCache, such that the SymbolPath shouldn't need to point at the extracted PDBCache.  In looking at this further, the merge was failing because `Directory.Move` cannot move across volumes.  I have updated this code to enumerate the files and move them individually.

This PR also reverts #1846.

### DIFF
--- a/src/PerfView/App.cs
+++ b/src/PerfView/App.cs
@@ -813,20 +813,6 @@ namespace PerfView
                         symPath.Insert("SRV*" + wprSymDir);
                     }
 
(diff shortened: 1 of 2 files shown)


pharring (approved):
LGTM

brianrob:
Thanks @pharring!

PR #1849, opened 2023-03-30
Add Additional Configuration Events to Minimal Rundown

### LINKED ISSUE #1831: Capture Tiered Compilation status as part of rundown
Per internal discussion:

This is not something that would be in a default PerfView collection – you would need to turn on the CompilationKeyword on the rundown provider to get the TieredCompilation/Settings event.  This event will tell you the configuration of tiered compilation, which has the following values:

```xml
<bitMap name="TieredCompilationSettingsFlagsMap">
  <map value="0x0" message="$(string.RuntimePublisher.TieredCompilationSettingsFlags.NoneMapMessage)"/>
  <map value="0x1" message="$(string.RuntimePublisher.TieredCompilationSettingsFlags.QuickJitMapMessage)"/>
  <map value="0x2" message="$(string.RuntimePublisher.TieredCompilationSettingsFlags.QuickJitForLoopsMapMessage)"/>
  <map value="0x4" message="$(string.RuntimePublisher.TieredCompilationSettingsFlags.TieredPGOMapMessage)"/>
  <map value="0x8" message="$(string.RuntimePublisher.TieredCompilationSettingsFlags.ReadyToRunMapMessage)"/>
</bitMap>

```
 
One thing that you might be able to do today though is to either:

1. Look at JIT events to see if any of the JIT operations are marked as initiated by tiered compilation – the logic that PerfView uses is here – it’s not perfect, but likely good enough: https://github.com/microsoft/perfview/blob/main/src/TraceEvent/Computers/TraceManagedProcess.cs#L3937
1. You could look at CPU stacks or thread names to see if any tiered compilation behavior was captured.
 
This feels like another event that we should be capturing so that we have a good handle on runtime configuration – it’s cheap and gives us some good configuration information.

brianrob:
Allow minimal rundown to collect the following events:
 - Microsoft-Windows-DotNETRuntimeRundown/TieredCompilationRundown/SettingsDCStart
 - Microsoft-Windows-DotNETRuntimeRundown/GC/SettingsRundown

Both of these events are emitted once per process and provide useful configuration information for the runtime.

Fixes #1831.
cc: @Maoni0, @brandonh-msft

### DIFF
--- a/src/PerfView/CommandProcessor.cs
+++ b/src/PerfView/CommandProcessor.cs
@@ -3526,7 +3526,8 @@ namespace PerfView
                         // - Runtime/Start event
                         // - TieredCompilation
                         var rundownKeywords = ClrRundownTraceEventParser.Keywords.ForceEndRundown |
-                            ClrRundownTraceEventParser.Keywords.Compilation;
+                            ClrRundownTraceEventParser.Keywords.Compilation |
+                            ClrRundownTraceEventParser.Keywords.GC;
 
                         // Only consider forcing suppression of these keywords if full rundown is enabled.
                         if (!parsedArgs.NoRundown && !parsedArgs.NoClrRundown)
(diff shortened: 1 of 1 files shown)


Maoni0:
cool, thanks!

Maoni0:
CC @cshung

PR #1861, opened 2023-04-17
Fix malformed XML documentation in TraceEventSession.cs

r0bfr:
While working with a `TraceEventSession` session, I pressed F12 on the `EnableKernelProvider()` method to see what exactly it did.  The documentation that was displayed to me above the decompiled method was a bunch of sentences without periods that had been mushed together into a giant paragraph that didn't make a lot of sense.  Inspection of the source code here revealed that the closing `</summary>` tag accidentally got placed after the parameter documentation (instead of before it).

This change fixes the malformed XML documentation, and fixes a couple other spots in the file as well while we're at it.


r0bfr:
@brianrob An easy one to send your way.

sharwell (approved):

brianrob:
👍  Thanks!

PR #1864, opened 2023-04-28
Change source view to include both the inclusive samples data as well as the exclusive samples data

davidwrighton:
I have a scenario where I am attempting to optimize the exlusive cost of a set of methods, and viewing exlusive costs in the source view has been quite helpful.


brianrob (approved):
LGTM.  Thanks @davidwrighton.

Issue #1829, opened 2023-03-09
Create TraceLog (ETLX) without writing to file


vaind:
I have an `EventPipe` data capture as a `MemoryStream`. 

I'm able to create an [`EventPipeEventSource`](https://github.com/microsoft/perfview/blob/601ca49eb882c7a4716465d40fec7cfff3ce8da3/src/TraceEvent/EventPipe/EventPipeEventSource.cs#L40) albeit only via the private constructor by setting `isStreaming=false` (otherwise passing this to TraceLog for conversion fails), but let's not dwelve on this for now:

I'm then able to use [`TraceLog.CreateFromEventPipeEventSources()`](https://github.com/microsoft/perfview/blob/8c78639176e31f57ce3b2aaf17419e6854cbd9b2/src/TraceEvent/TraceLog.cs#L816)

However, I don't want to write the .etlx to a file only to read it again in order to work with the `TraceLog` instance. Instead, I'd like to initialize the `TraceLog` directly from the `EventPipeEventSource`, letting it do the conversion and ending up with a usable instance. Is this somehow possible with existing APIs? If not, what part would be reasonable to change to make this work?

bruno-garcia:
@brianrob This would really help us doing profiling for .NET in-proc for production apps

PR #1867, opened 2023-05-04
TraceLog streaming/in-memory EventPipe support

vaind:
I'm trying to make changes to enable `TraceLog` use without creating an ETLX file from an EventPipe (nettrace). I've taken the existing ETW streaming support as a baseline and tried to adapt that. However, I'm having trouble getting the newly added tests to pass because the events coming from TraceLog don't match what I'd get from `EventPipeEventSource` (the previous test case, `Streaming` in the same file). 

@brianrob maybe if you could have a look at my changes, you could spot an issue what I'm doing wrong (or not doing and I should be) to get this working properly? Or is there some bigger problem I'm missing completely that prevents this from happening? 

The use-case I'm going after is a continuously running sampling profiler that only samples as needed (because actually starting the profiler takes a lot of time in the CLR so I want to do it only once).

This is a follow up, though not strictly an implementation of, #1829


vaind:

brianrob:
@vaind, thanks for working on this.  I think at a high level this should work.  A couple of thoughts:

1. It would be worthwhile to architect this similiar to how we do with `TraceEventSession` for live sessions.  This allows the `TraceLog` to know that it is being used in a live mode instead of a file-based mode.  There are some differences in how one can use the `TraceLog` when its live, because it hasn't seen the "end" of the trace, unlike in the file-based mode.
2. Can you share what differences you're seeing?  It's possible that it's just some piece of data that hasn't made it over to the `TraceLog`.

vaind:
> @vaind, thanks for working on this. I think at a high level this should work. A couple of thoughts:

Thanks for getting back to me, it indeed seems to work from my testing in the Sentry profiling SDK, with some quirks so far (I didn't get stack trace method resolution working yet, but from my understanding it should work because all the info has already come from the runtime). Also the ActivityComputer (and others) don't seem to work when streaming at the moment so that would also need some tuning I guess. I was able to get around them for now but I think they're useful so I'll likely get back to them in the future.

> 1. It would be worthwhile to architect this similiar to how we do with `TraceEventSession` for live sessions.  This allows the `TraceLog` to know that it is being used in a live mode instead of a file-based mode.  There are some differences in how one can use the `TraceLog` when its live, because it hasn't seen the "end" of the trace, unlike in the file-based mode.

Yes, I've made the change by reusing the same code, really, with a minor change of setting up the `kernelSession` in the `TraceEventSession` factory instead of the common constructor.

> 2. Can you share what differences you're seeing?  It's possible that it's just some piece of data that hasn't made it over to the `TraceLog`.

e.g the following line is missing from the captured output but is present in the `eventpipe-dotnetcore2.1-linux-x64-objver3.netperf.baseline.txt`:

```
System.Threading.Tasks.TplEventSource/SetActivityId, 1, <Event MSec=   "111.4517" PID="80749" PName="Process(80749)" TID="80749" EventName="SetActivityId" ProviderName="System.Threading.Tasks.TplEventSource" NewId="10000000-0000-0000-0000-000000000001"/>
```

Also, all the custom events are missing, as present in `eventpipe-dotnetcore2.1-linux-x64-tracelogging.netperf.baseline.txt`

I'm pretty sure it's a test-code issue, because as I've mentioned earlier, the overall approach seems to work in general, based on my changes to the profiling code in the Sentry SDK which now starts the profiler early in the process and slices profiles when actually needed (drops the other samples if not currently needed). I think this functionality will need better tests here, instead of trying to reuse what was already present in `EventPipeParsing/Streaming`.

vaind:
So the issue I'm having with stacktrace info not being available is due to missing modules & method info because the rundown on a session only happens when a session is stopped, while I'm trying to access this info while the sample-profiler is running (on each sample). I've raised an inquiry for this issue: https://github.com/dotnet/runtime/issues/86103

I was wondering whether I could start & stop a session, in the beginning, to force the rundown to happen and then use these events to feed TraceLog. Afterwards, the runtime information would rely on the runtime provider to get updates. @brianrob do you think this could work, conceptually or do you see any issues that would need to be overcome (or prevent this completely)?

vaind:
> I was wondering whether I could start & stop a session, in the beginning, to force the rundown to happen and then use these events to feed TraceLog. Afterwards, the runtime information would rely on the runtime provider to get updates. @brianrob do you think this could work, conceptually or do you see any issues that would need to be overcome (or prevent this completely)?

So I've verified this works fine, I'll just need to add test cases to this repo so that it's covered here.

vaind:
@brianrob Any chance you could have a first look at `TraceLog.cs` changes if they seem reasonable? If so, I'd update tests to cover the changes & mark this PR as Ready for review

brianrob:
@vaind, apologies for not getting to this sooner.  I am reviewing your code.

brianrob:
@vaind, the concept here looks good.  One thing I think we will need to resolve before this is ready for review is the rundown design.  Your instinct to trigger a rundown at the beginning of the trace is the right one.  I was going to recommend that you use the `StartRundown` keyword to do this, but it doesn't look like this is plumbed through the runtime, so it wouldn't have any impact.  If I recall correctly, EventPipe has a relatively hardcoded rundown setup, in which case we may not have many options here.

@davmason, do you know if rundown is configurable at all?  If I recall correctly, end rundown is hardcoded at the end of the session.  Is that right?  I'm trying to figure out the best option for @vaind to trigger a start rundown at the beginning of streaming so that `TraceLog` gets method rundown information and can resolve stacks.

@vaind, once this is resolved, then I think it will make sense to review this more.  I saw some diffs that I want to examine further, but I'm not on a great network connection today and so I don't have access to a quality diff tool.

vaind:
> @vaind, the concept here looks good. One thing I think we will need to resolve before this is ready for review is the rundown design. Your instinct to trigger a rundown at the beginning of the trace is the right one. I was going to recommend that you use the `StartRundown` keyword to do this, but it doesn't look like this is plumbed through the runtime, so it wouldn't have any impact. If I recall correctly, EventPipe has a relatively hardcoded rundown setup, in which case we may not have many options here.

Thanks for getting back to me @brianrob. And yes,  there doesn't seem a way to trigger a rundown on the same session. I've discussed that in the runtime issue, I've created two weeks ago: https://github.com/dotnet/runtime/issues/86103

davmason:
> @davmason, do you know if rundown is configurable at all? If I recall correctly, end rundown is hardcoded at the end of the session. Is that right? I'm trying to figure out the best option for @vaind to trigger a start rundown at the beginning of streaming so that `TraceLog` gets method rundown information and can resolve stacks.

Rundown is not currently configurable with EventPipe. You are remembering correctly, it is hardcoded to run at the end of the session. The internal code has the option to configure whether to do rundown or not, but we don't expose it over the IPC commands so customers can't turn it off or on. It would be fairly easy to expose the option to have rundown or not, and I don't think it would be that much work to have it at the start of a trace. I think we could reuse all the existing code

vaind:
> @vaind, once you're ready, go ahead and mark this as ready for review and I'll take a more thorough pass on it.

I've done so. It does need tests but I want to make sure the approach and APIs are OK-ish before trying to devise tests for them.

vaind:
P.S. I've tested it as a consumer in [Sentry .NET SDK](https://github.com/getsentry/sentry-dotnet/pull/2385). What I meant by needing tests is that it needs tests in this repo.

src/TraceEvent/TraceLog.cs
```diff
@@ -153,10 +153,227 @@ public static TraceLog OpenOrConvert(string etlOrEtlxFilePath, TraceLogOptions o
...
+        /// From a EventPipeEventSource, create a real time TraceLog Event Source. Like an EventPipeEventSource a TraceLogEventSource
+        /// will deliver events in real time. However an TraceLogEventSource has an underlying Tracelog (which you can access with
+        /// the .Log Property) which lets you get at aggregated information (Processes, threads, images loaded, and perhaps most
+        /// importantly TraceEvent.CallStack() will work. Thus you can get real time stacks from events).
+        /// </summary>
+        /// <param name="rundownSource">
+        /// If given, the rundownSource is used to initialize module and method information.
+        /// This only makes sense in realtime sessions when you need to resolve function names.
+        /// </param>
+        public static TraceLogEventSource CreateFromEventPipeEventSource(EventPipeEventSource source, EventPipeSession rundownSource = null)
```
brianrob, line 227: I think that this function should be private.  There is already a path to create a `TraceLog` from a file path, and I would like to make sure that we ensure that someone doesn't pass an `EventPipeEventSource` from a file into this method.

src/TraceEvent/TraceLog.cs
```diff
@@ -153,10 +153,227 @@ public static TraceLog OpenOrConvert(string etlOrEtlxFilePath, TraceLogOptions o
...
+                // However, we cannot stop it before starting to process the source.
+                // Therefore, we need attach to the first event and only then stop the session.
+                var completionSource = new TaskCompletionSource<bool>();
+                source.AllEvents += delegate (TraceEvent _)
+                {
+                    completionSource.TrySetResult(true);
+                };
+                var task = Task.Run(source.Process);
+                completionSource.Task.Wait();
+                session.Stop();
```
brianrob, line 256: @davmason, will this work to ensure that the entire end rundown is received?

src/TraceEvent/TraceLog.cs
```diff
@@ -543,88 +760,66 @@ private TraceLog()
         /// Functionality of TraceLog that does not depend on either remembering past EVENTS or require future
         /// knowledge (e.g. stacks of kernel events), will 'just work'.
         /// </summary>
-        private unsafe TraceLog(TraceEventSession session)
+        private unsafe TraceLog(TraceEventDispatcher source)
```
brianrob, line 763: I think we will want to do some real-time ETW testing here to ensure that nothing has regressed in that path.

src/TraceEvent/TraceLog.cs
```diff
@@ -1080,17 +1284,14 @@ private unsafe void SetupCallbacks(TraceEventDispatcher rawEvents)
...
-            _QPCFreq = rawEvents._QPCFreq;
-            sessionStartTimeQPC = rawEvents.sessionStartTimeQPC;
-            sessionEndTimeQPC = rawEvents.sessionEndTimeQPC;
-            cpuSpeedMHz = rawEvents.CpuSpeedMHz;
-            numberOfProcessors = rawEvents.NumberOfProcessors;
-            eventsLost = rawEvents.EventsLost;
-            osVersion = rawEvents.OSVersion;
+            if (rawEvents is EventPipeEventSource eventPipeEventSource)
+            {
+                eventPipeEventSource.HeadersDeserialized += delegate () { CopyHeadersFrom(rawEvents); };
```
brianrob, line 1289: It would be good to have a comment here that describes why this is necessary.

src/TraceEvent/TraceLog.cs
```diff
@@ -10632,10 +10831,7 @@ public static TraceProcess Process(this TraceEvent anEvent)
                 throw new InvalidOperationException("Attempted to use TraceLog support on a non-TraceLog TraceEventSource.");
             }
             TraceProcess ret = log.Processes.GetProcess(anEvent.ProcessID, anEvent.TimeStampQPC);
-            // When the trace was converted, a TraceProcess should have been created for
-            // every mentioned Process ID.
-            // When we care, we should ensure this is true for the RealTime case.
-            Debug.Assert(ret != null || log.IsRealTime);
+            Debug.Assert(ret != null);
```
brianrob, line 10834: It looks like there are a couple of these asserts that were changed to remove `log.IsRealTime`.  Can you tell me more about what changed here?

brianrob (commented):
Thanks for your patience.  Here's a first cut at the review.  I think we may have a few iterations, but overall, this is looking good.

src/TraceEvent/TraceEvent.Tests/Parsing/EventPipeParsing.cs
```diff
@@ -1,17 +1,22 @@
 ﻿using FastSerialization;
 using Microsoft.Diagnostics.Tracing;
+using Microsoft.Diagnostics.Tracing.AutomatedAnalysis;
```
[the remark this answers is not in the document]
vaind: It was unused in the end. Removed.

src/TraceEvent/TraceEvent.Tests/Parsing/EventPipeParsing.cs
```diff
@@ -1,17 +1,22 @@
...
 using Microsoft.Diagnostics.Tracing.Etlx;
 using Microsoft.Diagnostics.Tracing.EventPipe;
 using Microsoft.Diagnostics.Tracing.Parsers;
 using Microsoft.Diagnostics.Tracing.Parsers.Clr;
+using Microsoft.Diagnostics.Tracing.Parsers.Kernel;
```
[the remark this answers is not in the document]
vaind: It was unused in the end. Removed.

vaind (commented):
I've made some changes as requested and replied to some questions. 

Keeping the testing open until the questions are resolved.

src/TraceEvent/TraceLog.cs
vaind: There doesn't seem to be any test case for `CreateFromTraceEventSession()`, i.e. the ETW one - how do you think it would be best to go around testing this (and the EventPipeSession factory too)?

src/TraceEvent/TraceLog.cs
davmason: Yeah, just stopping the session is enough

src/TraceEvent/TraceEvent.Tests/Parsing/EventPipeParsing.cs
```diff
@@ -23,11 +24,34 @@ private class EventRecord
...
+                if (Records.ContainsKey(eventName))
+                {
+                    Records[eventName].TotalCount++;
+                }
+                else
+                {
+                    Records[eventName] = new EventRecord()
+                    {
+                        TotalCount = 1,
+                        FirstSeriazliedSample = new String(data.ToString().Replace("\n", "\\n").Replace("\r", "\\r").Take(1000).ToArray())
```
rapetum228, line 44: 1000 - magic number

vaind:
> I've made some changes as requested and replied to some questions.
> 
> Keeping the testing open until the questions are resolved.

@brianrob Have you been able to finish reviewing this? Any more concerns or do you think it would be acceptable after adding tests for the changes/new APIs? Also, do you have any preferences for the testing approach?

src/TraceEvent/TraceEvent.Tests/Parsing/EventPipeParsing.cs
bruno-garcia, line 44:
```suggestion
                        FirstSerializedSample = new String(data.ToString().Replace("\n", "\\n").Replace("\r", "\\r").Take(1000).ToArray())
```

vaind: Old code, moved from here: https://github.com/microsoft/perfview/pull/1867/files/eb41100b9dfd53913bea8a11a302630d4b407afb#diff-c94b1f053e78f716a557a5876766fc48558133ec18fe0fa752e5cf40c23a6fcfL61

src/TraceEvent/TraceEvent.Tests/Parsing/EventPipeParsing.cs
```diff
@@ -23,11 +24,34 @@ private class EventRecord
             public string FirstSeriazliedSample;
```
bruno-garcia, line 24: Oh the type was actually already in

bruno-garcia:
@brianrob this would really help unblock some work we want to do for .NET here at Sentry. Another round of review much appreciated

src/TraceEvent/TraceLog.cs
```diff
@@ -153,10 +153,199 @@ public static TraceLog OpenOrConvert(string etlOrEtlxFilePath, TraceLogOptions o
...
+            Action<ModuleLoadUnloadTraceData> onLoaderRundown = delegate (ModuleLoadUnloadTraceData data)
+            {
+                TraceProcess process = processes.GetOrCreateProcess(data.ProcessID, data.TimeStampQPC);
+                process.LoadedModules.ManagedModuleLoadOrUnload(data, false, true);
+            };
+
+            clrRundownParser.LoaderModuleDCStop += onLoaderRundown;
+            clrRundownParser.LoaderModuleDCStart += onLoaderRundown;
+
+            // TODO We don't seem to be getting any of these events, only MethodDCStopVerbose.
```
brianrob, line 258: This makes sense because rundown is limited to what's called "stop rundown" - rundown that is triggered when we stop tracing.  You can safely remove this commented out code.

src/TraceEvent/TraceLog.cs
```diff
@@ -153,10 +153,199 @@ public static TraceLog OpenOrConvert(string etlOrEtlxFilePath, TraceLogOptions o
...
+            {
+                codeAddresses.AddILMapping(data);
+            };
+
+            clrRundownParser.MethodDCStopVerbose += delegate (MethodLoadUnloadVerboseTraceData data)
+            {
+                // TODO we need this also for non-jitted methods, otherwise we won't resolve some frames, for example:
+                //      "System.Private.CoreLib.il" - "System.Threading.Tasks.Task.Wait()"
+                //      Is it OK to use InsertJITTEDMethod & FindJITTEDMethodFromAddress or do we need something else?
+                // if (data.IsJitted)
```
brianrob, line 285: It should be fine to use the JITTED method calls for all MethodDCStopVerbose events.

src/TraceEvent/TraceLog.cs
brianrob: Good question - there aren't any tests for this.  We don't really have an environment where we can run such tests elevated in an automated fashion.

I suspect the best thing to do here is to write some tests that emit events while a real-time session of each type is live, and then confirm that those events come through.

The EventPipe one can be automated if we so desire, because it doesn't require admin privileges.  The ETW one will likely just need to remain adhoc.

src/TraceEvent/TraceLog.cs
```diff
@@ -4287,13 +4462,15 @@ public override bool Process()
...
-                    kernelTask.Start();
                 }
                 TraceLog.rawEventSourceToConvert.Process();
                 if (kernelTask != null)
                 {
                     TraceLog.rawKernelEventSource.StopProcessing();
                     kernelTask.Wait();
+
+                    // Flush all outstanding events in the realTimeQueue.
+                    TraceLog.FlushRealTimeEvents();
```
brianrob, line 4473: Can you describe the reason for this?  Just want to make sure I understand if this is due to a bug or design change.

src/TraceEvent/TraceLog.cs
```diff
@@ -9064,11 +9241,7 @@ internal void AddILMapping(MethodILToNativeMapTraceData data)
         {
             var ilMap = new ILToNativeMap();
             ilMap.Next = ILMapIndex.Invalid;
-            var process = log.Processes.GetProcess(data.ProcessID, data.TimeStampQPC);
-            if (process == null)
-            {
-                return;
-            }
+            var process = log.Processes.GetOrCreateProcess(data.ProcessID, data.TimeStampQPC);
```
brianrob, line 9244: Same here - is this a bug?  I could believe that it is.

src/TraceEvent/TraceLog.cs
brianrob: Gotcha.  Go ahead and remove it then.  If we notice this isn't the case, we can always re-add it.

src/TraceEvent/TraceLog.cs
```diff
@@ -153,10 +153,227 @@ public static TraceLog OpenOrConvert(string etlOrEtlxFilePath, TraceLogOptions o
...
+        /// );
+        /// EventPipeProvider[] providers = new[]
+        /// {   new EventPipeProvider(ClrTraceEventParser.ProviderName, EventLevel.Informational, (long) ClrTraceEventParser.Keywords.Default),
+        ///     new EventPipeProvider(SampleProfilerTraceEventParser.ProviderName, EventLevel.Informational),
+        /// };
+        /// var session = client.StartEventPipeSession(providers, requestRundown: false);
+        /// var eventSource = TraceLog.CreateFromEventPipeSession(session, rundownSession);
+        /// eventSource.Process();
+        /// </example>
+        public static TraceLogEventSource CreateFromEventPipeSession(EventPipeSession session, EventPipeSession rundownSession = null)
```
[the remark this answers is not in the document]
brianrob: Yes, you're right - you can't re-use the session as far as I can tell.  Barring some change in EventPIpeSession or the DiagnosticsClient, I think we'll have to do something here that eventually changes.

It occurs to me that the rundown session needs to be configured properly.  I wonder if we should have another construct in TraceEvent that handles this.  We could do something like this:

Create an EventPipeRundownConfiguration class that has two static factories:
 - None (creates a configuration that does not trigger rundown)
 - Enable (takes a DiagnosticsClient and creates a rundown session)

This would allow us to control how rundown occurs later on, and would allow us to make changes to use StartRundown without users having to change their code.  It's not exactly what I was hoping for, but maybe OK.  What do you think?

brianrob:
> > I've made some changes as requested and replied to some questions.
> > Keeping the testing open until the questions are resolved.
> 
> @brianrob Have you been able to finish reviewing this? Any more concerns or do you think it would be acceptable after adding tests for the changes/new APIs? Also, do you have any preferences for the testing approach?

Just did a pass.  This is looking good.  I added a bit of comment on testing, but not sure if that's what you're looking for.  Let me know and I can think on this some more.  Thanks.

src/TraceEvent/TraceLog.cs
vaind: I'd say it was a bug here - after Process() exited, you wouldn't have received all the events yet because the realtime queue has a flush interval.

I've noticed this issue at a certain time when the EventPipe was also routed through this queue. On the other hand, it doesn't use it anymore so the fix here isn't really helping **me**, but it still is a fix, although unrelated to the changes in this PR...

src/TraceEvent/TraceLog.cs
vaind: OK, changed the TODO to a comment and cleaned up  there.

src/TraceEvent/TraceLog.cs
vaind: thx, removed

src/TraceEvent/TraceLog.cs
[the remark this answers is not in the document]
vaind: Sounds good, I've made the change

src/TraceEvent/TraceLog.cs
vaind: I can't confirm this is a bug in general, but for the realtime processing that happens with EventPipe session, it is necessary to let the process be created.

src/TraceEvent/TraceLog.cs
vaind: Alright, I'll implement the EventPipe session tests similar to what I've done in Sentry dotnet SDK - there should be no problem running them in CI.

vaind: I've added the test. After much headscratching why it was always freezing without an error I've found out that's what happens when you try starting an EventPipe session on a DiagnosticClient connection to a process that is older than .NET core 3.0... 
Since I was trying to connect to the test process itself to test the session, I had to add another `targetframework` to the test project and run the new test conditionally only on the newer .NET. I hope that's alright.

brianrob: Yes, that's great.  I don't think the IPC channel exists on .NET Core 3.x, so I wouldn't expect it to work there.

brianrob:
FYI, I am in the middle of reviewing this, but need to stop right now.  Things are looking good though - I need to look at the test changes and the CI failures.

src/TraceEvent/TraceLog.cs
vaind: It does work since .NET core 3.0. It's only missing on .NET framework.

Also, since I've found out the .netcoreapp3.1 is already being installed in the existing azure pipelines CI workflow, I've adapted the targetframework from net6.0 to netcoreapp3.1 (already in use in `PerfViewCollect` project

vaind:
> FYI, I am in the middle of reviewing this, but need to stop right now. Things are looking good though - I need to look at the test changes and the CI failures.

I've tried a couple of things to get this to work but am always hitting a roadblock.  
Compilation is fixed by switching to .NET 6.0 in the UseDotNet task - it still seems to compile for net462 just fine. Not sure what the task does exactly as I don't have experience with Azure pipelines.
However, tests then start to fail (test runner crashes). I've figured out, this is caused by WPF tests in `StackWindowTests.cs`. Since I can't find a clean solution without drawbacks, I think it will need some input from you which way you'd want to go from here.

brianrob:
FYI I am doing some experiments to see if I can help unblock the tests: https://github.com/microsoft/perfview/pull/1894.

brianrob:
In my testing over in #1894 I discovered that running each of the test assemblies separately works.  It seems that the combination of all of the test runs is an issue.  I'm wondering if the mix of .NET Framework and .NET Core is the issue - I don't have a clear picture of exactly how that's happening on the test side.  I wonder if we need to find a way to do an explicit .NET Core test run and an explicit .NET Framework test run, and choose the specific assemblies that should run for each.  This would allow us to avoid targeting the WPF tests at .NET Core, and would allow us to make sure that we're not attempting to run .NET Framework code against .NET Core and vice versa.  Thoughts?

vaind:
> In my testing over in #1894 I discovered that running each of the test assemblies separately works. It seems that the combination of all of the test runs is an issue. I'm wondering if the mix of .NET Framework and .NET Core is the issue - I don't have a clear picture of exactly how that's happening on the test side. I wonder if we need to find a way to do an explicit .NET Core test run and an explicit .NET Framework test run, and choose the specific assemblies that should run for each. This would allow us to avoid targeting the WPF tests at .NET Core, and would allow us to make sure that we're not attempting to run .NET Framework code against .NET Core and vice versa. Thoughts?

In that case, how about merging this PR first and then there's not so much duplication going on: https://github.com/microsoft/perfview/pull/1896

bruno-garcia:
@brianrob @vaind wondering if there's any update on this? Thanks

ericsampson:
bump?

vaind:
Hi @brianrob, I've updated this PR with the latest changes from the main branch and all the CI passes. Is there anything I should do to so that these changes can eventually land?

src/TraceEvent/TraceLog.cs
```diff
@@ -154,10 +154,192 @@ public static TraceLog OpenOrConvert(string etlOrEtlxFilePath, TraceLogOptions o
...
+                ))
+                {
+                    traceLog.ProcessInitialRundown(rundownSession);
+                }
+            }
+
             return traceLog.realTimeSource;
         }
 
+        public class EventPipeRundownConfiguration
```
brianrob, line 227: Please add comments to the public APIs here.

src/TraceEvent/TraceLog.cs
```diff
@@ -154,10 +154,192 @@ public static TraceLog OpenOrConvert(string etlOrEtlxFilePath, TraceLogOptions o
...
+
+            // currentID is used by the dispatcher to define the EventIndex.  Make sure at both sources have the
+            // same notion of what that is if we have two dispatcher.
+            if (rawEventSourceToConvert != null)
+            {
+                rawEventSourceToConvert.currentID = (EventIndex)eventCount;
+            }
+
+            // Skip samples from the idle thread.
+            if (data.ProcessID == 0 && data is SampledProfileTraceData)
```
brianrob, line 322: This should not happen because we'll always be attached to a real process, and there is no idle process for event pipe.  You can remove this check.

brianrob (commented):
@vaind thank you for your patience on this and for all of the work that you've done thus far.  I think we're getting close.  I've added a few comments here, but I don't think any are big.

I see the following as the path to merge:

1. @noahfalk, can I get you to give this a review please?  This is for livesession support for EventPipe.
2. I would like to validate that we don't regress live session support for ETW traces.  Have you done any testing in this space?  I know we don't have tests here in the repo, but presumably, we can do some adhoc testing, similar to https://github.com/brianrob/examples/tree/main/src/realtime-session-with-stacks.

Assuming we can resolve this, I think we'll be ready to merge (or at least know what we need to do).

noahfalk (approved):
This looked fine to me, but I warn my knowledge of the TraceLog portion of TraceEvent is limited so I wouldn't read too much into it. If there were anything subtly wrong I don't expect I would catch it without spending a while researching the internals of TraceLog.

bruno-garcia:
wooo thanks @noahfalk !

Getting there

src/TraceEvent/TraceLog.cs
vaind: right, thanks for the insight

src/TraceEvent/TraceLog.cs
vaind: docs added to the class & its public methods

vaind:
@brianrob thanks for the review, somehow I've missed it until now.

I've addressed your requests and updated the branch. Let me know if there's anything else needed.

vaind:
P. S. besides review changes, there was one more fix [`36d2e2c` (#1867)](https://github.com/microsoft/perfview/pull/1867/commits/36d2e2c28c1ebeaa780908dc6f831144872a37a5#diff-31ffbfdad96fc84f3aa7d25e1980cdca5a4d7870529c669ac9cc975cf24969c8R342) - the structures would grow indefinitely otherwise (https://github.com/getsentry/sentry-dotnet/issues/3375)

src/PerfView.Tests/StackViewer/StackWindowTests.cs
```diff
@@ -32,54 +32,54 @@ public StackWindowTests(ITestOutputHelper testOutputHelper)
         {
         }
 
-        [WpfFact]
```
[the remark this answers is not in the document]
vaind: Sorry about this, must have been some temp change when testing this and it slipped through because the file showed up as collapsed in the GH PR. reverted to the original

bruno-garcia:
This PR had its 1 year anniversary 😅 

It's got all the review points addressed. Could we please get some help pushing this through 🙏

ericsampson:
@bruno-garcia too bad that Elon tanked Twitter 😝

src/TraceEvent/TraceEvent.csproj
```diff
@@ -51,6 +51,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Diagnostics.NETCore.Client" Version="$(MicrosoftDiagnosticsNETCoreClientVersion)" />
```
brianrob, line 54: This will also require a corresponding addition to `<NuspecProperties>` above, as well as an entry in the dependencies section of the nuspec file.  This is what ensures that NuGet restores the dependency.

brianrob (commented):
Looking good.  One oustanding issue and then I think we're about ready to merge.

brianrob:
> This PR had its 1 year anniversary 😅
> 
> It's got all the review points addressed. Could we please get some help pushing this through 🙏

Yup!  Happy Anniversary?

src/TraceEvent/TraceEvent.csproj
vaind: missed this. it's done now

vaind:
FYI I'm investigating an unbound memory growth when streaming EventPipe. At first, I suspected the cause is this [`GC.KeepAlive`](https://github.com/getsentry/perfview/blob/18df88f900a562f38f6df1500c7ac0ab0a58be6c/src/TraceEvent/EventPipe/EventCache.cs#L198) but it seems there's (also) another issue. Maybe it has to do with pinned buffers in or in EventCache or something else entirely. In any case, I'm unable to pinpoint the issue, mainly because in while memory profiling, the managed memory actually doesn't grow. Or more specifically, managed memory grows, but when a GC is triggered, it is freed but the unmanaged memory stays the same. 
image

On the other hand, if I run the same app with ` procgov64 --maxmem 70M -- app-name.exe` (https://github.com/lowleveldesign/process-governor) the memory usage is stable (with an occasional crash which may be caused by some temporary memory pressure I guess). This, to me, sounds like it may have something to do with how runtime reuses (or doesnt) heaps when there are pinned objects inside... Any ideas? In any case I wouldn't want to block this PR so feel free to merge it if you're fine with the current status and any changes can come in  a followup.

brianrob:
I don't expect the `GC.KeepAlive` call to be a problem.  That's just ensuring that the JIT reports the object as live until the call to `GC.KeepAlive` completes.

It's hard to say where the unmanaged memory growth is coming from.  Does it repro in a minimal repro app?  It might be worth looking at a GC trace (GCStats in PerfView) to see if the GC heap is increasing in size and this is just fragmentation being held, or if there is truly a native leak.

Either way, I am OK taking this change as is, and we can address this in a follow-up.

brianrob (approved):
Thanks for your patience and for bringing this across the finish line!

PR #1868, opened 2023-05-09
Add new instrumented tiers, update html guide for tiering and pgo

AndyAyersMS:
.Net 8 introduces new instrumented tiers into Tiered Compilation. Add event parsing support for these.

Update the HTML guide section on Tiered Compilation, and add some information about OSR and PGO.


AndyAyersMS:
@brianrob PTAL
cc @EgorBo

brianrob (approved):
LGTM.  Thanks.

Issue #1767, opened 2022-12-12
System.Runtime.InteropServices.COMException (0x800700EA): More data is available. (Exception from HRESULT: 0x800700EA)


jiahuizheng1114:
System.Runtime.InteropServices.COMException (0x800700EA): **More data is available.** (Exception from HRESULT: 0x800700EA)
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHRInternal(Int32 errorCode, IntPtr errorInfo)
   at Microsoft.Diagnostics.Tracing.Session.TraceEventSession.**GetActiveSessionNames()**
   at Microsoft.Practices.EnterpriseLibrary.SemanticLogging.Etw.Utility.TraceEventUtil.CreateSession(String sessionName)
   at Microsoft.Practices.EnterpriseLibrary.SemanticLogging.Etw.TraceEventServiceWorker.Initialize()
   at Microsoft.Practices.EnterpriseLibrary.SemanticLogging.Etw.TraceEventService.<>c__DisplayClass21_1.<AddWorkers>b__0()
   at Microsoft.Practices.EnterpriseLibrary.SemanticLogging.Etw.TraceEventService.HandleException(String callerName, Action body)

brianrob:
Can you please put together a set of steps/repro code that causes this?

ryandle:
Hey @brianrob, hope you are well. I just hit this during some local testing. Repro information:

- Windows 11 build 22621.1555
- I have 65 ETW sessions active on the machine

What seems to happen is:

1. `TraceEventSession.GetActiveSessionNames()` gets `MAX_SESSIONS` from `GetETWMaxLoggers()` introduced in #1408 
2. `GetETWMaxLoggers` is checking `SYSTEM\\CurrentControlSet\\Control\\WMI` EtwMaxLoggers key, but that doesn't seem to exist in my registry, so it defaults to 64
3. `GetActiveSessionNames` calls `QueryAllTraces` and gets `ERROR_MORE_DATA`
4. It then throws on the next line with the call to `ThrowExceptionForHR`

So, a few ideas on how to fix:

1. The code seems to assume that EtwMaxLoggers is set in the registry if the system supports more than 64 loggers, but that doesn't seem to be the case on my windows 11 machine. Perhaps there is another way to check for current supported max loggers?
2. An alternative would be to check for the `ERROR_MORE_DATA` return code and call QueryAllTraces again with the new max session count value it has set in the `ref sessionCount` parameter.

Any other ideas?

[...]

PR #1869, opened 2023-05-10
Fix for issue #1767

ryandle:
This change updates GetActiveSessionNames to check for ERROR_MORE_DATA and retry once when calling QueryAllTraces.

I chose to implement this by modifying the current function as little as possible. To do that, I wrapped the core logic and made a recursive call to it for the retry, with a depth counter that maxes out at 1.

This change also has a unit test, but I don't think it is ready to be committed. I am including it so a discussion can be had about it.

In general it seems hard to reliably simulate this condition on a broad set of machines that will run this test suite (65+ active etw sessions while EtwMaxLoggers registry key is not set.) Just between my two developer machines I couldn't do it 🤣 

I'm open to ideas here on how the test can be improved to be more reliable, otherwise it probably needs to be removed.

### DIFF
--- a/src/TraceEvent/TraceEvent.Tests/Regression/Issue1767.cs
+++ b/src/TraceEvent/TraceEvent.Tests/Regression/Issue1767.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.Diagnostics.Tracing;
+using Microsoft.Diagnostics.Tracing.Parsers;
+using Microsoft.Diagnostics.Tracing.Session;
+using Xunit;
+
+namespace TraceEventTests
+{
+    public class Issue1767 : IDisposable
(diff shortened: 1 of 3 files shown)


brianrob (commented):
@ryandle, thank you for contributing this, and sorry for the delay in reviewing it.  Overall, the change to TraceEventSession looks good to me.  I did make a change in this space that is conflicting with your change - can you please update your code to consume the change that allocates `propertiesArray` on the heap?

As far as the unit test goes, I think you are going to have trouble getting this to work almost everywhere.  The only real option I see is to mock it.  I would say that at this point, you can just remove the unit test, but it is important to make sure that this actually works in practice.  Have you seen this work on machine that does have this many sessions?

brianrob:

brianrob:

Issue #1850, opened 2023-03-30
Failed to install perfcollect in Ubuntu 22.04 due to unable to locate package liblttng-ust0


t-bzhan:
perfcollect failed in Ubuntu 22.04 with below error message:
```
Setting up python3-lazr.restfulclient (0.14.4-1) ...
Setting up packagekit-tools (1.2.5-2ubuntu2) ...
Setting up python3-launchpadlib (1.10.16-1) ...
Setting up python3-software-properties (0.99.22.6) ...
Setting up software-properties-common (0.99.22.6) ...
Processing triggers for libc-bin (2.35-0ubuntu3.1) ...
Processing triggers for dbus (1.12.20-2ubuntu4.1) ...
Installing LTTng packages.
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package liblttng-ust0
```

OS info:
```
# cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.2 LTS"
```

From the package search result from Ubuntu: https://packages.ubuntu.com/search?keywords=lttng-ust&searchon=names&suite=jammy&section=all, It looks like we should use liblttng-ust1 for ubuntu 22.04
![image](https://user-images.githubusercontent.com/61817681/228884111-81952aff-2052-4029-94d2-eaf1436d9b9b.png)

PR #1874, opened 2023-05-26
Check whether LTTng is installed and if so, don't attempt to install it.

cincuranet:
The explicit check is to support scenario when somebody installed LTTng from non-package (i.e. compiling from sources).
That in turn is used in perf lab to use LTTng from 2.12 branch to ensure compatibility (at the moment) with runtime.

### DIFF
--- a/src/perfcollect/perfcollect
+++ b/src/perfcollect/perfcollect
@@ -1249,6 +1249,16 @@ InstallLTTng_Ubuntu()
 
 InstallLTTng()
 {
+    if command -v lttng &> /dev/null
+    then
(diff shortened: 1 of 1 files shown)


cincuranet:
@brianrob I used `command` command, because that should have widest compatibility. I'm checking for `lttng`, I think that's good enough estimation for this edge case.

cincuranet:
Related to https://github.com/microsoft/perfview/issues/1850.

brianrob (approved):
LGTM.  Thanks @cincuranet!

PR #1876, opened 2023-06-03
Handle uncompressed embedded source files

pharring:
Embedded source code is usually compressed with the deflate algorithm. However, very small files (up to 200 bytes) are not compressed.
This change handles the non-compressed case for very small files.
Tests were updated to exercise the new code path.


brianrob (approved):
LGTM.  Thanks @pharring!

PR #1886, opened 2023-07-23
Dynamic events

cshung:
This change is the PerfView counterpart of https://github.com/dotnet/runtime/pull/89371

Starting .NET 7, we make `clrgc.dll` compatible with older runtimes, so that you can use a .NET 8 `clrgc.dll` with the .NET 6 app.

However, older runtimes don't know about the new events in the new `clrgc.dll`, which limits our capability to use ETW events to diagnose issues.

To remedy that, we have a `GCDynamic` event that was introduced when `clrgc.dll` was built. The event features a name and an opaque payload. That allows us to serialize any data and fire that instead, and also use the name to disambiguate different events.

Starting .NET 8, we are using it for diagnosing issues related to dynamic heap tuning and memory utilization issues. 

This is the first PR on the PerfView side to handle these dynamic events, and therefore we don't have any prior art on how to expose them. Here I proposed a pattern that works in the sense that we can get the values in the raw XML view as well as on the `TraceGC` objects, but it is not currently available in the Events view and EventStats view.

### DIFF
--- a/src/PerfView/GcStats.cs
+++ b/src/PerfView/GcStats.cs
@@ -228,6 +228,7 @@ namespace Stats
             writer.Write("{0}<GCProcess", indent);
             writer.Write(" Process={0}", StringUtilities.QuotePadLeft(stats.Name, 10));
             writer.Write(" ProcessID={0}", StringUtilities.QuotePadLeft(stats.ProcessID.ToString(), 5));
+
             if (runtime.GC.GCSettings != null)
             {
                 writer.Write(" HardLimit=\"{0}\"", runtime.GC.GCSettings.HardLimit);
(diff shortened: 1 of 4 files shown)


src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -17,6 +18,31 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
     using Microsoft.Diagnostics.Tracing.Parsers.Clr;
     using System.Collections.Generic;
 
+    // TODO, AndrewAu, naming
+    public class CommittedUsageTraceData
```
mrsharm, line 22: Can we seal these classes?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -17,6 +18,31 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
...
+    {
+        public long totalCommittedInUse;
```
mrsharm, line 24: Probably best to add the units (KB/MB) to the names of each of these fields.

mrsharm, line 24: Worth encapsulating these into properties with private sets?

src/PerfView/GcStats.cs
```diff
@@ -228,6 +228,7 @@ public static void ToXml(TextWriter writer, TraceProcess stats, TraceLoadedDotNe
             writer.Write("{0}<GCProcess", indent);
             writer.Write(" Process={0}", StringUtilities.QuotePadLeft(stats.Name, 10));
             writer.Write(" ProcessID={0}", StringUtilities.QuotePadLeft(stats.ProcessID.ToString(), 5));
+
```
mrsharm, line 231: NIT: Should we get rid of this line?

src/PerfView/GcStats.cs
```diff
@@ -454,6 +455,37 @@ public static void ToXmlAttribs(TextWriter writer, TraceProcess stats, TraceLoad
             writer.Write(" AllocedSinceLastGC={0}", StringUtilities.QuotePadLeft(gc.AllocedSinceLastGCMB.ToString("n3"), 5));
             writer.Write(" Type={0}", StringUtilities.QuotePadLeft(gc.Type.ToString(), 18));
             writer.Write(" Reason={0}", StringUtilities.QuotePadLeft(gc.Reason.ToString(), 27));
+            if (gc.committedUsageBefore != null)
+            {
+                writer.Write(" BeforeTotalCommitedInUse={0}", StringUtilities.QuotePadLeft(gc.committedUsageBefore.totalCommittedInUse.ToString("n3"), 10));
```
mrsharm, line 460: Do we need 3 decimal places of precision?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
Maoni0, line 22: @brianrob what are your thoughts on where to put these?

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -844,6 +870,89 @@ internal ClrTraceEventParserState State
                 source.UnregisterEventTemplate(value, 38, ProviderGuid);
             }
         }
+        public event Action<GCDynamicTraceData> GCDynamic
```
Maoni0, line 873: we should certainly think about how to display these in the Events view since we use it heavily. @brianrob could you please give us an idea what work needs to happen for the Events view to display these dynamic events based on the payload (ie, have them on the left hand side so you can select them just like you do with traditional events)? thanks!

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -867,6 +867,31 @@ internal static void SetupCallbacks(TraceEventDispatcher source)
                     }
                 };
 
+                source.Clr.GCDynamic += delegate (GCDynamicTraceData data)
```
cshung, line 870: Get rid of this

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -4613,7 +4656,7 @@ internal void AddServerGCThreadFromMark(int ThreadID, int HeapNum)
         {
             if (IsServerGCUsed == 1)
             {
-                Debug.Assert(HeapCount > 1);
+                //Debug.Assert(HeapCount > 1);
```
cshung, line 4659: Here are some unrelated fixes for dynamic heap adaptation

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -3,6 +3,7 @@
 using Microsoft.Diagnostics.Tracing.Etlx;
 using Microsoft.Diagnostics.Tracing.Utilities;
 using System;
+using System.Collections.Generic;
```
cshung, line 6: Should not require this import

src/PerfView/GcStats.cs
mrsharm: Nvm - it's cleaner this way.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
mrsharm: Discerned that these are all in Bytes. @cshung, please correct me if this is not the case.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
mrsharm: Tracking: https://github.com/microsoft/perfview/issues/1930 - we plan to separate the ability to get the dynamic events in the TraceGC and GCStats and in the Events View.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
mrsharm: @brianrob, any opinions here? We simply added these classes outside the ClrTraceEventParser.

PR #1891, opened 2023-07-29
3 UI changes to the Events view

Maoni0:
1) increasing the columns displayed in the Events view to 10 before resolving to Rest.for someone who uses the Events view heavily only displaying 4 columns before resolving to Rest seems like a unnecessary limitation (was this because this was written at a time when we had very small monitors? 😄). 

2) displaying the Events view window maximized by default because most likely you will want to do that

3) changed the Rest column to fit the window instead of leaving a huge amount of space which means if you select that column you will mostly see just empty space and just lost most of your valuable columns on the left hand side - 

image

### DIFF
--- a/src/CSVReader/EventSource.cs
+++ b/src/CSVReader/EventSource.cs
@@ -173,13 +173,19 @@ namespace EventSources
         /// <summary>
         /// Displays fields as key-value pairs.  
         /// </summary>
-        public virtual string Rest { get { return m_displayFields[5]; } set { m_displayFields[5] = value; } }
+        public virtual string Rest { get { return m_displayFields[11]; } set { m_displayFields[11] = value; } }
         // The properties are for binding in the GUI.   
         // set property is a hack to allow selection in the GUI (which wants two way binding for that case)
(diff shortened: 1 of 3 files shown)


brianrob:
Thanks @Maoni0, these are great changes.  Overall, the code looks reasonable.  The one thing that I noticed during testing is that the horizontal scrollbar doesn't appear even if there is enough contents to warrant one.  This means that you can't scroll to the right to see the rest of the `Rest` column.  Can you take a look and see if you can fix this?

Maoni0:
right, I noticed that too. I just copy the cell and see the data that way. but it's completely rare that I do that because when the cell gets very long, I always resort to multiple columns (it seems quite inconvenient to read such a long line anyway IMO). I have no idea how to make the scroll bar appear due to my extremely limited WPF knowledge 😆 so I'm happy to leave that change out if you like.

brianrob:
Fair enough.  Let's leave that one out so that folks don't trip over it.  I have no idea how to fix it either.  :)

Maoni0:
done :)

brianrob (approved):
LGTM.  Thanks!

PR #1893, opened 2023-08-09
Added a fix for the Computation of the Pause Time % in GC for GCs by not including FGCs

mrsharm:
BGCs shouldn't accumulate the FGCs' Pause Times to compute their corresponding % Pause Time in GC. 
Also, fixes an error with the wrong index being used for the BGC. 

CC: @Maoni0 @cshung


Maoni0 (approved):
LGTM. this also fixes the problem of finding the right index for the BGC

mrsharm:
@brianrob, could you please approve and merge this PR? We found a bug in GC Stats related to BGCs. Thanks!

PR #1899, opened 2023-08-22
Implementation for GetCtfToETWMappings.

cincuranet:
This allows 3rd party parsers to register for known events and process trace file properly.

Current parsers (like `ClrTraceEventParser`) are left unoutched, because these are initialized lazily (i.e.
https://github.com/microsoft/perfview/blob/680ed266e842e49a3c5cd0af89eb7e4f9645b785/src/TraceEvent/TraceEvent.cs#L170).


src/TraceEvent/Ctf/CtfTraceEventSource.cs
```diff
@@ -566,6 +566,15 @@ internal override string ProcessName(int processID, long timeQPC)
             return base.ProcessName(processID, timeQPC);
         }
 
+        internal override void RegisterParserImpl(TraceEventParser parser)
+        {
+            base.RegisterParserImpl(parser);
+            foreach (var (key, mapping) in parser.GetCtfToETWMappings())
+            {
+                _eventMapping.Add(key, mapping);
```
brianrob, line 574: I would like to make sure that we have something that tests this mechanism.  Can I ask you to move the contents of `InitEventMap` to use this mechanism?  The change should be pretty quick and will ensure that this works for any tests that use `CtfTraceEventSource`.

src/TraceEvent/TraceEvent.cs
```diff
@@ -2810,6 +2810,14 @@ protected internal object StateObject
         /// </summary>
         protected internal abstract void EnumerateTemplates(Func<string, string, EventFilterResponse> eventsToObserve, Action<TraceEvent> callback);
 
+        /// <summary>
+        /// Returns the set of ETWMappings between CTF event names and the corresponding ETW Provider GUID, etc.
+        /// </summary>
+        protected internal virtual IEnumerable<(string, ETWMapping)> GetCtfToETWMappings()
```
brianrob, line 2816: Since this is a public API, I would like to avoid using Tuples here, and rather have an API that can sustain changes without all being a breaking change (e.g. additions).  I'm thinking something like:

 - Rename `ETWMapping` to `CtfEventMapping`.
 - Make `CtfEventMapping` a class instead of a struct to avoid copy overhead.
 - Add a first parameter to the constructor of `CtfEventMapping` for the event name.  Also add a public read-only (get) property for the name.
 - Convert the public fields into public read-only (get) properties so that the mappings are immutable.
 - Rename `GetCtfToETWMappings` to `EnumerateCtfEventMappings` to match existing naming conventions in the class.
 - Change the signature of `EnumerateCtfEventMappings` to return `IEnumerable<CtfEventMapping>`.

brianrob (commented):
Thanks @cincuranet!  Some recommendations below.

src/TraceEvent/Ctf/CtfTraceEventSource.cs
cincuranet: I'm not sure I follow here.

* I can't move the existing mappings into their respective classes into `RegisterParserImpl` (like `ClrTraceEventParser`), because these are initialized lazily (see link above) (or not by default like `ClrPrivateTraceEventParser`) and hence the mappings might not be mapped.

* I can't move content of `InitEventMap` into `RegisterParserImpl`, because `RegisterParserImpl` is not called always. For example:
```csharp
var source = new CtfTraceEventSource("...");
source.AllEvents += evt => { };
source.Process();
```

* I can create helper method that takes enumerable of mappings and does the _foreach_ + _add_. This can be called instead of `InitEventMap` from ctor (with all these well-known mappings) and also from `EnumerateCtfEventMappings`. Is that what you have in mind?

brianrob: You bring up a good point - this would likely be considered a breaking change, but does bring up the fact that `CtfTraceEventSource` can't handle arbitrary events without someone creating a parser for them and registering them using your new mechanism.  One way we could solve this is to move the contents of `InitEventMap` into the appropriate parsers, but then force the parsers into existence in the constructor of `CtfTraceEventParser`:

```
GC.KeepAlive(Clr);
GC.KeepAlive(new LinuxTraceEventParser(this));
```

This would allow these components to consume the new mechanism and build a bit of testing into the change while ensuring that we don't break the existing behavior of `AllEvents`.  What do you think?

cincuranet: Fine with me. Implemented.

I only have two concerns.

* The `GC.KeepAlive` looks confusing. I did put a comment there, but still feels like abuse of `GC.KeepAlive`. But I don't have better alternative either. :)
* The comment on top of the parsers says the code was generated by _TraceParserGen_. But looking into _TraceParserGen_ and what it likely produces, it does not match what is currently in these classes. Thus, I'm not sure whether it's really generated sometimes. If it's sometimes regenerated, we can simply create `partial` class with the mappings and all is good (or create (for `LinuxKernelEventParser`) `base` file or update existing (for the rest) `base` files).

Otherwise this should be good to go.

brianrob: This looks good to me.  Thanks for adding a comment near the calls to `GC.KeepAlive`.  Agreed it's not completely clear, but I can't think of something better.

Regarding the auto generated code - this file has been heavily edited, so this would be one of several things we'd need to address if we want to re-generate using `TraceParserGen`.

brianrob (approved):

PR #1906, opened 2023-09-08
Use previous stable release of VSTest

nohwnd:
Temporarily pin version of VSTest to 17.6.3 till we figure out why TRX is not produced on newer versions.


nohwnd:
This should fix you for now, I am not sure what the problem is, sorry :/

nohwnd:
Thanks @brianrob  for pointing out that I am only fixing one of the legs 😊

brianrob (approved):
Thanks @nohwnd!

Issue #901, opened 2019-03-22
Is there any difference between BufferSize and BufferSizeMB arguments?


felipepessoto:
The User Guide mentions both. They are the same thing?

vancem:
They are the same thing.    /BufferSize was the old form, and later PerfView adopted the convention of including units on most things to be clear.    We left the old form to avoid unnecessarily breaking scripts.

PR #1907, opened 2023-09-08
Fix calculation of buffer size to use MB

andd64:
Setting /BufferSize: to a value in MB does not actually set the correct buffer size for a provider, and ends up using the default 64 MB instead.

This fix ensure the buffersize correctly uses MB.

### DIFF
--- a/src/TraceEvent/TraceEventSession.cs
+++ b/src/TraceEvent/TraceEventSession.cs
@@ -121,7 +121,7 @@ namespace Microsoft.Diagnostics.Tracing.Session
                 }
                 if (properties->BufferSize != 0)
                 {
-                    m_BufferQuantumKB = (int)properties->BufferSize;
+                    m_BufferQuantumKB = (int)properties->BufferSize / 1024; //properties->BufferSize is defined in MB
                 }
 
                 m_BufferSizeMB = (int)(properties->MinimumBuffers * m_BufferQuantumKB) / 1024;
@@ -2333,7 +2333,7 @@ namespace Microsoft.Diagnostics.Tracing.Session
             properties->FlushTimer = 60;                // flush every minute for file based collection.  
 
             Debug.Assert(m_BufferQuantumKB != 0);
-            properties->BufferSize = (uint)m_BufferQuantumKB;
+            properties->BufferSize = (uint)m_BufferSizeMB;
             properties->MinimumBuffers = (uint)(m_BufferSizeMB * 1024 / m_BufferQuantumKB);
             properties->LogFileMode = TraceEventNativeMethods.EVENT_TRACE_INDEPENDENT_SESSION_MODE;
 


andd64:
@brianrob Can you please help review (not sure how to add reviewers)?

brianrob:
I think the math here is already correct.  Per https://learn.microsoft.com/en-us/windows/win32/api/evntrace/ns-evntrace-event_trace_properties#members, `BufferSize` is specified in kilobytes.  The value `m_BufferSizeMB` in PerfView gets translated into a calculation for ETW, which is a combination of the actual buffer size (`m_BufferQuantumKB`) and the minimum number of buffers (`properties->MinimumBuffers`).

andd64:
Thanks Brian.  Going by https://github.com/Microsoft/perfview/issues/901, I thought BufferSize is meant to be in MB?


Here are some experiments I did that I think proves it is in MB:
Per https://devblogs.microsoft.com/dotnet/work-flow-of-diagnosing-memory-performance-issues-part-2/:
PerfView /nogui /accepteula /KernelEvents=default+Memory+VirtualAlloc /ClrEvents:GC+Stack /MaxCollectSec:600 /BufferSize:3000 collect

BufferSize under 'Windows Kernel / Event Trace' remains 65535.:
image

Even if i make /BufferSize:3072000 (3000 * 1024), the Windows Kernel/Event Trace still shows 65535.

If I use Xperf with -BufferSize 1024, then Windows Kernel/Event Trace shows:
image

If I use my current commit and rebuild PerView, then /BufferSize:
image

andd64:
Friendly ping @brianrob  =)

brianrob:
You are right that BufferSize gets specified to PerfView in MB.  However, when we specify it to PerfView, it is specified with a quantum size (much smaller) and then a min/max number of buffers.  This allows ETW to allocate in chunks of quantum size instead of allocating one single buffer.  This is much more memory efficient, and is the design that ETW expects.

The xperf command line is different, and so its buffer size parameter translates directly to the quantum size.  It should generally be quite small (same as PerfView).

andd64:
> You are right that BufferSize gets specified to PerfView in MB. However, when we specify it to PerfView, it is specified with a quantum size (much smaller) and then a min/max number of buffers. This allows ETW to allocate in chunks of quantum size instead of allocating one single buffer. This is much more memory efficient, and is the design that ETW expects.
> 
> The xperf command line is different, and so its buffer size parameter translates directly to the quantum size. It should generally be quite small (same as PerfView).

Ah okay I understand now.  The /BufferSize given in MB is just split into 64kb chunks, which result in the minimum number of buffers to be used.  I guess I was confused because running GPUView's log.cmd uses a BufferSize of 1024, and I actually saw that set in the Windows Kernel/Event Trace event.

Thanks @brianrob for the explanation, I will close this PR.

BTW, for future PRs that I open, how do I properly add reviewers? I did not see anything about reviewers in the Contribution Guide (https://github.com/microsoft/perfview/blob/main/CONTRIBUTING.md).

brianrob:
> BTW, for future PRs that I open, how do I properly add reviewers? I did not see anything about reviewers in the Contribution Guide (https://github.com/microsoft/perfview/blob/main/CONTRIBUTING.md).

No need to do anything - we'll pick it up automatically.  Feel free to ping me if I miss it, often it's just me doing the reviews.  I'm working to change that.

PR #1916, opened 2023-09-21
Fix clean build of PerfView.sln

pharring:
TraceEvent doesn't build cleanly. The $OutDir$ variable used in the .nuspec doesn't expand correctly, so the build tries to reference C:\netstandard2.0 The fix is to switch from TargetFrameworks (plural) to TargetFramework.


brianrob:
@pharring thanks for this.  Question - I haven't seen this before, and I don't (think) I see this in the CI.  Am I missing this somewhere?

pharring:
Thanks for checking. Indeed, this looks like this was a machine-specific problem. After a "git clean" it builds cleanly without this patch. Abandoning the PR.

PR #1917, opened 2023-09-27
Provide a StackResolver so that InternStackSource can resolve stacks …

marshall-lerner:
…along the way

Provide an optional way to resolve the stacks as we are interning them, in case somebody wants a fully resolved stack.

Also, change InternFullStackFromSource from a private method to protected, in case there's a class that extends InternStackSource and wants to call the method directly


marshall-lerner:
Closed in favor of #1991

PR #1918, opened 2023-09-29
Add properties for accessing processor time & 

WAftring:
- Add LogfileHeader.[TimerResolution](https://learn.microsoft.com/en-us/windows/win32/api/evntrace/ns-evntrace-trace_logfile_header)
- Add UserTime and KernelTime fields for [ProcessorTime](https://learn.microsoft.com/en-us/windows/win32/api/evntcons/ns-evntcons-event_header)

### DIFF
--- a/src/TraceEvent/ETWTraceEventSource.cs
+++ b/src/TraceEvent/ETWTraceEventSource.cs
@@ -161,6 +161,12 @@ namespace Microsoft.Diagnostics.Tracing
         /// The name of the session that generated the data. 
         /// </summary>
         public string SessionName { get { return logFiles[0].LoggerName; } }
+
+        /// <summary>
+        /// Resolution of the hardware timer, in units of 100 nanoseconds.
+        /// </summary>
+        public ulong TimerResolution { get { return logFiles[0].LogfileHeader.TimerResolution; } }
+
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -161,6 +161,12 @@ public void ReprocessEvent(TraceEvent ev)
         /// The name of the session that generated the data. 
         /// </summary>
         public string SessionName { get { return logFiles[0].LoggerName; } }
+
+        /// <summary>
+        /// Resolution of the hardware timer, in units of 100 nanoseconds.
+        /// </summary>
+        public ulong TimerResolution { get { return logFiles[0].LogfileHeader.TimerResolution; } }
```
brianrob, line 168: The value in the header is of type `uint` is there a reason that it should be `ulong` here?

brianrob (commented):
@WAftring, thank you for your patience on this.  Looks good - one question below.

src/TraceEvent/ETWTraceEventSource.cs
WAftring: My mistake. Addressed.

brianrob (approved):

PR #1919, opened 2023-10-05
Reproduce vstest issue

nohwnd:
Reverting back the version of vstest to get another build that fails to produce TRX so I can debug this.

PR #1920, opened 2023-10-05
Enable running in isolation and --blame

nohwnd:
PerfView tests are running in process because they are .NET Framework tests, and the WPF code is exiting process with 0, which makes the test runner think all tests passed, but in reality only a small portion of tests run.

Make the tests run in a sub-process, where we can detect that it exited before it should, even when it exits with 0. I am also enabling blame so it shows you which test most likely crashed the process. This adds a little bit of overhead, feel free to remove that if you don't need it. It will still crash, but with less info.

### DIFF
--- a/.ado.yml
+++ b/.ado.yml
@@ -45,25 +45,17 @@ jobs:
         solution: PerfView.sln
         configuration: Debug
 
-    - task: VisualStudioTestPlatformInstaller@1
-      displayName: Install Latest Stable VSTest
-      inputs:
-        packageFeedSelector: 'nugetOrg'
-        versionSelector: 'specificVersion'
-        testPlatformVersion: '17.6.3'
-
     - task: VSTest@2
       displayName: 'Run Tests'
       inputs:
         testAssemblyVer2: |
-         **\LinuxTracing.Tests.dll
-         **\Tests.dll
-         **\TraceEventTests.dll
-         **\PerfViewTests.dll
-         !**\*TestAdapter.dll
-         !**\obj\**
+         **\bin\**\LinuxTracing.Tests.dll
+         **\bin\**\Tests.dll
+         **\bin\**\TraceEventTests.dll
+         **\bin\**\PerfViewTests.dll
         testRunTitle: 'PerfView - Debug'
-        vsTestVersion: 'toolsInstaller'
+        runTestsInIsolation: true
+        otherConsoleOptions: '--blame'
 
     - task: CopyFiles@2
       displayName: 'Copy Files to Artifacts Staging'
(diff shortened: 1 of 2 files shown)


nohwnd:
Releated to #1919, that I want to keep for investigation of why it errors out on 17.6.2 and not on 17.7.1, even though they both exit prematurely, and there is no trx in both cases.

nohwnd:
To show that this protects you from re-introducing the same problem, running in isolation reports: 

```
##[error]The active test run was aborted. Reason: Test host process crashed
Results File: D:\a\_temp\TestResults\VssAdministrator_fv-az318-905_2023-10-05_13_27_47.trx
##[error]The active Test Run was aborted because the host process exited unexpectedly. Please inspect the call stack above, if available, to get more information about where the exception originated from.
##[error]The test running when the crash occurred: 
##[error]PerfViewTests.StackViewer.StackWindowTests.TestSetTimeRangeInStartTextBoxRuRu02Async
##[error]This test may, or may not be the source of the crash.
Attachments:
  D:\a\_temp\TestResults\43605865-de85-4c2b-9698-117d1d390da1\Sequence_55ebb1391aba47ebb8ac4cd407c518dc.xml
##[error]Test Run Aborted.
Total tests: Unknown
     Passed: 58
    Skipped: 3
 Total time: 17.3678 Seconds
```

src/PerfView/MainWindow.xaml.cs
```diff
@@ -2055,7 +2055,7 @@ private void Window_Closed(object sender, EventArgs e)
                 DoAbort(null, null);
             }
 
-            Environment.Exit(0);        // TODO can we do this another way?
+            // Environment.Exit(0);        // TODO can we do this another way?
```
nohwnd, line 2058: I can see the process exits when tests run, and I could not figure out when it is needed to close the application explicitly.

.ado.yml
```diff
@@ -45,25 +45,17 @@ jobs:
...
-         **\LinuxTracing.Tests.dll
-         **\Tests.dll
-         **\TraceEventTests.dll
-         **\PerfViewTests.dll
-         !**\*TestAdapter.dll
-         !**\obj\**
+         **\bin\**\LinuxTracing.Tests.dll
+         **\bin\**\Tests.dll
+         **\bin\**\TraceEventTests.dll
+         **\bin\**\PerfViewTests.dll
```
nohwnd, line 55: Can also make this just `**\bin\**\*Tests.dll`, but wanted to give the UI tests a chance to run last, because they have the highest chance of killing the whole run.

brianrob (approved):
Thank you very much @nohwnd for investigating this!  I really appreciate it.

PR #1923, opened 2023-10-10
Add also AlignDown `long` version.

cincuranet:


brianrob:
@cincuranet this seems reasonable.  Can you clarify the intended use of this method?

cincuranet:
I'm using it in the CtfReader alignment fix. 

And also it's same overloads we already have for `AlignUp`. Kind of keeping it in-sync.

brianrob (approved):

PR #1929, opened 2023-10-17
Task Field of Individual Events not Correctly Set

### LINKED ISSUE #1928: Task Field of Individual Events not Correctly Set
This seems like a simple bug to fix.

We use ETLs in our project, recorded at multiple systems, processed and analyzed by self-written tools. We use the "Task" field of the event headers to group events of various types semantically during the analysis. While the results look correct in WPA, the task field is not correctly updated when loading the ETL with the `Microsoft.Diagnostics.Tracing.TraceEvent` library.

I wrote a small console app (c++ emitting events), record it as ETL (attached), and a console app (c#) loading the events with your library to repro the issue:
https://github.com/sgrottel/etl-demo

image
Works in WPA (top half).
Console app using TraceEvent Library does not work (lower half).

Debugging the TraceEvent Library revealed the inconsistency when dispatching the event (from the correctly selected template) to the callbacks:
image
First line in the "watch" window on the left shows the `Task` value in the original EventHeader.
Second line shows the not updated `Task` property in the Event object which is dispatched to the callbacks.

Generated ETL file repro'ing the issue:
[etl-demo.etl.zip](https://github.com/microsoft/perfview/files/12929028/etl-demo.etl.zip)

I got a simple fix working, just setting the `task` field, similar to the `eventRecord` and `userData`.
I'll setup a pull request.

sgrottel:
Explicitly set task field of the returned event template based on the EventHeader.

Closes #1928

### DIFF
--- a/src/TraceEvent/TraceEvent.cs
+++ b/src/TraceEvent/TraceEvent.cs
@@ -3584,6 +3584,7 @@ namespace Microsoft.Diagnostics.Tracing
                         curTemplate.eventRecord = eventRecord;
                         curTemplate.userData = eventRecord->UserData;
                         curTemplate.eventIndex = currentID;
+                        curTemplate.task = (TraceEventTask)eventRecord->EventHeader.Task;
                         currentID = currentID + 1;      // TODO overflow. 
 
                         if ((((int)currentID) & 0xFFFF) == 0) // Every 64K events allow Thread.Interrupt.  


brianrob:
@sgrottel, this is a good find.  I was hoping to scope this fix slightly to just `DynamicTraceEventData`, but the path that I followed to start writing down what the fix would look like led me to more questions.  In the case of your repro app, you're using TraceLogging, which I would have expected to result in use of the TdhParser: https://github.com/microsoft/perfview/blob/main/src/TraceEvent/RegisteredTraceEventParser.cs#L657.  Do you know where the instance of `DynamicTraceEventData` that you're using is being created?  If it gets created here, I would have expected the `TaskID` to be set properly already.

sgrottel:
Thanks @brianrob for your quick reply :+1:

You are absolutely right, that is the spot where the event object is created for the very first occurrence of the event.

The issue is that our analysis tool only logically groups specific event instances with task numbers, e.g. events related to a specific piece of data. The task value of the first event parsed is stored in the event template (`ExternalTraceEventParser.OnNewEventDefintion :L1294`). Later events of the exact same type might have different task values, which is not updated from the template. And then, this issue is triggered.

---

I can think of one alternative implementation:

* Make the `TraceEvent.Task` property `virtual`
* `override DynamicTraceEventData.Task` to not use the `task` field, but to fetch the value directly from `eventRecord->EventHeader.Task`

Then the `task` field would be orphaned for `DynamicTraceEventData` and might run into inconsistencies. That's why I decided for the fix I prepared here.

If you prefer that way, or if you have another solution in mind, please, tell me.
I happily update/replace the change in this PR.

brianrob:
@sgrottel, to confirm my understanding of your setup, are you saying that for a specific event (provider id / event name) you vary the TaskID?  Or am I mis-understanding?

sgrottel:
Yes, you understand correctly.

You can have a look at the example etl file I attached to the Issue #1928 in WPA.
There are events of Provider "EtlProviderApp" with Event Name "WriteEventInts" and one of them set "Task" to 1, while all other have "Task" on 0.

brianrob:
Thanks @sgrottel.  Can you tell me a bit more about why you are varying the task value?  The reason I ask is that I spoke with some folks on the ETW team, and while this is technically possible to do, it's not really expected behavior.  Perhaps we can recommend a different approach that would work with the existing tooling.  As you've seen, TraceLoggingProvider.h doesn't allow you to set the Task, so you have to build all of the metadata on your own in order to do this.

sgrottel:
Replaced by #1977

PR #1934, opened 2023-10-19
Protect Initialization of TraceEventSession Provider Lists

brianrob:
Protect initialization of TraceEventSession.ProviderNameToGuid and TraceEventSession.ProviderGuidToName with a lock to allow for multiple concurrent callers.

This does not change the concurrency guarantees for TraceEvent, but does allow for multiple concurrent callers of TraceEventSession.EnableProvider on different instances of TraceEventSession, where prior to this change it was possible for this to produce a crash on insertion into the Dictionary or SortedDictionary.

Reported via https://github.com/dotnet/runtime/issues/91769.


danmoseley (approved):
thanks!

PR #1935, opened 2023-10-21
Properly setting the process.Log to be used by the GC to discern whether the process is running with SVR GC Threads.

mrsharm:
By correctly getting the Etlx.TraceLog, we have the information necessary to discern whether the process has SVR GC threads. This is important as in DATAS, we start off with 1 heap and then change the heap count and the previous logic was failing to get us the BGC based GlobalHeapHistory event as it is fired after GC/Stop. 

## Solution 

We appropriately set process.Log to point to the pertinent Etlx.TraceLog (if available) by ensuring that the TraceProcess allows an Etlx.TraceLog in its constructor rather than the Analysis.TraceLog; the initial concern was that we'll be coupling the Etlx.TraceLog to the Analysis but this has already been done by setting the type of process.Log to Etlx.TraceLog vs. Analysis.TraceLog.

## Further Details on the Problem

The order of events for BGC based GlobalHeapHistory events were:

-> GC/Start
-> GC/Stop
-> GC/GlobalHeapHistory 

And since we previously didn't have details about the SVR GC threads, we were missing this event as the check to do this is:

```csharp
// This is the last GC in progress. We need this for server Background GC.
// See comments for lastCompletedGC.
private static TraceGC GetLastGC(TraceLoadedDotNetRuntime proc)
{
    TraceGC _event = TraceGarbageCollector.GetCurrentGC(proc);
    if ((proc.GC.m_stats.IsServerGCUsed == 1) &&
        (_event == null))
    {
        if (proc.GC.m_stats.lastCompletedGC != null)
        {
            Debug.Assert(proc.GC.m_stats.lastCompletedGC.Type == GCType.BackgroundGC);
            _event = proc.GC.m_stats.lastCompletedGC;
        }
    }

    return _event;
}
```

### DIFF
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -10,6 +10,7 @@ using Microsoft.Diagnostics.Tracing.Etlx;
 using Microsoft.Diagnostics.Tracing.Parsers;
 using Microsoft.Diagnostics.Tracing.Parsers.Clr;
 using Microsoft.Diagnostics.Tracing.Parsers.ClrPrivate;
+using Microsoft.Diagnostics.Tracing.Parsers.FrameworkEventSource;
 using Microsoft.Diagnostics.Tracing.Parsers.Kernel;
 using Microsoft.Diagnostics.Tracing.Parsers.Symbol;
 using Microsoft.Diagnostics.Tracing.Stacks;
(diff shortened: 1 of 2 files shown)


mrsharm:
CC: @Maoni0 @cshung @markples 

PTAL, @brianrob 

Thanks!

Maoni0:
so why is `process.Log` null? any reason why we can't make it the traceLog you are getting now?

mrsharm:
> so why is `process.Log` null? 

`process.Log` is null because it's hardcoded as a null here: https://github.com/microsoft/perfview/blob/a6c87911fe1aef8f59c9ce54aa4e16a1be6db91e/src/TraceEvent/Computers/TraceProcess.cs#L26

There was probably a reason why this was a null and @brianrob was planning to take a look as to why this is the case. 

> any reason why we can't make it the traceLog you are getting now?
I didn't want to change a property that probably has a reason for being null as it exists at the TraceProcess level and not at the TraceGC level. Will gladly set this once @brianrob confirms this is fine to do.

brianrob:
> There was probably a reason why this was a null and @brianrob was planning to take a look as to why this is the case.
> 
> > any reason why we can't make it the traceLog you are getting now?
> 
> I didn't want to change a property that probably has a reason for being null as it exists at the TraceProcess level and not at the TraceGC level. Will gladly set this once @brianrob confirms this is fine to do.

I think we can safely set the `TraceLog` in the `TraceProcesses` constructor.  From what I can see, this seems completely reasonable.  It's possible that there will be fallout from this, but I'd rather fix that than depend upon this being `NULL` when it doesn't need to be / shouldn't be.

brianrob:
> Excellent - thanks for the clarification, @brianrob! The only downside the solution affords is a potential circular dependency if a developer were to try to process the log inside a callback handler:
> 
> ```cs
> source.Clr.GCSuspendEEStart += delegate (GCSuspendEETraceData data)
> {
>     var process = data.Process();
>     var trace = process.Log;
>     var source = trace.Events.GetSource();
>     source.NeedLoadedDotNetRuntimes();
>     source.Process(); // This infinitely loops.
>    ....
> };
> ```
> 
> Not too important since the onus will be on the developer to test this but thought I'd still mention.
> 
> I can make the aforementioned change and test.

It's true, but that's definitely a developer bug. The possibility for this exists elsewhere too, so this is not a new possibility.

Maoni0 (approved):
looks much better!

brianrob (approved):

brianrob (approved):

PR #1936, opened 2023-10-21
Addressed Feedback for Consuming Dynamic Events 

mrsharm:
Addressed feedback for: https://github.com/microsoft/perfview/pull/1886 since I couldn't add my own changes.

### DIFF
--- a/src/TraceEvent/Parsers/ClrTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/ClrTraceEventParser.cs
@@ -15,38 +15,8 @@ using Address = System.UInt64;
 namespace Microsoft.Diagnostics.Tracing.Parsers
 {
     using Microsoft.Diagnostics.Tracing.Parsers.Clr;
+    using Microsoft.Diagnostics.Tracing.Parsers.GCDynamicData;
     using System.Collections.Generic;
-
-    // TODO, AndrewAu, Decide to convert this to a property.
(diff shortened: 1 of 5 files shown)


src/TraceEvent/Parsers/GCDynamicData.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace Microsoft.Diagnostics.Tracing.Parsers.GCDynamicData
```
mrsharm, line 1: Change the name to GCDynamicEvents

src/TraceEvent/Parsers/GCDynamicData.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    public sealed class CommittedUsageTraceData
```
mrsharm, line 3: Change the name of the classes; instead of Data, it should be Events

src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -7946,6 +8047,75 @@ internal GCBulkRCWValues(TraceEvent data, int baseOffset)
         #endregion
     }
 
+    public sealed class GCDynamicTraceData : TraceEvent
```
mrsharm, line 8050: Change this to GCDynamicTraceEvent

PR #1937, opened 2023-10-23
Fix datatype for AlignDown.

cincuranet:


src/TraceEvent/Ctf/CtfMetadataTypes.cs
```diff
@@ -51,7 +51,7 @@ public static int AlignDown(int val, int alignment)
             return val & ~(alignment - 1);
         }
 
-        public static long AlignDown(long val, long alignment)
+        public static long AlignDown(long val, int alignment)
```
brianrob, line 54: This seems fine, but just curious - what kind of issue did you run into here?

cincuranet: No issue. Just noticed the discrepancy (rest is using `int` as alignment and it makes sense) when working there.

brianrob (approved):

PR #1943, opened 2023-11-06
Reset FocusProcess when checkbox is unselected

ocoanet:
Once set, the "Focus process" option cannot be reset because `PopulateCommandLineArgs` never sets `FocusProcess` to null.

This is a very simple fix without unit tests. I can add tests for `RunCommandDialog` if needed, but it requires to slightly change the constructor to allow skipping the elevation.


brianrob:
@ocoanet thank you for your contribution.  Can you please share a set of repro steps for this?  I want to make sure that I understand the bug that you're fixing.

brianrob:
Got it.  Thank you @ocoanet.

brianrob (approved):

PR #1945, opened 2023-11-14
Add missing entries for TypeLoadStopTraceData

kunalspathak:
The 3 entries were missing from `TypeLoadStopTraceData` because of which we are getting `null` for `PayLoadByName("TypeName").


brianrob (approved):

kunalspathak:
@brianrob - is it possible to release a new version of PerfView with this change so @safihamid can consume it in their prod environment?

brianrob:
This should be in a release soon as we have one coming up shortly.  Please do let me know if this is urgent.

PR #1949, opened 2023-11-22
Replace images with Mermaid diagrams for TraceEvent Programming Guide

cincuranet:


cincuranet:
Not sure the change to Mermaid would be acceptable, but I'm giving it a shot. It looks bit different, but the benefit is that it's easier to modify/maintain/...

brianrob:
This seems very reasonable to me.  Thanks @cincuranet!

brianrob (approved):

PR #1953, opened 2023-11-30
Support ubuntu 22.04 in perfcollect

sebastienros:
Fixes https://github.com/microsoft/perfview/issues/1850

### DIFF
--- a/src/perfcollect/perfcollect
+++ b/src/perfcollect/perfcollect
@@ -1244,7 +1244,14 @@ InstallLTTng_Ubuntu()
     BlueText
     echo "Installing LTTng packages."
     ResetText
-    apt-get install -y lttng-tools lttng-modules-dkms liblttng-ust0
+
+    # Handle Ubuntu 22.04.
+    distro=$(lsb_release -c -s)
+    if [ ${distro} != 'jammy' ]; then
+        apt-get install -y lttng-tools lttng-modules-dkms liblttng-ust0
+    else
+        apt-get install -y lttng-tools lttng-modules-dkms liblttng-ust1
+    fi
 }
 
 InstallLTTng()


brianrob (approved):

cincuranet:
@brianrob I think we discussed this before. We didn't want to implicitly install `liblttng-ust1` because the user-space tracing is not going to work because of runtime incompatibility with ABI changes in LTTng. 

Is runtime side fixed now?

brianrob:
@cincuranet this sounds familiar now that you mention it - I was thinking that it was still useful to install because we can get kernel events, but I wasn't thinking because `lttng-ust` isn't used for kernel events.

I do worry a bit about installing this and then not having things work.  I don't think that this is fixed, as I don't think there are any plans to fix it.  I wonder if we should just avoid install `lttng-ust` on newer distros and just have `perfcollect` print a warning.  What do you think?

cincuranet:
I think we can look at it from two angles.

1. You know the story about `lttng-ust` and you're aware of the consequences and you simply want to conveniently install all that needs to be installed and do whatever you're trying to do.
2. You don't know about the ins and outs. The warning/error is probably a good idea to prevent unexpected surprises.

Not sure which of the two is more common. 

Now as I'm writing this, what about printing warning and then either asking whether to continue or relying on `force` (either reusing the existing one or having separate) switch to either stop (when not present) or continue (when present)?

brianrob:
@cincuranet, I like your idea about warning the user.  I think we should do that, but rather than prompting, let's just say "use -f to force installation".  And then, we will need to make sure that the collect code works properly so that it continues to work even if lttng isn't installed (it just skips that part).  I think this just means that we need to detect and set `$useLttng=0`.

brianrob:
FWIW, someone just reported this exact issue.  LTTng installed, but they never get any events because it's version 2.13.x.

cincuranet:
@brianrob, OK, I'll take it from here and prepare PR.

PR #1964, opened 2024-01-10
Explicit install of LTTng UST on Ubuntu 22 or higher.

cincuranet:
Based on discussion from #1953.

The `nolttng` switch comes handy here and should be a good path.

### DIFF
--- a/src/perfcollect/perfcollect
+++ b/src/perfcollect/perfcollect
@@ -1245,12 +1245,22 @@ InstallLTTng_Ubuntu()
     echo "Installing LTTng packages."
     ResetText
 
-    # Handle Ubuntu 22.04.
-    distro=$(lsb_release -c -s)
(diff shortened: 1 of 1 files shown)


brianrob:
@cincuranet thanks!  FYI, it is OK to not use `-nolttng` when capturing.  It's just that you still won't get any data from the .NET runtime, since the tracepoint provider won't get loaded.

@sebastienros, you will want to change your code to specify `-force` during installation, since I believe you are collecting kernel events (can't remember for sure).

brianrob (approved):

cincuranet:
@brianrob You're right about `nolttng`. I removed it from the message.

brianrob (approved):
LGTM.  Let's make sure @sebastienros can react to this if need be before we merge.

sebastienros (approved):

PR #1969, opened 2024-01-11
Update PerfCollect test containers to use Mariner.

cincuranet:
We can't use Ubuntu because that is causing warning from Secure Supply Chain Analysis. Mariner is basically only way to go.


brianrob:
I like this change, but think it might be worthwhile to run against multiple distros.  Thoughts on add Mariner 2.0 to Ubuntu?

cincuranet:
Updated the description as I marked this as _ready for review_.  If we want to continue using Ubuntu, I think we need to get an exception from Secure Supply Chain Analysis.

brianrob:
Oh, I see what you're referring to.  I think this is fine.

brianrob (approved):

PR #1975, opened 2024-01-19
Maintain stable root names for heap snapshots

brianrob:
The goal of this change is to ensure that we don't unexpectedly change root names.

 - Detect if the set of roots changes via a test.
 - Add code comments to document these changes via release notes.

### DIFF
--- a/src/HeapDumpCommon/GCHeapDumpNames.cs
+++ b/src/HeapDumpCommon/GCHeapDumpNames.cs
@@ -1,0 +1,31 @@
+internal static class GCHeapDumpNames
+{
+    internal const string GCHeapsTitle = "GC Heaps";
+    internal const string DotNetRootsTitle = ".NET Roots";
+    
+    internal const string StaticVarsRootTitle = "static vars";
+    internal const string ThreadStaticVarsRootTitle = "thread static vars";
+    internal const string OtherHandlesRootTitle = "other Handles";
+    internal const string DependentHandlesRootTitle = "Dependent Handles";
+    internal const string LocalVarsRootTitle = "local vars";
+    internal const string PinnedLocalVarsRootTitle= "Pinned local vars";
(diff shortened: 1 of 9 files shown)


cincuranet (dismissed):

leculver:
This looks right to me, though the reported issue of names changing feels different from what's being fixed here.  They were taking a dependency on some things like "Strong handle" vs "StrongHandle" vs etc, but the code as it exists here wouldn't produce different values.

In short, this change looks like a good addition, definitely check it in...but I feel like we are missing something.

leculver (approved):

brianrob:
You're right - I did a bit more digging and it looks like the path that creates a gdcump file by listening to ETW specifies a different set of strings.  Let me work to unify this path and then update this PR.

brianrob:
I just added another commit that unifies names for the two paths that create heap snapshots.  This is a breaking change, but should work for tools that are already trying to consume both cases.

src/HeapDumpCommon/GCHeapDumpNames.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    internal const string GCHeapsTitle = "GC Heaps";
+    internal const string DotNetRootsTitle = ".NET Roots";
+    
+    internal const string StaticVarsRootTitle = "static vars";
+    internal const string ThreadStaticVarsRootTitle = "thread static vars";
+    internal const string OtherHandlesRootTitle = "other Handles";
+    internal const string DependentHandlesRootTitle = "Dependent Handles";
+    internal const string LocalVarsRootTitle = "local vars";
+    internal const string PinnedLocalVarsRootTitle= "Pinned local vars";
```
cincuranet, line 11: nit: Missing space before `=`.

cincuranet (approved):

cincuranet (approved):

asundheim (approved):

PR #1976, opened 2024-01-22
Update Mariner version

cincuranet:


brianrob (approved):
LGTM.  If we keep getting these, we should chat with the dependabot folks about how to get these to show up on GitHub.

PR #1977, opened 2024-01-23
Opt-in Support for TaskID Lookup Directly From the Event Stream

brianrob:
Inspired by #1928 and #1929.

By default, TaskID is not looked up from the event stream for each event, and is instead pulled from event metadata or hardcoded in the parser. This change allows users to ask `ETWTraceEventSource` to lookup the TaskID from the event stream (present in the header) for each event.  This enables tools that vary the TaskID for a given event ID to function properly.

### DIFF
--- a/src/TraceEvent/ETWTraceEventSource.cs
+++ b/src/TraceEvent/ETWTraceEventSource.cs
@@ -84,6 +84,15 @@ namespace Microsoft.Diagnostics.Tracing
             this.fileNames = fileNames;
         }
 
+        /// <summary>
+        /// Allows users to change the characteristics of dispatch behavior.
+        /// Behavior is undefined if this field is changed while processing is in-progress.
+        /// </summary>
+        public TraceEventSourceDispatchBehavior DispatchBehavior
+        {
+            get; set;
+        }
+
         // Process is called after all desired subscriptions have been registered.  
         /// <summary>
         /// Processes all the events in the data source, issuing callbacks that were subscribed to.  See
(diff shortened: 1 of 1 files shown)


src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -84,6 +84,24 @@ public ETWTraceEventSource(IEnumerable<string> fileNames)
...
+        public TraceEventSourceDispatchBehavior DispatchBehavior
+        {
+            get
+            {
+                return dispatchBehavior;
+            }
+            set
+            {
+                dispatchBehavior = value;
+                lookupTaskIDFromEventStream = (dispatchBehavior & TraceEventSourceDispatchBehavior.FetchTaskIDFromEventStream)
```
cincuranet, line 100: I would use `dispatchBehavior.HasFlag(TraceEventSourceDispatchBehavior.FetchTaskIDFromEventStream)`. JIT will handle it properly.

It could maybe even completely replace the `lookupTaskIDFromEventStream` variable and go directly to L860, but I don't have a strong opinion here.

brianrob: You are right.  I did a bit of benchmarking, and I should be able to simplify this a bit further and remove the boolean.

```
BenchmarkDotNet v0.13.12, Windows 11 (10.0.22621.3007/22H2/2022Update/SunValley2)
Intel Core i7-8700 CPU 3.20GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET SDK 8.0.200-preview.24060.48
  [Host]     : .NET 8.0.0 (8.0.23.53103), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.0 (8.0.23.53103), X64 RyuJIT AVX2


| Method     | Mean      | Error     | StdDev    |
|----------- |----------:|----------:|----------:|
| BitwiseAnd | 0.4703 ns | 0.0211 ns | 0.0187 ns |
| HasFlag    | 0.4714 ns | 0.0116 ns | 0.0109 ns |
| BoolValue  | 0.4635 ns | 0.0159 ns | 0.0141 ns |
```

cincuranet (approved):

PR #1985, opened 2024-02-01
Update to current .NET SDK in pipelines

cincuranet:
Part of #1972.


brianrob:
Looks good.  We should also update `PerfViewCollect` to .NET 8, but that can be done separately.

brianrob (approved):

PR #1991, opened 2024-02-06
Provide a StackResolver so that InternStackSource can resolve stacks

marshall-lerner:
…along the way

Provide an optional way to resolve the stacks as we are interning them, in case somebody wants a fully resolved stack.

Also, change InternFullStackFromSource from a private method to protected, in case there's a class that extends InternStackSource and wants to call the method directly


brianrob (approved):

src/TraceEvent/Stacks/Stacks.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.  All rights reserved
 // This file is best viewed using outline mode (Ctrl-M Ctrl-O)
 
+using Microsoft.Diagnostics.Tracing.AutomatedAnalysis;
```
brianrob, line 4: @marshall-lerner, can you confirm that this using is needed?  Looking at the code diff, I don't think it is required (but could be wrong).

brianrob (commented):
LGTM - one follow-up question.

src/TraceEvent/Stacks/Stacks.cs
marshall-lerner: You're right! Removing it now :)

brianrob (approved):

PR #1999, opened 2024-02-29
Properly handle self-describing events.

cincuranet:


src/TraceEvent/TraceEvent.cs
```diff
@@ -1383,6 +1383,16 @@ public virtual unsafe TraceEvent Clone()
             }
             return ret;
         }
+
+        public virtual TraceEvent CloneToTemplate()
```
brianrob, line 1387: Does this need to be public?  If not, my preference is to keep this detail internal.

src/TraceEvent/TraceLog.cs
```diff
@@ -1115,6 +1125,11 @@ private unsafe void SetupCallbacks(TraceEventDispatcher rawEvents)
             rawEvents.Clr.GCFinalizeObject += doNothing;
             rawEvents.Clr.MethodJittingStarted += doNothing;
 
+            if (IsRealTime)
```
brianrob, line 1128: Would be good to add a quick comment here that this is required to ensure that self-describing metadata gets ingested before the event's extended data gets overwritten by the `TraceLog`.

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -1038,7 +1038,7 @@ protected internal override Delegate Target
             get { return m_target; }
             set
             {
-                Debug.Assert(m_target == null);
+                Debug.Assert(m_target == null || value == null);
```
brianrob, line 1041: Would be good to add a quick comment here that value can be set to `null` when cloning and converting to a template for live session dispatch.

src/TraceEvent/TraceLog.cs
```diff
@@ -642,6 +642,16 @@ private static void RemoveAllButLastEntries<T>(ref GrowableArray<T> growableArra
         /// </summary>
         private unsafe void DispatchClonedEvent(TraceEvent toSend)
         {
+            realTimeSource.RegisterUnhandledEventImpl(te =>
+            {
+                if (toSend.containsSelfDescribingMetadata)
```
brianrob, line 647: Would be good to have a comment here describing the flow of self-describing metadata.  Something along the lines of:

Self describing metadata doesn't make it to `realTimeSource` because `TraceLog` overwrites the event's extended data.  To address this, we use the event from the raw source, which contains the metadata, and convert it to a template that can be registered with `realTimeSource`.

brianrob (commented):
@cincuranet, this looks great.  A few small things.

src/TraceEvent/TraceEvent.cs
cincuranet: My mistake. As I was around the `Clone` I used the same access modifiers.

src/TraceEvent/TraceLog.cs
cincuranet: 👍

src/TraceEvent/DynamicTraceEventParser.cs
cincuranet: 👍

src/TraceEvent/TraceLog.cs
cincuranet: 👍

brianrob (approved):

PR #2000, opened 2024-02-29
Stack view for Contention events

ezsilmar:
Contention is an important factor that frequently prevents dotnet applications from scaling well.

Luckily, the dotnet runtime provides `Contention/Start` and `Contention/Stop` diagnostic events with all necessary data. However, the usual AnyStack view is not enough: it doesn't take into account lock duration and precise information about each contention case is impossible to get from the UI.

The new view solves this issue, providing a straightforward way to analyze contention.

- Metric is equal to the duration of contention in milliseconds
- `EventData DurationNs` is a fake frame that allows to distinguish separate events and get precise duration.

image
image


src/TraceEvent/Computers/ContentionTimeComputer.cs
```diff
@@ -0,0 +1,63 @@
...
+            eventSource.Process();
+            
+            _stackSource.DoneAddingSamples();
+        }
+
+        private void OnContentionStart(ContentionStartTraceData startData)
+        {
+            if (!_samplesPerThread.TryGetValue(startData.ThreadID, out var sample))
+            {
+                _samplesPerThread[startData.ThreadID] = sample = new StackSourceSample(_stackSource);
```
brianrob, line 39: Instead of using `startData.ThreadID` here, please use `startData.Thread().ThreadIndex`.  This will handle the case where thread IDs get reused without you having to keep track of process/thread creation/termination.

src/TraceEvent/Computers/ContentionTimeComputer.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (!_samplesPerThread.TryGetValue(stopData.ThreadID, out var sample))
+            {
+                // no corresponding start event
+                return;
+            }
+
+            sample.Metric = (float) (stopData.DurationNs / NanosInMillisecond);
+            sample.StackIndex = _stackSource.Interner.CallStackIntern(_stackSource.Interner.FrameIntern($"EventData DurationNs {stopData.DurationNs:N0}"), sample.StackIndex);
+
+            _stackSource.AddSample(sample);
```
brianrob, line 60: After adding the sample, please clear the `StackIndex`, `Metric`, and `TimeRelativeMSec`.  Then, you should also look for these unset values in `OnContentionStop` and bail out early.  This will handle the case where you see a previous `Contention/Start` that paired properly with a `Contention/Stop`, but then where a `Contention/Start` was missed.  In that case, the current code would add the time range from the first `Contention/Start` to the first `Contention/Stop` and then the time from the first `Contention/Start` to the second `Contention/Stop`.

src/TraceEvent/Computers/ContentionTimeComputer.cs
```diff
@@ -0,0 +1,63 @@
...
+        private void OnContentionStop(ContentionStopTraceData stopData)
+        {
+            if (!_samplesPerThread.TryGetValue(stopData.ThreadID, out var sample))
+            {
+                // no corresponding start event
+                return;
+            }
+
+            sample.Metric = (float) (stopData.DurationNs / NanosInMillisecond);
+            sample.StackIndex = _stackSource.Interner.CallStackIntern(_stackSource.Interner.FrameIntern($"EventData DurationNs {stopData.DurationNs:N0}"), sample.StackIndex);
```
brianrob, line 58: I am thinking that adding the `DurationNs` to the top frame can create a very large number of top-level frames, and the time won't aggregate together.  Have you found that we should aggregate these together by default, maybe with a FoldPat, or does this look OK in your testing?

src/TraceEvent/Computers/WaitHandleWaitTimeComputer.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System.Collections.Generic;
+using Microsoft.Diagnostics.Tracing.Etlx;
+using Microsoft.Diagnostics.Tracing.Parsers.Clr;
+using Microsoft.Diagnostics.Tracing.Stacks;
+
+namespace Microsoft.Diagnostics.Tracing.Computers
+{
+    public class WaitHandleWaitTimeComputer
```
brianrob, line 8: Please apply the same comments from the `ContentionTimeComputer` to this one as well.

src/PerfView/PerfViewData.cs
```diff
@@ -9144,6 +9171,36 @@ protected internal override StackSource OpenStackSourceImpl(string streamName, T
                         stackSource.DoneAddingSamples();
                         return stackSource;
                     }
+                case "Contention":
+                    {
```
brianrob, line 9175: It is safe to add both of these views for ETW traces as well.  Would you mind adding these to the `Advanced` group there as well?  You can also remove the setting for `OnlyManagedCodeStacks` when adding for ETW traces, since they can capture both native and managed frames.

brianrob (commented):
@ezsilmar, thank you very much for your contribution.  I'd be happy to accept this.  A few comments/questions below.

src/TraceEvent/Computers/WaitHandleWaitTimeComputer.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            if (!_samplesPerThread.TryGetValue(stopData.ThreadID, out var sample))
+            {
+                // no corresponding start event
+                return;
+            }
+
+            var durationMs = stopData.TimeStampRelativeMSec - sample.TimeRelativeMSec;
+            sample.Metric = (float)durationMs;
+            sample.StackIndex = _stackSource.Interner.CallStackIntern(_stackSource.Interner.FrameIntern($"EventData DurationNs {durationMs*NanosInMillisecond:N0}"), sample.StackIndex);
```
verdie-g, line 59: What do you think about using milliseconds with maybe 3 decimals instead of nanoseconds? It would be easier to read and I think perfview already uses milliseconds in several case (e.g. gcstats)

src/TraceEvent/Computers/ContentionTimeComputer.cs
ezsilmar: I did that but there are other cases that I'm not sure how to treat meaningfully. For example, we could receive Start -> Start sequence and the current code will just silently drop the first Start. I wonder if I should signal it somehow to the user to indicate the data is flawed? However in this case it'd mean that something is very wrong either with clr or with parsers, not sure what user can do about that.

src/TraceEvent/Computers/ContentionTimeComputer.cs
ezsilmar: I added the FoldPat that hides it by default. 

The nice thing about these durations is that if you look at individual pauses you get the largest on top, so you can instantly estimate how badly it impacts latency (or whatever other thing the user investigates). But indeed in many cases one just wants to know which path is the most prone to contention, so folding by default makes sense.

src/PerfView/PerfViewData.cs
ezsilmar: Added them but I didn't test that it works

src/TraceEvent/Computers/WaitHandleWaitTimeComputer.cs
ezsilmar: I think nanoseconds are easier because they don't have a decimal point. I use `N` format so 1.2 ms duration looks something like `1'234'556`. It also allows to visually compare individual pauses by string length + show the real duration that is stored with ns precision instead of `0.000`

ezsilmar:
@brianrob please have a look at the update. I also added the help entry for both views, hopefully I didn't write something totally incorrect there 😅.

src/TraceEvent/Computers/ContentionTimeComputer.cs
brianrob: Yes, that is a risk.  It can show up in a couple of ways:

1. The trace ends before the End event gets fired.
2. The trace is missing events.

Generally speaking, I don't worry about this, and take one of two approaches:

1. Ignore the incomplete time ranges.
2. Annotate the incomplete time ranges in a way that lets them be filtered out separately if need be via ExcPats.

You can also partially solve the problem by looking for starts that never have another start or stop event during the trace, and setting the stop time to be the end of the trace.  This accounts for starts that are missing stops due to completion of the trace (case 1 above).

For this PR, I'm OK either way.  Missing events aren't super common, and we can often capture new traces when this happens.  It's the cases where the trace starts or ends in the middle of an operation that's much more common.

src/TraceEvent/Computers/ContentionTimeComputer.cs
brianrob: Totally agree with you on this, and I think this is a nice way to keep more detailed information.

src/PerfView/PerfViewData.cs
brianrob: Yes, that's right.

src/TraceEvent/Computers/ContentionTimeComputer.cs
ezsilmar: I rewrote the code to handle all of these cases and also to remove the duplication between Contention and WaitHandleWait. 

Special cases are handled as following:
* Stop is the first event of a thread: not an error, set metric to `timestamp of stop - start of the trace`
* Start is the last event of a thread: not an error, set metric to `end of trace - timestamp of start`
* Start -> Start sequence: error, add `BROKEN_EVENT NO_CORRESPONDING_STOP` to the stack, set metric to `float.Epsilon`
* Stop -> Stop sequence: error, add `BROKEN_EVENT NO_CORRESPONDING_START` to the stack, set metric to `float.Epsilon`

src/PerfView/PerfViewData.cs
ezsilmar: I did the testing, it works! However, I had a lot of trouble collecting stacks on Windows with PerfView (it works fine with `dotnet-trace` for the same process). I used this command line at the end, it does collect the stacks but for some reason the symbols are not resolved correctly; I get `?!?`:

```
PerfView.exe  "/DataFile:PerfViewData.etl" /BufferSizeMB:256 /StackCompression /CircularMB:500 /KernelEvents:None /ClrEvents:None /Providers:"Microsoft-Windows-DotNETRuntime:0x40000004000:Verbose:@StacksEnabled=true;@ProcessIDFilter=1234" /NoGui /NoNGenRundown collect
```

src/TraceEvent/Computers/StartStopThreadTimeComputer.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        protected override void RecordAdditionalStopData(StackSourceSample sample, TraceEvent data)
+        {
+            var stopData = (WaitHandleWaitStopTraceData) data;
+            var durationMs = stopData.TimeStampRelativeMSec - sample.TimeRelativeMSec; // recompute the duration to get better precision than float
+            sample.StackIndex = _stackSource.Interner.CallStackIntern(_stackSource.Interner.FrameIntern($"EventData DurationNs {durationMs*NanosInMillisecond:N0}"), sample.StackIndex);
+        }
+    }
+    
+    public abstract class StartStopThreadTimeComputer
```
brianrob, line 82: I like the idea of what you're doing here.  I would recommend calling this something different, since it's not actually a thread time computer, and we have a thread time computer already.  I think I would call this something like `StartStopLatencyComputer`, since it calculates the latency between starts and stops.

src/TraceEvent/Computers/StartStopThreadTimeComputer.cs
```diff
@@ -0,0 +1,229 @@
...
+            _samplesPerThread = new StackSourceSample[eventLog.Threads.Count];
+        }
+
+        protected abstract void Subscribe(TraceLogEventSource eventSource);
+        protected abstract void RecordAdditionalDataOnMissingStartAtTraceStart(StackSourceSample sample);
+        protected abstract void RecordAdditionalStartData(StackSourceSample sample, TraceEvent data);
+        protected abstract void RecordAdditionalDataOnMissingStopAtTraceEnd(StackSourceSample sample);
+        protected abstract void RecordAdditionalStopData(StackSourceSample sample, TraceEvent data);
+        
+        public void GenerateStartStopThreadTimeStacks()
```
brianrob, line 108: Same here - I'd call this `GenerateStacks`.

src/TraceEvent/Computers/StartStopThreadTimeComputer.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        protected StartStopThreadTimeComputer(TraceLog eventLog, MutableTraceEventStackSource stackSource)
+        {
+            _eventLog = eventLog;
+            _stackSource = stackSource;
+            _interner = stackSource.Interner;
+            _samplesPerThread = new StackSourceSample[eventLog.Threads.Count];
+        }
+
+        protected abstract void Subscribe(TraceLogEventSource eventSource);
```
brianrob, line 102: Can you please add comments for each of these methods to make clear what can/should happen in their implementations?

src/PerfView/PerfViewData.cs
brianrob: Glad to hear that it worked.  The reason stacks didn't work is because you're missing kernel and CLR events.  If you remove `/KernelEvents:None` and `/ClrEvents:None` it should work.

brianrob (commented):
Looking good.  Some additional comments.

src/PerfView/SupportFiles/UsersGuide.htm
```diff
@@ -2437,6 +2437,27 @@ <h2>
...
+                            <strong><a id="ContentionStacks">Contention Stacks</a></strong> - This view aggregates Contention events.&nbsp;
+                            Contention event is fired when a thread tries to acquire a managed lock that is currently owned
+                            by another thread. Note that each event provides DurationNs event data that
+                            is folded by default. Unfolding it reveals individual pauses of each thread: this can be useful 
+                            to correlate a particular long wait to another events in the trace. Note that not all contention events
+                            are real OS-level waits: the runtime will first spin wait to try acquire the lock fast. The metric
+                            represents the amount of time spent to acquire the lock in milliseconds.
+                        </li>
+                        <li>
+                            <strong><a id="WaitHandleWaitStacks">WaitHandleWait Stacks</a></strong> - This view is similar to <a href="#ContentionStacks">Contention Stacks</a>
```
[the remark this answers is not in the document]
kouvel: Looks good to me. A couple of suggestions:
- `Contention` events would also be raised by the `Lock` class (`System.Threading.Lock`)
- It may be useful to mention that the `WaitHandleWait` events may not cover all waits, only those that wait on a wait handle (and only alertable ones, which are the typical waits issued from managed code). Waits that bypass the runtime or wait on other synchronous APIs would not raise the event.

src/PerfView/PerfViewData.cs
ezsilmar: I tried with your suggestions and indeed it fixed the symbols however it now collects a lot of other events (like CPU samples) and it's system-wide (guess because my process filter only worked for dotnet events and not kernel ones?)

src/PerfView/SupportFiles/UsersGuide.htm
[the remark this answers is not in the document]
ezsilmar: I rephrased it to include your comments @kouvel , thanks for the review! I didn't know about `System.Threading.Lock`: will it eventually replace `Monitor.Enter` for the usual `lock` construct?

src/TraceEvent/Computers/StartStopThreadTimeComputer.cs
ezsilmar: Done

src/PerfView/SupportFiles/UsersGuide.htm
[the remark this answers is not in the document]
kouvel: Thanks, looks good to me. `Lock` is a new type being introduced, it's separate from `Monitor` and can be used with the `lock` statement. More of the implementation may be shared in the future too, as it is in NativeAOT.

src/PerfView/PerfViewData.cs
brianrob: Yes, that's expected.  The filter will only impact the provider that specifies it, and I don't believe there is a way to do so for the kernel provider.

src/PerfView/SupportFiles/UsersGuide.htm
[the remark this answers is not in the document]
brianrob: Thanks @kouvel!

brianrob (approved):
@ezsilmar, this looks great to me.  Thank you!  Confirming - is this ready to merge?  It is from my perspective.

ezsilmar:
Hi @brianrob yep it’s ready! Thanks for the review

PR #2001, opened 2024-03-02
Add ASP.NET Core Stats view

mahamr:
This PR adds the _ASP.NET Core Stats_ view to the "Advanced" folder (placement was based on where the current IIS Stats and ASP.NET Stats views are located, but can definitely be changed). At the moment it relies on the presence of **Microsoft.AspNetCore.Hosting** events, such as _Request/Start_ and _Request/Stop_. 
This view is added for both regular ETL files, as well as EventPipe files - it's only built/generated in one place so it looks the same from either type of dataset.

While right now it only consumes those Hosting events, the goal for future output is to be able to consume Microsoft.Extensions.Logging events as well to supplement output, or even base the calculations on those events instead of Hosting (say, when Hosting events aren't available but ILogger events are). Another possibility is using DiagnosticSource events if available, such as the Activity Start/Stop events that correspond to Hosting events. One benefit to those latter events, if available, is those contain the status code of the response, whereas the regular Hosting's Request/Stop events do not. 

I was using the IIS Stats view as reference output here, although, one feature I have not currently implemented is the ability to open the Events view and feed it specific ActivityIDs like what IIS Stats does in the Slow Requests section. I did try this initially but the opening of the Events window in IIS Stats view seems to have been built specifically for ETL files, and it didn't work for EventPipe files. The PR code does contain some remnants of this testing in the PerfViewAspNetCoreStats.DoCommand method, but right now nothing triggers DoCommand. This is something I hope to fix sometime soon in the near-ish future, as it would be useful to have IMO.

In a PerfView-captured and compressed file, here's how it's surfaced when the Hosting events are present:
image

And in a dotnet-trace EventPipe file:
image

Some example output:
image
image

A bit more complex output with multiple processes emitting events:
image

### DIFF
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -3401,7 +3401,8 @@ table {
         private const string _eventHostStop = "Host/Stop";
         private const string _eventUnhandledException = "UnhandledException";
         private const string _eventServerReady = "ServerReady";
-        //private const string _iloggerProvider = "Microsoft-Extensions-Logging";
+        private const string _iloggerProvider = "Microsoft-Extensions-Logging";
+        private const string _iloggerProviderFormattedMessage = "Microsoft-Extensions-Logging/FormattedMessage";
 
         // master requests dictionary
         private Dictionary<Guid, ANCHostingRequest> requests = new Dictionary<Guid, ANCHostingRequest>();
@@ -3461,6 +3462,9 @@ table {
                                     request.StartTimeRelativeMSec = traceEvent.TimeStampRelativeMSec;
                                     request.Method = (string)traceEvent.PayloadByName("method");
                                     request.Path = (string)traceEvent.PayloadByName("path");
+                                    if (String.IsNullOrEmpty(request.Path))
+                                        request.Path = "/";
+
                                 }
                                 else
                                 {
@@ -3475,6 +3479,8 @@ table {
                                 request = new ANCHostingRequest(traceEvent);
                                 request.Method = (string)traceEvent.PayloadByName("method");
                                 request.Path = (string)traceEvent.PayloadByName("path");
+                                if (String.IsNullOrEmpty(request.Path))
+                                    request.Path = "/";
                                 request.StartTimeRelativeMSec = traceEvent.TimeStampRelativeMSec;
                                 requests.Add(request.ActivityId, request);
                             }
@@ -3542,10 +3548,31 @@ table {
 
             }
 
-            sw.Stop();
-            log.WriteLine($"AspNetCoreStats: took {sw.ElapsedMilliseconds} milliseconds to iterate through all events.");
+            // fixup requests without a start or a stop
+            if (requests.Count > 0)
+            {
+                foreach (ANCHostingRequest request in requests.Values)
+                {
+                    // DurationMsec only returns -1 if the request is not complete (i.e. does not have both a start and a stop event associated with it)
+                    if (request.DurationMsec == -1)
+                    {
+                        if (request.HasStart)
+                        {
+                            // if the request is not complete but has a start, then it must not have a stop, so mark the duration as such
+                            request.DurationMsec = dataFile.SessionEndTimeRelativeMSec - request.StartTimeRelativeMSec;
+                        }
+                        else
+                        {
+                            request.DurationMsec = request.StartTimeRelativeMSec;
+                        }
+                    }
+                }
+            }
 
-            // at this point we have all the relevant events stored, need to start processing and building the HTML output
+            sw.Stop();
+            log.WriteLine($"AspNetCoreStats: took {sw.ElapsedMilliseconds} milliseconds to process events.");
+
+            // at this point we have all the relevant information, now start processing and creating the HTML output
             sw.Restart();
 
             // main stats
@@ -3556,17 +3583,22 @@ table {
             outputWriter.WriteLine($"<li>RPS (Requests/Sec): {requests.Count / dataFile.SessionDuration.TotalSeconds:N2}</li>");
             outputWriter.WriteLine($"</ul>");
 
+            // for commands:
+            // string detailedRequestCommandString = $"detailedrequestevents:{request.ContextId};{request.StartTimeRelativeMSec};{request.EndTimeRelativeMSec}";
+            // <A HREF=\"command:{detailedRequestCommandString}\">{requestPath}</A>
+
             // output the general/non-request-related events if any
             if (otherEvents.Count > 0)
             {
                 outputWriter.WriteLine("<h3>General Events</h3>");
                 outputWriter.WriteLine("<table border=\"1\">");
-                outputWriter.Write("<tr>");
+                outputWriter.Write("<thead><tr>");
                 outputWriter.Write("<th align='center' title='Raw time of the event in UTC'>Time (UTC)</th>");
                 outputWriter.Write("<th align='center' title='How many milliseconds after the trace started this event was emitted'>Trace Relative Time (msec)</th>");
                 outputWriter.Write("<th align='center' title='The PID that emitted the event'>Process ID</th>");
                 outputWriter.Write("<th align='center' title='The event that was emitted'>Event</th>");
-                outputWriter.WriteLine("</tr>");
+                outputWriter.WriteLine("</tr></thead>");
+                outputWriter.WriteLine("<tbody>");
                 foreach (OtherHostingEvent hostingEvent in otherEvents)
                 {
                     outputWriter.Write("<tr>");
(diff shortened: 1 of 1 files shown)


src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,447 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+
+            // fixup requests without a start or a stop
+            if (requests.Count > 0)
+            {
+                foreach (ANCHostingRequest request in requests.Values)
+                {
+                    // DurationMsec only returns -1 if the request is not complete (i.e. does not have both a start and a stop event associated with it)
+                    if (request.DurationMsec == -1)
+                    {
+                        if (request.HasStart)
```
brianrob, line 3545: Recommend flagging these incomplete requests in the UI somehow so that it's clear to users how the latency was calculated.  This is especially useful in the case of lost events where it might be confusing and cause folks to go after a long request that wasn't actually long.

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,447 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+             * ServerReady
+             */
+
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
+
+            // using the pull/foreach model for iterating through the TraceLog events
+            foreach (TraceEvent traceEvent in dataFile.Events)
+            {
+                if (traceEvent.ProviderName.Equals(_hostingProvider))
```
brianrob, line 3437: Recommend that rather than using `foreach` here, you call `dataFile.Events.GetSource()` and then call `AddCallbackForProviderEvent`, passing in the provider name, and `null` for the event name.  This will give you all events for the provider, but will do in a more efficient manner than iterating over all events in the trace.

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,447 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+            log.WriteLine($"AspNetCoreStats: took {sw.ElapsedMilliseconds} milliseconds to build and flush HTML.");
+        }
+
+        protected override string DoCommand(string command, StatusBar worker)
+        {
+            // string detailedRequestCommandString = $"detailedrequestevents:{request.ActivityId};{request.StartTimeRelativeMSec};{request.EndTimeRelativeMSec}";
+
+            try
+            {
+                if (command.StartsWith("detailedrequestevents:"))
```
brianrob, line 3715: Looks like this is not used.

brianrob (commented):
@mahamr, this is awesome!  Thanks for contributing this.  Some comments below.

Also, this looks like it will work with events that are emitted via nettrace, so we should add this to the nettrace path as well.

brianrob:
FYI @davidfowl

src/PerfView/PerfViewData.cs
mahamr: Will remove this

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,447 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+                            }
+                            break;
+
+                        case _eventManifestData:
+                            // ignore
+                            break;
+
+                        case _eventHostStart:
+                        case _eventHostStop:
+                        case _eventServerReady:
```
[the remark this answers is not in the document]
mahamr: will do

mahamr:
> Also, this looks like it will work with events that are emitted via nettrace, so we should add this to the nettrace path as well.

Thanks @brianrob! This part should be done already in the submitted code unless there's something I missed - check the stuff in `EventPipePerfViewData`.

src/PerfView/PerfViewData.cs
brianrob: Yes, I would still recommend using the callback model.  This allows the dispatch machinery to bail out earlier and not do as much work for events that you don't care about, which are likely to be a large chunk of the total events.

brianrob:
> > Also, this looks like it will work with events that are emitted via nettrace, so we should add this to the nettrace path as well.
> 
> Thanks @brianrob! This part should be done already in the submitted code unless there's something I missed - check the stuff in `EventPipePerfViewData`.

You're right - I missed that this was done for ETW and EventPipe.  Thanks!

src/PerfView/PerfViewData.cs
mahamr: understood and will do

mahamr:
Pushing a commit soon with lots of changes to add more output, as well changes to address requests and concerns so far. 

Some updated screenshots are below.

Incomplete requests (1 each where the request was started before the trace but ended during the trace, and another where the trace captures a request start but not the end):
image

full page of a bit more complex output with multiple processes and some slow requests:
image

and another:
image

src/PerfView/PerfViewData.cs
mahamr: done in latest commit

src/PerfView/PerfViewData.cs
mahamr: redid some of this output and also have some screenshots of the latest commit with some examples of this

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,447 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+                                requests.Add(request.ActivityId, request);
+                            }
+
+                            break;
+
+                        case _eventRequestStop:
+                            if (requests.TryGetValue(traceEvent.ActivityID, out request))
+                            {
+                                // request exists, sanity PID check and create if necessary
+                                if (request.ProcessID == traceEvent.ProcessID)
```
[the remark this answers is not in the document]
mahamr: incomplete and complete requests are now stored separately and marked in the output, we can discuss that implementation and any of its changes here if needed

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,447 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+                if (traceEvent.ProviderName.Equals(_hostingProvider))
+                {
+                    // determine which event we're dealing with
+                    switch (traceEvent.EventName)
+                    {
+                        case _eventRequestStart:
+                            if (requests.TryGetValue(traceEvent.ActivityID, out ANCHostingRequest request))
+                            {
+                                // request exists, sanity PID check and update
+                                if (request.ProcessID == traceEvent.ProcessID)
```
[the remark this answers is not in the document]
mahamr: done, I utilize ProcessIndex under-the-hood, while showing PID in the user-facing output

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,447 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+
+                            break;
+
+                        case _eventUnhandledException:
+                            if (traceEvent.ActivityID != null && requests.TryGetValue(traceEvent.ActivityID, out request))
+                            {
+                                // request exists, sanity PID check and create if necessary
+                                if (request.ProcessID == traceEvent.ProcessID)
+                                {
+                                    request.HasUnhandledException = true;
```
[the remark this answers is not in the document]
mahamr: this has definitely been thought about and I will hopefully get to working on it once this initial implementation is done. There's a few things I want for v2 :)

mahamr:
added a quick/small commit to prevent _EventCounters_ events from spamming the "General Events" section

mahamr:
@brianrob I noticed when opening a trace and opening this stats view, then closing it and reopening it, the **WriteHtmlBody** is run again causing the main data structures to be populated with a second set of events, doubling everything. If closed and reopened again, a new/third set is added, etc. etc. I notice the _PerfViewAspNetStats_ view declares the main **m_requests** data structure within its **WriteHtmlBody** method, effectively doing all the work again:
```
   public class PerfViewAspNetStats : PerfViewHtmlReport
   {
       public PerfViewAspNetStats(PerfViewFile dataFile) : base(dataFile, "Asp.Net Stats") { }
       protected override void WriteHtmlBody(TraceLog dataFile, TextWriter writer, string fileName, TextWriter log)
       {
           var dispatcher = dataFile.Events.GetSource();
           var aspNet = new AspNetTraceEventParser(dispatcher);

           m_requests = new List<AspNetRequest>();
```

What's the best way to approach this in the new view? Should I do the same thing and just re-process everything? Or perhaps set a flag that is used to check whether **WriteHtmlBody** has already run and skip the trace processing part if so? Or is there some other better solution?

Here's what happens right now, first run:
image

Then close the view and reopen it:
image

and so on

Thanks!

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,436 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+                // determine which event we're dealing with
+                switch (traceEvent.EventName)
+                {
+                    case _eventRequestStart:
+                        inboundRequestCount++;
+                        if (incompleteRequests.TryGetValue(ANCHostingRequest.GetIndexingKeyFromEvent(traceEvent), out request))
+                        {
+                            // request exists, update its properties
+                            request.StartTimeRelativeMSec = traceEvent.TimeStampRelativeMSec;
+                            request.HasStart = true;
```
brianrob, line 3452: I think other than right after construction `request.HashStart` always equals `true`.  I think (but do check my work), that you don't need this property.

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,436 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+                            // if the request is complete then move it to the completeRequests List
+                            if (request.IsComplete)
+                            {
+                                incompleteRequests.Remove(request.IndexingKey);
+                                completeRequests.Add(request);
+                            }
+                        }
+                        else
+                        {
+                            // request does not exist yet, create and track it
```
brianrob, line 3496: If you see a stop event, but there isn't a matching request in `incompleteRequests`, is there a reason to start tracking it?  I would not expect you to see any more events associated with this request.

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,436 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+                    case _eventUnhandledException:
+                        if (incompleteRequests.TryGetValue(ANCHostingRequest.GetIndexingKeyFromEvent(traceEvent), out request))
+                        {
+                            // request exists, update it
+                            request.HasUnhandledException = true;
+                        }
+                        else
+                        {
+                            // request does not yet exist, create and track it
+                            request = new ANCHostingRequest(traceEvent);
```
brianrob, line 3514: If you hit this path, then you don't have any information to associate with the request - is it worth showing this, or should we just leave this request out?

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,436 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+                    {
+                        // used for opening the request in the Events view
+                        //string detailedRequestCommandString = $"detailedrequestevents:{request.ActivityId};{request.StartTimeRelativeMSec};{request.EndTimeRelativeMSec}";
+
+                        outputWriter.Write("<tr>");
+                        outputWriter.Write($"<td>{request.ProcessID}</td>");
+                        outputWriter.Write($"<td>{request.Method}</td>");
+                        outputWriter.Write($"<td>{request.Path}</td>");
+                        outputWriter.Write($"<td>{request.DurationMsec:N2}</td>");
+                        //outputWriter.Write($"<td><a href=\"command:{detailedRequestCommandString}\">{StartStopActivityComputer.ActivityPathString(request.ActivityId)}</a></td>");
```
brianrob, line 3708: There are a few extra commented out lines here.  Are these needed?

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,436 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+                outputWriter.Write("<th align='center' title='Request Path - may not be available'>Path</th>");
+                outputWriter.Write("<th align='center' title='Duration of the request in milliseconds'>Duration (msec)</th>");
+                //outputWriter.Write("<th align='center' title='ActivityId of the request - click to open Events view for this ID'>ActivityID</th>");
+                outputWriter.Write("<th align='center' title='ActivityId of the request - this is a good request-specific text filter for the Events view'>ActivityID</th>");
+                outputWriter.WriteLine("</tr></thead>");
+                outputWriter.WriteLine("<tbody>");
+                foreach (ANCHostingRequest request in completeRequests)
+                {
+                    // used for opening the request in the Events view
+                    //string detailedRequestCommandString = $"detailedrequestevents:{request.ActivityId};{request.StartTimeRelativeMSec};{request.EndTimeRelativeMSec}";
```
brianrob, line 3740: Same with these comments.

brianrob (commented):
@mahamr, this is looking good.  A few more comments.

src/PerfView/PerfViewData.cs
mahamr: removed

src/PerfView/PerfViewData.cs
mahamr: not needed, removed

src/PerfView/PerfViewData.cs
mahamr: HasStart and HasStop are both set to _false_ on construction -- they are individually set to _true_ when the corresponding event is found in the TraceLog. This is one area that I did change in the recent commit to help with keeping track of timing vs. event-related completeness. HasStart and HasStop used to return _true_ if the request had a nonzero start or stop time. Now, since I set start/stop time manually for incomplete requests (for timing purposes), I want to ensure that HasStart/HasStop are only true IF an actual start or stop event were encountered. This is then used to help the user see which scenario an incomplete request is.

src/PerfView/PerfViewData.cs
mahamr: This could be useful if there's a request that was started, let's say 1 second, before the trace was started. Then, if a stop event is encountered 10 seconds later, we know that particular request was active for at least 10 seconds. This info is surfaced to the user along with the ActivityId so they can go inspect any related events for it, perhaps to see why it took at least 10 seconds to complete. So while there probably won't be any more events in this particular view, allowing the user to see it allows them to decide whether or not it's useful to them, IMO.

src/PerfView/PerfViewData.cs
mahamr: It's possible there won't be any other events, but there _should_ be a correlated **Request/Stop** event that comes after. 

However, as of right now, the view does not use this information or surface it in any way to the user. Initially, this was more future-prep for plans when CLR events would be checked and any correlated exceptions can be surfaced with it. I can, however, see a quick enhancement to surface this by way of a boolean output to the request tables, something like an unhandled exception column with a yes/no output. Thoughts?

src/PerfView/PerfViewData.cs
```diff
@@ -3388,6 +3388,436 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+                Path = "N/A";
+                HasUnhandledException = false;
+            }
+
+            /// <summary>
+            /// Returns a string that *should* be unique per request, and is thus good for use as a unique index.
+            /// </summary>
+            public static string GetIndexingKeyFromEvent(TraceEvent theEvent)
+            {
+                return theEvent.Process().ProcessIndex.ToString() + theEvent.ActivityID.ToString();
```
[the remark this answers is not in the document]
mahamr: Good point -- to build on top of that, if Activity.ID is empty, then I don't see any way of correlating any events together with any degree of certainty. Perhaps if that's the case, then maybe recognizing this scenario and surfacing this to the user would be best? Some stats like incoming request rate could still be calculated, but there's no way of knowing which requests finished and which ones didn't.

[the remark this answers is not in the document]
brianrob: Yes, agreed.  Without the ActivityID, there isn't really a good way to do this, as requests are not synchronous.

src/PerfView/PerfViewData.cs
brianrob: What you're describing makes sense to me, but I see `request.HasStart = true` being set when a request stop event shows up regardless of whether or not there is an incomplete request.  That's where my thought that this isn't necessary comes from.  Is that just maybe unintentional?  I wouldn't expect to see this when a stop event shows up.

src/PerfView/PerfViewData.cs
brianrob: Sure.  Seems reasonable to me.  I think the key is that we want to make it clear to users that the request time is trustworthy (or isn't).  My suspicion is that the number of times we hit this case is low, or the trace is missing lots of events and the data is generally untrustworthy and should be recollected.

mahamr: I have added some output (a table column) to the request tables to indicate via "yes/no" value whether or not a request has an associated unhandled exception and pushed it. Looks like this:
image

src/PerfView/PerfViewData.cs
mahamr: Hey Brian, unless I missed it, request.HasStart is only set to _true_ in 2 places, both inside of the code for processing a request/start event. In the request/stop event, only the HasStop is set to true:
image
Let me know if I missed anything there.

src/PerfView/PerfViewData.cs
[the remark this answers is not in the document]
mahamr: Sounds great, thanks for bringing this to light. Will update the PR in the next couple of days with these changes.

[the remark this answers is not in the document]
mahamr: new commit pushed that adds the check for an empty GUID - I added a counter for the number of request-related events (request/start, /stop, and /unhandledexception) so if it is nonzero the user would be informed that those were dropped, and only events with ActivityIds were being tracked in the tables

src/PerfView/PerfViewData.cs
brianrob: You're right - looking at this again, I do not see this issue anymore (or I just completely mis-read it).

src/PerfView/PerfViewData.cs
brianrob: This makes sense to me.  If that's the case, then I recommend removing the `if(request.IsComplete)` branch under the `_eventRequestStart` case.  In this case, it would mean that you saw the stop event before the start event, which shouldn't happen.  If it does, I don't think we want to match them, as likely this is a trace issue or a bug.  If an ActivityID gets re-used, we'd definitely want this to reset things for that ActivityID .

src/PerfView/PerfViewData.cs
brianrob: Looks great!

brianrob (commented):
@mahamr, this is looking good.  I think there is just one outstanding thing, and a merge conflict that needs to be addressed.

mahamr:
I'll work on those this week - thanks! There's one other issue I brought up before that I'd like to get your thoughts on the best way to fix:
https://github.com/microsoft/perfview/pull/2001#issuecomment-1998507979

Thanks!

brianrob:
> I'll work on those this week - thanks! There's one other issue I brought up before that I'd like to get your thoughts on the best way to fix: [#2001 (comment)](https://github.com/microsoft/perfview/pull/2001#issuecomment-1998507979)
> 
> Thanks!

Ah, sorry, I missed that one.  There are a couple of patterns that get employed here, but the most common one is just to re-process all of the data for each invocation of the report.  This is what I would recommend that you do.

You'll also see that some of the subcommands will cache their files.  Here's an example: https://github.com/microsoft/perfview/blob/main/src/PerfView/PerfViewData.cs#L3679.  This could be done with the html file as well, but given that none of them do this as far as I can tell, feel free to follow the pattern, and perhaps at some point, we'll fix this for all.

src/PerfView/PerfViewData.cs
mahamr: > In this case, it would mean that you saw the stop event before the start event, which shouldn't happen

To confirm, are you're saying it's guaranteed that:

1. the events are written to the trace in the order they were emitted, and
2. the callbacks are guaranteed to happen in the order of events in the trace

For # 1 - what if the events left the source process out-of-order and were thus received out-of-order by the tracing tool (i.e. PerfView, dotnet-trace, etc.)? 

Thanks!

mahamr:
Hey @brianrob also, any comment on the location of the view? This was mentioned as something to discuss previously. Right now it's in the _Advanced_ view for both ETL (next to IIS and regular ASP.NET stats view) and EventPipe traces -- it this appropriate or should it be changed?

src/PerfView/PerfViewData.cs
brianrob: Point 2 is definitely true - the process of dispatching the events is single threaded.

Point 1 is generally true, but not guaranteed.  If they do leave the source out of order, then yes, I would expect them to be out of order.  Generally though, when we have events that show up out of order, we generally toss them unless we can be sure that the order doesn't matter.  Otherwise, we end up telling a story with the data that isn't correct.

brianrob:
> Hey @brianrob also, any comment on the location of the view? This was mentioned as something to discuss previously. Right now it's in the _Advanced_ view for both ETL (next to IIS and regular ASP.NET stats view) and EventPipe traces -- is this appropriate or should it be changed?

Thanks for checking.  I think Advanced is fine for now.  I can see about organization of the Advanced group separately.

src/PerfView/PerfViewData.cs
mahamr: understood - thanks! I will remove the `if(request.IsComplete)` block as mentioned.

mahamr:
@brianrob I believe everything has been addressed, including the merge conflict that existed. I do have continued plans to add to this view some enhancements, but where it's at I feel is a good v1. Let me know if there's anything else you'd like to see here - thanks!

src/PerfView/PerfViewData.cs
```diff
@@ -3402,6 +3402,454 @@ private void CreateCSVFile(List<AspNetRequest> requests, string csvFileName)
...
+                                incompleteRequests.Remove(request.IndexingKey);
+                                completeRequests.Add(request);
+                            }
+                        }
+                        else
+                        {
+                            // request does not exist yet, create and track it
+                            request = new ANCHostingRequest(traceEvent);
+                            request.EndTimeRelativeMSec = traceEvent.TimeStampRelativeMSec;
+                            request.HasStop = true;
```
brianrob, line 3534: Just thinking about this a little further - I think that we want this else block to not add to incompleteRequests.  If you want to try and track requests that have a stop but no start, that seems OK, but this code would allow a start to show up after a stop and I still think we don't want that.

brianrob (approved):
@mahamr, thank you for building this!

Issue #1700, opened 2022-08-29
Perfview option to display/export actual time/ticks instead of/in addition to the time offset


Malcolm-Stewart:
This is important when comparing PerfView logs with network traces or other logs and trying to line them up.

brianrob:
This seems reasonable.  Where specifically would you want to see this?  Would you be looking to have this information in the events view?

olegsych:
I also would like to have this option to facilitate analysis of long trace logs. Specifically, I would like the Events view to display a column with the absolute (UTC or local) time instead of or in addition to the existing time from the start of trace capture.  The time range controls would also accept absolute time.

brianrob:
This seems reasonable to have.  Would you be interested in submitting a PR for this?

olegsych:
I would, but I’m not familiar with the codebase and would need detailed guidance on where to make the changes, test requirements, etc.
________________________________

iaalm:
Hey, this feature is very useful to me too. Seems there're not update since 2022, I will try make a PR for it. Just FYI. 😀

PR #2008, opened 2024-03-18
Display timestamp while view ETW event

iaalm:
Addressing #1700 
As the `Time MSec` is quite useful for filter, I just adding one extra field.
Manually tested worked.

### DIFF
--- a/src/PerfView/EventViewer/EventWindow.xaml
+++ b/src/PerfView/EventViewer/EventWindow.xaml
@@ -79,11 +79,13 @@
             <Menu DockPanel.Dock="Top" Background="{DynamicResource ControlDefaultBackground}">
                 <MenuItem Header="_File">
                     <MenuItem Header="_Parent Window" Click="DoOpenParent"/>
-                    <MenuItem Header="_Show Local Time" IsCheckable="True" IsChecked="False" Checked="DoUseLocalTime" Unchecked="DoUseOriginTime"/>
                     <MenuItem Header="_Close" Click="DoClose"/>
                 </MenuItem>
+                <MenuItem Header="_View">
+                    <MenuItem Header="_Show Local Time" IsCheckable="True" IsChecked="False" Checked="DoUseLocalTime" Unchecked="DoUseOriginTime"/>
+                </MenuItem>
                 <MenuItem Header="_Help">
-                    <MenuItem Header="_Help on Event Viewer" Command="Help" CommandParameter="EventViewerQuickStart" />
+                        <MenuItem Header="_Help on Event Viewer" Command="Help" CommandParameter="EventViewerQuickStart" />
                     <MenuItem Header="_Users Guide" Command="{x:Static src:EventWindow.UsersGuideCommand}" CommandParameter="UsersGuide"/>
                 </MenuItem>
             </Menu>
--- a/src/PerfView/SupportFiles/UsersGuide.htm
+++ b/src/PerfView/SupportFiles/UsersGuide.htm
@@ -3649,7 +3649,7 @@
         <li>
             <strong>Show Local Time</strong> - The timestamp is displayed in the time zone where
             events are collected (Trace Local). You can switch to the local machine's time zone
-            by clicking on the 'File - Show Local Time' option.
+            by clicking on the 'View - Show Local Time' option.
         </li>
         <li>
             <strong>Launch in Excel </strong>- Right clicking and selecting &#39;Open in Excel&#39;


iaalm:

iaalm:
@brianrob could you help review?

src/PerfView/EventViewer/EventWindow.xaml
```diff
@@ -242,6 +242,13 @@
                         MinWidth="60"
                         IsReadOnly="True" />
                         <WPFToolKit:DataGridTextColumn
+                        x:Name="TimeStampColumn"
+                        Header="TimeStamp"
+                        Binding="{Binding TimeStamp, StringFormat=u}"
```
pharring, line 247: StringFormat u (Universal) is compact, but it lacks precision (nearest second).
Consider `StringFormat=o` (ISO8601) which has more precision.

brianrob, line 247: I would like to make sure that the column description contains the time zone - specifically either "UTC" or "Trace Local".  It looks like `u` is UTC, while `o` contains the time zone information.

brianrob (commented):
Thanks @iaalm for contributing this.  A comment below - also it looks like @pharring has a comment as well.

src/PerfView/EventViewer/EventWindow.xaml
iaalm: Fixed

iaalm: Sure, adding `Trace Local`

pharring:
Time zones are always tricky. My experience with services tells me to convert everything to UTC. That tends to be a safe approach -- avoiding all question about time zone -- but, it's probably not appropriate for PerfView (expect, perhaps as an option).
With PerfView, there are two local time zones to consider: The time zone on the machine where the data was gathered (if known) and the time zone on the machine where PerfView is running. When we say "Trace local", I assume that means that the times will be displayed using the time zone on the machine where the trace was gathered (including any adjustment for daylight saving that was in effect when the trace started).
I seem to remember that ETW sessions include a time zone offset in the header. I don't know if netperf does.

I wanted to call this out, so we are clear and precise about what "Trace Local" means.

Oh, now that I think of it, please update the Help documentation to describe the new column. Add it to the section headed "Column Descriptions". You'll then get a chance to explain what a "Trace Local" timestamp means.

pharring (approved):
See my comment about being precise about time zones. Would be nice to have it precisely defined in the help doc.

iaalm:
@pharring That's a good point. Seems these two time zones are both accessible, so I'm thinking about make an option for it. Maybe default using the time zone of gathered machine.

brianrob:
> Time zones are always tricky. My experience with services tells me to convert everything to UTC. That tends to be a safe approach -- avoiding all question about time zone -- but, it's probably not appropriate for PerfView (expect, perhaps as an option). With PerfView, there are two local time zones to consider: The time zone on the machine where the data was gathered (if known) and the time zone on the machine where PerfView is running. When we say "Trace local", I assume that means that the times will be displayed using the time zone on the machine where the trace was gathered (including any adjustment for daylight saving that was in effect when the trace started). I seem to remember that ETW sessions include a time zone offset in the header. I don't know if netperf does.
> 
> I wanted to call this out, so we are clear and precise about what "Trace Local" means.
> 
> Oh, now that I think of it, we should update the Help documentation to describe the columns in the Event Viewer. There's a section for the Event Viewer, but it doesn't describe the various columns. (There is a "Column Description" section for the Stack Viewer, but not for the Event Viewer) We should add a "Column Description" section to the "The Event Viewer" section. We'll then get a chance to explain what "Time MSec" and "Timestamp (Trace Local)" mean.

Agreed with @pharring's comments here.  We should make sure to clarify time zone - I like the idea of a "Trace Local" option, but if that is too complicated, I am OK with UTC as the only option.  I just want to make sure that we make it clear which one the user is looking at.

Also agreed on updating the help and description.

iaalm:
@brianrob Please help review again.

src/PerfView/EventViewer/EventWindow.xaml
```diff
@@ -79,6 +79,7 @@
             <Menu DockPanel.Dock="Top" Background="{DynamicResource ControlDefaultBackground}">
                 <MenuItem Header="_File">
                     <MenuItem Header="_Parent Window" Click="DoOpenParent"/>
+                    <MenuItem Header="_Show Local Time" IsCheckable="True" IsChecked="False" Checked="DoUseLocalTime" Unchecked="DoUseOriginTime"/>
```
pharring, line 82: Suggestion: Create a new "View" or "Options" menu for this. It's not a "File" thing.

pharring (approved):

iaalm:
@brianrob  for review😀

iaalm:
@brianrob gentle ping..

brianrob (approved):
LGTM.  Thanks @iaalm!

PR #2027, opened 2024-04-25
Expose times array

cshung:
In .NET 6, @Maoni0 introduced the `GCGlobalHeapHistory_V4` in https://github.com/dotnet/runtime/pull/55888 that expose some timing information for each phase of the GC and I did the parsing of it in https://github.com/microsoft/perfview/pull/1758. It turns out just parsing it is not enough, we need to expose it as property on `TraceGC` in order to be consumed, so this PR expose that array.

Unfortunately, the compact time can often be wrong because a bug I fixed in https://github.com/dotnet/runtime/pull/100599, practically that means for some traces, the compact time will be wrong (often seen as 0)

### DIFF
--- a/src/TraceEvent/Computers/TraceManagedProcess.cs
+++ b/src/TraceEvent/Computers/TraceManagedProcess.cs
@@ -2420,7 +2420,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
             /// <summary>
             /// This field records the time spent for marking roots (except objects pointed by sizedref handle and their descendents)
             ///
-            /// This happen for any type of GC, and it should not be zero for all GCs
+            /// This happen for any type of GC, and it should not be zero for any GCs
             ///
             /// Available in .NET 6+
             /// </summary>
@@ -2429,7 +2429,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
             /// <summary>
             /// This field records the time spent for nulling out short weak references.
             /// GC traced through strong references and any object found dead at this point, if they are targets of a short weak ref, that ref's target will be null-ed.
-            /// This happens for any type of GC, and it should not be zero for all GCs.
+            /// This happens for any type of GC, and it should not be zero for any GCs.
             ///
             /// Available in .NET 6+
             /// </summary>
@@ -2438,7 +2438,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
             /// <summary>
             /// This field records the time spent for promoting finalizable objects that were found dead (so their finalizers can run). 
             ///
-            /// This happens for any type of GC, and it should not be zero for all GCs.
+            /// This happens for any type of GC, and it should not be zero for any GCs.
             ///
             /// Available in .NET 6+
             /// </summary>
@@ -2451,7 +2451,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
             ///
             /// This include the work for scanning the sync blocks, so this value could indicate sync block usage issues.
             ///
-            /// This happens for any type of GC, and it should not be zero for all GCs.
+            /// This happens for any type of GC, and it should not be zero for any GCs.
             ///
             /// Available in .NET 6+
             /// </summary>
@@ -2461,7 +2461,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
             /// This field records the time spent for planning.
             /// Planning refers to the activity of calculating new addresses for objects if this was going to be a compacting GC.
             ///
-            /// This happen only for blocking GCs.
+            /// This happen only for blocking GCs, and it should not be zero for any blocking GCs.
             ///
             /// Available in .NET 6+
             /// </summary>
@@ -2471,7 +2471,7 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
             /// This fields records the time spent for relocating.
             /// Relocation refers to the activity of changing all the pointers to objects to their new location.
             ///
-            /// This happen only for blocking compacting GCs.
+            /// This happen only for blocking compacting GCs, and it should not be zero for any blocking compacting GCs.
             ///
             /// Available in .NET 6+
             /// </summary>
@@ -2481,7 +2481,9 @@ namespace Microsoft.Diagnostics.Tracing.Analysis.GC
             /// This fields records the time spent for compacting.
             /// Compacting refers to the activity of copying the objects from their old location to the new location.
             ///
-            /// This happen only for blocking compacting GCs.
+            /// This happen only for blocking compacting GCs, and it should not be zero for any blocking compacting GCs.
+            ///
+            /// Note: Due to a bug in the runtime, its value can be 0 for .NET 6 to .NET 8
             ///
             /// Available in .NET 6+
             /// </summary>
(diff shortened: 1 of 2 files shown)


src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -2414,6 +2414,101 @@ public GCCondemnedReasons[] PerHeapCondemnedReasons
...
+        public enum TimingType
+        {
+            /// <summary>
+            /// This field records the time spent for marking a type of GCHandle named HNDTYPE_SIZEDREF.
+            /// This happens only for blocking GC, and for majority of the time, this does not happen (because there are no sizedref handles)
+            /// so you get a 0 for blocking GC and a Null for background GC.
+            ///
+            /// Available for .NET 6+
+            /// </summary>
+            MarkSizedRef,
```
Maoni0, line 2427: since this is not in use, we could just leave it out. it's fine to have the code in GC but in tooling it's probably better to leave it out since you'll never get a non zero value.

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -2414,6 +2414,101 @@ public GCCondemnedReasons[] PerHeapCondemnedReasons
...
+            /// This happen for any type of GC, and it should not be zero for all GCs
+            ///
+            /// Available for .NET 6+
+            /// </summary>
+            MarkRoot,
+
+            /// <summary>
+            /// This field records the time spent for nulling out short weak references.
+            /// Short weak references should be null out if the only reason an object is alive is because of finalization (or dependent handle?)
+            /// This happens for any type of GC, and it should not be zero for all GCs.
```
Maoni0, line 2440: not dependent handles - those are strong handles. 

I would explain short and strong weak references together. 

GC will trace through strong references and any object found dead at this point, if they are targets of a short weak ref, that ref's target will be null-ed.

GC then promotes any finalizable objects that are dead at this point, making them live again. This is what MarkScanFinalization reports.

GC looks at the long weak references' targets and null out any whose targets are still dead at this point (ie, were not promoted due to finalization).

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -2414,6 +2414,101 @@ public GCCondemnedReasons[] PerHeapCondemnedReasons
...
+            ///
+            /// Available for .NET 6+
+            /// </summary>
+            MarkShortWeak,
+
+            /// <summary>
+            /// This field records the time spent for scanning the finalization queue and remarking the secondaries for dependent handles.
```
Maoni0, line 2447:
```suggestion
            /// This field records the time spent for promoting finalizable objects that were found dead (so their finalizers can run). if any promotion happens because these objects are primary targets of dependent handles, this time also includes rescanning those handles.
```

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -2414,6 +2414,101 @@ public GCCondemnedReasons[] PerHeapCondemnedReasons
...
+            /// This include the work for scanning the sync blocks, so this value could indicate sync block usage issues.
+            /// This happens for any type of GC, and it should not be zero for all GCs.
+            ///
+            /// Available for .NET 6+
+            /// </summary>
+            MarkLongWeak,
+
+            /// <summary>
+            /// This field records the time spent for planning.
+            /// Planning refers to the activity of determining where the contigous block of life memory (aka plugs) are and where they should be
```
Maoni0, line 2466:
```suggestion
            /// Planning refers calculating new addresses for objects if this was going to be a compacting GC
```

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -2414,6 +2414,101 @@ public GCCondemnedReasons[] PerHeapCondemnedReasons
...
+            ///
+            /// Available for .NET 6+
+            /// </summary>
+            Plan,
+
+            /// <summary>
+            /// This fields records the time spent for relocating.
+            /// Relocation refers to the activity of changing all the pointers to objects to their new location.
+            ///
+            /// This happen only for blocking compacting GC.
```
Maoni0, line 2477:
```suggestion
            /// This happen only for blocking compacting GCs.
```

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -2414,6 +2414,101 @@ public GCCondemnedReasons[] PerHeapCondemnedReasons
...
+            ///
+            /// Available for .NET 6+
+            /// </summary>
+            Relocate,
+
+            /// <summary>
+            /// This fields records the time spent for compacting.
+            /// Compacting refers to the activity of copying the ojbects from their old location to new location.
+            ///
+            /// This happen only for blocking compacting GC.
```
Maoni0, line 2487:
```suggestion
            /// This happen only for blocking compacting GCs.
```

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -2414,6 +2414,101 @@ public GCCondemnedReasons[] PerHeapCondemnedReasons
...
+            ///
+            /// Available for .NET 6+
+            /// </summary>
+            Relocate,
+
+            /// <summary>
+            /// This fields records the time spent for compacting.
+            /// Compacting refers to the activity of copying the ojbects from their old location to new location.
```
Maoni0, line 2485:
```suggestion
            /// Compacting refers to the activity of copying the objects from their old location to the new location.
```

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -2414,6 +2414,101 @@ public GCCondemnedReasons[] PerHeapCondemnedReasons
...
+            ///
+            /// This happen only for blocking compacting GC.
+            ///
+            /// Available for .NET 6+
+            /// </summary>
+            Compact,
+
+            /// <summary>
+            /// This field records the time spent for sweeping.
+            /// Sweeping refers to the activity of inserting the free memory blocks into the size seggregated free
```
Maoni0, line 2495:
```suggestion
            /// Sweeping refers to the activity of making the dead spaces inbetween live objects into a free list
```

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -2414,6 +2414,101 @@ public GCCondemnedReasons[] PerHeapCondemnedReasons
...
+            ///
+            /// Available for .NET 6+
+            /// </summary>
+            Sweep,
+        };
+
+        /// <summary>
+        /// In .NET 6+, we have a poor man's profiler that capture the timing information
+        /// for various phases in GC. This provides a lightweight way to know how much time
+        /// is spent on each phase of the GC.
```
Maoni0, line 2508:
```suggestion
        /// A lightweight way to know how much time is spent on each phase of the GC.
        /// Available in .NET 6+, 
```

Maoni0: I would also change all the "Available for .NET 6+" to "Available in .NET 6+" from above.

Maoni0 (approved):
Thanks!

cshung:
@brianrob, this PR is ready.

cincuranet (approved):

PR #2028, opened 2024-04-28
Add clickable ActivityIDs to requests in ASP.NET Core Stats view

mahamr:
Added ability to click ActivityIds for requests in all request tables that opens the Events window with that ActivityId as a text filter (along with start/end times set). This is similar to what IIS Stats does.

From a zipped PerfView trace:
image

and clicking one of the request IDs:
image

Behavior is identical for a nettrace file as well.


brianrob (approved):
LGTM.  Thanks @mahamr!

PR #2039, opened 2024-05-27
Add Support for Symweb Authentication to PerfView

brianrob:
- Add support for authentication to Symweb.
- Add missing dependencies that previously broke existing ADO authentication and are also needed for Symweb.
- Increase timeout for downloads.
- Tell `HttpClient.GetAsync` to return immediately after the response headers have been downloaded.  This allows `SymbolReader` to show download progress and not timeout for larger symbol files.

### DIFF
--- a/src/PerfView/Authentication.cs
+++ b/src/PerfView/Authentication.cs
@@ -214,6 +214,9 @@ namespace PerfView
 
             handler.ClearHandlers();
 
+            // Always add Symweb authentication.
+            handler.AddSymwebAuthentication(log);
+
             // The order isn't critical, but we chose to put GCM last
             // because the user might want to use GCM for GitHub and
             // Developer identity for Azure DevOps. If GCM were first,
--- a/src/PerfView/PerfView.csproj
+++ b/src/PerfView/PerfView.csproj
@@ -37,7 +37,6 @@
     <NeutralLanguage>en</NeutralLanguage>
 
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-
     <NuspecFile>PerfView.nuspec</NuspecFile>
     <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);SetNuspecProperties</GenerateNuspecDependsOn>
   </PropertyGroup>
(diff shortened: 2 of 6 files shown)


src/PerfView/PerfView.csproj
```diff
@@ -510,6 +511,13 @@
       <Link>Microsoft.Identity.Client.Extensions.Msal.dll</Link>
       <Visible>False</Visible>
     </EmbeddedResource>
+    <EmbeddedResource Include="$(PkgMicrosoft_IdentityModel_Abstractions)\lib\net461\Microsoft.IdentityModel.Abstractions.dll">
```
pharring, line 514: @brianrob Not on this PR specifically, but the embedding of dependencies can be automated so that you don't have to do the "GeneratePathProperty" thing and we never have missing embedded dependencies. My (increasingly fallible) memory tells me I've showed you how to do this before and you weren't interested. Are you interested now?

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -1072,7 +1072,7 @@ internal bool GetPhysicalFileFromServer(string serverPath, string pdbIndexPath,
                         {
                             m_log.WriteLine("FindSymbolFilePath: In task, sending HTTP request {0}", fullUri);
 
-                            var responseTask = HttpClient.GetAsync(fullUri);
+                            var responseTask = HttpClient.GetAsync(fullUri, HttpCompletionOption.ResponseHeadersRead);
```
pharring, line 1075: I'm not against it, but I'm curious what problem ResponseHeaerRead solves.

src/PerfView/PerfView.csproj
brianrob: If I recall correctly, the issue I ran into with this last time was that the set of binaries was much larger than we actually need for PerfView (though probably not if we're being exhaustive with regard to TraceEvent).  Can you remind me how this works?

src/TraceEvent/Symbols/SymbolReader.cs
brianrob: This ensures that the call to `HttpClient.GetAsync` returns quickly and allows the code in `SymbolReader` to show status for the download.  If we don't do this, then symbol servers that download the entire blob (e.g. new symweb, msdl, azdo, etc.) will not show any download status in the logs until the call to `HttpClient.GetAsync` completes.  For large files, this could cause a timeout.  This is especially true for `msedge` PDBs which are greater than 1GB in size.

The reason that you may not have noticed this (as I didn't) is that the existing symweb just downloads a redirect to a UNC path, and so the download from the symbol server is tiny and super quick.  Then we go copy the file from the UNC path.  When this happens, it's a 2 step operation rather than for a blob-based symbol server, where the file comes in the response of the original request.

src/PerfView/Authentication.cs
```diff
@@ -214,6 +214,9 @@ public static void Configure(this SymbolReaderHttpHandler handler, Authenticatio
 
             handler.ClearHandlers();
 
+            // Always add Symweb authentication.
+            handler.AddSymwebAuthentication(log);
```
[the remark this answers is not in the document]
brianrob: PerfView doesn't do anything to prevent, this, but users would have to login.  This will come in the form of an authentication prompt.  If the user doesn't specify the new symweb in their symbol path, then they won't ever hit this.  Additionally, we don't specify the new symweb in the default symbol path, so users will need to specify it.  I intend to send mail to the internal users group alias on how to specify the new symweb in the symbol path once this is released.  Does that address your concern?

cincuranet (dismissed):

src/PerfView/PerfView.csproj
pharring: It was some extra MSBuild logic that we used in the Snapshot Collector for the same purpose.
Yes, it does result in a larger set. It's actually the complete, transitive set of dependencies -- according to the build. As you say, that's probably larger than strictly needed at runtime. I reason it is better to carry the complete set than to hope you can discover missing references by manual testing.

pharring: I looked for the commit that had it in, but I can't find it. I must've deleted the old branch where I did the work. I can try and recreate it in a different PR.

src/TraceEvent/Symbols/SymbolReader.cs
pharring: Thanks for the explanation. Makes sense to me.

src/PerfView/Authentication.cs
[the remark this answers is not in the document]
pharring: I see. Thanks.

pharring (approved):

cincuranet (approved):

PR #2046, opened 2024-06-11
Add Symbols Authentication Support for TraceEvent

brianrob:
- Creates a new symbols authentication NuGet package to be used alongside TraceEvent.  This ensures that non-Microsoft internal users are not subjected to all the new authentication-related dependencies.
 - Moves symweb authentication into the new NuGet package.  Additional auth mechanisms can be moved to be available to TraceEvent, but this is beyond the scope of this PR.
 - Creates a samples application to show users how to setup authentication when using SymbolReader.


src/PerfView/Authentication.cs
```diff
@@ -183,19 +186,19 @@ private void UpdateState(AuthProviderFlags newState, string propertyName)
     }
 
     /// <summary>
-    /// Extension methods for <see cref="SymbolReaderHttpHandler"/>.
+    /// Extension methods for <see cref="SymbolReaderAuthenticationHandler"/>.
     /// </summary>
     internal static class SymbolReaderHttpHandlerExtensions
```
pharring, line 191: Rename to `SymbolReaderAuthenticationHandlerExtensions`

src/SymbolsAuth/SymwebHandler.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            authority = builder.Uri;
+            return true;
+        }
+    }
+
+    /// <summary>
+    /// A handler that adds authorization for Symweb.
+    /// </summary>
+    public sealed class SymwebHandler : SymbolReaderAuthHandler
```
pharring, line 71: There's a lot of similarity between this and the AzureDevOpsHandler. I wonder if there could be a common base class.

pharring (approved):
Looks great. Couple of minor comments, but not showstoppers.

src/PerfView/Authentication.cs
brianrob: Good catch.

src/SymbolsAuth/SymwebHandler.cs
brianrob: You're right.  I'm not sure.  Let me take this as a follow-up to this PR.

cincuranet (approved):

PR #2051, opened 2024-06-14
Truly Dynamic Event

cshung:
This PR is ready to be reviewed. (It is easier to review by commits)

The idea of truly dynamic event is to make `GCDynamicEvent` truly dynamic, in the sense that the GC team can modify the runtime to emit any dynamic event, and then we can consume them in a reasonable API without modifying `TraceEvent` at all.

The primary use case for this is through a .NET Interactive notebook, where: 

1. The notebook user specifies a list of `DynamicEventSchema` that specify the names of the events the schemas handle as well as the field and types, and then
2. Be able to access them through the `TraceGC.DynamicEvent()` extension method as if they were parsed, using the C# dynamic keyword magic.

Beside the obvious advantage to reduce work, it also allow a more agile practice. Previously, whenever we release a new event, we have to stay backward compatible with respect to things like field name and types and how values are interpreted, etc. With the design right now, we can selectively only expose event in the traditional way if they were meant to be stable and useful to others, and expose it the dynamic way if it is otherwise. Customers are warned that these event schemas are subjected to change without notice, as a matter of fact we won't even release them. They aren't meant to be private (you can easily skim them from the source), they merely meant without back compat guarantees.

This PR is meant to be used in https://github.com/dotnet/performance/pull/4274

### DIFF
--- a/src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
@@ -64,18 +64,16 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
                     callback(template);
         }
 
-        /// <summary>
-        /// Do not use.  This is here to avoid asserts that detect undeclared event templates.
-        /// </summary>
-        public event Action<GCDynamicTraceEvent> GCDynamicTraceEvent
+        private event Action<RawDynamicTraceEvent> _gcDynamicTraceEvent;
+        public event Action<RawDynamicTraceEvent> GCDynamicTraceEvent
         {
             add
             {
-                throw new NotSupportedException();
+                _gcDynamicTraceEvent += value;
             }
             remove
             {
-                throw new NotSupportedException();
+                _gcDynamicTraceEvent -= value;
             }
         }
 
@@ -141,6 +139,12 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
             {
                 _gcHeapCountSample(data.EventPayload as HeapCountSampleTraceEvent);
             }
+
+            else if (_gcDynamicTraceEvent != null &&
+                data.EventPayload is RawDynamicTraceEvent)
+            {
+                _gcDynamicTraceEvent(data.EventPayload as RawDynamicTraceEvent);
+            }
         }
 
         private static GCDynamicTraceEvent GCDynamicTemplate(Action<GCDynamicTraceEvent> action, GCDynamic.GCDynamicEvent eventTemplate)
@@ -199,7 +203,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.GCDynamic
         private readonly HeapCountTuningTraceEvent _heapCountTuningTemplate = new HeapCountTuningTraceEvent();
         private readonly CommittedUsageTraceEvent _committedUsageTemplate = new CommittedUsageTraceEvent();
         private readonly HeapCountSampleTraceEvent _heapCountSampleTemplate = new HeapCountSampleTraceEvent();
-        private readonly RawDynamicTraceData _rawTemplate = new RawDynamicTraceData();
+        private readonly RawDynamicTraceEvent _rawTemplate = new RawDynamicTraceEvent();
 
         /// <summary>
         /// Contains the fully parsed payload of the dynamic event.
@@ -295,7 +299,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.GCDynamic
         /// <summary>
         /// The list of specific event templates.
         /// </summary>
-        internal static readonly RawDynamicTraceData RawDynamicTemplate = new RawDynamicTraceData();
+        internal static readonly RawDynamicTraceEvent RawDynamicTemplate = new RawDynamicTraceEvent();
         internal static readonly HeapCountTuningTraceEvent HeapCountTuningTemplate = new HeapCountTuningTraceEvent();
         internal static readonly CommittedUsageTraceEvent CommittedUsageTemplate = new CommittedUsageTraceEvent();
         internal static readonly HeapCountSampleTraceEvent HeapCountSampleTemplate = new HeapCountSampleTraceEvent();
@@ -340,7 +344,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.GCDynamic
         }
     }
 
-    internal sealed class RawDynamicTraceData : GCDynamicEvent
+    public sealed class RawDynamicTraceEvent : GCDynamicEvent
     {
         internal override TraceEventID ID => (TraceEventID)39;
         internal override string TaskName => "GC";
@@ -624,5 +628,19 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.GCDynamic
         public double ElapsedTimeBetweenGCsMSec { get; internal set; }
         public double GCPauseTimeMSec { get; internal set; }
         public double MslWaitTimeMSec { get; internal set; }
+    }
+
+    public sealed class DynamicEvent
+    {
+        public DynamicEvent(string name, DateTime timeStamp, byte[] payload)
+        {
+            Name = name;
+            TimeStamp = timeStamp;
+            Payload = payload;
+        }
+
+        public string Name { get; internal set; }
+        public DateTime TimeStamp { get; internal set; }
+        public byte[] Payload { get; internal set; }
     }
 }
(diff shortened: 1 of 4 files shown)


src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
```diff
@@ -64,18 +66,16 @@ protected internal override void EnumerateTemplates(Func<string, string, EventFi
                     callback(template);
         }
 
-        /// <summary>
-        /// Do not use.  This is here to avoid asserts that detect undeclared event templates.
-        /// </summary>
-        public event Action<GCDynamicTraceEvent> GCDynamicTraceEvent
+        private event Action<RawDynamicTraceEvent> _gcDynamicTraceEvent;
+        public event Action<RawDynamicTraceEvent> GCDynamicTraceEvent
```
Maoni0, line 70: this is a public class, is it exposed to users of TraceEvent at all?

Maoni0:
the dynamic events should be in a List, not a Dictionary. all you need to communicate to the notebooks side is for this GC, here are timestamps for the dynamic event it observed and here's the info for these events.

"info" could even mean just a blob of data that's fired in this event and the notebooks side can parse everything including name/version/payload. do you see a problem with this approach? is there a need for TraceEvent to even parse the name/version?

I would also suggest to try out using this in a notebook just to make sure it actually works, if you haven't.

src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
cshung: Is there a particular concern here about the exposure? Are you worried about the signature change breaking existing apps, or something else?

cshung:
> the dynamic events should be in a List, not a Dictionary. all you need to communicate to the notebooks side is for this GC, here are timestamps for the dynamic event it observed and here's the info for these events.
> 
> "info" could even mean just a blob of data that's fired in this event and the notebooks side can parse everything including name/version/payload. do you see a problem with this approach? is there a need for TraceEvent to even parse the name/version?
> 
> I would also suggest to try out using this in a notebook just to make sure it actually works, if you haven't.

Not at all, the parsing is really all meant for usability so that you can write:

```c#
gc.DynamicEvents.SizeAdaptationTuning.new_n_heaps
```

instead of 

```c#
BitConverter.ToUInt16(gc.dynamicEventIndex.Single(r => r.Name.Equals("SizeAdaptationTuning")).Payload, 4)
```

I still think that the parsing is valuable, but `TraceEvent` might not be the right home for it. I was thinking about moving them to  `GC.Infrastructure` instead, which allows for quick iteration. For that I would talk to @MokoSan first.

The parsing of the names are currently needed because we want to filter out the known events (e.g. `CommittedUsage`) regardless, but `TraceEvent` don't need to index them by name, we can totally just do a plain list of `RawDynamic` objects (a good name is wanted for that)

src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
Maoni0: yes, I'm asking if changing a public class would break any existing usage.

Maoni0:
> I still think that the parsing is valuable, but TraceEvent might not be the right home for it. 

parsing is of course valuable and the question is exactly where it should be. I was suggesting we didn't need to do any parsing in TraceEvent but you pointed out we needed to parse the event name on the TraceEvent side to filter out supported dynamic events so it sounds like you are saying only the name parsing is needed on the TraceEvent side and everything else should be done on the GC analysis side (we can decide if that's in the c# library or the notebooks).

brianrob:
Just looking through this PR, and I want to make sure that I understand the goal correctly.  Is the goal here to keep the existing "known" dynamic events, but otherwise allow access to the unknown dynamic events without having TraceEvent understand their metadata (and you just parse them in your own code)?

Maoni0:
> Just looking through this PR, and I want to make sure that I understand the goal correctly. Is the goal here to keep the existing "known" dynamic events, but otherwise allow access to the unknown dynamic events without having TraceEvent understand their metadata (and you just parse them in your own code)?

@brianrob it's to keep the _supported_ dynamic event (right now we have the commit usage event which we do support, meaning tools can use it and depend on it - if we change it we'll bump up the version; and _we might add more later_) and allow access to the unsupported ones that may change at any time which we only consume for the GC internal analysis.

cshung:
Just for context, this work is inspired because we wanted additional diagnostics information for DATAS as in this PR.

https://github.com/dotnet/runtime/pull/103405

With that change, the runtime will never emit `HeapCountTuning` and `HeapCountSample` events anymore.

For that, I would lilke to depreciate `HeapCountTuning` and `HeapCountSample` in favor of this new approach. Ideally I will just delete them. While theoretically that will be a breaking change, I really doubt if anyone actually depends on them in practice, considering that:

1. These events are emitted only when DATAS is turned on in .NET 8 or 9 early previews, 
2. These events emits rather cryptic values that a general person who don't work in the GC probably doesn't know what they are, and
3. As a general person, they can probably do nothing about them. It is not like they can tune the heuristic implementation or modify any configuration for that.

brianrob:
> Just for context, this work is inspired because we wanted additional diagnostics information for DATAS as in this PR.
> 
> [dotnet/runtime#103405](https://github.com/dotnet/runtime/pull/103405)
> 
> With that change, the runtime will never emit `HeapCountTuning` and `HeapCountSample` events anymore.
> 
> For that, I would lilke to depreciate `HeapCountTuning` and `HeapCountSample` in favor of this new approach. Ideally I will just delete them. While theoretically that will be a breaking change, I really doubt if anyone actually depends on them in practice, considering that:
> 
> 1. These events are emitted only when DATAS is turned on in .NET 8 or 9 early previews,
> 2. These events emits rather cryptic values that a general person who don't work in the GC probably doesn't know what they are, and
> 3. As a general person, they can probably do nothing about them. It is not like they can tune the heuristic implementation or modify any configuration for that.

I don't have a problem with you removing them entirely if they are no longer relevant.  I agree that I am not super concerned with such a break.

brianrob:
> rted ones that may change at any time which we only consume for the GC i

Gotcha, thanks.

brianrob:
If you do end up removing the existing supported dynamic events, and don't add more, I am wondering if any change to TraceEvent other than the removal is required?  If you do all the parsing outside of TraceEvent, then I suspect you just need access to the raw event payload itself.

src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
cshung: Earlier I missed the fact that I was changing the generic type argument.

I don't think any "useful" usage could be broken with this change. We can certainly write a contrived scenario such that this change broke something.

cshung:
> If you do all the parsing outside of TraceEvent, then I suspect you just need access to the raw event payload itself.

Yes, the new iteration I pushed today do just that. For the unparsed events, fire them through the `GCDynamicTraceEvent` and so TraceGC can expose them, that's all.

cshung:
@brianrob, this change is ready to be reviewed and merged. Can you please take a look?

src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
```diff
@@ -556,73 +404,17 @@ public sealed class CommittedUsage
...
-                {
-                    _payloadNames = new string[] { "Version", "GCIndex", "ElapsedTimeBetweenGCs", "GCPauseTime", "MslWaitTime" };
-                }
-
-                return _payloadNames;
-            }
-        }
-
-        internal override object PayloadValue(int index)
+        public DynamicEvent(string name, DateTime timeStamp, byte[] payload)
```
brianrob, line 409: Can you share a bit of detail about the differences between the RawDynamicTraceEvent class and the DynamicEvent class?

src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
```diff
@@ -340,7 +278,7 @@ internal GCDynamicEvent Bind(GCDynamicTraceEvent underlyingEvent)
         }
     }
 
-    internal sealed class RawDynamicTraceData : GCDynamicEvent
+    public sealed class RawDynamicTraceEvent : GCDynamicEvent
```
brianrob, line 281: If you are going to make this public, and we name it GCDynamicTraceEvent to match the event name, and scope it appropriately?

src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
brianrob, line 409: Can we rename this to GCDynamicEvent to scope it appropriately?  It would also be good to add some comments around why this object exists vs. just using RawDynamicTraceEvent.  I assume this is to avoid the Clone and finalization/disposal cost.

brianrob (approved):

PR #1674, opened 2022-07-31
Fix a few bugs related to object generation detection

cshung:
@Maoni0 reported that generational aware analysis in PerfView is not working for full core dumps. After debugging for a while, I have found a few bugs related to generation detection. This PR fixes some of them, the remaining ones are in ClrMD and are fixed [there](https://github.com/microsoft/clrmd/pull/1035).

**Note** - I changed the serialization format by adding information about POH, which I need help with backward compatibility. The CI failures are related to this.

@tommcdon - This might need to be replicated for `dotnet-gcdump`

PR #2052, opened 2024-06-20
Update to ClrMD version 4.0.0-beta.24314.3

brianrob:
- Enable ClrMD to authenticate to a symbol server if required.
 - Update ClrMD API calls.

### DIFF
--- a/src/HeapDump/GCHeapDumper.cs
+++ b/src/HeapDump/GCHeapDumper.cs
@@ -20,6 +20,9 @@ using System.Threading;
 using Address = System.UInt64;
 using Microsoft.Diagnostics.Utilities;
 using Microsoft.Diagnostics.HeapDump;
+using Azure.Core;
+
+
 #if CROSS_GENERATION_LIVENESS
 using Microsoft.Diagnostics.CrossGenerationLiveness;
 #endif
@@ -35,9 +38,10 @@ public class GCHeapDumper
     /// to dump a heap.  
     /// </summary>
     /// <param name="log"></param>
-    public GCHeapDumper(TextWriter log)
+    public GCHeapDumper(TextWriter log, TokenCredential symbolServerAuthCredential = null)
     {
         m_origLog = log;
+        m_symbolServerAuthCredential = symbolServerAuthCredential;
         m_copyOfLog = new StringWriter();
         m_log = new TeeTextWriter(m_copyOfLog, m_origLog);
 
@@ -274,11 +278,11 @@ public class GCHeapDumper
         {
             try
             {
-                dataTarget = DataTarget.CreateSnapshotAndAttach(processID);
+                dataTarget = DataTarget.CreateSnapshotAndAttach(processID, m_symbolServerAuthCredential);
             }
             catch
             {
-                dataTarget = DataTarget.AttachToProcess(processID, Freeze);
+                dataTarget = DataTarget.AttachToProcess(processID, Freeze, m_symbolServerAuthCredential);
             }
         }
         else
@@ -288,7 +292,7 @@ public class GCHeapDumper
                 UseOSMemoryFeatures = false // disable AWE
             };
 
-            dataTarget = DataTarget.LoadDump(processDumpFile, cacheOptions);
+            dataTarget = DataTarget.LoadDump(processDumpFile, cacheOptions, m_symbolServerAuthCredential);
         }
 
         if (dataTarget.DataReader.PointerSize != IntPtr.Size)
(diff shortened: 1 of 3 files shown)


src/HeapDump/GCHeapDumper.cs
```diff
@@ -956,7 +960,7 @@ private void DumpDotNetHeapDataWorker(DataTarget dataTarget, ClrRuntime[] runtim
             gcHeapDumpSegments.Add(gcHeapDumpSegment);
 
             total += seg.Length;
-            m_log.WriteLine("Segment: Start {0,16:x} Length: {1,16:x} {2,11:n3}M LOH:{3}", seg.Start, seg.Length, seg.Length / 1000000.0, seg.IsLargeObjectSegment);
+            m_log.WriteLine("Segment: Start {0,16:x} Length: {1,16:x} {2,11:n3}M LOH:{3}", seg.Start, seg.Length, seg.Length / 1000000.0, seg.Kind == GCSegmentKind.Large);
```
leculver, line 963: Does anything rely on this log's format?  If not, I would suggest just logging the segment kind:

```suggestion
            m_log.WriteLine("Segment: Start {0,16:x} Length: {1,16:x} {2,11:n3}M Kind:{3}", seg.Start, seg.Length, seg.Length / 1000000.0, seg.Kind);
```

src/HeapDump/GCHeapDumper.cs
```diff
@@ -928,15 +932,15 @@ private void DumpDotNetHeapDataWorker(DataTarget dataTarget, ClrRuntime[] runtim
             // ClrMD returns UOH segments with seg.Generation2.Start == seg.Start and seg.Generation2.End == seg.End
             // To make it easy to determine the real generation of the object, augmenting the object with Gen3End and
             // Gen4End as follows such that DotNetHeapInfo.GenerationFor works.
-            if (seg.IsPinnedObjectSegment)
+            if (seg.Kind == GCSegmentKind.Pinned)
```
leculver, line 935: While not something you should tackle here, this set of `if` statements are questionable with the new GC segments feature.

I guess this all works correctly, but this takes a real dependency on the ordering of segment chunks.  Things don't really work that way when GC segments are enabled.

leculver (approved):
One suggestion, and a note of something to investigate later.  The TokenCredential changes look good to me.

src/HeapDump/GCHeapDumper.cs
brianrob: Nope, this is just for logging.  Thanks.

src/HeapDump/GCHeapDumper.cs
brianrob: @cshung, was this part of the area that you looked at when updating TraceEvent to support regions?

cshung: Yes, I did update ClrMD and HeapDump to support regions.
The PRs are https://github.com/microsoft/clrmd/pull/1035 and https://github.com/microsoft/perfview/pull/1674 respectively.
I am not sure I understand what might have gone wrong here, @leculver can you elaborate?

leculver: There's not anything in particular _wrong_ here.  I'm only flagging to check whether this particular code was reviewed and looks correct for GC_REGIONS.  Since you reviewed this and submitted #1674, I guess we already covered perfview.  This code just looks _odd_ to me in terms of how it handles segments/regions at all (the original code, not your updates).

In short, if it's already been reviewed from a GC_REGIONS perspective and you are good with it, then nothing to do.  I'm just double checking. :)

cshung: Thanks for confirming. Here is how I see it.

In the old days, a segment could have multiple generations in it, so it makes sense to start with searching a containing region, and then check if it lands before gen2 ends, if it does, it is gen2, otherwise, check if it lands before gen 1, and so on. That's a pretty simple linear search.

Then it comes the large object heap, we could have check a flag and declare it is gen 3, but instead, PerfView chose to add a fake Gen3 end to the all segments, so that there is a fake gen 3 portion in the ephemeral segments as well, of length 0, starts and ends at the very beginning of the ephemeral segment (and also symetrically, add the fake gen 2/gen 1/gen 0 portions) that start and ends at the end of the large object segment, that will fit the linear search algorithm.

```txt
Under segments, an ephemeral segment looks like this:

|222222|11111|000|
|      |     |   +-------> Gen 0 Ends
|      |     +-----------> Gen 0 Start
|      |     +-----------> Gen 1 Ends
|      +-----------------> Gen 1 Start
|      +-----------------> Gen 2 Ends
+------------------------> Gen 2 Start
+------------------------> Gen 3 End
+------------------------> Gen 3 Start

Under regions, an ephemeral segment looks like this:

|1111111111111111|
|                +------> Gen 0 Ends
|                +------> Gen 0 Start
|                +------> Gen 1 Ends
+-----------------------> Gen 1 Start
+-----------------------> Gen 2 End
+-----------------------> Gen 2 Start
+-----------------------> Gen 3 End
+-----------------------> Gen 3 Start
```

My change to POH and regions simply follow suit. For regions, there is only one generation per region, so in fact all other generations in the same region is fake. 

I agree It is _odd_, the whole thing is bent backward to fit the original linear search algorithm, but I think it should work. In particular, it might look like there is some dependency on the ordering because of the linear search, but in fact, it doesn't, all that linear search does is going through the fakes. There will be exactly one region containing the object, and for each region, there will only be exactly one generation with a non zero length.

leculver: Thanks for taking a look!

brianrob:
Thanks everyone for reviewing!

PR #2066, opened 2024-07-15
Support Concurrent Access to LinuxPerfScriptProcessNameBuilder

brianrob:
Required to support ParallelLinuxPerfScriptStackSource.

Address unlocked access to `Dictionary` objects in `LinuxPerfScriptProcessNameBuilder` that can occur when called from `ParallelLinuxPerfScriptStackSource`, which splits the processing work across multiple threads.

### DIFF
--- a/src/TraceEvent/Stacks/Linux/LinuxPerfScriptProcessNameBuilder.cs
+++ b/src/TraceEvent/Stacks/Linux/LinuxPerfScriptProcessNameBuilder.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Text;
+using System.Threading;
 
 namespace Microsoft.Diagnostics.Tracing.StackSources
 {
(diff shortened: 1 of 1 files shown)


src/TraceEvent/Stacks/Linux/LinuxPerfScriptProcessNameBuilder.cs
```diff
@@ -16,36 +17,45 @@ internal sealed class LinuxPerfScriptProcessNameBuilder
                 ".NET Server GC"
             };
 
+        private readonly object _lock = new object();
```
cincuranet, line 20: Maybe more descriptive name instead of `_lock` would be nicer?

brianrob: Sure.  How about `_dictionaryLock`?  I was going for a lock for a of the structures.

cincuranet: `_dictionariesLock`, because it is guarding multiple dictionaries?

brianrob: Yes.  It's basically guarding all of the dictionaries, rather than having multiple locks.

cincuranet (approved):

ivberg (approved):

PR #2068, opened 2024-07-16
fix: arm64/kerneltracecontrol.dll file name

vaind:
Fixes builds on case-sensitive systems. The actual DLL in the package is lowercase.

image

```
/usr/share/dotnet/sdk/8.0.303/Microsoft.Common.CurrentVersion.targets(5270,5): error MSB3030: Could not copy the file "/home/runner/.nuget/packages/microsoft.diagnostics.tracing.traceevent.supportfiles/1.0.28/lib/native/arm64/KernelTraceControl.dll" because it was not found. [/home/runner/work/sentry-dotnet/sentry-dotnet/modules/perfview/src/TraceEvent/TraceEvent.csproj::TargetFramework=netstandard2.0]
```

### DIFF
--- a/src/TraceEvent/TraceEvent.csproj
+++ b/src/TraceEvent/TraceEvent.csproj
@@ -121,7 +121,7 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Visible>False</Visible>
     </None>
-    <None Include="$(TraceEventSupportFilesBase)native\arm64\KernelTraceControl.dll">
+    <None Include="$(TraceEventSupportFilesBase)native\arm64\kerneltracecontrol.dll">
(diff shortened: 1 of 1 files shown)


cincuranet:
@brianrob Do we maybe want to rename the file in the package to have proper casing? Do you see any risk in doing that?

brianrob:
> @brianrob Do we maybe want to rename the file in the package to have proper casing? Do you see any risk in doing that?

I think this would be fine.

cincuranet:
This has been fixed in the package directly. New release (and hence new support files) will contain proper casing. Thanks for bringing that up!

cincuranet:
https://github.com/microsoft/perfview/pull/2115

PR #2112, opened 2024-09-13
Update building docs + add a .vsconfig file

caaavik-msft:
With the recent change to ensure that the C++ libraries are built with spectre mitigations enabled, it requires you to install an additional component to the VS installation to build the project which wasn't mentioned in the building instructions. This PR updates the building instructions to mention this, and also makes it a bit easier to read. I also used this as an opportunity to add a .vsconfig file which will get VS to prompt you if it notices any required components are missing:

image

### DIFF
--- a/.vsconfig
+++ b/.vsconfig
@@ -1,0 +1,15 @@
+{
+  "version": "1.0",
+  "components": [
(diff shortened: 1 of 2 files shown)


cincuranet (approved):
Thanks @caaavik-msft! LGTM.

PR #2117, opened 2024-10-09
Add support for EVENT_TRACE_PRIVATE_LOGGER_MODE and EVENT_TRACE_PRIVATE_IN_PROC flags for file mode logging

jaholme:
This changes adds support for two additional EVENT_TRACE flags, EVENT_TRACE_PRIVATE_LOGGER_MODE and EVENT_TRACE_PRIVATE_LOGGER_MODE.

These flags are not compatible with real time logging, but can be used with file mode based logging which is useful for applications which are trying to capture locally the events that their process is generating for later diagnosis.

### DIFF
--- a/src/TraceEvent/TraceEventSession.cs
+++ b/src/TraceEvent/TraceEventSession.cs
@@ -2,8 +2,8 @@
 // This file is best viewed using outline mode (Ctrl-M Ctrl-O)
 //
 // This program uses code hyperlinks available as part of the HyperAddin Visual Studio plug-in.
-// It is available from http://www.codeplex.com/hyperAddin 
-// 
+// It is available from http://www.codeplex.com/hyperAddin
+//
 using Microsoft.Diagnostics.Tracing.Compatibility;
 using Microsoft.Diagnostics.Tracing.Parsers;
 using Microsoft.Diagnostics.Utilities;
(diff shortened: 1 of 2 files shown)


cincuranet:
Link for reviewing without whitespace changes: https://github.com/microsoft/perfview/pull/2117/files?w=1

cincuranet (approved):
LGTM.

src/TraceEvent/TraceEventSession.cs
```diff
@@ -2817,27 +2845,36 @@ public enum TraceEventSessionOptions
         /// for sessions that expect more than 1K events per second.
         /// </summary>
         NoPerProcessorBuffering = 8,
+        /// <summary>
+        /// Creates a user-mode event tracing session that runs in the same process as its event trace provider.
+        /// </summary>
+        PrivateLogger = 16,
+        /// <summary>
+        /// Use in conjunction with the EVENT_TRACE_PRIVATE_LOGGER_MODE mode to start a private session.
```
brianrob, line 2853: Should we switch `EVENT_TRACE_PRIVATE_LOGGER_MODE` to `PrivateLogger` since that's the value we expose here?

brianrob (commented):
Overall, LGTM.  One question below.

src/TraceEvent/TraceEventSession.cs
jaholme: Makes sense, this was the MSDN docs quoted verbatim.

brianrob (approved):

PR #2120, opened 2024-10-15
Address accessibility issues

caaavik-msft:
An accessibility review was completed on PerfView which identified a number of accessibility issues with PerfView, this PR addresses as many of the Sev 1 and Sev 2 accessibility issues that were possible without too much work. There are some other issues which were not addressed in this PR because they would require a significant amount of work to fix or it needed some more investigation.

## Replaced ActiveX with WebView2 for all web browser components
The ActiveX-based WebBrowser control has been replaced with WebView2 controls. A large number of accessibility issues that were identified were in places we were using the WebBrowser as it wasn't generating a very accessible UI. WebView2 was a (mostly) drop-in replacement that was able to address every accessibility issue that was reported. The WebView2 package includes inside it the driver/dll, so it is all self-contained and won't require the user to install anything to get it to work. Some of the subscribed event handlers on the browser control had to be updated to support the different event handler argument types as well.

It is worth mentioning that the latest version of WebView2 does not officially support Windows 7/8 ([source](https://blogs.windows.com/msedgedev/2022/12/09/microsoft-edge-and-webview2-ending-support-for-windows-7-and-windows-8-8-1/)), however it will still likely work it is just that they aren't testing that it works (see https://github.com/MicrosoftEdge/WebView2Feedback/issues/3404). If we want to be extra sure, we can roll back the WebView2 SDK version to `1.0.1519.0` though.

## Added AutomationProperties.Name for many elements
This property is used by accessibility software such as Narrator to describe what the currently highlighted element is. If an element is a text block or contains some content, the narrator will use that content by default as this name, but for other elements such as input elements, we need to specify it manually as it isn't able to extract it automatically from any adjacent labels. 

There are also some cases like TreeView or ListBox where the items inside the control use the ToString() method on the object to get the AutomationProperties.Name, which in some cases was undesirable as it was not user-friendly. For these cases we make use of a Setter to override the AutomationProperties.Name binding to a better value. There were also some other places that needed some specific work done to get the UI name to show up when WPF was unable to determine the correct value from the content or if the content was too long and needed to be shortened.

## Other changes
- There were a handful of places where `KeyboardNavigation.IsTabStop="False"` was set where it should have been true as it contained useful information that a user using accessibility tools may need.
- The border around the selected element when using tab navigation in the dark theme was very hard to see. I have changed it so that instead of being a dashed line it is a solid line.
- I disabled legacy accessiblity features in the app.config that were causing some false-positive accessibility issues to be reported. You can read more about this [here](https://learn.microsoft.com/en-us/dotnet/framework/whats-new/whats-new-in-accessibility#accessibility-switches).

### DIFF
--- a/src/PerfView/GuiUtilities/WebBrowser/WebBrowser.xaml.cs
+++ b/src/PerfView/GuiUtilities/WebBrowser/WebBrowser.xaml.cs
@@ -1,11 +1,11 @@
 ﻿using System;
 using System.ComponentModel;
-using System.Runtime.InteropServices;
-using System.Security.Policy;
-using System.Threading.Tasks;
+using System.IO;
 using System.Windows;
 using System.Windows.Controls;
+using Microsoft.Web.WebView2.Core;
 using Microsoft.Web.WebView2.Wpf;
+using Utilities;
 
 
 namespace PerfView.GuiUtilities
@@ -19,6 +19,7 @@ namespace PerfView.GuiUtilities
         {
             InitializeComponent();
         }
+
         /// <summary>
         /// If set simply hide the window rather than closing it when the user requests closing. 
         /// </summary>
@@ -28,24 +29,33 @@ namespace PerfView.GuiUtilities
         public bool CanGoBack { get { return Browser.CanGoBack; } }
         public WebView2 Browser { get { return _Browser; } }
 
-        /// <summary>
-        /// LIke Browser.Navigate, but you don't have to be on the GUI thread to use it.  
-        /// </summary>
-        public void Navigate(string uri)
+        public static readonly DependencyProperty SourceProperty = DependencyProperty.Register(
+            nameof(Source),
+            typeof(Uri),
+            typeof(WebBrowser),
+            new PropertyMetadata(OnSourceChanged));
+
+        public Uri Source
         {
-            // Make sure we do it on the GUI thread.
-            Dispatcher.BeginInvoke((Action)delegate
-            {
-                WebBrowserWindow.Navigate(Browser, uri);
-            });
+            get { return (Uri)GetValue(SourceProperty); }
+            set { SetValue(SourceProperty, value); }
+        }
+
+        private static void OnSourceChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
+        {
+            (d as WebBrowserWindow)?.Navigate();
         }
 
         /// <summary>
-        /// A simple helper wrapper that translates some exceptions nicely.  
+        /// If WebView2 has been initialized, navigate to current source. If WebView2 is not initialized yet, it will 
+        /// be navigated to once initialization has completed.
         /// </summary>
-        public static void Navigate(WebView2 browser, string url)
+        private void Navigate()
         {
-            browser.Source = new Uri(url);
+            if (Source != null && _Browser.CoreWebView2 != null)
+            {
+                _Browser.CoreWebView2.Navigate(Source.ToString());
+            }
         }
 
         #region private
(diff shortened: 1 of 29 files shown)


src/PerfView/GuiUtilities/WebBrowser/WebBrowser.xaml.cs
```diff
@@ -23,9 +26,10 @@ public WebBrowserWindow(Window parentWindow) : base(parentWindow)
 
         public bool CanGoForward { get { return Browser.CanGoForward; } }
         public bool CanGoBack { get { return Browser.CanGoBack; } }
-        public WebBrowser Browser { get { return _Browser; } }
+        public WebView2 Browser { get { return _Browser; } }
+
         /// <summary>
-        /// LIke Broswer.Navigate, but you don't have to be on the GUI thread to use it.  
+        /// LIke Browser.Navigate, but you don't have to be on the GUI thread to use it.
```
cincuranet, line 32: I would also fix the spelling of "LIke". ;)

cincuranet (approved):
Thanks @caaavik-msft. That looks great. Given that both Windows 7 and 8 are EOL, I don't see problem with using latest WebView2, but I'll defer to @brianrob for final decision.

src/PerfView/PerfView.csproj
```diff
@@ -89,6 +92,7 @@
     <PackageReference Include="Microsoft.IdentityModel.Abstractions" Version="$(MicrosoftIdentityModelAbstractionsVersion)" GeneratePathProperty="true" />
     <PackageReference Include="Microsoft.IdentityModel.Tokens" Version="$(MicrosoftIdentityModelTokensVersion)" GeneratePathProperty="true" />
     <PackageReference Include="Microsoft.IdentityModel.JsonWebTokens" Version="$(MicrosoftIdentityModelJsonWebTokensVersion)" GeneratePathProperty="true" />
+    <PackageReference Include="Microsoft.Web.WebView2" Version="$(MicrosoftWebWebView2Version)" GeneratePathProperty="true" />
```
brianrob, line 95: The reason that we set `GeneratePathProperty=true` for most of these package references is because we copy files from them into the self-extracting PerfView.exe.  I see that we aren't doing that with WebView2.  Are we guaranteed to have WebView2 on the machine already?

brianrob (commented):
Overall, this looks good to me.  We're already on the plan of supporting Windows 10+, so I'm not concerned about adding the WebView2 dependency.  A couple of questions/comments below.

src/PerfView/app.config
```diff
@@ -6,6 +6,7 @@
   </system.diagnostics>
   <runtime>
     <gcAllowVeryLargeObjects enabled="true"/>
+    <AppContextSwitchOverrides value="Switch.UseLegacyAccessibilityFeatures=false;Switch.UseLegacyAccessibilityFeatures.2=false;Switch.UseLegacyAccessibilityFeatures.3=false;Switch.UseLegacyAccessibilityFeatures.4=false;" />
```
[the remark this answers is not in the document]
caaavik-msft: I have moved these switches to instead be set programatically inside App.cs

src/PerfView/PerfView.csproj
caaavik-msft: This should now be all fully resolved, it now includes the WebView DLLs + the drivers as embedded resources. I've verified it is working with the self-extracting PerfView.exe on my machine and it seems to be all working fine.

brianrob (approved):
LGTM.  Thank you @caaavik-msft!

Issue #2134, opened 2024-12-04
Azure Linux 3 perfcollect install script not supported


AguirreMoy:
Reproduction steps:

Run the following on AzureLinux 3 in accordance with [Microsoft learn instructions](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/trace-perfcollect-lttng).

`curl -OL https://aka.ms/perfcollect
chmod +x perfcollect
sudo ./perfcollect install
`

Fatal error occurs:

`Auto install unsupported for this distribution.  Install perf manually to continue.`

Suggested fix:
Update erfview/main/src/perfcollect/perfcollect for Azure Linux 3 support.

Screenshot:
![Image](https://github.com/user-attachments/assets/b96b1a22-f7bc-47cf-b3eb-bd1f59d09c9d)

PR #2135, opened 2024-12-04
FIX: perfcollect install script Azure Linux 3

AguirreMoy:
WHY:
Azure Linux and Mariner are the same image with different branding however the `DISTRIB_ID` on Azure Linux 3.0 onwards is changed from "mariner" to "azurelinux"

FIX:
Include checking for "azurelinux" when setting the Ismariner flag.

Bug/Fix discovered with @faluciano

Fixes https://github.com/microsoft/perfview/issues/2134

CleanShot 2024-12-03 at 19 26 11@2x


brianrob (approved):
LGTM.  Thanks @AguirreMoy for reporting and fixing!

PR #2139, opened 2025-01-03
Update start stop activity computer to support OpenTelemetry-Sdk EventSource

xiaomi7732:
### Description

The OpenTelemetry SDK event source is emitting request start/stop information, but the format differs from what TraceEvent expects for start/stop operations. This mismatch prevents proper recognition of activities in the CallTree view.

This change introduces an exception to accommodate the OpenTelemetry SDK’s event format, ensuring that start/stop activities are correctly recognized and displayed in the CallTree view.

**Details:**  
For reference, see the implementation in the OpenTelemetry SDK:  
[OpenTelemetrySdkEventSource.cs](https://github.com/ualehosaini/opentelemetry-dotnet/blob/1e89b39d72236285081c727956439af6715ab782/src/OpenTelemetry/Internal/OpenTelemetrySdkEventSource.cs)

**Before the change:**
image

**After the change:**
image

Tagging @brianrob for awareness.

### DIFF
--- a/src/TraceEvent/Computers/StartStopActivityComputer.cs
+++ b/src/TraceEvent/Computers/StartStopActivityComputer.cs
@@ -149,6 +149,13 @@ namespace Microsoft.Diagnostics.Tracing
                     }
                 }
 #endif
+                // Special case - OpenTelemetry-Sdk
+                else if (data.Opcode == TraceEventOpcode.Info && data.providerGuid == OpenTelemetrySdkProvider)
+                {
+                    FixAndProcessAzureMonitorOpenTelemetryEvents(data);
+                    return;
+                }
+
                 // TODO decide what the correct heuristic for deciding what start-stop events are interesting.  
                 // Currently I only do this for things that might be an EventSource 
                 if (!TraceEventProviders.MaybeAnEventSource(data.ProviderGuid))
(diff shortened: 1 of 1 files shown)


src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -149,6 +149,13 @@ public StartStopActivityComputer(TraceLogEventSource source, ActivityComputer ta
                     }
                 }
 #endif
+                // Special case - OpenTelemetry-Sdk
+                else if (data.Opcode == TraceEventOpcode.Info && data.providerGuid == OpenTelemetrySdkProvider)
```
pharring, line 153: Lower down there are already special cases specifically for fixing up providers that don't follow the start/stop conventions completely (e.g. FixAndProcessAppInsightsEvents). Should OTel be handled there instead?

xiaomi7732: @pharring Thanks for the ask. That will cause more changes. On line 154 (before change), there's an algorithm to check if it is **maybe** event source, and the `OpenTelemetry-Sdk` event source won't satisfy that check:

```csharp
if (!TraceEventProviders.MaybeAnEventSource(data.ProviderGuid))
{
    return;
}
```

We could update the algo to special case `OpenTelemetry-Sdk` there, and some other, but I don't see a why. It will only end up making it complicated.

pharring: The OTel SDK Provider GUID _is_ an EventSource and `MaybeAnEventSource` will be true.

xiaomi7732: Yes, you are right. I must oversight something. New iteration pushed.

src/TraceEvent/Computers/StartStopActivityComputer.cs
```diff
@@ -1131,6 +1138,20 @@ private bool TryProcessDiagnosticSourceStartEvents(TraceEvent data)
             return false;
         }
 
+        private void FixAndProcessAzureMonitorOpenTelemetryEvents(TraceEvent data)
```
pharring, line 1141: Consider removing "AzureMonitor" from the name. This EventSource comes from the "base" OpenTelemetry SDK -- not the Azure Monitor distro.

pharring (changes requested):

src/TraceEvent/Computers/StartStopActivityComputer.cs
xiaomi7732: Of course.

xiaomi7732: Fixed.

pharring (approved):

cincuranet (approved):

PR #2140, opened 2025-01-08
Added missing time info fields in the Raw XML View in GCStats

mrsharm:
Updated the following in the Raw XML View in GCStats:

1. Added the `MarkShortWeak` entry to the GlobalHeapHistory node that was missing.
2.  Added the `MarkSizedRef`, `MarkOverflow`, `MarkDependentHandles`, `MarkNewFQ`, `MarkSteal` and `MarkBGCRoots` entries to the PerHeapHistory node.


cincuranet (approved):

cincuranet:
Thanks!

PR #2143, opened 2025-01-13
Fixed timings associated with the MarkTimes and MarkPromoted fields for the per heap mark times.

mrsharm:
CC: @Maoni0


cincuranet:
@mrsharm Updated the title, for now. Once ready, let me know and I'll review and merge.

src/PerfView/GcStats.cs
```diff
@@ -542,43 +543,31 @@ public static void ToXmlAttribs(TextWriter writer, TraceProcess stats, TraceLoad
...
+                                writer.Write(" {0} =\"{1:n3}", type, mt.MarkTimes[(int)type]);
                                 if (mt.MarkPromoted != null)
                                 {
-                                    writer.Write("({0})", mt.MarkPromoted[(int)MarkRootType.MarkOverflow]);
+                                    writer.Write("({0})", mt.MarkPromoted[(int)type]);
                                 }
+                                writer.Write("\"");
                             }
+
+                            WriteDetailsAboutMarkRootType(MarkRootType.MarkStack);
```
Maoni0, line 561: can we just call this in a loop instead of having to duplicate the call many times?

src/PerfView/GcStats.cs
```diff
@@ -542,43 +543,31 @@ public static void ToXmlAttribs(TextWriter writer, TraceProcess stats, TraceLoad
 
                         if (mt != null)
                         {
-                            writer.Write(" MarkStack =\"{0:n3}", mt.MarkTimes[(int)MarkRootType.MarkStack]);
-                            if (mt.MarkPromoted != null)
+                            void WriteDetailsAboutMarkRootType(MarkRootType type)
```
Maoni0, line 546: nit - maybe call this something like WriteMarkRootTypeInfo instead of something that reads like a sentence?

src/TraceEvent/Computers/TraceManagedProcess.cs
```diff
@@ -3177,36 +3177,31 @@ private void ConvertMarkTimes()
             {
                 foreach (KeyValuePair<int, MarkInfo> item in PerHeapMarkTimes)
                 {
-                    if (item.Value.MarkTimes[(int)MarkRootType.MarkSizedRef] == 0.0)
+                    List<(MarkRootType, double)> perRootMarkTime = new List<(MarkRootType, double)>();
+
+                    // Accumulate all the times that have a corresponding mark event.
+                    for (int i = 0; i < item.Value.MarkTimes.Length; i++)
```
Maoni0, line 3183: it seems like it wouldn't take much work to improve the perf here? you could preallocate `perRootMarkTime` instead of allocating one over and over again in per iteration per GC. you could also sort in place instead of creating new tuples every time.

brianrob: I think these are good recommendations.  One of the things that we're seeing more often these days is larger traces from machines with larger heaps.  I think there are a few simple perf/memory improvements here that can make a difference here without you having to write too much code.

brianrob (approved):

PR #2145, opened 2025-01-14
Set the timestamp as a QPC value based on UTC time

andd64:
When using relogger to writes an event that did not exist previously into the data stream, the context data and timestamp is given explicitly.  The newEvent.SetTimeStamp expects a QPC time based on UTC time.

### DIFF
--- a/src/TraceEvent/ETWReloggerTraceEventSource.cs
+++ b/src/TraceEvent/ETWReloggerTraceEventSource.cs
@@ -157,7 +157,14 @@ namespace Microsoft.Diagnostics.Tracing
                 newEvent.SetEventDescriptor(ref *ptrDescr);
                 newEvent.SetProviderId(ref providerId);
                 _LARGE_INTEGER fileTimeStamp = new _LARGE_INTEGER();
-                fileTimeStamp.QuadPart = timeStamp.ToFileTimeUtc();
+                if (DateTimeKind.Utc != timeStamp.Kind)
+                {
+                    fileTimeStamp.QuadPart = UTCDateTimeToQPC(timeStamp.ToUniversalTime());
+                }
+                else
+                {
+                    fileTimeStamp.QuadPart = UTCDateTimeToQPC(timeStamp);
+                }
                 newEvent.SetTimeStamp(ref fileTimeStamp);
                 newEvent.SetProcessId((uint)processId);
                 newEvent.SetProcessorIndex((uint)processorIndex);
--- a/src/TraceEvent/ETWReloggerTraceEventSource.cs
+++ b/src/TraceEvent/ETWReloggerTraceEventSource.cs
@@ -156,16 +156,16 @@ namespace Microsoft.Diagnostics.Tracing
 
                 newEvent.SetEventDescriptor(ref *ptrDescr);
                 newEvent.SetProviderId(ref providerId);
-                _LARGE_INTEGER fileTimeStamp = new _LARGE_INTEGER();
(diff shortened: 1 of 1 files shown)


brianrob:
@andd64 thanks for the contribution.  Do you have a pointer to the documentation?  From what I can find from the [documentation](https://learn.microsoft.com/en-us/windows/win32/api/relogger/ns-relogger-event_header):

```
TimeStamp

Contains the time that the event occurred. The resolution is system time unless the ProcessTraceMode member of 
[EVENT_TRACE_LOGFILE](https://learn.microsoft.com/en-us/windows/desktop/ETW/event-trace-logfile) contains the 
PROCESS_TRACE_MODE_RAW_TIMESTAMP flag, in which case the resolution depends on the value of the Wnode.ClientContext 
member of [EVENT_TRACE_PROPERTIES](https://learn.microsoft.com/en-us/windows/desktop/ETW/event-trace-properties) at the 
time the controller created the session.
```

andd64:
Hi @brianrob , thanks for the prompt reply.
Originally I tried to use system time, as per the documentation, but when I view the events back, it had the system time as the QPC time itself, which messed up the order.

For example, with the original code which simply converts timeStamp.ToFileTimeUtc(), this snippet of code writes some payload data using the event data's timestamp itself.

```

DateTime currTimeStamp = data.TimeStamp;
currTimeStamp = currTimeStamp.AddMilliseconds(500);

var manifestDescr2 = new _EVENT_DESCRIPTOR();
manifestDescr2.Id = (ushort)data.ID;
manifestDescr2.Task = (ushort)data.Task;
manifestDescr2.Opcode = (byte)data.Opcode;

float[] test_arr = { 0x69, 0x70, 0x71 };
object[] payloads2 = test_arr.Cast<object>().ToArray();

relogger.WriteEvent(data.ProviderGuid, ref manifestDescr2, currTimeStamp, 1, 1, 1, activityID, payloads2);

```

Here's an example where we trigger this event write, and view it back:

The event get's injected with this timestamp info:
TimeStamp="01/24/49 08:59:55.841114" ID="15505" Version="0" Keywords="0x00000000" TimeStampQPC="133,813,634,114,243,789" QPCTime="0.100us"


image

image


But with my change, it get's injected at the correct timestamp.

brianrob:
Yeah, that definitely seems wrong.  And in fact, looking at how we handle time stamps in event parsing, we treat the time stamp in the event header elsewhere as QPC too.  Let's go with what you have here.

src/TraceEvent/ETWReloggerTraceEventSource.cs
```diff
@@ -157,7 +157,14 @@ public unsafe void WriteEvent(Guid providerId, ref _EVENT_DESCRIPTOR eventDescri
...
                 newEvent.SetProviderId(ref providerId);
                 _LARGE_INTEGER fileTimeStamp = new _LARGE_INTEGER();
-                fileTimeStamp.QuadPart = timeStamp.ToFileTimeUtc();
+                if (DateTimeKind.Utc != timeStamp.Kind)
+                {
+                    fileTimeStamp.QuadPart = UTCDateTimeToQPC(timeStamp.ToUniversalTime());
+                }
+                else
+                {
+                    fileTimeStamp.QuadPart = UTCDateTimeToQPC(timeStamp);
```
brianrob, line 166: Since this is no longer a file timestamp, will you please rename `fileTimeStamp` to `qpcTimeStamp`?

brianrob (approved):
LGTM.  Thank you!

PR #2156, opened 2025-01-30
[WIP] Adding OOM Diagnostics to GCStats

mrsharm:
This PR adds a separate section for OOM details emitted as a dynamic event from the runtime. 

Example of the details (if an OOM exists in the trace) in GCStats:
image

Interpretation of the Dynamic Event in the Events View:
image


brianrob:
Closing after offline conversation with @mrsharm.

PR #2159, opened 2025-02-02
Fixes that repeatedly adding unhandled event handler to TraceLog to real time session conversion

wwh1004:
Before this pr the unhandled event handler will be registered repeatedly as long as DispatchClonedEvent is called causing very high cpu usage


brianrob (approved):
LGTM.  Thank you!

PR #2161, opened 2025-02-02
Add support for configurable real-time delay in TraceLogEventSource creation

wwh1004:


src/TraceEvent/TraceLog.cs
```diff
@@ -154,11 +154,32 @@ public static TraceLog OpenOrConvert(string etlOrEtlxFilePath, TraceLogOptions o
...
+        /// will deliver events in real time.   However an TraceLogEventSource has an underlying Tracelog (which you can access with
+        /// the .Log Property) which lets you get at aggregated information (Processes, threads, images loaded, and perhaps most
+        /// importantly TraceEvent.CallStack() will work.  Thus you can get real time stacks from events).
+        ///
+        /// Note that in order for native stacks to resolve symbolically, you need to have some Kernel events turned on (Image, and Process)
+        /// and only windows 8 has a session that allows both kernel and user mode events simultaneously.   Thus this is most useful
+        /// on Win 8 systems.
+        /// </summary>
+        /// <param name="realTimeDelayMSec">The delay in milliseconds between when an event is received in TraceLog and when it is dispatched to the real time event source.</param>
+        public static TraceLogEventSource CreateFromTraceEventSession(TraceEventSession session, int realTimeDelayMSec)
```
brianrob, line 171: Please rename to something like `minDispatchDelayMSec`.  From a quick read of the code, this is a minimum delay based on the time the event spent in the queue.  It may spend longer in the queue.

brianrob (commented):
Thank you for the contribution.  One request below.

src/TraceEvent/TraceLog.cs
wwh1004: Fixed

brianrob (approved):
LGTM.  Thanks!

PR #2166, opened 2025-02-04
Don't do FlushRealTimeEvents if already in FlushRealTimeEvents

wwh1004:


src/TraceEvent/TraceLog.cs
```diff
@@ -861,47 +861,68 @@ private unsafe void DispatchClonedEvent(TraceEvent toSend)
...
             {
-                var nowTicks = Environment.TickCount;
-                // TODO review.
-                while (realTimeQueue.Count > 0)
+                lock (realTimeQueue)
                 {
-                    QueueEntry entry = realTimeQueue.Peek();
-                    // If it has been in the queue less than 1 second, we we wait until next time) & 3FFFFFF does wrap around subtraction.
-                    if (minimumAgeMs > 0 && ((nowTicks - entry.enqueueTick) & 0x3FFFFFFF) < minimumAgeMs)
+                    if (!isFlushingRealTimeEvents)
```
brianrob, line 870: If I'm reading this correctly, if you get here, then by definition `isFlushingRealTimeEvents` is false because the lock won't allow any callers into this section of the code when `isFlushingRealTimeEvents` is true.  I think you can get rid of the check inside the `lock` block.

brianrob (commented):
Thanks for contributing this.  I support this change, but I think you can simplify it just a bit.  Please see my comment below.

src/TraceEvent/TraceLog.cs
wwh1004: I have reread the relevant code. It seems that the purpose of this code is to ensure that any event is not dispatched until after minimumAgeMs. Assuming minimumAgeMs is 1000, an event enqueue at 900ms, and FlushRealTimeEvents is triggered at 1000ms. This code can guarantee that the event is postponed until the next invocation of FlushRealTimeEvents.

brianrob: Just to clarify, I'm referring to removing line 870 in the new diff.  If you get into the lock `isFlushingRealTimeEvents` is guaranteed to be false, and so you don't need to check it again, you can just set it to true and call `FlushRealtimeEventsNoLock`.

wwh1004: The github unified diff view caused my misunderstanding 😂. Fixed.

brianrob (approved):
LGTM.  Thank you!

Issue #1446, opened 2021-06-18
ThreadID is uint64 in nettrace v4, but handled as int in EventPipeEventSource.


lateralusX:
There is currently a downcast of ThreadID read from nettrace v4 into a int. On platforms using pointers as thread id's, there is a risk we will get negative thread id's showing up in SpeedScope and Perfview. It doesn't seem to break anything, but looks "wrong" when displayed and could in theory cause collisions on 64-bit systems.

What we probably need to do is changing ThreadID to a long (unlong to match nettrace v4) from here, https://github.com/microsoft/perfview/blob/8a34d2d64bc958902b2fa8ea5799437df57d8de2/src/TraceEvent/EventPipe/EventPipeEventSource.cs#L1485, all the way up the parsing logic.

lateralusX:
//CC @brianrob @josalem

lateralusX:
Took a quick look at this and doing it thoroughly will be a very invasive change that probably will break a lot of scenarios, there are hundreds of assumptions across the repro that thread id is of type int, and thread id's might be emitted not just into the record header but might also be part of specific event payloads, and in order for all to work we would need to go through all of this. Most of tools also depend on ETW structs that also handles thread id as int and since it is part of sequential structs, we would need to extend it with an additional 64-bit thread id. If we are going to do this we might need to start with some specific scenario where we have full overview of all usage of thread id, bringing it into full perfview repro seems like a bigger task with big regression risks.

brianrob:
Yup, unfortunately, TraceEvent uses integers for thread IDs because it was written when it only handled Windows traces, and thread IDs on Windows are 4 bytes wide.  Agreed this would have fairly large ramifications.

brianrob:
@lateralusX thanks for bringing this up again.  Agreed that it's worth considering this again as we look to support the next version of nettrace which would enable cross-platform multi-process traces.

PR #2170, opened 2025-02-10
Discuss new NetTrace V6 format

noahfalk:
[EDIT]: I have updated this PR and now I am aiming to get it checked in as long as everyone is happy with it.

This PR adds support for parsing a new V6 version of the NetTrace file format. The design of the format is described in the included NetTraceFormat.md. At the end of that doc is a summary of the changes between V5 and V6.

NOTE: The parser is capable of reading things that the higher level TraceEvent API doesn't yet expose such as event descriptions, optional trace metadata, 8 byte thread ids, new correlation identifiers, etc. I think Brian and I will need to work together outside the scope of this PR to decide if/how/when we expose parts of that data in TraceEvent's public API surface. As-is the unit tests are calling into private APIs to confirm those extra data elements were parsed correctly.


src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,235 @@
...
+
+The content of a thread block is:
+
+- Count - uint16
+- Concatenated sequence of Count entries, each of which is
+  - Index - varuint
+  - ThreadId - varuint
+  - ProcessId - varuint
+
+When referencing a ThreadBlock entry by index, the reference must occur no later in the stream than the content of the next SequencePoint block and no earlier than ThreadBlock itself.
```
beaubelgrave, line 127: What are the order requirements then? Does a ThreadBlock need to be in the file before Events that utilize it? Or just before another SequencePoint? What if there are no SequencePoints? (I had this same question about the StackBlock, since the current format.md just states they can be in any order).

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,235 @@
...
+  - BeginPrivateObject Tag
+  - NullReference Tag
+  - version (int)
+  - minimum reader version (int)
+  - type name (string)
+  - EndPrivateObject Tag
+
+This header is 16 bytes + a variable length UTF8 string indicating the actual type of the object. The length of the object content was not included in the header. The EventBlock, MetadataBlock, and SPBlock objects included a BlockSize field as the first 4 bytes of the object content and the Trace object had fixed size.
+
+In V6 the object header is replaced by a simplified block header:
```
beaubelgrave, line 67: Love it

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,235 @@
...
+
+```
+struct BlockHeader
+{
+  uint24 BlockSize
```
beaubelgrave, line 72: This only gives us 16MB, I'll overflow that and don't want to have to always make multiple blocks. Is there a reason to be so "cheap" on the header size? Why not 56 bits for size and 8 bytes for kind?

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+In V6 the EventBlock, MetadataBlock, StackBlock and SPBlock no longer have the initial BlockSize and padding fields. The value previously stored in the BlockSize object field is now in the BlockHeader. The semantics of the BlockSize header field are very similar to before, it represents the number of bytes in the remainder of the block. Specifically if file offset X points to the start of a block, then X + sizeof(BlockHeader) + BlockSize points to the start of the next block.
+
+Readers should skip over any blocks with an unrecognized BlockKind. New blocks might be introduced in minor version updates to store new types of data and this is expected to be a back-compatible change because old readers skip over it.
+
+
+#### Removed block footer
+
+In V5 all objects ended with a trailing EndPrivateObject tag byte. In V6 this trailing byte serves no purpose and is removed.
```
beaubelgrave, line 101: Love it

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,235 @@
...
+
+The SequencePoint block in V5 has a list of (ThreadId,SequenceNumber) tuples. All the ThreadIds in that list are now replaced with indexes into the ThreadBlock table and the field is renamed ThreadIndex.
+
+#### Updated semantics for Trace fields
+
+The TraceBlock has a ProcessId field that was historically set to the ID of the process being traced. When doing a trace that may contain more than one process this field should be set to zero.
+
+The TraceBlock PointerSize field represents the machine pointer size for traces that may contain more than one process. For processes whose bitness is smaller than the machine bitness (such as when using an emulator), memory addresses should be zero extended to the machine pointer size.
+
+The ExpectedCPUSamplingRate field has been unused for a long time. Writers are recommended to set its value to zero and readers are recommended to ignore it.
```
beaubelgrave, line 141: Guessing we'll put this in metadata if it's important to folks to know. IE: Sampling CPU might result in a value of 1, but not sure how much time that value of 1 represents.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,235 @@
...
+                                     // The reader should skip Size bytes of data after the Size field to get to the next optional metadata element
+
+```
+enum OptionalMetadataKind
+{
+  OpCode = 1,
+  // 2 is no longer used. In format V5 it was V2Params
+  KeywordLevelVersion = 3,
+  MessageTemplate = 4,
+  Description = 5,
```
beaubelgrave, line 232: I'd suggest having a general name value pair entry, so we can stamp stuff like "SamplingRate", "1Khz". A human or machine could then understand it. We would then have some standard key value pairs that we expect good producers of data to stamp, so tooling can look for general details. However, all tools are expected to be able to show all name value pairs to the user when asked/requested. PPROF does this, and it allows a lot of details to fit into the format for engineers to see. Like maybe we want to store the machine name or ID that the data was captured from, or some environment name that only makes sense to a subset of engineers.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: The order needs to be ThreadBlock before EventBlock. Any number/kind of blocks can be in between except for a SequencePointBlock.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: There is probably some missing context about the expected sizes of these things in order to give good performance. Typically the runtime produces Blocks that are no larger than 100KB and aims to have SequencePointBlocks about 5MB apart in the file. While there is no hard correctness constraint on those numbers, the size of these things affects memory consumption needed by the reader to parse the file. So the 24 bit size on the blocks is within the context that you won't want really large blocks to begin with.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: The CPU sampling events all have timestamps on them. The current viewer tools observe the timestamps to see how far apart they are and how many samples occurred in a given time range.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: I'm happy to add a Key/Value pair.

> Like maybe we want to store the machine name or ID that the data was captured from

That sounds more like metadata describing a trace rather than metadata describing an event. Do you want me to add an OptionalMetadata list onto the TraceBlock so that you can throw key/value pairs in there?

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
brianrob: If we want to use other metadata, I think that's OK, but I can say that there is generally value in knowing the expected sampling rate at analysis time.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
brianrob: Clarifying question here - I see that indices into the ThreadBlock are only valid between the ThreadBlock and a SequencePointBlock.  Do we expect there to be another ThreadBlock immediately after a SequencePointBlock that may contain data from the previous ThreadBlock if subsequent events need to reference previously used process/thread values?

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,239 @@
...
+- varuint Count
+- Contenated sequence of Count field descriptions, each of which is:
+  - Type;                     // The type of a field - described below
+  - string FieldName;         // varuint length-prefixed UTF8 string (no null terminator)
+
+Type format is:
+
+- uint8 TypeCode;             // Type is a discriminated union and this is the discriminant. 
+- optional Type ElementType;  // Only present if TypeCode == Array (19).
+- optional FieldDescriptions ObjectFields; // Only present if TypeCode == Object (1).
```
brianrob, line 176: In the case of more complex types (e.g. objects), the format seems to support nested objects (objects within objects).  I'd expect this to be supported given that this is used in TraceLogging events.  Just want to make sure that you do expect this to be supported.

brianrob (commented):
Thanks for putting this together @noahfalk!  A few comments/questions.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
beaubelgrave: I think it would be useful to have on TraceBlock as well, for sure. This is where we can put sampling frequency, etc.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: Yep, I'm expecting it to be supported. FieldDescriptions contains 1 or more Types, and a Type could recursively be an object with its own nested FieldDescriptions.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: Yeah, I expect it will be common that a (thread,process) tuple is long-lived so many ThreadBlocks will re-introduce it after each sequence point. Its a tradeoff where the file format gets a little more verbose in exchange for not asking the reader to cache every thread it has ever seen in the trace.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: Let me know if you guys don't think we should make this recommendation. As far as I could tell PerfView doesn't care what the value of this field is right now so practically it seemed unused. From a file format perspective it felt limiting if the metadata constrains us to only defining CPU sampling and only one rate for the entire file. For example a trace might have the OS doing CPU sampling at rate R1 and .NET runtime doing thread-time sampling at rate R2 and any number of other things also creating events from their own independent sampling mechanisms. Including expected rates as event metadata or as some initial event in the trace appeared like a more flexible mechanism.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: Key/Value pairs are added to the spec in the latest PR update. I also took the liberty of removing a few of the Trace object fields converting them to semantic conventions on the key/value pairs. I can return any or all of them to be explicit format fields if folks think that is preferable. My rationale for converting them is that they weren't required for parsing anything else in the format and I could imagine traces where those pieces of data weren't relevant for a user interpreting the trace. It seemed nice not to force writers to fill in data if the data didn't matter to them.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
brianrob: Great!  Thanks.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
brianrob: I think it's reasonable to not have the format care about this particular field - I'd just be inclined to have one-collect write out an event with the metadata that it wants to export, and then tools like PerfView and TraceEvent can expose it.  Thinking a bit further, my point here is more that I think knowing the expected rate is interesting - I agree that it doesn't need to be/probably shouldn't be tied to the serialization format.

noahfalk:
Just a heads up, I've been tweaking the design a bit while implementing the parser. I'll have an updated spec out soon as part of the PR that implements the functionality.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,288 @@
...
+- PointerSize - int
+- KeyValueCount - int
+- A sequence of KeyValueCount entries, each of which is:
+  - key - string      // varuint prefixed UTF8 string
+  - value - string    // varuint prefixed UTF8 string
+
+For the three fields that were removed, they can be optionally encoded as key/value pairs where the key names are:
+
+- NumberOfProcessors -> "HardwareThreadCount"
+- ProcessId -> "ProcessId"
```
lateralusX, line 189: This TraceBlock is currently emitted at the beginning of the stream. If the idea is to include multiple processes in the same stream, then I believe it wouldn't make much sense to have any process specific information in this record, only stuff that is shared between all processes included in the trace. Since some of the information is tied specifically to a process, like process id, CPU sampling rate (could in theory be different for each dotnet process), maybe it make sense to emit a ProcessBlock in the stream with additional process specific metadata, I believe that block only need to be emitted once in the stream, before first use of the process id, in for example the ThreadBlock. It's an option to put an index representing the process id in that record and then use that in the ThreadBlock, but that will probably not be necessary, pid_t is normally 32-bit and currently an uint32_t in EventPipe, it will mainly be referenced through the new ThreadBlock that will appear once per thread between sequence points, so using the full uint32_t process id is probably good enough.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,288 @@
...
+
+#### Removed block footer
+
+In V5 all objects ended with a trailing EndPrivateObject tag byte. In V6 this trailing byte serves no purpose and is removed.
+
+#### Changed end of stream marker
+
+In V5 the file ended with a NullReferenceTag byte. In V6 the file ends with a new EndOfStream block. This block consists only of a BlockHeader and has BlockSize=0.
+
+## Multi-process support
```
lateralusX, line 107: I understand that this is just making the format multi-process aware, but how do we envision this to be used? Currently each EventPipe session owns the whole stream, and there is plenty of state included serializing the stream, like caching stack traces, sequence point management, compression etc. If we would like to include multiple processes in the same session stream, then I would assume each EventPipe session would just store the raw events into memory, similar to how buffer manager currently does it internally and then have an external trace process that could trace multiple dotnet processes at the same time, meaning that the trace process will take care of most of the serialization currently done in the per session EventPipe serialization layer?

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,288 @@
...
+
+#### New ThreadBlock
+
+A new top-level block, BlockHeader.Kind=6, contains a set of thread id/process id tuples referenced by index. This allows many events to refer the same (thread id, process id) without needing to repeatedly encode the same pair of potentially large IDs.
+
+The content of a thread block is:
+
+- Count - uint16
+- Concatenated sequence of Count entries, each of which is
+  - Index - varuint
```
lateralusX, line 123: Since this index must be unique for all events emitted between SequencePoints and if we could include events from multiple processes inside the same SequencePoint range, then the creation and handling of this block must be done by the process responsible to merge all events coming in from all traced processes into the nettrace stream. Each EventPipe session can't handle this since then there will be collision in indexes between processes included in the same nettrace file.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: I wanted to balance support for having multiple processes while still letting things be simple/efficient for traces that do only have one process. My expectation is that when the runtime emits a single process trace it will set the ProcessId key, but in a multi-process trace the key will be absent and each thread will specify the process id separately. I expect it will be helpful for some readers to know up-front if they are dealing with a multi-process trace rather than needing to scan every thread in the trace to discover that only one process ever appears.

I've not seen an example where CPU sampling rate varied by process but I also made that one optional for similar reasoning that maybe there won't be a single global rate. If the rate did vary I think varying by process is one possibility but there are others such as vary by time or multiple distinct sampling rates are occurring simultaneously. If we create a ProcessBlock and put CPU sampling rate it would handle per-process rates well, but wouldn't handle any of those other cases. It seemed easier to leave our choice of how to encode this to our future selves. I think we have options that wouldn't require any format changes such as emitting a well-defined event once-per-process that sets the rate or encoding multiple rates in a differently named Key-Value pair.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: The background context here is that we are planning to create a Linux profiling tool (based on perf_events and user_events) that will be doing full machine profiling. We wanted an open data format to write the trace info into that PerfView already had decent support for so we thought nettrace would be a good option. In this scenario it won't be EventPipe creating the data at all. At the moment there is no plan for EventPipe to create a multi-process trace.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
```diff
@@ -0,0 +1,288 @@
...
+  - ThreadId - varuint
```
lateralusX, line 124: In EventPipe the thread id is a uint64_t sized value, but currently our TraceEvent handles it as int, https://github.com/microsoft/perfview/issues/1446. This has caused issues with negative thread id's showing up in perfview, there are other platforms that typedef pthread_t to a pointer (iOS/tvOS/MacCatalyst/MacOS), so thread id will be truncated on these platforms as well. If we change the file format, I also believe we should try to enhance libraries parsing this data to either switch to a wider data type for thread id, or utilize the index as an indirection to get to the full thread id and process id.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: No plan right now to have EventPipe do any merging or generate a multi-process trace. I agree that if we did ever want a scenario that took multiple single-process traces and merged them together it would require the merge to do substantial work with the current format. In addition to thread indexes there would also be collisions on metadata ids and stack ids + all the events would have to spliced together with specific ordering to obey the format's timestamp ordering rules.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
lateralusX: OK, so that means that this new profiling tool will handle the full serialization of the nettrace file and rely on rutnime user_events to get the data out.

noahfalk:
This PR is updated with a complete implementation and some adjustments to the design as I considered it more and got some experience with implementing it. 

Unforetunately due to a personal issue I'll be away for likely at least a week but others are welcome to discuss, build the source if you need something to test with, or merge the code if its urgent to do that.

I did have one other design piece I was considering putting in V6 that isn't included right now - changing the current encoding of ActivityId/RelatedActivity into an extensible set of correlation identifiers. In particular it might useful to have a place to store a TraceId/SpanId from the W3C TraceParent spec but it could also allow for open-ended storage of other identifiers. I was thinking of something like

new ConstantPoolBlock that stores a list of indexed binary blobs:
index - varuint
size - uint16
blob - size bytes of data

In the event header repurpose the slot/bit current used for activityId and store an optional reference into the ConstantPool there. The data that would go in the blob would be a concatenation of correlation identifiers similar to the optional thread data or optional metadata info. Something like:
kind - uint8
if (kind==1)
  ActivityId Guid
If (kind==2)
  RelatedActivityId guid
If (kind==3)
  TraceId
If (kind==4)
  SpanId
If (kind==5)
  KeyValue string,string

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: Is this a suggestion for a change to the format, or a suggestion to change TraceEvent? From what I can tell both the current and new format will happily support a 64-bit ThreadId. The parser for NetTrace also supports reading/storing a 64 bit thread in its own internal datastructure. The problem arises because the interface between the NetTrace parser and the rest of TraceEvent forces the parser to downcast the thread ID to a 32 bit value. Fixing that is the work tracked by #1446. I couldn't tell that anything we do to change the format or the NetTrace parser changes the work involved to solve #1446 ?

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: We have many different rundown events that include lots of process specific information - not sure if that is what you were refering to? I'd lean towards keeping that data as events unless there is a specific benefit we'd get by putting it in the format? My overall thought is that its much easier to evolve the set of events emitted into the file and the contents of their payloads whereas file format data structures are quite resistant to change. I think things are a good candidate to go in the format when we expect them to be very stable, broadly applicable, and/or its challenging to get encoding size efficiency we'd want via events.

lateralusX: We could of course continue with what we already have, but I believe it would make sense to collect the most obvious process specific data into a block in the beginning of the file instead of having tools read through the full event stream hunting down specific events emitted somewhere in the stream to collect this information. One event that we always emit when closing the session is process info, that event includes process command line, os info, arch info, I also saw some other special handling in the header block around pointer size that might make sense to have in a process block if we would end up with both 32-bit and 64-bit traces in the same file, other parts of the data currently encoded in the Trace object might make sense to "override" it in a process block in case the specific process use different values. Tools will probably also be interested to build a list of all processes included in the trace file (with process name and pid). Have specific process blocks in the beginning of the file representing the process collected in the file will also simplify for tools to collect this information without scanning through the complete trace looking for different process related events.

lateralusX: What does ETW do with respect to process metadata in a multi process trace file?

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
lateralusX: This was mainly just a reminder of the issue in the TraceEvent library and if we implement a new stuff to parse format, we should consider adding a new API to handle thread id as a 64-bit unisgned data type all the way through the public API. Its not directly related to the format, since we already handle it as 64-bit in the format its not directly related to this issue. I make a note in #1446.

src/TraceEvent/EventPipe/NetTraceFormat_v6.md
noahfalk: I don't know the details of ETW's format but I do know that ETW traces have ProcessStart and ProcessDCStart kernel events. Events also have a Process Id header field. PerfView uses that data to populate it's own understanding of processes in a trace. If ETW has some other special treatment of processes in the format I'm not aware of it and I don't think PerfView makes use of it.
https://github.com/microsoft/perfview/blob/b0c027a3a16edc44847d8032f46d6432541f81b8/src/TraceEvent/Computers/TraceProcess.cs#L130

For a tool that is doing a live trace with processes starting and stopping it wouldn't necessarily know about all processes in advance and so couldn't guarantee that ProcessBlocks are at the start of the stream. Presumably we'd have to allow ProcessBlocks to be inserted at any point so multi-process streams would still need to search for them. For streams where all process information is known up-front we could certainly put the data into ProcessBlocks at the front, but we could also put the data into events at the front or in key-value pair data hanging off the TraceBlock.

@brianrob @beaubelgrave: For the pointer size aspect, are you expecting to collect traces where process bitness is going to vary and its important that individual events can specify different pointer sizes? The only place pointer sized data shows up directly in the .nettrace format today are the IPs for stacktraces and my current suggestion was to zero-extend 32 bit addresses out to 64 bits if the trace is globally labeled as 64 bit. ETW presumably also supports events where pointer sized data is present in the event payload area in which case understanding process bitness is required for correct decoding. .nettrace doesn't offer any metadata to describe a pointer sized type but it also doesn't stop people from writing arbitrary payload data that is not described by metadata.

noahfalk: OK, since it doesn't sound like we have a scenario that needs to distinguish bitness on a per-event basis within the trace and other process data can continue to be conveyed in events or TraceBlock info it doesn't seem like ProcessBlock is needed right now.

One place where I imagine we might want it in the future (as part of a future version of the format) is if we start supporting symbolication of callstacks as a file-format level concern rather than requiring rundown events layered on top. The 'event' portion of a rundown event such as timestamps, thread ids, sequence numbers, etc. isn't meaningful so there is some marginal file format efficiency that could be gained by giving them an explicit non-event representation. Rundown events as-is also create some ambiguity whether they are a 1st class part of the user's mental model or they are just an encoding convenience that should be abstracted away from the user. Our command-line interfaces, UI, and documentation tend to treat rundown events inconsistently.

noahfalk:
I updated the format with a support for label lists and updated the docs. If folks are happy with this then I think we could check it in. Now that we aren't racing to get this in for the next VS update we do have more time to adjust it, but aside from the label list part which is brand new I'm hoping most feedback elsewhere has already been resolved. I'm not aiming to make any significant further changes.

cincuranet:
@noahfalk Can you check the conflicts? Thanks.

noahfalk:
yeah, getting that merged up is in my todo list. The ball is in my court :)

noahfalk:
@cincuranet - I've rebased so it should be merging cleanly now

Based off discussions with @beaubelgrave there are a few small adjustments in this most recent update:
- ThreadIndex in the compressed header is restored to the same behavior ThreadId had in V5: it defaults to the ThreadIndex in the previous event when not specified. This makes header encoding more efficient for writers that set a single CaptureThreadIndex throughout the trace.
- SequencePoint blocks now have a flags field. Setting flag & 1 means that the reader should flush all ThreadBlock data, similar to what it already did for stack and label list references. This can be used by writers that don't want to redefine threads after every sequence point rather than managing their lifetimes explicitly. Since I was already editing the SPBlock format I also opportunistically converted the thread index and sequence number encodings in the list to use variable sized integers.

Bug fixes and spec clarification:
- I fixed a bug in the parser where it was misreporting the number of dropped events in cases where it was non-zero.
- I clarified the thread list in the sequence point block doesn't have to contain all (or any) of the threads. Its an optional mechanism just for dropped event detection.

cincuranet (approved):

PR #2172, opened 2025-02-25
Fix Stack Handling for NetTrace V4

brianrob:
When processing each event in an EventBlock, TraceEvent reads each event header and then uses the header to setup and dispatch the event.  To hook-up stacks, the system uses the StackId emitted as part of the header.  When converting from StackId to the ExtendedData representation, there is a bug where the StackSizeInBytes and StackBytes fields don't get reset when StackId == 0.

Fix this issue by resetting the StackSizeInBytes and StackBytes on every event to ensure that values from a previous event are not re-used.

cc: @beaubelgrave


noahfalk (approved):
👍

Issue #2158, opened 2025-01-31
Viewing Events with multi-line event payload is not ideal


rauhs:
Hi all,

We have accumulated more and more of our custom ETW events over the year (they're great!). For us the payload of the events can be quite large. We often dump multi-line strings as event payloads.

The are tough to display: The export to Excel/CSV/XML does't show them well

What we have used is "Dump Event": That works okay: The XML attributes break the line on newlines.

However: The XML attribute is XML-Escaped. E.g. here is part of an SQL explain that is one of our events:

```xml
Predicate Information (identified by operation id):
    ---------------------------------------------------
     
       3 - filter(&quot;tbl0&quot;.&quot;USERID&quot; IS NULL AND &quot;tbl0&quot;.&quot;CLIENTID&quot; IS NULL AND &quot;tbl0&quot;.&quot;USERGROUPID&quot; IS NULL AND 
                  &quot;tbl0&quot;.&quot;ORGANIZATIONID&quot; IS NULL OR &quot;tbl0&quot;.&quot;USERID&quot; IS NULL AND &quot;tbl0&quot;.&quot;USERGROUPID&quot;=11 OR &quot;tbl0&quot;.&quot;USERID&quot; IS 
                  NULL AND &quot;tbl0&quot;.&quot;USERGROUPID&quot; IS NULL AND &quot;tbl0&quot;.&quot;CLIENTID&quot;=1 AND &quot;tbl0&quot;.&quot;ORGANIZATIONID&quot; IS NULL OR 
                  &quot;tbl0&quot;.&quot;USERID&quot;=10021517062 OR &quot;tbl0&quot;.&quot;USERID&quot; IS NULL AND &quot;tbl0&quot;.&quot;ORGANIZATIONID&quot;=10052802872 AND 
                  &quot;tbl0&quot;.&quot;USERGROUPID&quot; IS NULL)
```

This is very tough to read. We have to manually copy that and unescape to just read it.

Ideas:

- If the payload was not an **Attribute** but an XML **Tag**: We could use `CDATA` and avoid the escaping.

Question:

1. Is the "dump event" considered some standard? I imagine we can't change from attributes to tags, right?
2. Any way we could extend this so we change the format? Maybe a second command (`ALT+SHIFT+D`?)?
3. Or just use a User Setting to change the Dump Event format?

Any other ideas? I'd certainly be open to send a pull request.

Thanks for reading!

rauhs:
Thanks for the quick reply!

[...]

PR #2175, opened 2025-03-10
Add multi line viewer for Event Window

paulusakademius:
This will add a multi line viewer to the Event Viewer, which can be turned on and off with F2.

Issue #2158

### DIFF
--- a/src/PerfView/EtwEventSource.cs
+++ b/src/PerfView/EtwEventSource.cs
@@ -554,12 +554,8 @@ namespace PerfView
 
                 m_timeStampRelativeMSec = data.TimeStampRelativeMSec;
                 m_idx = data.EventIndex;
-                m_payloads = new Payload[data.PayloadNames.Length];
-                for (int i = 0; i < m_payloads.Length; i++)
-                {
-
-                    m_payloads[i] = new Payload(data.PayloadNames[i], data.PayloadString(i));
-                }
+                m_payloads = new List<Payload>();
+                
                 // Compute the data column 
                 var restString = new StringBuilder();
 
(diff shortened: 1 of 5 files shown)


paulusakademius:

paulusakademius:
Hi, I am a student and I don't have much experience yet. I hope my code is fine and I would be happy about some feedback.

src/PerfView/EtwEventSource.cs
```diff
@@ -552,7 +554,12 @@ internal ETWEventRecord(ETWEventSource source, TraceEvent data, Dictionary<strin
 
                 m_timeStampRelativeMSec = data.TimeStampRelativeMSec;
                 m_idx = data.EventIndex;
+                m_payloads = new Payload[data.PayloadNames.Length];
```
brianrob, line 557: I think it would be good to change the type of `m_payloads` to a `List<Payload>` so that you can add new items to it after iterating over the fields in the event.  If you look at `ETWEventRecord..ctor`, there are a number of conditional calls to `AddField`.  It would be ideal to add these fields to `m_payloads` as well, so that what's in the single line view is also in the multi-line view.  Bonus points if the order matches.

src/PerfView/EtwEventSource.cs
```diff
@@ -666,7 +673,8 @@ internal ETWEventRecord(ETWEventSource source, TraceEvent data, Dictionary<strin
             public DateTime OriginTimeStamp { get { return TimeZoneInfo.ConvertTime(LocalTimeStamp, this.m_source.OriginTimeZone); } }
             public override string Rest { get { return m_asText; } set { } }
             public EventIndex Index { get { return m_idx; } }
-
+            public override Payload[] Payloads { get { return m_payloads; }  }
```
brianrob, line 676: If you change `m_payloads` to be a `List<Payload>`, you can call `.ToArray()` here if an array is required for binding (not sure).  If not, you can keep it as a `List` and let it be enumerated by the GUI framework.

src/PerfView/EventViewer/EventWindow.xaml
```diff
@@ -17,6 +17,7 @@
     </Window.Resources>
     <DockPanel Background="{DynamicResource BackgroundColour}">
         <DockPanel.CommandBindings>
+            <CommandBinding Command="{x:Static src:EventWindow.ToggleMultiLineViewPaneCommand}" Executed="DoToggleMultiLineViewPane" />
```
brianrob, line 20: I think we need a way to make sure users know how to toggle this view.  Would it be possible to show some text in the window that says something like "Press F2 to hide." whenever there is not any data shown?

brianrob, line 20: Would it be possible to make the new window take up the bottom 25% of the window by default?  I think 50% is likely too much.

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -237,6 +240,33 @@ public void SaveDataToCsvFile(string csvFileName, int maxNonRestFields = int.Max
...
+            set
+            {
+                if (value == m_MultiLineViewPaneHidden)
+                {
+                    return;
+                }
+
+                if (value)
+                {
+                    App.UserConfigData["MultiLineViewPaneHidden"] = "true";
```
brianrob, line 255: This is great.  Thanks for keeping track of the user's preference across runs.

brianrob (commented):
This is looking great!  Some comments and questions below.

brianrob:
@caaavik-msft, would you mind reviewing this and see if you think there might be any changes necessary from an accessibility perspective (or otherwise)?

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -1804,6 +1840,17 @@ private void SelectedCellsChanged(object sender, SelectedCellsChangedEventArgs e
             }
 
             m_maxColumnInSelection = null;
+
+            if (!cells.IsNullOrEmpty())
```
brianrob, line 1844:
```suggestion
            if (cells != null && cells.Count > 0)
```
`IsNullOrEmpty` is fairly heavy for this check because it goes through a LINQ abstraction and results in construction of an enumerator and an attempted enumeration operation, when all we really need to know if is there are any items in the collection  It's preferrable to avoid this sort of call when we can and check a bit more explicitly.

brianrob:
This is looking good.  @caaavik-msft do you want to see any additional changes?

src/PerfView/EventViewer/EventWindow.xaml.cs
paulusakademius: Yes you're right. Just checking for null and any items makes more sense.

caaavik-msft (approved):
Looks good to me

brianrob (approved):
LGTM.  Thanks for contributing this @paulusakademius!

PR #2176, opened 2025-03-18
Further accessibility fixes

caaavik-msft:
This PR resolves some more accessibility bugs with PerfView. 

There were some major changes to the flame graph:
- Fixed a bug where only two of the 100 randomly generated flame graph colors were used, now we have our random colors again
- Updated the color generator to ensure that all colors have sufficient contrast with the text and background
- Added a border around flame boxes to allow for light coloured flame boxes (e.g. yellow) to have sufficient contrast against the background
- Fixed some bugs regarding rendering at different zoom levels. We now re-render every time the zoom level is updated. This might need some additional testing on some large traces to ensure it can handle it, but it seemed to handle fine on my machine with some test traces I made.
- Allowed keyboard focus on the entire flame graph canvas and ensured that it would get announced as a flame graph to screen readers, however there is still no ability to perform keyboard navigation inside the flame graph. 
- The help link for the flame graph is now keyboard navigable, so users with screen readers will be able to use that and understand that the flame graph is an alternate visualization of data that can be accessed using the other windows in the stack viewer. This should be sufficient for us to not have to implement a full automation peer for the flame graph (although that would be nice as a future PR) and comply with accessibility requirements.

New flame graph look with color contrasting and borders for both colour schemes:
image
image

In addition, there were also the following changes:
- Ensured that all help links are keyboard navigable, some were disabled as a tab stop.
- Ensured that all help links have appropriate automation names as some were just a wrapper around a "?"
- Fixed keyboard focusing issues with the Cpu Ctrs box in the Run Command dialog
- Ensured that DataGrid column headers are focusable (and can be navigated to with tabs)
- Changed default text formatting to use "Display" mode, which works better for small text which we have a lot of.
- Changed some of the background colours to be a bit lighter to add more contrast.

There are three more bugs I want to resolve before I merge in this PR, but am making this PR now to get the review process started:
- The border that shows up when hovering over a menu item does not have sufficient contrast with the background
  - It seems the only way to fix this is to clone the >600 line ControlTemplate for `MenuItem` just so that we can change the BorderBrush when highlighted or selected. But want to make sure there is not a cleaner way to do that first.
- Some controls do not handle window resizing or text size 200% well and it leads to text being truncated
- `HistoryComboBox` controls do not announce the currently selected item when using the keyboard.
  - I believe this might be a bug with Windows Narrator, but will confirm with WPF team to be sure and if so will not need to address this in this PR.

### DIFF
--- a/src/PerfView/Dialogs/RunCommandDialog.xaml
+++ b/src/PerfView/Dialogs/RunCommandDialog.xaml
@@ -42,14 +42,14 @@
             </RichTextBox.Document>
         </RichTextBox>
         <!-- CommandToRun section for running specific process -->
-        <TextBlock Grid.Row="1" Grid.Column="0" x:Name="CommandToRunLabel" Margin="5,0" VerticalAlignment="Center" KeyboardNavigation.IsTabStop="False" ToolTip="The command line of the process to run while collecting data.">
+        <TextBlock Grid.Row="1" Grid.Column="0" x:Name="CommandToRunLabel" Margin="5,0" VerticalAlignment="Center" ToolTip="The command line of the process to run while collecting data.">
                 <Hyperlink Command="Help" CommandParameter="CommandToRunTextBox">Command:</Hyperlink>
         </TextBlock>
         <controls:HistoryComboBox Grid.Row="1" Margin="0,5,0,0"  Grid.Column="1" x:Name="CommandToRunTextBox" VerticalAlignment="Center" KeyDown="CommandToRunKeyDown" AutomationProperties.Name="Command to run" />
 
         <!-- FocusProcess section for Collect command-->
         <StackPanel Grid.Row="1" Grid.Column="0" Orientation="Horizontal">
-            <TextBlock Name="FocusProcessLabel" Margin="5,0" VerticalAlignment="Center" KeyboardNavigation.IsTabStop="False" ToolTip="Name or PID of the process to collect user-mode events for.">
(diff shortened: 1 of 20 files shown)


caaavik-msft:
Just pushed some new changes:

The border contrast issue is resolved, however the only option was to duplicate the entire MenuItem control template, I was not able to find a cleaner solution so if anyone has one then that would be preferable. 

I have also fixed the issues around some of the text boxes and dialogs showing as truncated when at 200% text scaling, this was done by replacing some Grids with WrapPanels where there was no need for grid alignment. I also made it so that the RunCommandDialog now displays a scrollbar if the screen contents can't fit, which happens on smaller screen resolutions at 200% text scaling.

This should be all the fixes complete now, so we are good to merge this now after review and with this we will have resolved all Sev 1 and 2 accessibility bugs.

src/PerfView/Dialogs/RunCommandDialog.xaml
```diff
@@ -6,507 +6,494 @@
         Style="{DynamicResource CustomWindowStyle}"
         Background="{StaticResource ControlDarkerBackground}"
         Title="Collecting ETW Data while running a command"
-        Height="304" Width="900" MinHeight="210" MinWidth="900">
+        Height="304" Width="950" MinHeight="210" MinWidth="950">
```
cincuranet, line 9: Is this change od width(s) intentional?

caaavik-msft: Yes, this is being done because I've now made the run command dialog scrollable and the scrollbar needs an extra 50 pixels. Below I compare new (on top), and previous (on bottom)

image

cincuranet (approved):

PR #2180, opened 2025-04-08
Change NetTrace format version support, eliminate pinning, and improve UnsupportedVersionException

noahfalk:
- Re-enabling support for parsing format V6 and re-enabling the tests
- Removed support for ancient netperf format version 1 and 2 that were only used during .NET Core 2.0 (not even 2.1).
- The implementation of the Nettrace reader was doing a bunch of pinning because the TraceEvent internal model of events uses unmanaged pointers, however the pinning was creating bad GC performance. TraceEvent builds for .NET standard so we don't have guaranteed access to the pinned heap and I didn't want to start doing lots of copying and/or managing pools of pinned arrays. Instead I switched the stream reading over to using spans backed by native heap allocations and explicit block lifetime tracking. As a side benefit it also brings the V3-5 block iteration code a bit closer to the V6 block iteration and removed all the dependencies on the existing FastSerialization library.
- Add MinSupportedVersion and MaxSupportedVersion to the UnsupportedVersionException

@wiktork - This may fix the GC issues you were seeing in dotnet-monitor
@karpinsn - Exception improvements should help with version error messages in Dev18

@mdh1418, @brianrob, @lateralusX, @beaubelgrave - aside from turning back on V6 support hopefully the behavior should be nearly identical to before despite all the code churn in there.

### DIFF
--- a/src/TraceEvent/EventPipe/EventPipeEventSource.cs
+++ b/src/TraceEvent/EventPipe/EventPipeEventSource.cs
@@ -1,4 +1,3 @@
-using FastSerialization;
 using Microsoft.Diagnostics.Tracing.EventPipe;
 using Microsoft.Diagnostics.Tracing.Parsers;
 using Microsoft.Diagnostics.Tracing.Parsers.Clr;
@@ -7,7 +6,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Reflection.PortableExecutable;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
@@ -180,31 +178,6 @@ namespace Microsoft.Diagnostics.Tracing
         internal LabelListCache LabelListCache { get; private set; }
 
         internal override string ProcessName(int processID, long timeQPC) => string.Format("Process({0})", processID);
-
-
-#if SUPPORT_V1_V2
-        internal void ReadTraceObjectV1To2(Block block)
-        {
-            SpanReader reader = block.Reader;
-            // The start time is stored as a SystemTime which is a bunch of shorts, convert to DateTime.
-            short year = reader.ReadInt16();
-            short month = reader.ReadInt16();
-            short dayOfWeek = reader.ReadInt16();
-            short day = reader.ReadInt16();
-            short hour = reader.ReadInt16();
-            short minute = reader.ReadInt16();
-            short second = reader.ReadInt16();
-            short milliseconds = reader.ReadInt16();
-            _syncTimeUTC = new DateTime(year, month, day, hour, minute, second, milliseconds, DateTimeKind.Utc);
-            _syncTimeQPC = reader.ReadInt64();
-            _QPCFreq = reader.ReadInt64();
-
-            sessionStartTimeQPC = _syncTimeQPC;
-            _processId = 0; // V1 && V2 tests expect 0 for process Id
-            pointerSize = 8; // V1 EventPipe only supports Linux which is x64 only.
-            numberOfProcessors = 1;
-        }
-#endif
 
         internal void ReadTraceObjectV3To5(Block block)
         {
@@ -768,12 +741,12 @@ namespace Microsoft.Diagnostics.Tracing
 
     internal class V6BlockParser : BlockParser
     {
-
         public V6BlockParser(EventPipeEventSource source, RewindableStream stream) : base(source, stream, 4) { }
 
         protected override BlockHeader ReadBlockHeader()
         {
             int headerInt = _stream.Read<int>();
+            // See NetTraceFormat.md#Blocks for the description of the block header format.
             return new BlockHeader((BlockKind)(headerInt >> 24), headerInt & 0xFFFFFF);
         }
 
@@ -817,9 +790,13 @@ namespace Microsoft.Diagnostics.Tracing
                     break; // ignore unknown block types so that the file format can be extended in the future
             }
         }
-
     }
 
+    /// <summary>
+    /// This parses older versions of the format which used the FastSerialization format. However this implementation
+    /// doesn't use the FastSerialization library to implement it. This gives us more control over how the IO and memory
+    /// management is done and keeps the code structurally more similar to how parsing is done V6.
+    /// </summary>
     internal class FastSerializationObjectParser : BlockParser
     {
         bool _isNetTrace;
@@ -866,7 +843,6 @@ namespace Microsoft.Diagnostics.Tracing
             Span<byte> typeName = stackalloc byte[typeLen];
             _stream.Read(typeName);
             AssertTag(FastSerializationTag.EndObject);
-
 
             int blockSize;
             BlockKind type;
@@ -939,8 +915,6 @@ namespace Microsoft.Diagnostics.Tracing
             switch (block.Kind)
             {
                 case BlockKind.Trace:
-
-                    
                     _source.ReadTraceObjectV3To5(block);
                     break;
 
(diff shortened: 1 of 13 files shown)


src/TraceEvent/EventPipe/RewindableStream.cs
```diff
@@ -0,0 +1,167 @@
...
+    {
+        delegate int StreamReadDelegate(Stream stream, Span<byte> buffer);
+
+        private static readonly StreamReadDelegate s_streamReadDelegate = InitSpanRead();
+
+        private static StreamReadDelegate InitSpanRead()
+        {
+            // Opportunistically use it if reflection shows it is available, otherwise fallback to
+            // the slower Read(byte[], int, int) method + a copy.
+            MethodInfo method = typeof(Stream).GetMethod("Read", new Type[] { typeof(Span<byte>) });
```
brianrob, line 123: It would be good to have a test that ensures that this works properly for platforms that have the `Read(Span<byte>)` method.  TraceEventTests targets and runs on both .NET Framework and .NET 8, so we should be able to cover both.

src/TraceEvent/EventPipe/RewindableStream.cs
```diff
@@ -0,0 +1,167 @@
+﻿using PEFile;
```
brianrob, line 1: I think this is unused?

brianrob (commented):
Thanks @noahfalk for putting this together!  A few comments/questions below.

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -666,351 +657,389 @@ private DynamicTraceEventData CreateTemplate(EventPipeMetadata metadata)
...
-                    case BlockKind.LabelList:
-                        _source.ReadLabelListBlock(blockBytes, blockStartOffset);
-                        break;
-                    case BlockKind.EndOfStream:
-                        _source.EventCache.Flush();
-                        return;
-                    default:
-                        break; // ignore unknown block types so that the file format can be extended in the future
+                    ParseBlock(block);
+                    if (block.Kind == BlockKind.EndOfStream) break;
```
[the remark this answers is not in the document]
noahfalk: The existing behavior, which I intended to be unchanged, is that an exception gets thrown saying the end-of-file was unexpectedly reached. I believe PerfView's --continueOnError option ignores that exception and allows the trace to work. This PR should be throwing the same exception from a slightly different callstack but now that you mention it I realize there probably isn't a test case which verifies that. I'll look into adding one.

I'm certainly open to adjusting that behavior, though if its a breaking change we might want to separate it in its own PR.

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -176,8 +183,9 @@ public override bool Process()
 
 
 #if SUPPORT_V1_V2
```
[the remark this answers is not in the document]
noahfalk: oh yeah, I missed this, thanks. I'll take care of it.

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -666,351 +657,389 @@ private DynamicTraceEventData CreateTemplate(EventPipeMetadata metadata)
...
-            // This is only here for V2 and V1.  V3+ should use the name EventTrace, it can be removed when we drop support.
-            _deserializer.RegisterFactory("Microsoft.DotNet.Runtime.EventPipeFile", delegate { return this; });
-#endif
-            _deserializer.RegisterFactory("Trace", delegate { return this; });
-            _deserializer.RegisterFactory("EventBlock", delegate { return new EventPipeEventBlock(_source); });
-            _deserializer.RegisterFactory("MetadataBlock", delegate { return new EventPipeMetadataBlock(_source); });
-            _deserializer.RegisterFactory("SPBlock", delegate { return new EventPipeSequencePointBlock(_source); });
-            _deserializer.RegisterFactory("StackBlock", delegate { return new EventPipeStackBlock(_source); });
+            int headerInt = _stream.Read<int>();
+            return new BlockHeader((BlockKind)(headerInt >> 24), headerInt & 0xFFFFFF);
```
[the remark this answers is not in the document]
noahfalk: sure thing

src/TraceEvent/EventPipe/EventPipeEventSource.cs
```diff
@@ -666,351 +657,389 @@ private DynamicTraceEventData CreateTemplate(EventPipeMetadata metadata)
...
-        /// An EventPipeSequencePointBlock represents a stream divider that contains
-        /// updates for all thread event sequence numbers, indicates that all queued
-        /// events can be sorted and dispatched, and that all cached events/stacks can
-        /// be flushed.
-        /// </summary>
-        internal class EventPipeSequencePointBlock : EventPipeBlock
+        enum FastSerializationTag : byte
         {
-            public EventPipeSequencePointBlock(EventPipeEventSource source) : base(source) { }
+            NullReference = 1,
```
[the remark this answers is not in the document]
noahfalk: I'll add comments to that effect.

src/TraceEvent/EventPipe/RewindableStream.cs
noahfalk: yeah, it should be unused. I think VS's automatic adding of namespaces must have gotten a little over-zealous and I failed to notice :)

src/TraceEvent/EventPipe/RewindableStream.cs
noahfalk: Sure, I should be able to do something like that.

src/TraceEvent/EventPipe/EventPipeEventSource.cs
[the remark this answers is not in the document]
brianrob: Thanks.  Maintaining the current behavior is great.  For PerfView, yes --continueOnError is the escape hatch.  Just wanted to make sure we aren't inadvertently changing the behavior.

src/TraceEvent/EventPipe/RewindableStream.cs
brianrob: Happens to me all the time.  I'm tempted to do a repo-wide pass.  I'm afraid of how many of these would show up.  :)

noahfalk:
Thanks for the feedback @brianrob! Hopefully the latest commit addresses all of it.

src/TraceEvent/EventPipe/FixedBuffer.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Microsoft.Diagnostics.Tracing.EventPipe
+{
+    internal sealed class FixedBuffer : MemoryManager<byte>
+    {
+        private IntPtr _pointer;
+        private int _length;
+
+        internal FixedBuffer(int length)
+        {
+            _pointer = Marshal.AllocHGlobal(length);
```
wiktork, line 14: Would NativeMemory be more appropriate here?

noahfalk: Unforetunately TraceEvent targets .NET Standard 2.0 and NativeMemory was first supported in .NET 6 so it isn't available.

src/TraceEvent/EventPipe/EventPipeEventSource.cs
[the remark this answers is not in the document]
brianrob: Sounds great.  Thank you.

brianrob (approved):
LGTM.  Thanks @noahfalk!

PR #2181, opened 2025-04-09
Add /OSHeapMaxMB to set a max size for OS heap sessions

zachcmadsen:
The OS heap provider technically supports circular buffering, but it's very sensitive to lost events (dropped or overwritten). So, it doesn't make sense to use circular buffering to limit its size. This PR adds a new flag to set a max file size on the heap session, which enforces the size even if PerfView is killed before the session is stopped.

I tested the flag on my machine by killing PerfView in the middle of a collection and confirming that the heap file stopped at the specified max size.


brianrob (approved):
LGTM.  Thanks @zachcmadsen!

PR #2182, opened 2025-04-09
Implement Nettrace Support for Traces with Universal Providers

brianrob:
Adds support for consuming nettrace files that contain usage of the universal providers.  These are:

1. UniversalSystem - Provides a view of the state of the system (processes, mappings, symbols, etc.)
2. UniversalEvents - Provides sampled based events (e.g. CPU, CSwitch)

cc: @beaubelgrave, @noahfalk, @mdh1418

### DIFF
--- a/src/TraceEvent/EventPipe/UniversalProviders.md
+++ b/src/TraceEvent/EventPipe/UniversalProviders.md
@@ -5,7 +5,7 @@ This document describes the definitions of a set of universal machine-wide traci
 
 ## Format Restrictions
  - Any format that can be converted to a TraceLog (ETLX) by TraceEvent can use these event definitions.  At this time, PerfView will only consume them from nettrace files.
- - Formats must provide a common set of fields with each event.  These fields are used by in addition to the payloads listed in this document to interpret the events.  The fields do not explicitly required data types.  If they don't match what TraceEvent uses for storage, then TraceEvent will need to convert them.  It is recommended to use the types specified in the field descriptions below (if specified).
+ - Formats must provide a common set of fields with each event.  These fields are used by in addition to the payloads listed in this document to interpret the events.  The fields do not explicitly require data types.  If they don't match what TraceEvent uses for storage, then TraceEvent will need to convert them.  It is recommended to use the types specified in the field descriptions below (if specified).
  - Formats must provide the QPC frequency to ensure that times that are stored in the Value fields of the Universal Events Provider can be interpreted.
 
 ### Required Fields
--- a/src/TraceEvent/EventPipe/UniversalProviders.md
+++ b/src/TraceEvent/EventPipe/UniversalProviders.md
@@ -1,0 +1,63 @@
+# Universal Providers Tracing Events
+
+## Overview
+This document describes the definitions of a set of universal machine-wide tracing events that are consumed by TraceEvent and exposed in PerfView.
+
+## Format Restrictions
+Any format that can be converted to a TraceLog (ETLX) by TraceEvent can use these event definitions.  At this time, PerfView will only consume them from nettrace files.
+
+## Common Format Information
+ - All integers are of type varint encoded in ULEB64/7-bit encoded format.
+ - All strings are length-prefixed (16-bit unsigned integer) followed by UTF8 bytes.  Not null-terminated.
+
+## Universal System Provider
+The universal system provider is named "Universal.System". It's purpose is to provide system-level events for reconstructing the state of a machine and its processes during a period of time.
+
+### Event Definitions
+- **Event Name**: ProcessCreate
+    - **Description**: Identifies a newly created processes. The process ID is inferred from the format's process ID field.
+    - **Field: NamespaceId**: varint - The actual process ID on the system within the running namespace.
+    - **Field: Name**: string - The name of the process.
+    - **Field: NamespaceName**: string - Namespace of the process. Can be used to convey a set of related processes.
+
+- **Event Name**: ProcessExit
+    - **Description**: Identifies a process that has exited. The process ID is inferred from the format's process ID field.
+
+- **Event Name**: ProcessMapping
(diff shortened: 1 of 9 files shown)


noahfalk (commented):
This looks good to me. My only suggestion would be to add some documentation to the PerfView repo which describes UniversalProvider so that we don't have to infer the event schema from code. Not sure if we expect any other collector aside from OneCollect to ever produce it?

brianrob:
> This looks good to me. My only suggestion would be to add some documentation to the PerfView repo which describes UniversalProvider so that we don't have to infer the event schema from code. Not sure if we expect any other collector aside from OneCollect to ever produce it?

Yeah, that's a good idea.  I'll do that.  Thanks!

src/TraceEvent/EventPipe/UniversalProviders.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Universal Events Provider
+The universal events provider is named "Universal.Events". It's purpose is to provide events that contain a value and have a callstack. There are no stable event IDs, but there will be a set of stable names.
+
+### Current Stable Event Names
+ - cpu - Represents a CPU sample.
+ - cswitch - Represents a machine-level context switch.
+
+### Event Definition
+ - **Field: Value**: varint - Value of the event.
```
noahfalk, line 58: What does value represent for a cswitch?

I was assuming a typical cswitch event would have a few parameters such as the thread that was previously on the processor core and which processor core is being rescheduled. I take it we aren't planning to include that information?

src/TraceEvent/Parsers/UniversalEventsTraceEventParser.cs
```diff
@@ -0,0 +1,128 @@
...
+            sb.Append("/>");
+            return sb;
+        }
+
+        public override string[] PayloadNames
+        {
+            get
+            {
+                if (payloadNames == null)
+                    payloadNames = new string[] { "Value" };
```
cincuranet, line 105: nit: This could use `??=`.

src/TraceEvent/Parsers/UniversalSystemTraceEventParser.cs
```diff
@@ -0,0 +1,336 @@
...
+            sb.Append("/>");
+            return sb;
+        }
+
+        public override string[] PayloadNames
+        {
+            get
+            {
+                if (payloadNames == null)
+                    payloadNames = new string[] { "NamespaceId", "Name", "NamespaceName" };
```
cincuranet, line 150: nit: This could use `??=`.

src/TraceEvent/Parsers/UniversalSystemTraceEventParser.cs
```diff
@@ -0,0 +1,336 @@
...
+            sb.Append("/>");
+            return sb;
+        }
+
+        public override string[] PayloadNames
+        {
+            get
+            {
+                if (payloadNames == null)
+                    payloadNames = new string[] { "Id", "StartAddress", "EndAddress", "FileOffset", "FileName", "MetadataId" };
```
cincuranet, line 226: nit: This could use `??=`.

src/TraceEvent/Parsers/UniversalSystemTraceEventParser.cs
```diff
@@ -0,0 +1,336 @@
...
+            sb.Append("/>");
+            return sb;
+        }
+
+        public override string[] PayloadNames
+        {
+            get
+            {
+                if (payloadNames == null)
+                    payloadNames = new string[] { "Id", "MappingId", "StartAddress", "EndAddress", "Name" };
```
cincuranet, line 305: nit: This could use `??=`.

cincuranet (dismissed):
LGTM, minus small nits.

src/TraceEvent/EventPipe/UniversalProviders.md
beaubelgrave: We only currently have the thread that took the cswitch, the time, and the stack. Anyone can add more sample types, so these I assume are just the canonical ones we use.

We can have specific events for thread switching details, if we require them.

brianrob: Yup.  I just added this and some additional context to the documentation.

brianrob:
> LGTM, minus small nits.

Thanks.  I think I want to keep these as is since they match the format used by all of the other parsers.

src/TraceEvent/EventPipe/UniversalProviders.md
noahfalk: Cool, just wanted to confirm :)

src/TraceEvent/EventPipe/UniversalProviders.md
```diff
@@ -4,7 +4,15 @@
 This document describes the definitions of a set of universal machine-wide tracing events that are consumed by TraceEvent and exposed in PerfView.
 
 ## Format Restrictions
-Any format that can be converted to a TraceLog (ETLX) by TraceEvent can use these event definitions.  At this time, PerfView will only consume them from nettrace files.
+ - Any format that can be converted to a TraceLog (ETLX) by TraceEvent can use these event definitions.  At this time, PerfView will only consume them from nettrace files.
+ - Formats must provide a common set of fields with each event.  These fields are used by in addition to the payloads listed in this document to interpret the events.  The fields do not explicitly required data types.  If they don't match what TraceEvent uses for storage, then TraceEvent will need to convert them.  It is recommended to use the types specified in the field descriptions below (if specified).
```
cincuranet, line 8: Shouldn't the "fields do not explicitly required" be "fields do not explicitly require" (not "required")?

brianrob: Yes, you're right.  I'll fix that.

cincuranet (approved):

PR #2183, opened 2025-04-17
Implement R2R Symbol Lookup for Linux Traces

brianrob:
- Capture information required for remote symbol information from universal traces and store in the `TraceLog`.
 - Implement support for downloading r2rmap files from symbol servers.
 - Validate that r2rmap file matches the requested file prior to use.
 - Update `TraceCodeAddresses.LookupSymbolsForModule` to lookup r2rmap files in addition to PDBs.

cc: @beaubelgrave, @noahfalk, @mdh1418


mdh1418 (approved):

src/TraceEvent/Symbols/R2RPerfMapSymbolModule.cs
```diff
@@ -0,0 +1,246 @@
...
+                    break;
+            }
+        }
+
+        private void FinalizeSymbols()
+        {
+            _symbols.Sort((a, b) => a.StartAddress.CompareTo(b.StartAddress));
+        }
+
+        private int BinarySearch(uint rva)
```
noahfalk, line 171: Array.BinarySearch() instead?

noahfalk (approved):
Looks good to me (albeit I'm unfamiliar with this part of PerfView and I'd be unlikely to spot any non-obvious issue)

src/TraceEvent/Symbols/R2RPerfMapSymbolModule.cs
brianrob: Thanks @noahfalk!  I looked into this, but it looks like you have to give it an element that matches the type of the array to compare to, so it's a bit unwieldy because we'd have to construct symbol a partial symbol object with the address of interest and then know how to compare it.  Ultimately, this seems a bit cleaner, but if I'm misunderstanding how Array.BinarySearch works, please let me know as it is exposed on List<T>, which is the type being used here.

noahfalk: Yeah, I took a closer look at the interface Array.BinarySearch() provides and I agree its not an exact fit. You could make it work as you said with a partially constructed element + an IComparable implementation + some extra logic to deal with non-exact matches and range overlap checks but at that point its around the same complexity as what you already have.

PR #2186, opened 2025-04-24
Fix NetTrace parsing for Start/Stop event names

noahfalk:
The work to add V6 NetTrace parsing regressed the special name handling for names that end with "Start" or "Stop". Prior to this fix the suffix was correctly removed when the opcode in the trace was 0, but it was not removed if the opcode reported in the file was Start or Stop.

The fixed code removes the "Start" and "Stop" suffixes when the opcode is either 0 or when it is already set to start/stop respectively. An old V4 nettrace produced by the runtime would have opcode = 0 whereas a newer V5 nettrace from the runtime would have the opcode set.

Ideally the runtime would be doing this event name transform on its own and the parser would not be in the business of making these changes but for backwards compat we must.

@wiktork - this PR looks like it should fix the issue you were hitting
@brianrob @mdh1418 - PTAL, thanks!

### DIFF
--- a/src/TraceEvent/EventPipe/EventPipeMetadata.cs
+++ b/src/TraceEvent/EventPipe/EventPipeMetadata.cs
@@ -771,6 +771,8 @@ namespace Microsoft.Diagnostics.Tracing
             {
                 ExtractImpliedOpcode();
             }
+            
+            StripStartStopInEventName();
         }
 
         public static Guid GetProviderGuidFromProviderName(string name)
@@ -875,7 +877,7 @@ namespace Microsoft.Diagnostics.Tracing
 
         /// <summary>
         /// If the event doesn't have an explicit Opcode and the event name ends with "Start" or "Stop",
-        /// then we strip the "Start" or "Stop" from the event name and set the Opcode accordingly.
+        /// then we set the Opcode to Start or Stop respectively.
         /// </summary>
         private void ExtractImpliedOpcode()
         {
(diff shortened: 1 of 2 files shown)


src/TraceEvent/EventPipe/EventPipeMetadata.cs
```diff
@@ -884,11 +886,28 @@ private void ExtractImpliedOpcode()
...
+
+        /// <summary>
+        /// If the event has a Stop/Start opcode and also ends with the word "Start" or "Stop", then
+        /// remove the "Start" or "Stop" from the event name.
+        /// </summary>
+        private void StripStartStopInEventName()
+        {
+            if (EventName != null)
+            {
+                if (Opcode == (int)TraceEventOpcode.Start && EventName.EndsWith("Start", StringComparison.OrdinalIgnoreCase))
```
wiktork, line 905: Do we know that ActivityJson/Start and Activity1Start and Activity2Start also use the TraceEventOpcode.Start? (and respective stops use TraceEventOpcode.Stop)?

wiktork: They will all end with Start and Stop so should work correctly.

wiktork (approved):

mdh1418 (approved):
Looks good to me

src/TraceEvent/EventPipe/EventPipeMetadata.cs
noahfalk: EventSource internally attempts to set the Start/Stop opcode for any event that has "Start" or "Stop" as a name suffix and didn't explicitly set the opcode to something else. I'd expect the events you mentioned to all have the Start/Stop opcode set when emitted from a recent runtime version. I think old runtimes like .NET 3 and 5 didn't include the opcode when the data was serialized (the nettrace V4 format) but the parser should still handle it correctly regardless. The unit tests cover both when the opcode is set explicitly and when it is left at the default value.

cincuranet (approved):

PR #2192, opened 2025-05-06
Switch to pipeline artifacts

rainersigwald:
Per https://learn.microsoft.com/azure/devops/pipelines/tasks/reference/publish-build-artifacts-v1 this is recommended and faster.

### DIFF
--- a/.ado.yml
+++ b/.ado.yml
@@ -18,11 +18,12 @@ jobs:
       parameters:
         flavor: 'Debug'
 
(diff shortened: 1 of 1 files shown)


rainersigwald:
1:19/0:41 for this, 4:32/4:28 before.

brianrob:
Thanks for this one @rainersigwald.  Would you mind taking a look - it looks like the GitHub UI doesn't like something about this PR:

image

PR #2193, opened 2025-05-06
Try parallel VSTest in PRs

rainersigwald:


rainersigwald:
Looks like tests fail; never mind.

PR #2198, opened 2025-05-07
Switch to pipeline artifacts

rainersigwald:
Per https://learn.microsoft.com/azure/devops/pipelines/tasks/reference/publish-build-artifacts-v1 this is recommended and faster.

(#2192 from a "clean" branch name)


brianrob (approved):
LGTM. Thanks!

PR #2200, opened 2025-05-14
Update NetTrace parameter parsing

noahfalk:
BREAKING CHANGE: NetTrace V5 files that use the Array parameter type with character elements were previously marshalled into a .NET char[], but now TraceEvent converts them to a .NET System.String instead.

- Updated support for parameter types in NetTrace V6
  - Added new parameter types FixedLengthArray, DataLoc, RelLoc, and UTF8CodeUnit
  - Parameter type UTF16Char was renamed UTF16CodeUnit in docs and code for clarity (chars are variable size in UTF16, code units are always 2 bytes). Behaviorly there is no change from before.
  - LengthPrefixedUTF8/16String were removed because they were duplicative with Array\<UTF16CodeUnit> and Array\<UTF8CodeUnit>. It didn't seem worthwhile to have a bunch of different string types when we could just interpret any of the 4 array variants with UTF8/16 elements as strings.

In order to add the new array types I needed to increase the functionality of the underlying PayloadFetch that TraceEvent uses to parse the parameters. PayloadFetch is serializable in the Etlx format so I wanted to added some tests that could round-trip the payload fetch through that serialization. That in turn led me to refactor TraceLog so that I could do it in-memory. I also found a variety of Debug asserts weren't true for the test data so I removed or relaxed them.

PTAL @brianrob 
Fyi @beaubelgrave @mdh1418 @lateralusX @wiktork

Although most of the work here only impacts the new V6 NetTrace format heads up about the (hopefully minor) breaking change for V5. I could do more to make the change optional or preserve the existing behavior, but it just feels awkward to have character arrays marshalled as .NET char[] rather than string. In any case, if you are concerned about it do let me know.


src/TraceEvent/TraceLog.cs
```diff
@@ -2163,7 +2179,6 @@ private unsafe void CopyRawEvents(TraceEventDispatcher rawEvents, IStreamWriter
             // While scanning over the stream, copy all data to the file.
             rawEvents.AllEvents += delegate (TraceEvent data)
             {
-                Debug.Assert(_syncTimeQPC != 0);         // We should have set this in the Header event (or on session start if it is read time
```
brianrob, line 2166: Do you know what changed `_syncTimeQPC` to be 0?  I see this in a few places in this change, but I don't see what actually changed it.

brianrob (commented):
Overall LGTM.  A few questions/comments.

src/TraceEvent/TraceLog.cs
noahfalk: Oops, that was the assert above. This is the line that changed the test data: https://github.com/microsoft/perfview/pull/2200/files#diff-c94b1f053e78f716a557a5876766fc48558133ec18fe0fa752e5cf40c23a6fcfR2850

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -1155,52 +1199,45 @@ internal int OffsetOfNextField(ref PayloadFetch payloadFetch, int offset, int pa
...
+                return offset + elemSize;
+            }
+            else if (size == VARINT)
+            {
+                return SkipVarInt(offset);
             }
             else
             {
-                return offset + size;
+                return ushort.MaxValue;     // Something sure to fail
```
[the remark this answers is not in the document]
noahfalk: Its fine by me. As-is the code is trying to preserve the same error behavior we had executing line 1198 before, but I'm happy to change it if that wasn't important. There were already other exceptions being thrown from this method so its odd that some errors are exceptional and others aren't.

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -1384,22 +1421,118 @@ public PayloadFetch(ushort offset, RegisteredTraceEventParser.TdhInputType inTyp
...
+                for (int i = 0; i < fields.FieldFetches.Length; i++)
+                {
+                    PayloadFetch field = fields.FieldFetches[i];
+                    if (field.IsFixedSize)
+                    {
+                        size += field.Size;
+                    }
+                    else
+                    {
+                        size = DynamicTraceEventData.UNKNOWN_SIZE; // We don't know the size of the field.
```
noahfalk, line 1526:
```suggestion
                        size = DynamicTraceEventData.UNKNOWN_SIZE; // We don't know the size of the field.
                        break;
```

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -1155,52 +1199,45 @@ internal int OffsetOfNextField(ref PayloadFetch payloadFetch, int offset, int pa
...
+                return offset + elemSize;
+            }
+            else if (size == VARINT)
+            {
+                return SkipVarInt(offset);
             }
             else
             {
-                return offset + size;
+                return ushort.MaxValue;     // Something sure to fail
```
noahfalk, line 1240:
```suggestion
                // We should never reach this point unless there was a bug initializing the
                // PayloadFetch object.
                Debug.Fail("Unexpected PayloadFetch size");
                throw new InvalidOperationException("Unexpected PayloadFetch size");
```

noahfalk:
@brianrob - I think I've responded to all your feedback. Take a look when you get the chance and let me know if you think we should make more changes. Thanks!

src/TraceEvent/TraceLog.cs
brianrob: Gotcha, thanks.

brianrob (approved):

PR #2201, opened 2025-05-15
Fix tab navigation to hyperlinks inside help ribbons

caaavik-msft:
After the last round of accessibility bug fixes, only one bug was sent back as unresolved, so this should fix that last accessibility bug. Essentially, the hyperlinks inside the help ribbons were not navigable by the keyboard using tab navigation. It seems that this is a bug with RichTextBox or FlowDocument as I was not able to get it to work, so I just converted it to a TextBlock and added a border to match the previous styling.

Comparing the help ribbons before and after the change, the styling is identical.


brianrob (approved):
LGTM.  Thanks @caaavik-msft!

PR #2214, opened 2025-05-28
Implement PredefinedDynamicTraceEventParser For Known Dynamic Events

### LINKED ISSUE #2173: Better Support for Known Tracelogging Events
Right now, all tracelogging (aka self-describing) events can't really have strong typing information. This makes usage of them very hard. I would love to have a way to create strongly typed version of them, similar to what is supported for Manifest events.

brianrob:
Implement a new parser type called `PredefinedDynamicTraceEventParser` that implements a pre-compiled C# parser that targets dynamic events such as TraceLogging events.

The fundamental difference between this new type of parser and the existing manifest-based pre-compiled parser is that the
`PredefinedDynamicTraceEventParser` does not have the same event identity requirements as the manifest-based pre-compiled parser.  Instead of requiring a durable provider ID and event ID, the
`PreddefinedDynamicTraceEventParser` requires a durable provider ID and event name.  This enables users to build parsers for more flexible event providers with the same parsing performance as manifest-based pre-compiled parsers.

Fixes #2173


src/TraceEvent/PredefinedDynamicTraceEventParser.cs
```diff
@@ -0,0 +1,203 @@
...
+        public void AddTemplate(Guid providerGuid, string eventName, PredefinedDynamicEvent template)
+        {
+            var key = new TemplateKey(providerGuid, eventName);
+            templates[key] = template;
+        }
+
+        /// <summary>
+        /// Get a template from the dictionary
+        /// </summary>
+        public PredefinedDynamicEvent GetTemplate(Guid providerGuid, string eventName)
```
cincuranet: Would it make sense to switch this to `bool TryGetTemplate(..., out PredefinedDynamicEvent template)`? This would clearly communicate that the `template` could be `null` and the `OnUnhandledEvent` would have bit simpler code.

brianrob: Yeah, sounds good to me.  Just pushed the change.

cincuranet (approved):

PR #2242, opened 2025-06-17
Add support for native AOT gcdumps

MichalStrehovsky:
With this, PerfView will be able to create native AOT gcdumps. It activates the same codepaths we had for Project N, so it will also properly decode type names.

The changes to DotNetHeapDumpGraphReader match https://github.com/dotnet/diagnostics/pull/5506.

### DIFF
--- a/src/EtwHeapDump/DotNetHeapDumpGraphReader.cs
+++ b/src/EtwHeapDump/DotNetHeapDumpGraphReader.cs
@@ -114,12 +114,29 @@ public class DotNetHeapDumpGraphReader
                 return;
             }
 
-            if (!m_moduleID2Name.ContainsKey((Address)data.ModuleID))
+            if ((data.ModuleFlags & ModuleFlags.Native) != 0)
             {
-                m_moduleID2Name[(Address)data.ModuleID] = data.ModuleILPath;
-            }
+                if (!m_modules.ContainsKey((Address)data.ModuleID))
+                {
+                    Module module = new Module((Address)data.ModuleID);
+                    module.Path = data.ModuleNativePath;
+                    module.PdbGuid = data.NativePdbSignature;
+                    module.PdbAge = data.NativePdbAge;
+                    module.PdbName = data.NativePdbBuildPath;
+                    m_modules[module.ImageBase] = module;
+                }
 
-            m_log.WriteLine("Found Module {0} ID 0x{1:x}", data.ModuleILFileName, (Address)data.ModuleID);
+                m_log.WriteLine("Found Native Module {0} ID 0x{1:x}", data.ModuleNativePath, data.ModuleID);
+            }
+            else
+            {
+                if (!m_moduleID2Name.ContainsKey((Address)data.ModuleID))
+                {
+                    m_moduleID2Name[(Address)data.ModuleID] = data.ModuleILPath;
+                }
+
+                m_log.WriteLine("Found Module {0} ID 0x{1:x}", data.ModuleILFileName, (Address)data.ModuleID);
+            }
         };
         source.Clr.AddCallbackForEvents<ModuleLoadUnloadTraceData>(moduleCallback); // Get module events for clr provider
         // TODO should not be needed if we use CAPTURE_STATE when collecting.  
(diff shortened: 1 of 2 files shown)


src/EtwHeapDump/DotNetHeapDumpGraphReader.cs
```diff
@@ -540,12 +557,11 @@ internal unsafe void ConvertHeapDataToGraph()
...
                 var typeName = typeData.TypeName;
-                if (IsProjectN)
+                if ((typeData.Flags & TypeFlags.ModuleBaseAddress) != 0)
                 {
-                    // For project N we only log the type ID and module base address.  
+                    // For native modules we only log the type ID and module base address.
                     Debug.Assert(typeName.Length == 0);
-                    Debug.Assert((typeData.Flags & TypeFlags.ModuleBaseAddress) != 0);
-                    var moduleBaseAddress = typeData.TypeID - (ulong)typeData.TypeNameID;   // Tricky way of getting the image base. 
+                    ulong moduleBaseAddress = typeData.ModuleID;
```
brianrob, line 564: What you've done here makes sense to me, but I am not sure why we didn't just use `ModuleID` for ProjectN.  Do you know if this will work for ProjectN as well?  Perhaps this was just an oversight?

src/EtwHeapDump/DotNetHeapDumpGraphReader.cs
```diff
@@ -540,12 +557,11 @@ internal unsafe void ConvertHeapDataToGraph()
             {
                 GCBulkTypeValues typeData = data.Values(i);
                 var typeName = typeData.TypeName;
-                if (IsProjectN)
```
brianrob, line 543: I see that you're removing usages of `IsProjectN` where it makes sense.  Canyou confirm that the instance on line 382 should remain?  If not, I think we are safe to remove `IsProjectN` entirely.

src/HeapDump/GCHeapDumper.cs
```diff
@@ -171,6 +171,12 @@ private CollectionMetadata CaptureLiveHeapDump(int processID)
 
         m_log.WriteLine("Process Has DotNet: {0} Has JScript: {1} Has ClrDll: {2} HasMrt {3} HasCoreClr {4}", hasDotNet, hasJScript, hasClrDll, hasMrt, hasCoreClr);
 
+        if (!hasDotNet && !hasJScript && !hasMrt && !hasCoreClr)
+        {
+            m_log.WriteLine("No supported runtime type detected, going to assume native AOT.");
```
brianrob, line 176: Is there another way to detect that we're dealing with Native AOT?  I suppose that trying this as a last resort is OK if necessary, since we can currently detect the other cases.

brianrob (commented):
Thanks for implementing this @MichalStrehovsky.  A few questions.

src/EtwHeapDump/DotNetHeapDumpGraphReader.cs
MichalStrehovsky: It will work for Project N too. Project N started populating this with the OS module handle as part of TFS changeset 1217253 ("ETW changes to unblock Project N GC Heap Dump feature inside PerfView."). I assume PerfView didn't work before that change, so the first builds that have this changeset should have the correct thing in the ModuleID.

src/EtwHeapDump/DotNetHeapDumpGraphReader.cs
MichalStrehovsky: Yep, I believe that should remain, see https://github.com/dotnet/diagnostics/pull/5506#discussion_r2155469148.

src/HeapDump/GCHeapDumper.cs
MichalStrehovsky: We cannot detect it based on loaded modules.

The only externally visible marker we have is `DotnetRuntimeDebugHeader` exported symbol that is used by the debugger extension. I wasn't sure if we wanted to go as far as reading process memory and parsing the exports directory.

brianrob: Gotcha, thanks.  I'd say we can defer on that for now.  If we end up with another fully native scenario, we can cross that bridge then.

brianrob (approved):

PR #2251, opened 2025-07-05
Make TraceEvent and FastSerialization AOT-compatible

agocke:
Most cases were fairly simple, but a feature switch was needed for deserialization from a stream. I've audited all the call sites and it doesn't look like there are many types (if any) currently supported for event tracing that would be affected by disabling the feature switch.


agocke:
cc @hoyosjs @tommcdon I think this is needed for making dotnet-gcdump AOT-compatible

brianrob:
I've had a chance to look through this PR.  I'd like to have a conversation about what the goal is here and have some follow-up questions on several of the changes.  My initial thought is that this change brings in several things that I don't want to maintain and/or I don't believe will work for TraceEvent all-up.  To preview a few things:

- I really do not want to add a new TFM for these libraries.  We used to target multiple TFMs and this was much more complex for a variety of reasons.
- I will need help understanding the requirement for PolySharp and how this impacts runtime execution.  PerfView embeds its own dependencies, and so we'd need to understand the impact here.
- Primitive-only deserialization will not work - This will break arbitrary trace files.

I will likely have more questions if we can resolve these issues, but these are the big rocks.

src/TraceEvent/Ctf/CtfChannel.cs
```diff
@@ -162,23 +163,6 @@ private void ReadHeader()
             int bytes = _ctfStream.EventHeader.GetSize();
         }
 
-        private bool ReadStruct<T>(out T result) where T : struct
```
agocke, line 165: This method appears unused

src/TraceEvent/Symbols/NativeSymbolModule.cs
```diff
@@ -561,10 +562,8 @@ protected override string GetSourceFromSrcServer()
                     if (!File.Exists(target) && fetchCmdStr != null)
                     {
                         _log.WriteLine("Trying to generate the file {0}.", target);
-                        var toolsDir = Path.GetDirectoryName(typeof(SourceFile).GetTypeInfo().Assembly.ManifestModule.FullyQualifiedName);
-                        var archToolsDir = Path.Combine(toolsDir, NativeDlls.ProcessArchitectureDirectory);
```
agocke, line 565: This variable looks unused

src/TraceEvent/TraceEvent.csproj
```diff
@@ -55,6 +60,10 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles" Version="$(MicrosoftDiagnosticsTracingTraceEventSupportFilesVersion)" />
     <PackageReference Include="Microsoft.Win32.Registry" Version="$(MicrosoftWin32RegistryVersion)" />
+    <PackageReference Include="PolySharp" Version="1.14.1">
```
agocke: This package polyfills attributes for netstandard2.0

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -2007,10 +2061,15 @@ public ProviderManifest(Stream manifestStream, int manifestLen = int.MaxValue)
             id = "Stream";
             int len = Math.Min((int)(manifestStream.Length - manifestStream.Position), manifestLen);
             serializedManifest = new byte[len];
+#if NET
+            manifestStream.ReadExactly(serializedManifest, 0, len);
+#else
             manifestStream.Read(serializedManifest, 0, len);
```
agocke: This is almost certainly a bug

agocke:
> I'd like to have a conversation about what the goal is here 

The main goal is to support Native AOT for a number of diagnostic tools that use this library. Most choices are downstream of that goal:

> I really do not want to add a new TFM for these libraries. 

There isn't really a choice here. netstandard2.0 isn't annotated for trimming/AOT, so in order to see the incompatibilities you have to multi-target.

> I will need help understanding the requirement for PolySharp and how this impacts runtime execution. PerfView embeds its own dependencies, and so we'd need to understand the impact here.

Polysharp just adds attributes so we can avoid ifdef'ing every usage of `RequiresUnreferencedCode` or `SupportedOS`. It shouldn't have any runtime impact.

> Primitive-only deserialization will not work - This will break arbitrary trace files.

This is the key part of a "feature switch" -- it changes behavior when intent to trim (via PublishAot/PublishTrimmed/IsTrimmable/IsAotCompatible) is enabled. This will produce different behavior when trimming, but the old behavior is categorically impossible to support when trimming, so it's kind of the best middle-ground on offer. The alternative is to simply disallow use of the API entirely, which is problematic because it is pervasive in this library.

brianrob:
> The main goal is to support Native AOT for a number of diagnostic tools that use this library. Most choices are downstream of that goal:

I think that we should get together to discuss the goals and priority here.  As it stands, I am not inclined to take this change - this puts a burden on TraceEvent and its maintainers that isn't free and creates another row in the test matrix (AOT).  I think we need to understand the business need here.

agocke:
@tommcdon @steveisok @noahfalk  can you provide some guidance here? My understanding was that it’s a long-term goal to move the diagnostics tools to aot to eliminate the dependency on a shared runtime in diagnostics collection scenarios. Should trace event be part of that scenario, or is removing the dependency an option?

noahfalk:
> @tommcdon @steveisok @noahfalk can you provide some guidance here? My understanding was that it’s a long-term goal to move the diagnostics tools to aot to eliminate the dependency on a shared runtime in diagnostics collection scenarios. Should trace event be part of that scenario, or is removing the dependency an option?

I'd agree its a long-term goal. So far it hasn't been high enough priority that we've put any effort into it and I wasn't thinking of it as pressing. Is there anything creating urgency or could we just park this and revisit when converting the tools is a higher priority?

In terms of the dependency I think we have a mixture of:
- some tools don't use TraceEvent at all
- some tool features need potentially constrained subsets like NetTrace parsing and gcdump formatting
- some tool features like 'dotnet-trace report topN' probably have much broader code usage within TraceEvent. We could potentially cut or modify features like this to remove the dependency but that has user impacts and development cost.
It would probably be hard to remove the dependency if we want to keep all the features.

agocke:
I'm skeptical that there's going to be a future time where we have more time and resources to devote to this compared to now.

Can you say a bit more about what you would expect to change in the future? What new information do we need? What kind of thing would change that would shift our decision?

agocke:
One more thing: almost every library in .NET multi-targets. It is a recommended strategy for most of our modern .NET scenarios and will likely continue to be popular far into the future. I don't see a reason why EventTrace would be different in this regard. What makes it special compared to every other .NET library?

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -1710,14 +1712,66 @@ public void ToStream(Serializer serializer)
...
+            internal static class PrimitiveOnlyDeserialization
+            {
+                private const string FeatureSwitchName = "PrimitiveOnlyDeserialization.IsDisabled";
+
+
+                [FeatureSwitchDefinition(FeatureSwitchName)]
+                [FeatureGuard(typeof(RequiresUnreferencedCodeAttribute))]
+#pragma warning disable IL4000 // The feature switch should be make the return value match RequiresUnreferencedCode
+                // We reverse the logic here because FeatureGuards must always return false when the target type is not available.
+                public static bool IsDisabled => AppContext.TryGetSwitch(FeatureSwitchName, out bool disabled) ? disabled : true;
```
agocke: @sbomer Could you make sure I'm doing this right?

noahfalk:
> I'm skeptical that there's going to be a future time where we have more time and resources to devote to this compared to now.

> Can you say a bit more about what you would expect to change in the future? What new information do we need? What kind of thing would change that would shift our decision?

I largely think of it as priority and available devs to work on it. Andy, was it your plan that you were going to convert all the dotnet-* tools to NativeAOT yourself? If yes then we should try to figure this out now. If not and it fell to the diagnostics team to do the conversion then I am guessing it wouldn't happen until either:
- priority of converting the tools went up (more customer dissat about acquisition experience or some broader scenario requires it)
- other work priorities go down/get completed (SFI, engineering, multiple runtime convergence efforts, playing catchup to mandatory runtime features, new diagnostic tools and features, numerous bugs)

Of course if there is something I'm overlooking that makes the conversion more urgent/valuable than I realize do let me know. So far it just wasn't looming particularly large on my radar.

agocke:
> Andy, was it your plan that you were going to convert all the dotnet-* tools to NativeAOT yourself?

I was going to see how much work it would be. I started with dotnet-gcdump. Looks like this was the only blocker -- after I fixed TraceEvent I think dotnet-gcdump just works. I haven't yet looked at the rest of the tools

brianrob:
@noahfalk and I chatted about this offline yesterday.  When I look at this change, I am most worried about it from the perspective of support cost.  That's both in terms of additional development-time and testing complexity (e.g. new TFM) and behavior complexity (e.g. different behavior across different runtimes).

As a path forward, I think that I would like to step back and understand what the various options are for carving out the set of functionality that is capable of being AOT'd.  My read of the current PR (I could be wrong), is that everything would be available for use in NativeAOT, though some behavior would be different.  I'd like to take a different approach where we assert that all APIs available in NativeAOT have consistent behavior with non-AOT usage.  If that's not possible for an API, then by default, I'd like it to not be available for usage in NativeAOT'd apps.  This may or may not produce an issue for the current scenario (diagnostic tools), but I'd much rather have a conversation and find a path forward, knowing that we have API behavior consistency.

I would like to start this exercise by assuming that no APIs are available for NativeAOT, and then let's bring in the minimal set required for the diagnostic tools.  As that's happening, if there are APIs that can't be supported, that's when we have a discussion about how to move forward.  I can imagine potentially building new APIs that are more NativeAOT friendly so that we can be very clear about what works and what doesn't, and we'd be forcing folks that want to build NativeAOT'd apps that use TraceEvent to make changes to their code when an API that they currently use doesn't work with NativeAOT.  I imagine that we have a good amount of experience doing this kind of work, and so I'm hoping that we won't be re-inventing the wheel here.

Regarding adding a new TFM, I'd like to understand what other options exist here if any.  Is the issue that we're not targeting netstandard21, or just that no version of netstandard is sufficient for NativeAOT?  I'm thinking that if we need to add a new TFM, I'm not sure if I prefer to just add the TFM to TraceEvent, or if I'd rather create a new DLL that contains the AOT'able subset of TraceEvent and allow folks to consume that.

I'm sure that we can find a good path forward here so that we end up in the right place.  If we aren't sure how far we're going to get with converting the diagnostic tools to NativeAOT, then it might also be worth sticking with what you have for now, continuing down the path of converting the tools against a private branch of TraceEvent, and if all goes well, then we can come back to how we implement NativeAOT in TraceEvent.  If that's too burdensome, we can certainly start to delve into the answers to these questions now.

src/TraceEvent/TraceEvent.csproj
MagicAndre1981: Why not use latest version [1.15.0](https://www.nuget.org/packages/PolySharp/1.15.0)? Any breaking change that prevents usage?

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -525,7 +526,7 @@ private object GetPayloadValueAt(ref PayloadFetch payloadFetch, int offset, int
                 }
 
                 // TODO this is very inefficient for blitable types. Optimize that.
-                var ret = Array.CreateInstance(elementType, arrayCount);
+                var ret = new object[arrayCount];
```
noahfalk: This changes the dynamic return type of the API and likely breaks consumers. Although the method return is statically typed as object I don't think it is really free to return an arbitrary type.

I believe arrayInfo.Element.Type is a small closed set of potential types so a helper method with a switch case could create this correctly.

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -1710,14 +1712,66 @@ public void ToStream(Serializer serializer)
                     serializer.Write((byte)0);
                 }
             }
+
+#if NET9_0_OR_GREATER
+            /// <summary>
+            /// Feature switch that turns off deserialization of any types that are not primitive types.
+            /// Specifically, the <see cref="PayloadFetch.Type"/> property must either be a primitive type,
+            /// or null.
```
noahfalk: I'm not the expert in this TraceEvent code but I suspect there may not be any code path that attempts to serialize a non-primitive type so this switch is moot in practice. If you can confirm with Brian that this isn't actually needed that might reduce the perception that NativeAOT is changing the existing behavior.

src/TraceEvent/TraceEvent.csproj
agocke: Probably works fine? I just did `nuget add package` and it did this

src/TraceEvent/DynamicTraceEventParser.cs
agocke: Yeah, I looked through all the code and could only find one place where we check the type in a general manner. That case calls Activator.CreateInstance(type) -- and it's only used to check the size.

I made this a feature switch because i wanted to remove all possibility of changing behavior in the existing code paths. If we're comfortable making a change here in all TFMs, that would likely be easier to handle. Do we have tests that I can run for this?

src/TraceEvent/TraceEvent.csproj
MagicAndre1981: wired that this command adds on older version and not the most recent one. Try the update and look what happens.

agocke:
OK, I've refactored this to completely remove the usage of `System.Type`. @noahfalk you were right. There was no dynamic usage of Type in this code and it could be completely replaced with simple switch statements.

agocke:
@noahfalk @brianrob I think this is ready to go. I've narrowed any breaking change to two internal helper methods (GetInt16At and GetByteAt) that now return short and byte, respectively, instead of int. That's needed to avoid annoying reflection.

Even separate from trimming/AOT I think this will be faster and safer due to the removal of reflection.

brianrob:
@agocke, thanks.  I will re-review this.  One thing that I noticed is that this change has a very large number of whitespace-only changes.  Can I ask you to revert these so that it's easier to review this?  Thanks.

AaronRobinsonMSFT:
> @agocke, thanks. I will re-review this. One thing that I noticed is that this change has a very large number of whitespace-only changes. Can I ask you to revert these so that it's easier to review this? Thanks.

github has the option to remove whitespace in the diffs. It is in the upper right hand area and toggles them.

agocke:
VS Code also has a setting if you're reviewing locally: `"diffEditor.ignoreTrimWhitespace": true`

hoyosjs:
For some reason, GH fails to ignore the whitespace changes properly. This LGTM other than the breaking change GetByteAt/GetInt16At for object inheriting from `TraceEvent`. This feels pretty small though. Also, for polysharp - can we reduce the list of includeassets to build and analyzer? That way it's clear that it's all sourcegen and you don't need to bundle more assets.

src/TraceEvent/TraceEvent.csproj
```diff
@@ -70,6 +74,10 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent.SupportFiles" />
     <PackageReference Include="Microsoft.Win32.Registry" />
+    <PackageReference Include="PolySharp">
```
brianrob, line 77: Do we use `PolySharp` elsewhere?  It looks like it's MIT licensed - just want to make sure this doesn't require a review.

src/TraceEvent/TraceEvent.csproj
```diff
@@ -70,6 +74,10 @@
...
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
```
brianrob, line 79: Agreed with @hoyosjs - I'd like to make sure we're clear that this is only used for source generation at build time.  Do we need to include all of these assets?

src/FastSerialization/FastSerialization.cs
```diff
@@ -1678,20 +1679,20 @@ public void RegisterFactory(string typeName, Func<IFastSerializable> factory)
...
         /// <summary>
         /// Registers a type that can be created by instantiating the parameterless constructor.
-        /// 
+        ///
         /// When the <code>Deserializer</code>encounters a serialized type, it will look for a registered factory or type registration
         /// so that it knows how to construct an empty instance of the type that can be filled.  All non-primitive types
         /// must either be registered by calling RegisterFactory or RegisterType.
         /// </summary>
-        public void RegisterType(Type type)
+        public void RegisterType([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] Type type)
```
brianrob, line 1687: Is there some documentation I can read about how I should think about maintaining a library that supports NativeAOT?  For example, I see these annotations, but I don't know how to approach decorating code with them, or even the full set of attributes that I should know about and use.  What happens if I don't add them?

src/FastSerialization/FastSerialization.csproj
```diff
@@ -14,6 +14,8 @@
     <Version>$(FastSerializationVersion)</Version>
     <FileVersion>$(FastSerializationVersion)</FileVersion>
     <InformationalVersion>$(FastSerializationVersion)</InformationalVersion>
+    <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</IsAotCompatible>
```
brianrob: If we're targeting net8.0, should this be net8.0 as well?

src/TraceEvent/Ctf/CtfChannel.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
```
brianrob: Looks like this is unused.

src/TraceEvent/EventPipe/EventPipeMetadata.cs
```diff
@@ -462,112 +463,112 @@ private DynamicTraceEventData.PayloadFetch ParseType(
             {
                 case EventPipeTypeCode.Boolean32:
                     {
-                        payloadFetch.Type = typeof(bool);
+                        payloadFetch.Type = FetchType.System_Boolean;
```
brianrob, line 466: Can you please explain why this change is necessary?  I'm not arguing that it isn't necessary - I just don't understand enough about NativeAOT to know and would like to educate myself.

src/TraceEvent/Parsers/UniversalSystemTraceEventParser.cs
```diff
@@ -532,7 +539,39 @@ public override ProcessMappingSymbolMetadata Read(ref Utf8JsonReader reader, Typ
...
+    [JsonSerializable(typeof(ProcessMappingSymbolMetadata))]
+    internal partial class UniversalSystemJsonContext : JsonSerializerContext
+    { }
+}
+
+internal static class SourceGenJson
+{
+    /// <summary>
+    /// Deserialize a JSON string into an object of type T using the provided JsonSerializerOptions.
+    /// Assumes that the options are already configured with the appropriate TypeInfoResolver.
```
brianrob, line 557: Same here - can you describe why this intermediate class is necessary for NativeAOT?

src/TraceEvent/Symbols/NativeSymbolModule.cs
```diff
@@ -9,21 +9,22 @@
 using System.Text;
 using System.Text.RegularExpressions;
 using Microsoft.Diagnostics.Utilities;
+using System.Diagnostics.CodeAnalysis;
```
brianrob: This looks unused.

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -875,10 +880,10 @@ public bool ContainsKey(string key)
             #endregion
         }
 
-        // Return the default value for a given type  
-        private object GetDefaultValueByType(Type type)
+        // Return the default value for a given type
+        private object GetDefaultValueByType([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] Type type)
         {
-            if (type == typeof(string))     // Activator.CreateInstance does not work on strings.  
+            if (type == typeof(string))     // Activator.CreateInstance does not work on strings.
```
brianrob, line 894: Are we not updating this site because we've marked this as having dynamically accessed members and so this code path is already limited to non-NativeAOT?

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -1567,13 +1572,225 @@ internal PayloadFetchClassInfo Class
...
+                        FetchType.System_Byte => System.Convert.ToByte(value),
+                        FetchType.System_UInt16 => System.Convert.ToUInt16(value),
+                        FetchType.System_UInt32 => System.Convert.ToUInt32(value),
+                        FetchType.System_UInt64 => System.Convert.ToUInt64(value),
+                        FetchType.System_Single => System.Convert.ToSingle(value),
+                        FetchType.System_Double => System.Convert.ToDouble(value),
+                        FetchType.System_Decimal => System.Convert.ToDecimal(value),
+                        FetchType.System_DateTime => System.Convert.ToDateTime(value),
+                        FetchType.System_Guid => System.Guid.Parse(System.Convert.ToString(value)),
+                        FetchType.System_IntPtr => new IntPtr(System.Convert.ToInt64(value)),
```
brianrob, line 1767: I believe this may throw on 32-bit: [Doc](https://learn.microsoft.com/en-us/dotnet/api/system.intptr.-ctor?view=net-10.0#system-intptr-ctor(system-int64))

```
OverflowException
In a 32-bit process, value is too large or too small to represent as an [IntPtr](https://learn.microsoft.com/en-us/dotnet/api/system.intptr?view=net-10.0).
```

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -1659,13 +1876,13 @@ public void ToStream(Serializer serializer)
             {
                 serializer.Write((short)Offset);
                 serializer.Write((short)Size);
-                if (Type == null)
+                if (Type is not { } t)
```
brianrob, line 1887: Are these changes required for NativeAOT, or is this just using newer syntax?

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -2015,10 +2233,17 @@ public ProviderManifest(Stream manifestStream, int manifestLen = int.MaxValue)
             id = "Stream";
             int len = Math.Min((int)(manifestStream.Length - manifestStream.Position), manifestLen);
             serializedManifest = new byte[len];
-            manifestStream.Read(serializedManifest, 0, len);
+#if NET
+            manifestStream.ReadExactly(serializedManifest, 0, len);
```
brianrob, line 2245: Is there a reason to bifurcate this path based on the TFM?  I'm wondering if we can just use the else path?

src/TraceEvent/TraceEvent.cs
```diff
@@ -1836,14 +1840,14 @@ protected internal int[] GetInt32ArrayAt(int offset, int size)
         /// <summary>
         /// Returns a byte value that was serialized at 'offset' in the payload bytes
         /// </summary>
-        protected internal int GetByteAt(int offset)
+        protected internal byte GetByteAt(int offset)
```
brianrob, line 1843: This is a breaking change.  Do we need to make this change?

src/TraceEvent/TraceEvent.cs
```diff
@@ -1836,14 +1840,14 @@ protected internal int[] GetInt32ArrayAt(int offset, int size)
...
         {
             return TraceEventRawReaders.ReadByte(DataStart, offset);
         }
         /// <summary>
         /// Returns a short value that was serialized at 'offset' in the payload bytes
         /// </summary>
-        protected internal int GetInt16At(int offset)
+        protected internal short GetInt16At(int offset)
```
brianrob, line 1850: This is a breaking change.  Do we need to make this change?

src/TraceEvent/TraceEvent.csproj
```diff
@@ -25,6 +25,10 @@
 
     <NuspecFile>Microsoft.Diagnostics.Tracing.TraceEvent.nuspec</NuspecFile>
     <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);SetNuspecProperties</GenerateNuspecDependsOn>
+    <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</IsAotCompatible>
```
brianrob: Should this be net8.0 since that's what we're targeting?

src/TraceEvent/WPPTraceEventParser.cs
```diff
@@ -86,10 +88,26 @@ private string GetTmfPathForTaskGuid(Guid taskGuid)
...
             public TypeAndFormat(Type Type, IDictionary<long, string> Map) { this.Type = Type; this.Map = Map; }
+
             public Type Type;
             public IDictionary<long, string> Map;
         }
 
+        /// <summary>
+        /// This is only meant to be used in <see cref="CreateTemplatesForTMFFile"/>
+        /// </summary>
+        private static FetchType TypeToFetchType(Type type) => type switch
```
brianrob, line 99: This is not an area that I've dug into before.  Do we know that this is the full set of types involved here?

agocke:
Great questions!

Here are some useful resources:
- https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trimming-concepts Conceptual doc on how trimming works
- https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/fixing-warnings Doc on fixing trim warnings

For validation, I can create a project that publishes as Native AOT and we can verify that the AOT compiler doesn't produce any warnings. Unfortunately we cannot use the existing unit tests as: 1) the existing unit tests are also likely not AOT-compatible and 2) xunit itself is not AOT-compatible.

I'm working on (2) in spare time, but that's an XXL project. Also, the static analysis is generally very reliable. In many cases we consider it more important than testing because code which is not truly compatible can sometimes work by accident.

src/TraceEvent/DynamicTraceEventParser.cs
agocke: I believe it works, in the sense that this does throw on x86, but only if the actual underlying value is out of bounds. It's not based on the runtime type.

src/TraceEvent/DynamicTraceEventParser.cs
agocke: Other way around -- we're not updating it because the DAM annotation makes it AOT-compatible. DAM annotations are for making things compatible. RequiresUnreferencedCode are for marking things as incompatible.

src/TraceEvent/DynamicTraceEventParser.cs
agocke: That particular line is just using newer syntax to avoid any potential race conditions between field reads.

src/TraceEvent/Parsers/UniversalSystemTraceEventParser.cs
agocke: Reflection-based serialization is not compatible with AOT. We need to use source-generated serialization instead. These intermediate data structures are required by System.Text.Json for source-generated JSON serialization.

src/TraceEvent/WPPTraceEventParser.cs
agocke: This method is only used in one place in the code (below in the file) and it has a big if block of all the types that are supported. I copied support for all those types into this method.

We could make this airtight by using another enum instead of a System.Type variable, or changing TypeAndFormat to take a FetchType instead of a System.Type.

agocke (commented):
I'll respond to the questions about FetchType here.

The main change here (and in most of the PR) is to replace uses of System.Type, a reflection construction, with an enum.

The main problem with reflection constructions and trimming/AOT is that the code referenced via reflection may not be available after trimming, or it may be available in a limited form (e.g., accessing members might throw or return different results).

To guarantee that you have the same behavior between AOT and non-AOT, the best thing to do is avoid reflection where not necessary.

In this code, the use of System.Type in these payloads is not necessary. In fact, it's pretty risky. Using reflection on input opens the door to potential security risks where the application can be tricked into executing code or leaking data. This is the exact situation we ended up in with BinaryFormatter. 

TraceEvent really only wants to deal with a set of 10 or so primitive types, however, so there's no reason to use reflection here. We can instead just use switch statements to identify which type we want.

By getting rid of the reflection (in particular, the call to Type.GetType()), we can make the code AOT-compatible, improve performance, and get rid of potential security risks.

src/TraceEvent/TraceEvent.cs
agocke: I can avoid it by introducing more reflection in the deserialization routine to convert ints to bytes when deserializing byte arrays. This will make deserialization slower and more complex. I'll leave the decision up to you.

roee88:
We're also interested in this functionality (thanks @agocke!). @brianrob are there any opens here?

agocke:
@brianrob ping

brianrob:
Thanks for pinging.  I have this one on my list to come back to.

agocke:
@brianrob what can i do to move this along?

brianrob:
Apologies, this one requires me to think through some of the long-term shipping questions and to really understand more about NativeAOT.  I've been having trouble getting this to fit into my schedule - let me give this another shot.  Thank you for the reminder.

agocke:
Happy to walk through the code with you in a call if that would help

agocke:
@brianrob This has been stalled for quite a long time. Should we start a conversation about moving TraceEvent out of the perfview repo? To reduce bottlenecks?

cc @steveisok

brianrob:
Yes, this is one of those things that just keeps getting preempted.  It's definitely on my list of todos, but unfortunately, the list keeps growing and this one keeps getting pushed down. Last I spoke with @noahfalk on this, it sounded like there wasn't a strong push towards NativeAOT for our diagnostic apps, and so this just hasn't made it to the top.  Has anything around that changed, in which case I might be mis-prioritizing this?

PR #2272, opened 2025-08-04
Use themed background brush for menu and filter

JeremyKuhne:
Without this the menu and filter are white on white in dark mode.

Before:

image

After:

image

image


brianrob:
Thank you @JeremyKuhne!

brianrob (approved):

PR #2274, opened 2025-08-06
Improve rendering and dark mode

JeremyKuhne:
- Update manifest to opt into PMV2 scaling (eliminates blurring on high DPI)
- Tweak dark mode scroll bars to be more visible
- Update to C# 11
- Update most packages to latest (updating Azure packages cause runtime failures)
- Add Packages.props to the solution folder for simplicity
- Hook various XAML files to pick up the theme
- Set the WebView2 control preferred color scheme to match
- Remove outdated Internet Explorer related <meta> tag
- Change html sources to indicate that they support light and dark themes
- Generate html with theme specific code

I generally used two root entries `:root[data-theme="light"]` as it enables doing things like letting you flip the theme with a button (say to print). I didn't put that in this PR as I'm not super clear on the value.

For future reference, here is how you would do that:

```html
  <style>
@media print {
  .no-print {
    display: none !important;
  }
  </style>

<button type="button" class="no-print" onclick="toggleTheme()">🌗 Toggle Theme</button>

  <script>
    // Simple theme toggle script: switches the data-theme attribute on <html>
    function toggleTheme() {
      const root = document.documentElement;
      const newTheme = root.getAttribute('data-theme') === 'light' ? 'dark' : 'light';
      root.setAttribute('data-theme', newTheme);
    }
  </script>
```

I gave up on updating the Azure packages as they cause PerfView to fail at runtime. Would require a bit more time.

### DIFF
--- a/src/PerfView/GuiUtilities/WebBrowser/WebBrowser.xaml.cs
+++ b/src/PerfView/GuiUtilities/WebBrowser/WebBrowser.xaml.cs
@@ -52,9 +52,9 @@ namespace PerfView.GuiUtilities
         /// </summary>
         private void Navigate()
         {
-            if (!_disposed && Source != null && _Browser.CoreWebView2 != null)
+            if (!_disposed && Source?.ToString() is { } source)
             {
-                _Browser.CoreWebView2.Navigate(Source.ToString());
+                Browser?.CoreWebView2.Navigate(source);
             }
         }
 
@@ -89,9 +89,9 @@ namespace PerfView.GuiUtilities
             else
             {
                 // Dispose WebView2 to prevent finalizer crashes
-                if (!_disposed && _Browser != null)
+                if (!_disposed)
                 {
-                    _Browser.Dispose();
+                    Browser?.Dispose();
                     _disposed = true;
                 }
             }
@@ -109,6 +109,7 @@ namespace PerfView.GuiUtilities
 
             var userDataFolder = Path.Combine(SupportFiles.SupportFileDir, "WebView2");
             Directory.CreateDirectory(userDataFolder);
+
             var environmentAwaiter = CoreWebView2Environment
                 .CreateAsync(userDataFolder: userDataFolder)
                 .ConfigureAwait(true)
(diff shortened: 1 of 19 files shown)


JeremyKuhne:
image

JeremyKuhne:
image

JeremyKuhne:
There might be value in looking at using a more current "fluent" style library to theme. There are still a lot of rough edges. :)

src/PerfView/PerfView.csproj
```diff
@@ -72,6 +72,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <None Include="app.manifest"/>
```
brianrob, line 75: Does this manifest need to be next to the exe when it starts?  If so, we'll need to pack this as a resource in `PerfView.exe` so that it gets extracted on first run.

src/PerfView/GuiUtilities/WebBrowser/WebBrowser.xaml.cs
```diff
@@ -52,9 +52,9 @@ private static void OnSourceChanged(DependencyObject d, DependencyPropertyChange
         /// </summary>
         private void Navigate()
         {
-            if (!_disposed && Source != null && _Browser.CoreWebView2 != null)
```
brianrob, line 55: I might be a bit curmudgeonly here, but I had to have copilot explain how this syntax works.  I'm inclined to say that I'd like to keep this as it was for simplicity of understanding.  Is there a strong benefit to this change that I'm not aware of?

src/PerfView/PerfView.csproj
JeremyKuhne: It is already embedded, this was just to see it in the solution explorer.

src/PerfView/GuiUtilities/WebBrowser/WebBrowser.xaml.cs
JeremyKuhne: This puts it in a local so that you don't end up checking for null on one thread then having it get set to null on another before you use it. It is a little weird the first few times you see it, but it is a much safer pattern and becomes second nature pretty quickly.

brianrob: Ok, let's assume I'm being curmudgeony here and leave it in.

brianrob (approved):

PR #2286, opened 2025-08-21
Fix parsing issues and add support for additional events to the Linux perf text file parser

thooot:
This improves parsing support for more complex perf traces including ones where:
- some events have stacks and some do not (results in events on subsequent lines without a blank line between)
- some events have per-event modifiers (e.g. cpu-clock/freq=999/,block:block_rq_issue/call-graph=no/), in this case some linux distros will output the /modifier/ in the perf text trace

It also adds support for the following event types: 
- sched:sched_process_exec
- block:block_rq_issue
- block:block_rq_complete
- sched:sched_wakeup
- irq:irq_handler_entry
- irq:irq_handler_exit


src/TraceEvent/Stacks/Linux/LinuxPerfScriptEventParser.cs
```diff
@@ -432,36 +426,68 @@ private IEnumerable<LinuxEvent> NextEvent(Regex regex, FastStream source)
...
-                    if(eventKind == EventKind.ThreadExit)
+                    // Go back to the event so we can process it
+                    source.RestoreToMark(markedPosition);
+                    if (eventDetails.Length >= SchedulerEvent.Name.Length && eventDetails.Substring(0, SchedulerEvent.Name.Length) == SchedulerEvent.Name)
                     {
-                        source.RestoreToMark(markedPosition);
-                        exit = ReadExit(source);
-                        source.SkipUpTo('\n');
+                        ScheduleSwitch schedSwitch = ReadScheduleSwitch(source);
+                        IEnumerable<Frame> frames = ReadFramesForSample(processCommand, pid, tid, threadTimeFrame, source);
```
brianrob, line 441: No need to change in this PR - I think we should consider if there is a nice pattern we can employ here to ensure that the call to read the stack happens automatically for all event types.

brianrob (commented):
@thooot, thank you very much.  This looks great.

brianrob (approved):

PR #2291, opened 2025-08-29
Only Dispose Non-Null Handles in ETWTraceEventSource

brianrob:
If an ETL file is malformed, it's possible to end up with a null handle in `ETWTraceEventSource.handles`.  Ensure that we only dispose of non-null handles.

### DIFF
--- a/src/TraceEvent/ETWTraceEventSource.cs
+++ b/src/TraceEvent/ETWTraceEventSource.cs
@@ -879,7 +879,10 @@ namespace Microsoft.Diagnostics.Tracing
                 {
                     foreach (TraceEventNativeMethods.SafeTraceHandle handle in handles)
                     {
-                        handle?.Dispose();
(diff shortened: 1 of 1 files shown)


src/TraceEvent/ETWTraceEventSource.cs
```diff
@@ -879,7 +879,7 @@ protected override void Dispose(bool disposing)
                 {
                     foreach (TraceEventNativeMethods.SafeTraceHandle handle in handles)
                     {
-                        handle.Dispose();
+                        handle?.Dispose();
```
mconnew: Should you also check if the handle is valid? Is this a standard SafeHandle? If so, calling Dispose on an invalid handle can throw.

brianrob: Yes, that's a good idea.  Updated.

mconnew (approved):

PR #2318, opened 2025-10-16
Add support for Boolean8 to NetTrace V6.

noahfalk:
EventSource writes booleans differently in manifest mode vs self-describing mode. Manifest mode uses a 4 byte boolean event field which NetTrace has always supported but Self-describing wrote a 1 byte boolean which NetTrace had no metadata encoding for. Adding support for Boolean8 to NetTrace gives us a path forward to properly describe booleans in metadata regardless of mode.


brianrob (approved):
LGTM.  Thanks @noahfalk!

PR #2320, opened 2025-10-16
Implement A Thread Time View for Universal Traces

brianrob:
Traces generated by https://github.com/microsoft/one-collect can provide a comprehensive view of system behavior through CPU and context switch events.

Modify the `ThreadTimeStackComputer` to support aggregation of CPU and blocked time from the `Universal.Events` provider defined in one-collect.

Expose the new view for nettrace files that contain these events.

### DIFF
--- a/src/PerfView/PerfViewData.cs
+++ b/src/PerfView/PerfViewData.cs
@@ -9593,6 +9593,7 @@ namespace PerfView
                     }
                     else if (eventStats.ProviderGuid == UniversalSystemTraceEventParser.ProviderGuid)
                     {
+                        m_hasUniversal = true;
                         hasUniversalSystem = true;
                         m_supportsProcesses = true;
                     }
@@ -9712,7 +9713,6 @@ namespace PerfView
                 m_Children.Add(advanced);
             }
 
-            m_hasUniversal = hasUniversalSystem || hasUniversalCPU || hasUniversalCSwitch;
             return null;
         }
 
--- a/src/TraceEvent/Computers/ThreadTimeComputer.cs
+++ b/src/TraceEvent/Computers/ThreadTimeComputer.cs
@@ -455,17 +455,7 @@ namespace Microsoft.Diagnostics.Tracing
 
         private void InitializeForUniversal()
         {
-            bool initialize = false;
-            foreach (TraceEventCounts counts in m_eventLog.Stats)
-            {
-                if (counts.ProviderGuid == UniversalEventsTraceEventParser.ProviderGuid &&
-                    counts.EventName.Contains("cswitch"))
-                {
-                    initialize = true;
-                }
-            }
-
-            if (initialize)
+            if (m_eventLog.Stats.Contains(UniversalEventsTraceEventParser.ProviderGuid, "cswitch"))
             {
                 // Set has cswitches, since we don't get thread start events, and the computer needs to know
                 // that there will be cswitch events.
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -4735,6 +4735,19 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
             return sb.ToString();
         }
 
+        internal bool Contains(Guid providerId, string eventNameSearchVal)
+        {
+            foreach (var counts in m_counts.Values)
+            {
+                if (counts.ProviderGuid == providerId &&
+                    counts.EventName.Contains(eventNameSearchVal))
+                {
+                    return true;
+                }
+            }
+            return false;
+        }
+
         #region private
         /// <summary>
         /// Given an event 'data' look up the statistics for events that type.


src/PerfView/PerfViewData.cs
```diff
@@ -9700,6 +9712,7 @@ protected override Action<Action> OpenImpl(Window parentWindow, StatusBar worker
                 m_Children.Add(advanced);
             }
 
+            m_hasUniversal = hasUniversalSystem || hasUniversalCPU || hasUniversalCSwitch;
```
marklio: [cautious Mark] This looks correct and reasonable. However, it highlights a peril that any future change that adds an "early return" will miss setting this and this method is large enough and this setting is removed enough from the has* sets above that this could easily happen. Consider setting m_hasUniversal to true at the same time as in individual items.

Also, why the heck does this always return null, lol? I thought for sure I was going to find an earlier return when reviewing this.

src/TraceEvent/Computers/ThreadTimeComputer.cs
```diff
@@ -423,6 +453,37 @@ public void GenerateThreadTimeStacks(MutableTraceEventStackSource outputStackSou
             m_threadState = null;
         }
 
+        private void InitializeForUniversal()
+        {
+            bool initialize = false;
```
marklio: CONSIDER: changing the name of initialize to setHasCSwitches.

My personal preference would be to render the foreach below differently using LINQ or an extension method that does the search for cswitch events (if you don't like LINQ) so that the meat of this method is about the initialization and the check is inline in the if clause. That's just the way my eyes want to read this code.

src/TraceEvent/Computers/ThreadTimeComputer.cs
```diff
@@ -648,6 +709,87 @@ private void OnSampledProfile(TraceEvent data)
...
         }
 
+        private void OnUniversalCSwitch(SampleTraceData data)
+        {
+            // This event fires when a thread is switched back in.
+            // The timestamp is the time at which the thread was switched back in.
+            // The value is the amount of time that thread was switched out.
+
+            TraceThread thread = data.Thread();
+            if (thread == null)
```
marklio, line 709: Is this hardening against a plausible scenario? Or is this representative of a "do not crash" principle in this codebase? That is, would a null thread here be indicative of a bug elsewhere that we are making hard to diagnose by simply writing to Debug?

src/TraceEvent/Computers/ThreadTimeComputer.cs
```diff
@@ -648,6 +709,87 @@ private void OnSampledProfile(TraceEvent data)
...
+            {
+                Debug.WriteLine("Warning, no thread at " + data.TimeStampRelativeMSec.ToString("f3"));
+                return;
+            }
+
+            // Get the blocking stack.
+            StackSourceCallStackIndex stackIndex = GetCallStack(data, thread);
+
+            // If the thread wasn't previously marked as blocked, then mark it.
```
marklio, line 718: Similar question to above, would the thread not being marked as blocked indicate a bug? Does "making up" this blocking time represent a corruption of data? Or is this necessary to overcome some other weakness of the system to deliver all the data to render a perfect representation?

src/TraceEvent/Computers/ThreadTimeComputer.cs
```diff
@@ -648,6 +709,87 @@ private void OnSampledProfile(TraceEvent data)
...
+            if (!m_threadState[(int)thread.ThreadIndex].ThreadBlocked)
+            {
+                double startTimeStampRelMsec = m_eventLog.QPCTimeToRelMSec(data.TimeStampQPC - (long)data.Value);
```
marklio, line 721: Below, you use data.TimeStampRelativeMSec. This is calculation different? The answer to this may answer my question above

src/TraceEvent/Computers/ThreadTimeComputer.cs
```diff
@@ -648,6 +709,87 @@ private void OnSampledProfile(TraceEvent data)
...
+            // The value is the amount of time that thread was switched out.
+
+            TraceThread thread = data.Thread();
+            if (thread == null)
+            {
+                Debug.WriteLine("Warning, no thread at " + data.TimeStampRelativeMSec.ToString("f3"));
+                return;
+            }
+
+            // Get the blocking stack.
```
marklio, line 715: MUSING: Is "block" a standard term in this context? It took my brain a bit of time to realize we're simply talking about tracking thread scheduling/preemption and not about locking or thread synchronization. I'm just not used to blocking being used in this context. I guess what I'm asking is, would this code be more clear if different wording were used, or am I just old and pedantic?

src/PerfView/PerfViewData.cs
brianrob: This is a good call.  I will change this to detect if any events from the Universal.System provider are present.  There isn't a case where they shouldn't be present, but the Universal.Events provider is.  And if that's the case the trace is busted and won't work anyway.

Also, the return value is potential chain of delegates to be run after the UI is initialized and on the UI thread.  Here's the relevant comment:

```
        /// <summary>
        /// Implements the open operation.   Executed NOT on the GUI thread.   Typically returns null
        /// which means the open is complete.  If some operation has to be done on the GUI thread afterward
        /// then  action(doAfter) continuation is returned.  This function is given an addition action 
        /// that must be done at the every end.   
        /// </summary>
        protected virtual Action<Action> OpenImpl(Window parentWindow, StatusBar worker)
```

src/TraceEvent/Computers/ThreadTimeComputer.cs
brianrob: Seems reasonable to me.  This pattern is pretty pervasive, but we can certainly simplify it.  I created an internal method that we can try out.

src/TraceEvent/Computers/ThreadTimeComputer.cs
brianrob: For universal traces this would be a bug.  We create all of the threads that we know about when opening the trace for the first time.  It's important to check for null here since we don't want to break the tool if we introduce a bug elsewhere.  That said, we could potentially do something better here.  From a layering perspective, there's not a good log to write to, but that's something that we could fix going forward.  Thoughts?

src/TraceEvent/Computers/ThreadTimeComputer.cs
brianrob: Yes, block is referring to a thread either yielding or being pre-empted.  It might be clearer with a different word, but this one is used pretty heavily throughout the codebase.

src/TraceEvent/Computers/ThreadTimeComputer.cs
brianrob: Good question - this one is necessary to overcome missing events.  Depending on how the trace is written there are various ways that an event (or many) doesn't make it into the trace.  Code like this does its best to clean-up in this scenario without completely trashing the data.  This is fairly common behavior in profilers because you often don't know that you lost an event, and even when you do, you often want to see what you can knowing that the data isn't perfect.

Expect that the user will have already been told about missing events - PerfView will pop-up a dialog on trace open telling the user how many events were reported as lost.

src/TraceEvent/Computers/ThreadTimeComputer.cs
brianrob: TraceEvent does most of its calculations in milliseconds, though the data type is a double so we get fractional milliseconds.  The reason this one is in QPC is because the entity that writes the event is writing it in QPC and not milliseconds.

brianrob:
Thank you for the review @marklio.  Answered your questions and pushed some changes.

src/TraceEvent/Computers/ThreadTimeComputer.cs
marklio: This is a hard call when a tool is so widely used on a diversity of inputs we can't hope to reasonably cover in validation. Do we use Watson to look at crashes? If this were my tool, I'd want to be set up to get the very best diagnostic artifacts. What is the experience of this "log and ignore" approach? Is it _actually_ less annoying than an actionable crash for the type of user that runs the tool? Would be sweet to have the UI be isolated from this processing in a way that allowed actionable diagnostics to flow _and_ provide a good UI experience, but that's a lot of refactoring.

src/TraceEvent/Computers/ThreadTimeComputer.cs
marklio: I'm sure I'm missing something. On line 731 (with the same data), it uses .TimeStampRelativeMSec. And further below on 754, which seems like the same "mark as blocked" scenario, it also uses .TimeStampRelativeMSec. What is different about THIS callsite that it does the calculation from QPC?

src/TraceEvent/Computers/ThreadTimeComputer.cs
brianrob: On Watson - I don't usually look at it too much, but on occasion yes.  I would say that yes, the current behavior is less annoying than the crash.  I usually see this fall into one of two buckets:

1. There is a single or very small number of events that get ignored because the thread for whatever reason didn't get created.
2. We're missing something important in the trace and when you open the view it's effectively empty.

In the first case, the user's analysis is almost always unaffected.  In the second case, it's very obvious something went wrong and the user will re-collect and/or report the issue.

I do agree with you - there is a pattern missing here around how we egress diagnostics.  Some of the very core pieces of TraceEvent have logging capabilities that PerfView just has to turn on.  That pattern likely just needs to be extended here (and frankly to many other places).

src/TraceEvent/Computers/ThreadTimeComputer.cs
marklio: I don't think this should block this particular improvement. I'm more just curious in what our story is for reacting to customer pain with this tool.

PR #2323, opened 2025-10-29
Remove Incorrect Argument Description

brianrob:
A non-circular buffer is not supported in PerfView, but the documentation states that it is.  Remove this from the description.


brianrob:
@microsoft-github-policy-service rerun

brianrob:
@microsoft-github-policy-service rerun

PR #2327, opened 2025-11-06
Special-Case BitMask Parsing

brianrob:
When an event contains a bitmask, it is sometimes described in metadata as an array of pointers.  ETW differentiates between input and output types, and so there is a possibility for metadata to specify the input type as an array of UInt64 and the output type as an array of IntPtr. Existing code can't handle this properly, so we must special case it.

The MSNT_SystemTrace/SystemConfig/ProcessorGroup event is an example of one such event.

### DIFF
--- a/src/TraceEvent/DynamicTraceEventParser.cs
+++ b/src/TraceEvent/DynamicTraceEventParser.cs
@@ -533,9 +533,8 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
                     // input=Array of ulong
                     // output=Array of IntPtr
                     // This is common for bitmasks.  To address this, we special case it here.
-                    if (value is ulong && elementType == typeof(IntPtr))
+                    if (elementType == typeof(IntPtr) && value is ulong uintVal)
                     {
-                        ulong uintVal = (ulong)value;
                         value = new IntPtr(unchecked((long)uintVal));
                     }
                     else if (value.GetType() != elementType)


MagicAndre1981:
ok, the InvalidCastException is gone with the artifact build from this PR:

```
<Event MSec= "42988.4096" PID=  "-1" PName=        "" TID=  "-1" EventName="SystemConfig/ProcessorGroup"
  TimeStamp="11/07/25 10:57:24.957906" ID="Illegal" Version="2" Keywords="0x00000000" TimeStampQPC="2,246,767,051" QPCTime="0.100us"
  Level="Always" ProviderName="MSNT_SystemTrace" ProviderGuid="9e814aad-3204-11d2-9a82-006008a86939" ClassicProvider="True" ProcessorNumber="4"
  Opcode="26" TaskGuid="01853a65-418f-4f36-aefc-dc0f1d2fd235" Channel="0" PointerSize="8"
  CPU="4" EventIndex="9091131" TemplateType="DynamicTraceEventData">
  <PrettyPrint>
    <Event MSec= "42988.4096" PID=  "-1" PName=        "" TID=  "-1" EventName="SystemConfig/ProcessorGroup" ProviderName="MSNT_SystemTrace" Affinity="[255]"/>
  </PrettyPrint>
  <Payload Length="12">
       0:   1  0  0  0 ff  0  0  0 |  0  0  0  0               ........ ....
  </Payload>
</Event>
```

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -529,7 +529,16 @@ private object GetPayloadValueAt(ref PayloadFetch payloadFetch, int offset, int
                 for (int i = 0; i < arrayCount; i++)
                 {
                     object value = GetPayloadValueAt(ref arrayInfo.Element, offset, payloadLength);
-                    if (value.GetType() != elementType)
+                    // Some events have metadata of the form:
+                    // input=Array of ulong
+                    // output=Array of IntPtr
+                    // This is common for bitmasks.  To address this, we special case it here.
+                    if (value is ulong && elementType == typeof(IntPtr))
```
leculver: I'd combine the unboxing operation with pattern matching.

```
                    if (elementType == typeof(IntPtr) && value is ulong uintVal)
                    {
                        value = new IntPtr(unchecked((long)uintVal));
                    }
```

leculver (approved):
Small nitpick, otherwise looks fine to me.

src/TraceEvent/DynamicTraceEventParser.cs
brianrob: Oh yes, I like that a lot better.  Pushing that change.

Issue #2350, opened 2026-01-13
SymbolReader fails to parse SourceLink mappings with wildcard patterns


ivberg:
Description

  SymbolReader.SourceLocationForRva() fails to parse SourceLink JSON mappings that use wildcard patterns (e.g., "path\\*": "url"). This prevents source file retrieval for PDBs that use this common SourceLink format.

  Error Message

  Error: Could not parse SourceLink Mapping:
      "c:\\src\\myproject\\*": "https://dev.azure.com/myorg/myproject/_apis/git/repositories/myrepo/items?path=%2f*&versionDescriptor.versionType=commit&versionDescriptor.version=abc123&download=true&api-version=2.0",
      "external\\sdk\\inc\\header.h": "https://example.com/blobs/ABC123?download=true&filename=header.h"

  Reproduction Steps

  1. Load a PDB that contains SourceLink information with wildcard path mappings
  2. Call SymbolReader.SourceLocationForRva() for an RVA that maps to a source file
  3. Observe the parsing error in the log output

  Expected Behavior

  SourceLink JSON with wildcard patterns should be parsed correctly, allowing the source file URL to be constructed by:
  1. Matching the source path against the wildcard pattern (e.g., c:\src\myproject\*)
  2. Substituting the matched portion into the URL template

  Actual Behavior

  The parser fails with "Could not parse SourceLink Mapping" and source file retrieval fails.

  SourceLink JSON Format

  The SourceLink format supports two types of mappings:

  {
    "documents": {
      "c:\\src\\myproject\\*": "https://raw.githubusercontent.com/org/repo/commit/*",
      "c:\\external\\file.h": "https://example.com/direct-link-to-file.h"
    }
  }

  The wildcard * in the key matches any path suffix, and the * in the URL is replaced with the matched portion.

  Environment

  - Microsoft.Diagnostics.Tracing.TraceEvent: 3.1.x
  - .NET: 8.0
  - OS: Windows 11

  Impact

  This affects source file viewing for any PDB built with standard MSBuild SourceLink tooling, which commonly produces wildcard mappings. Many Microsoft and third-party PDBs use this format.

  Suggested Fix

  The SourceLink parsing logic in SymbolReader should:
  1. Handle wildcard patterns in the document keys
  2. Perform pattern matching against the requested source path
  3. Construct the final URL by substituting the matched path suffix

  References

  - https://github.com/dotnet/sourcelink/blob/main/docs/README.md
[...]

PR #2355, opened 2026-01-22
Fix SourceLink parsing to support both wildcard and exact path mappings

ivberg:
- Fixed regex bug that prevented parsing multiple SourceLink mappings
- Updated ParseSourceLinkJson to accept both wildcard patterns and exact file paths
- Enhanced GetUrlForFilePathUsingSourceLink with two-phase matching (exact then prefix)
- Added comprehensive test for both mapping types
- Resolves microsoft/perfview#2350

I did not test. This is semi-automated fix by remote codespace and Claude Sonnet 4.5 fix attempts.

### DIFF
--- a/src/TraceEvent/TraceEvent.Tests/Symbols/SymbolReaderTests.cs
+++ b/src/TraceEvent/TraceEvent.Tests/Symbols/SymbolReaderTests.cs
@@ -158,43 +158,43 @@ namespace TraceEventTests
         {
             // Create a test symbol module that returns SourceLink JSON with both wildcard and exact path mappings
             var testModule = new TestSymbolModuleWithSourceLink(_symbolReader);
-            
+
             // Test wildcard pattern matching
             bool result1 = testModule.GetUrlForFilePathUsingSourceLink(
-                @"C:\src\myproject\subfolder\file.cs", 
-                out string url1, 
+                @"C:\src\myproject\subfolder\file.cs",
+                out string url1,
                 out string relativePath1);
-            
+
             Assert.True(result1, "Should match wildcard pattern");
             Assert.Equal("https://raw.githubusercontent.com/org/repo/commit/subfolder/file.cs", url1);
             Assert.Equal("subfolder/file.cs", relativePath1);
-            
+
             // Test exact path matching
             bool result2 = testModule.GetUrlForFilePathUsingSourceLink(
-                @"c:\external\sdk\inc\header.h", 
-                out string url2, 
+                @"c:\external\sdk\inc\header.h",
+                out string url2,
                 out string relativePath2);
-            
+
             Assert.True(result2, "Should match exact path");
             Assert.Equal("https://example.com/blobs/ABC123?download=true&filename=header.h", url2);
             Assert.Equal("", relativePath2);
-            
+
             // Test another wildcard pattern with escaped characters
             bool result3 = testModule.GetUrlForFilePathUsingSourceLink(
-                @"C:\src\myproject\some folder\another file.cs", 
-                out string url3, 
+                @"C:\src\myproject\some folder\another file.cs",
+                out string url3,
                 out string relativePath3);
-            
+
             Assert.True(result3, "Should match wildcard pattern with spaces");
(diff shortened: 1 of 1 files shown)


brianrob:
@ivberg, thank you for submitting this.  I am in the process of reviewing this and doing some testing against the SourceLink spec.

brianrob:
@ivberg I pushed an additional commit based on some analysis of the spec.  The tests you added pass.  Is there any local testing that you might want to do on this before merging?

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -1726,15 +1726,36 @@ internal bool GetUrlForFilePathUsingSourceLink(string buildTimeFilePath, out str
 
             if (_sourceLinkMapping != null)
             {
-                foreach (Tuple<string, string> map in _sourceLinkMapping)
+                // First, try to find an exact match for the buildTimeFilePath (non-wildcard entries)
+                foreach (Tuple<string, string, bool> map in _sourceLinkMapping)
                 {
                     string path = map.Item1;
-                    string urlReplacement = map.Item2;
+                    string urlTemplate = map.Item2;
```
marklio, line 1733: SUGGESTION: deconstruct these in the foreach.

marklio: same below.

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -1726,15 +1726,36 @@ internal bool GetUrlForFilePathUsingSourceLink(string buildTimeFilePath, out str
...
+                        // Exact match found - this is a direct file mapping without wildcards
+                        relativeFilePath = "";
+                        url = urlTemplate; // Use the URL as-is for exact matches
+                        return true;
+                    }
+                }
+
+                // If no exact match, try prefix matching (wildcard patterns only)
+                // Per spec rule 2: only paths that had a wildcard (*) should be used for prefix matching
+                foreach (Tuple<string, string, bool> map in _sourceLinkMapping)
```
marklio, line 1747: COMMENT: It's at least a _little_ weird to swing through these twice and operate on disjoint sets. I get there's a simplicity to looking through once for the "exact match" and once for the others. Is this something that can be simplified in the data structure so this code isn't so weird? Some example approaches:
* Two different maps for isWildCard
* Sort for isWildcard first

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -1746,11 +1767,12 @@ internal bool GetUrlForFilePathUsingSourceLink(string buildTimeFilePath, out str
         }
 
         /// <summary>
-        /// Parses SourceLink information and returns a list of filepath -> url Prefix tuples.  
+        /// Parses SourceLink information and returns a list of (filepath, url, isWildcard) tuples.  
+        /// Supports both wildcard patterns ("path\\*" -> "url/*") and exact path mappings ("path\\file.h" -> "url").
         /// </summary>  
-        private List<Tuple<string, string>> ParseSourceLinkJson(IEnumerable<string> sourceLinkContents)
+        private List<Tuple<string, string, bool>> ParseSourceLinkJson(IEnumerable<string> sourceLinkContents)
```
marklio, line 1773: QUESTION: Is this a good opportunity to move to ValueTuple (and the built-in syntax for them that supports names?)

marklio (approved):
In general, this seems fine. I left a few comments to consider at this time or the future.

src/TraceEvent/Symbols/SymbolReader.cs
brianrob: I think this is possible, but I'm not sure it's worth it.  The set size is usually quite small and so it's likely more complex to the change the data structure to iterate through them one less time.

brianrob:
Thank you for the feedback @marklio.  I'm going to merge this one and consider these in the future.

ivberg:
Thanks for fixing this and merging! Any hope for a new TraceEvent release with this fix?

PR #2364, opened 2026-02-15
Strip .il and .ni suffixes from TraceModuleFile.Name

leculver:
Fix dotnet/diagnostics#3102: EventPipe module names have unexpected .il suffix.

TraceLog.ManagedModuleLoadOrUnload creates synthetic .il.dll paths for ReadyToRun assemblies to maintain separate IL and native PDB entries. This is an internal implementation detail that was leaking into the user-visible Name property, causing tools like dotnet-stack, PerfView, and WPA to display module names like 'System.Private.CoreLib.il' instead of 'System.Private.CoreLib'.

Strip the .il and .ni suffixes in TraceModuleFile.Name, matching the same logic already used in GetSourceLine() for source resolution.

### DIFF
--- a/src/TraceEvent/AutomatedAnalysis/StackView.cs
+++ b/src/TraceEvent/AutomatedAnalysis/StackView.cs
@@ -297,8 +297,8 @@ namespace Microsoft.Diagnostics.Tracing.AutomatedAnalysis
 
             // Allow the found name to be of equivalent or greater length than the requested length.
             // Example:
-            //  Requested: System.Private.CoreLib.il!System.String.Concat
-            //  Found:     System.Private.CoreLib.il!System.String.Concat(class System.String,class System.String)
+            //  Requested: System.Private.CoreLib!System.String.Concat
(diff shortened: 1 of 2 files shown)


leculver:
TraceEvent tests pass (well, they match the baseline of main).

brianrob:
Thanks @leculver.  I'm wondering if we might be able to just remove the ".il" suffix entirely, rather than stripping it off in various places.  It doesn't actually represent a file on disk, so there isn't a risk of not finding the file.  What if we removed [this](https://github.com/microsoft/perfview/blob/5409507b37bcab33bfca753c9e421d44491d6dfd/src/TraceEvent/TraceLog.cs#L7054) code and then we could even remove other code that strips it before presentation, such as [this](https://github.com/microsoft/perfview/blob/5409507b37bcab33bfca753c9e421d44491d6dfd/src/TraceEvent/TraceLog.cs#L8333-L8336).

leculver:
Will take a look, thanks!

leculver:
Done, @brianrob can you take a look?

PR #2368, opened 2026-02-20
Fix Debug.Assert failures in SpeedScope tests and DynamicTraceEventParser

brianrob:
The SpeedScope `CanConvertProvidedTraceFiles` tests were clearing `Trace.Listeners` to suppress `Debug.Assert` failures that fired during trace parsing. This masked two real bugs:

## Fix 1: Duplicate template enumeration in DynamicTraceEventParser

`DynamicTraceEventParser.EnumerateTemplates` enumerates templates from three sources (manifest-based, `registeredParser`, and `eventPipeTraceEventParser`) but only deduplicated between manifest-based templates and `registeredParser`. When both `registeredParser` and `eventPipeTraceEventParser` knew about the same event, the template was enumerated twice, causing `Subscribe` to assert on the duplicate with `mayHaveExistedBefore=false`.

**Fix:** Filter `eventPipeTraceEventParser` templates against `registeredParser`, matching the existing dedup pattern for manifest-based templates.

## Fix 2: Skip ConfirmAllEventsAreInEnumeration for ApplicationServerTraceEventParser

`ApplicationServerTraceEventParser` is auto-generated and has many events that share the same task name and opcode, producing duplicate computed event names (100+ duplicates). The `ConfirmAllEventsAreInEnumeration` DEBUG validation cannot handle this pattern.

**Fix:** Skip this parser from the validation, similar to how `PredefinedDynamicTraceEventParser` is already excluded.

## Cleanup

- Remove the listener-clearing hack from SpeedScope tests (no longer needed since the underlying asserts are fixed).
- Add documentation to `ThrowingTraceListener` and `DebugAssertionTests` explaining that on .NET 5+, the `DefaultTraceListener` already throws on assert failures, so no additional listener configuration is needed.

### DIFF
--- a/src/TraceEvent/TraceEvent.cs
+++ b/src/TraceEvent/TraceEvent.cs
@@ -2882,7 +2882,12 @@ namespace Microsoft.Diagnostics.Tracing
             }
 
 #if DEBUG
-            if (GetProviderName() != null && !m_ConfirmedAllEventsAreInEnumeration && !(this is PredefinedDynamicTraceEventParser))
+            // ApplicationServerTraceEventParser is auto-generated and has many events sharing the same
+            // task name, producing duplicate computed event names that ConfirmAllEventsAreInEnumeration
+            // cannot handle.
+            if (GetProviderName() != null && !m_ConfirmedAllEventsAreInEnumeration &&
+                !(this is PredefinedDynamicTraceEventParser) &&
+                !(this is Parsers.ApplicationServerTraceEventParser))
             {
                 ConfirmAllEventsAreInEnumeration();
                 m_ConfirmedAllEventsAreInEnumeration = true;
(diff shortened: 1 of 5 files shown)


src/TraceEvent/TraceEvent.cs
```diff
@@ -2882,7 +2882,12 @@ protected TraceEventParser(TraceEventSource source, bool dontRegister = false)
             }
 
 #if DEBUG
-            if (GetProviderName() != null && !m_ConfirmedAllEventsAreInEnumeration && !(this is PredefinedDynamicTraceEventParser))
+            // ApplicationServerTraceEventParser is auto-generated and has many events sharing the same
+            // task name, producing duplicate computed event names that ConfirmAllEventsAreInEnumeration
+            // cannot handle.
+            if (GetProviderName() != null && !m_ConfirmedAllEventsAreInEnumeration &&
+                !(this is PredefinedDynamicTraceEventParser) &&
+                !(this is Parsers.ApplicationServerTraceEventParser))
```
leculver, line 2890: This seems fragile.  Fine for now, but if you find you need another class it might be time to think through the design.

brianrob: Yep.  I agree that if this gets bigger we should look to make this a real pattern.

PR #2372, opened 2026-03-04
Add ELF symbol metadata to TraceModuleFile

zachcmadsen:
I'm not sure what what the plan is for resolving ELF symbols in `TraceEvent`. I tried to make the smallest change necessary to allow users to resolve symbols themselves.

I tested the changes on traces collected with `dotnet-trace collect-linux`. With the added metadata, I was able to download and resolve symbols for native binaries.


brianrob:
@zachcmadsen, thanks for your contribution.  Are you saying that you were able to resolve ELF native symbols in PerfView with this change?  I would expect that you'd need an ELF symbol parser for this.

zachcmadsen:
Sorry, that bit was unclear. The symbol downloading and parsing code was outside of TraceEvent/PerfView. I was just trying to say that this patch worked in an E2E scenario which used TraceEvent.

brianrob:
Superceded by #2397.

PR #2376, opened 2026-03-09
Fix NRE in AddUniversalDynamicSymbol for invalid symbol address ranges

### LINKED ISSUE #2373: NullReferenceException in AddUniversalDynamicSymbol when opening nettrace with invalid symbol ranges
Opening certain EventPipe .nettrace files containing universal trace format symbols crashes with a NullReferenceException during ETLX conversion.

Suspected root cause:
AddUniversalDynamicSymbol in TraceLog.cs passes the symbol size to ForAllUnresolvedCodeAddressesInRange using an (int) cast, while the downstream UniversalMapping uses a (long) cast:
https://github.com/microsoft/perfview/blob/46f2cfa164c2507c67bab8f5ce8d6338bfe537a8/src/TraceEvent/TraceLog.cs#L8546

https://github.com/microsoft/perfview/blob/46f2cfa164c2507c67bab8f5ce8d6338bfe537a8/src/TraceEvent/TraceLog.cs#L7134

  When a ProcessSymbol event has an invalid address range (e.g., StartAddress=0x0, EndAddress=0xFFFFFFFFFFFFFFFF), the (int) truncation produces -1, which wraps to 0xFFFFFFFFFFFFFFFF as a ulong — causing the loop to iterate over every code
  address in the process. Meanwhile, UniversalMapping correctly interprets the (long) cast as negative, leaving imageSize at 0. The resulting zero-size module cannot be found by FindModuleAndIndexContainingAddress, leaving loadedModule null and
  causing the NRE at loadedModule.ModuleID.

  This was observed with nettrace files generated by one-collect's record-trace without root access on Linux, where /proc/kallsyms returns zeroed addresses, causing the kernel symbol startup_64 to be emitted as [0x0, 0xFFFFFFFFFFFFFFFF).

```
  Started: Opening with-rbp-chain-walk.nettrace
  Creating ETLX file C:\Users\mihw\AppData\Local\Temp\PerfView (4)\with-rbp-chain-walk_d0b9d067.etlx from \\wsl.localhost\Ubuntu-24.04\home\mihw\repo\mihw-notes\stack-investigation\traces\with-rbp-chain-walk.nettrace
  [Opening a log file of size 0 MB.]
  [ELAPSED  0 seconds.     READ    131,072 events.     TIMESTAMP       0ms.     WRITTEN    19MB.     ]
  [ELAPSED  1 seconds.     READ    262,144 events.     TIMESTAMP       0ms.     WRITTEN    39MB.     ]
  [ELAPSED  1 seconds.     READ    393,216 events.     TIMESTAMP       0ms.     WRITTEN    58MB.     ]
  [ELAPSED  1 seconds.     READ    524,288 events.     TIMESTAMP       0ms.     WRITTEN    77MB.     ]
  [ELAPSED  1 seconds.     READ    655,360 events.     TIMESTAMP       0ms.     WRITTEN    97MB.     ]
  [ELAPSED  2 seconds.     READ    786,432 events.     TIMESTAMP       0ms.     WRITTEN   116MB.     ]
  [ELAPSED  2 seconds.     READ    917,504 events.     TIMESTAMP       0ms.     WRITTEN   135MB.     ]
  [ELAPSED  2 seconds.     READ  1,048,576 events.     TIMESTAMP       0ms.     WRITTEN   155MB.     ]
  [ELAPSED  3 seconds.     READ  1,179,648 events.     TIMESTAMP       0ms.     WRITTEN   174MB.     ]
  [ELAPSED  3 seconds.     READ  1,310,720 events.     TIMESTAMP       0ms.     WRITTEN   193MB.     ]
  [ELAPSED  4 seconds.     READ  1,441,792 events.     TIMESTAMP       0ms.     WRITTEN   213MB.     ]
  [ELAPSED  4 seconds.     READ  1,572,864 events.     TIMESTAMP       0ms.     WRITTEN   232MB.     ]
  [ELAPSED  4 seconds.     READ  1,703,936 events.     TIMESTAMP       0ms.     WRITTEN   252MB.     ]
  [ELAPSED  5 seconds.     READ  1,835,008 events.     TIMESTAMP       0ms.     WRITTEN   271MB.     ]
  [ELAPSED  5 seconds.     READ  1,966,080 events.     TIMESTAMP       0ms.     WRITTEN   290MB.     ]
  [ELAPSED  5 seconds.     READ  2,097,152 events.     TIMESTAMP       0ms.     WRITTEN   310MB.     ]
  [ELAPSED  6 seconds.     READ  2,228,224 events.     TIMESTAMP       0ms.     WRITTEN   329MB.     ]
  WARNING: Events occurred > 10msec before process 12975 start at 0.000 msec
  WARNING: Events occurred > 10msec before process 12977 start at 0.000 msec
  WARNING: Events occurred > 10msec before process 12976 start at 0.000 msec
  WARNING: Events occurred > 10msec before process 12978 start at 0.000 msec
  WARNING: Events occurred > 10msec before process 12979 start at 0.000 msec
  WARNING: Events occurred > 10msec before process 12980 start at 0.000 msec
  WARNING: Events occurred > 10msec before process 12981 start at 0.000 msec
  WARNING: Events occurred > 10msec before process 12982 start at 0.000 msec
  Error: Exception EventPipe conversion: System.NullReferenceException: Object reference not set to an instance of an object.
     at Microsoft.Diagnostics.Tracing.Etlx.TraceCodeAddresses.<>c__DisplayClass33_0.<AddUniversalDynamicSymbol>b__0(CodeAddressInfo& info)
     at Microsoft.Diagnostics.Tracing.Etlx.TraceCodeAddresses.ForAllUnresolvedCodeAddressesInRange(TraceProcess process, UInt64 start, Int64 length, Boolean considerResolved, ForAllCodeAddrAction body)
     at Microsoft.Diagnostics.Tracing.Etlx.TraceCodeAddresses.AddUniversalDynamicSymbol(ProcessSymbolTraceData data, TraceProcess process)
     at Microsoft.Diagnostics.Tracing.SourceConverters.NettraceUniversalConverter.AfterProcess(TraceLog traceLog)
     at Microsoft.Diagnostics.Tracing.Etlx.TraceLog.CopyRawEvents(TraceEventDispatcher rawEvents, IStreamWriter writer)
     at Microsoft.Diagnostics.Tracing.Etlx.TraceLog.<>c__DisplayClass134_0.<FastSerialization.IFastSerializable.ToStream>b__0()
     at FastSerialization.DeferedRegion.Write(Serializer serializer, Action toStream)
     at Microsoft.Diagnostics.Tracing.Etlx.TraceLog.FastSerialization.IFastSerializable.ToStream(Serializer serializer)
     at FastSerialization.Serializer.WriteObjectData(IFastSerializable obj, Tags beginTag)
     at FastSerialization.Serializer.WriteObjectRef(IFastSerializable obj, Boolean defered)
     at FastSerialization.Serializer..ctor(IStreamWriter writer, IFastSerializable entryObject)
     at Microsoft.Diagnostics.Tracing.Etlx.TraceLog.CreateFromEventPipeEventSources(TraceEventDispatcher source, IOStreamStreamWriter streamWriter, TraceLogOptions options)
     at Microsoft.Diagnostics.Tracing.Etlx.TraceLog.CreateFromEventPipeEventSources(TraceEventDispatcher source, String etlxFilePath, TraceLogOptions options)
     at Microsoft.Diagnostics.Tracing.Etlx.TraceLog.CreateFromEventPipeDataFile(String filePath, String etlxFilePath, TraceLogOptions options)
     at PerfView.EventPipePerfViewData.GetTraceLog(TextWriter log, Action`3 onLostEvents)
  [Error: exception while opening EventPipe data.]
  Completed: Opening with-rbp-chain-walk.nettrace   (Elapsed Time: 6.332 sec)
```

Possible fixes:
1. Skip symbols with EndAddress <= StartAddress (invalid range)
2. Change (int) to (long) cast for consistency with UniversalMapping
3. Add null guard on loadedModule as defense-in-depth

### Expected

.nettrace file opens and shows views, possibly with an indicator that the symbols had invalid ranges

### Actual

When double clicking the nettrace, no views appeared, and just the message `Error: exception while opening EventPipe data.` appeared.

brianrob:
When a ProcessSymbol event has an invalid address range (e.g., StartAddress=0x0, EndAddress=0xFFFFFFFFFFFFFFFF from zeroed /proc/kallsyms on Linux without root), the ETLX conversion crashed with a NullReferenceException.

Three fixes applied:
1. Skip symbols where EndAddress <= StartAddress (invalid range)
2. Change (int) to (long) cast for the size parameter to match ForAllUnresolvedCodeAddressesInRange's parameter type
3. Add null guard on loadedModule as defense-in-depth

Fixes #2373

### DIFF
--- a/src/TraceEvent/TraceEvent.Tests/Universal/InvalidSymbolRangeTest.cs
+++ b/src/TraceEvent/TraceEvent.Tests/Universal/InvalidSymbolRangeTest.cs
@@ -1,0 +1,190 @@
+using System;
+using System.IO;
+using FastSerialization;
+using Microsoft.Diagnostics.Tracing;
+using Microsoft.Diagnostics.Tracing.Etlx;
+using Microsoft.Diagnostics.Tracing.EventPipe;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace TraceEventTests
+{
+    /// <summary>
+    /// Regression tests for https://github.com/microsoft/perfview/issues/2373
+    /// Verifies that ProcessSymbol events with invalid address ranges don't crash ETLX conversion.
+    /// </summary>
+    public class InvalidSymbolRangeTest : TestBase
+    {
+        private static readonly Guid UniversalSystemProviderGuid = new Guid("8c107b6c-79f8-5231-4de6-2a0e20a3f562");
+
+        // Nettrace V6 metadata IDs for the events we'll emit.
+        private const int ProcessMappingMetadataId = 1;
+        private const int ProcessSymbolMetadataId = 2;
+
+        // Universal.System event IDs.
+        private const int ProcessMappingEventId = 3;
+        private const int ProcessSymbolEventId = 4;
+
+        public InvalidSymbolRangeTest(ITestOutputHelper output)
+            : base(output)
+        {
+        }
+
+        /// <summary>
+        /// When a ProcessSymbol has StartAddress=0 and EndAddress=0xFFFFFFFFFFFFFFFF (e.g., from
+        /// zeroed /proc/kallsyms on Linux without root), ETLX conversion must not throw.
+        /// </summary>
+        [Fact]
+        public void ConvertNettrace_WithInvalidSymbolRange_DoesNotThrow()
+        {
+            byte[] nettrace = BuildNettraceWithSymbol(
+                symbolStartAddress: 0,
+                symbolEndAddress: ulong.MaxValue,
+                symbolName: "startup_64");
+
+            ConvertAndVerify(nettrace);
+        }
+
+        /// <summary>
+        /// A zero-size symbol range (StartAddress == EndAddress) should also be skipped gracefully.
+        /// </summary>
+        [Fact]
+        public void ConvertNettrace_WithZeroSizeSymbolRange_DoesNotThrow()
+        {
+            byte[] nettrace = BuildNettraceWithSymbol(
+                symbolStartAddress: 0x1000,
+                symbolEndAddress: 0x1000,
+                symbolName: "empty_symbol");
+
+            ConvertAndVerify(nettrace);
+        }
+
+        /// <summary>
+        /// Builds a minimal nettrace (V6) containing one ProcessMapping and one ProcessSymbol with the given addresses.
+        /// </summary>
+        private static byte[] BuildNettraceWithSymbol(ulong symbolStartAddress, ulong symbolEndAddress, string symbolName)
+        {
+            var writer = new EventPipeWriterV6();
+            writer.WriteHeaders();
+
+            // Define metadata for Universal.System ProcessMapping (eventId=3) and ProcessSymbol (eventId=4).
+            var mappingMeta = new EventMetadata(ProcessMappingMetadataId, "Universal.System", "ProcessMapping", ProcessMappingEventId);
+            mappingMeta.ProviderId = UniversalSystemProviderGuid;
+
+            var symbolMeta = new EventMetadata(ProcessSymbolMetadataId, "Universal.System", "ProcessSymbol", ProcessSymbolEventId);
+            symbolMeta.ProviderId = UniversalSystemProviderGuid;
+
+            writer.WriteMetadataBlock(mappingMeta, symbolMeta);
+
+            // Thread entry: thread index 1, OS thread ID 100, process ID 1234.
+            writer.WriteThreadBlock(w =>
+            {
+                w.WriteThreadEntry(1, 100, 1234);
+            });
+
+            // Emit one ProcessMapping (valid range) and one ProcessSymbol (potentially invalid range).
+            writer.WriteEventBlock(w =>
+            {
+                // ProcessMapping payload: Id=1, StartAddress=0x7F000000, EndAddress=0x7F001000, FileOffset=0, FileName="test.so", MetadataId=0
+                w.WriteEventBlob(ProcessMappingMetadataId, 1, 1, WriteProcessMappingPayload(
+                    id: 1,
+                    startAddress: 0x7F000000,
+                    endAddress: 0x7F001000,
+                    fileOffset: 0,
+                    fileName: "test.so",
+                    metadataId: 0));
+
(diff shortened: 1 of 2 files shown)


src/TraceEvent/TraceEvent.Tests/Universal/InvalidSymbolRangeTest.cs
```diff
@@ -0,0 +1,191 @@
...
+                    // This must not throw (previously threw NullReferenceException).
+                    TraceLog.CreateFromEventPipeEventSources(
+                        eventSource,
+                        new IOStreamStreamWriter(etlxStream, SerializationSettings.Default, leaveOpen: true),
+                        null);
+
+                    etlxStream.Position = 0;
+                    using (var traceLog = new TraceLog(etlxStream))
+                    {
+                        Assert.NotNull(traceLog);
```
mconnew: This Assert is useless as traceLog isn't a value being returned from a method, it's an object being new'd up. The only way traceLog would be null is if the constructor throws, in which case the Assert statement wouldn't get hit.

brianrob: This is effectively a call to `GC.KeepAlive`.  The goal is to make sure that the jit doesn't determine that the `traceLog` isn't used and eliminates the construction of the object entirely.  I don't think it does that now, but I believe that would be allowed.

mconnew: I don't believe it would eliminate the call as the JIT isn't free to eliminate void returning method calls because the return value isn't used. And the return value is used here as the using statement is syntactic sugar for wrapping it in a try/finally block where traceLog.Dispose is called. There's also a bunch of criteria around what a JIT is allowed to optimize away. If any exceptions can be thrown by the method call (which a constructor call is a method call), then it's not allowed to optimize it away. Basically there are many reasons why the JIT can't optimize away the construction of TraceLog.

brianrob: Thanks.  Pushed a change to address this.

PR #2392, opened 2026-03-21
Add MethodDetails CTF event mapping for LTTng trace support

benaadams:
## Summary
- Add missing `DotNETRuntime:MethodDetails` CTF event mapping to `ClrTraceEventParser.EnumerateCtfEventMappings()`

## Problem
`MethodDetailsTraceData` (event ID 72, task Method, opcode 43) was added in .NET 5 but its CTF event mapping was never included in `EnumerateCtfEventMappings()`. This causes `CtfTraceEventSource` to skip `DotNETRuntime:MethodDetails` events when parsing perfcollect `.trace.zip` files.

`dotnet-pgo` checks for `MethodDetailsTraceData` events at [Program.cs line 1115](https://github.com/dotnet/runtime/blob/main/src/coreclr/tools/dotnet-pgo/Program.cs#L1115) and fails with:
```
Error: No MethodDetails
Was the trace collected with provider at least "Microsoft-Windows-DotNETRuntime:0x4000080018:5"?
```

The events ARE in the LTTng CTF trace (verified via CTF metadata inspection), but `CtfTraceEventSource.TryGetEventMapping()` returns false because the mapping is not registered, so the events are silently dropped.

This blocks SPGO (Sample-based Profile-Guided Optimization) on Linux when using perfcollect for trace collection.

## Fix
One line: add the CTF mapping for `DotNETRuntime:MethodDetails` alongside the existing Method event mappings.

## Test plan
- [x] Verified `DotNETRuntime:MethodDetails` events appear in perfcollect `.trace.zip` CTF metadata
- [x] Confirmed the event is defined in `ClrEtwAll.man` as event ID 72, task `CLRMethod`, opcode 43, keyword `MethodDiagnosticKeyword`
- [x] Confirmed the ETW/EventPipe path works (`.nettrace` files parse `MethodDetailsTraceData` correctly)
- [ ] Verify `dotnet-pgo create-mibc --spgo` succeeds with a perfcollect trace after this fix

### DIFF
--- a/src/TraceEvent/Parsers/ClrTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/ClrTraceEventParser.cs
@@ -2613,7 +2613,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers
             yield return new CtfEventMapping("DotNETRuntime:MethodJittingStarted", Parsers.ClrTraceEventParser.ProviderGuid, 42, 145, 0);
             yield return new CtfEventMapping("DotNETRuntime:MethodJittingStarted_V1", Parsers.ClrTraceEventParser.ProviderGuid, 42, 145, 1);
             yield return new CtfEventMapping("DotNETRuntime:MethodDetails", Parsers.ClrTraceEventParser.ProviderGuid, 43, 72, 0);
-            yield return new CtfEventMapping("DotNETRuntime:MethodILToNativeMap_V1", Parsers.ClrTraceEventParser.ProviderGuid, 87, 190, 1);
+            yield return new CtfEventMapping("DotNETRuntime:MethodILToNativeMap_V1", Parsers.ClrTraceEventParser.ProviderGuid, 87, 190, 0);
             yield return new CtfEventMapping("DotNETRuntime:AppDomainUnload", Parsers.ClrTraceEventParser.ProviderGuid, 42, 157, 0);
             yield return new CtfEventMapping("DotNETRuntime:AppDomainUnload_V1", Parsers.ClrTraceEventParser.ProviderGuid, 42, 157, 1);
             yield return new CtfEventMapping("DotNETRuntime:DomainModuleLoad", Parsers.ClrTraceEventParser.ProviderGuid, 45, 151, 0);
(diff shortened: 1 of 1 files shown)


src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -2612,6 +2612,8 @@ protected internal override IEnumerable<CtfEventMapping> EnumerateCtfEventMappin
             yield return new CtfEventMapping("DotNETRuntime:AppDomainLoad_V1", Parsers.ClrTraceEventParser.ProviderGuid, 41, 156, 1);
             yield return new CtfEventMapping("DotNETRuntime:MethodJittingStarted", Parsers.ClrTraceEventParser.ProviderGuid, 42, 145, 0);
             yield return new CtfEventMapping("DotNETRuntime:MethodJittingStarted_V1", Parsers.ClrTraceEventParser.ProviderGuid, 42, 145, 1);
+            yield return new CtfEventMapping("DotNETRuntime:MethodDetails", Parsers.ClrTraceEventParser.ProviderGuid, 43, 72, 0);
+            yield return new CtfEventMapping("DotNETRuntime:MethodILToNativeMap_V1", Parsers.ClrTraceEventParser.ProviderGuid, 87, 190, 0);
```
brianrob:
```suggestion
            yield return new CtfEventMapping("DotNETRuntime:MethodILToNativeMap_V1", Parsers.ClrTraceEventParser.ProviderGuid, 87, 190, 1);
```

brianrob:
@benaadams its been a while since I used `dotnet-pgo`, but if I recall correctly, it depends on the output coming from a nettrace file via `dotnet-trace`.  Does it know how to use the output from `lttng`?

benaadams:
> @benaadams its been a while since I used `dotnet-pgo`, but if I recall correctly, it depends on the output coming from a nettrace file via `dotnet-trace`. Does it know how to use the output from `lttng`?

We got it to; but had to jump though a bunch of hoops https://github.com/NethermindEth/nethermind/pull/10877

benaadams:
Not sure using the data improved things; but need the data to improve the improvements 😅

brianrob:
Sounds good.  This is a pretty innocuous change since it's just two event mappings - happy to take this.

brianrob (dismissed):

brianrob (commented):
Looks like I got ahead of myself and missed that I had caught one issue.  If you can fix the event version, this should be good to go.

brianrob:
@benaadams are you still wanting to get this change in?  If so, there is one small change and then we should be OK to merge.  If you just went the path of using `dotnet-trace`, totally fine to close this one.

brianrob:
I'm going to go ahead and close this PR.

benaadams:
> Ben {chmark} Adams (@benaadams) are you still wanting to get this change in? If so, there is one small change and then we should be OK to merge. If you just went the path of using `dotnet-trace`, totally fine to close this one.

Sorry been caught up with other things and missed your reply. Yes please do want this change

brianrob:
> Sorry been caught up with other things and missed your reply. Yes please do want this change

No worries.  Can you please take a look at the version change above?  If you can fix that up, think this is good to merge.

src/TraceEvent/Parsers/ClrTraceEventParser.cs
benaadams: Done

brianrob:
I'm seeing these CI failures elsewhere too.  I'm investigating.  Stand by.

benaadams:
I've enabled edits by maintainers if that helps

brianrob (approved):

PR #2403, opened 2026-04-01
Optimize nettrace-to-TraceLog Conversion

brianrob:
## Summary

Optimizes the nettrace-to-TraceLog/ETLX conversion pipeline, reducing conversion time for a 1.68 GB nettrace file from **~22 minutes to ~40 seconds** (34x speedup).

## Problem

Converting large nettrace files (e.g., from OneCollect Linux traces with many threads) to TraceLog/ETLX format was extremely slow. Profiling revealed three bottlenecks:

1. **`EventCache.SortAndDispatch`** consumed ~75% of CPU — O(N×T) linear scan over all thread queues per event block, plus per-call LINQ `List` allocations.
2. **`ParsedSymbolMetadata`** consumed ~10% of CPU — JSON deserialization repeated on every property access.
3. **`FileName` string allocations** consumed ~12% of CPU — `GetShortUTF8StringAt()` called 3 times per mapping event.

### EventCache min-heap (`EventCache.cs`)
- Replaced the O(N×T) linear scan in `SortAndDispatch` with an O(N×log T) min-heap merge.
- Introduced a generic `MinHeap<TValue>` class with full unit test coverage (13 tests).
- Added `_activeThreadQueues` HashSet to track only threads with pending events, avoiding iteration over the full thread dictionary.

### Cache ParsedSymbolMetadata (`UniversalSystemTraceEventParser.cs`)
- Cached the result of `ProcessMappingSymbolMetadataParser.TryParse()` to avoid repeated JSON deserialization.
- Reset cache in `Dispatch()` to prevent stale data when TraceEvent objects are reused.
- Added `Clone()` override to preserve cached values in cloned events.

### Reduce FileName allocations (`NettraceUniversalConverter.cs`)
- Read `data.FileName` once into a local variable instead of accessing the property 3 times (each access allocates a new string via `GetShortUTF8StringAt`).
- Call `UniversalMapping(string, ...)` overload directly, passing the cached string.

## Performance Results

Measured with a 1.68 GB nettrace file (8.6M events, 11.5K processes) using Release builds:

| Metric | Before | After |
|---|---|---|
| **Conversion time** | **22:26 (1346s)** | **00:40 (40s)** |
| **Speedup** | — | **34×** |
| Events | 8,583,800 | 8,583,800 ✓ |
| Processes | 11,509 | 11,509 ✓ |
| Unique code addresses | 174,476 | 174,476 ✓ |
| Unique stacks | 206,467 | 206,467 ✓ |
| ETLX file size | 3,051,574,460 | 3,051,574,460 ✓ |

All semantic statistics match. ETLX files are the same size but not byte-identical due to different tie-breaking order for same-timestamp events (min-heap vs linear scan).

## Testing

- All existing tests pass (2,274 on net8.0 + 2,289 on net462).
- 13 new MinHeap unit tests added.

### DIFF
--- a/src/TraceEvent/EventPipe/EventCache.cs
+++ b/src/TraceEvent/EventPipe/EventCache.cs
@@ -196,7 +196,7 @@ namespace Microsoft.Diagnostics.Tracing.EventPipe
             // Merge events in timestamp order using the min-heap.
             while (_heap.Count > 0)
             {
-                Queue<EventMarker> minQueue = _heap.PeekQueue;
+                Queue<EventMarker> minQueue = _heap.PeekValue;
                 EventMarker eventMarker = minQueue.Dequeue();
                 OnEvent?.Invoke(ref eventMarker.Header);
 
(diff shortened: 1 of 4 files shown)


brianrob:
cc: @zachcmadsen

leculver (approved):

PR #2407, opened 2026-04-09
Fix PDB Symbol Resolution for Unmerged Windows Traces

brianrob:
The ELF symbol work introduced a switch on `BinaryFormat` in `LookupSymbolsForModule` that only handled `PE` and `ELF` cases. For unmerged Windows traces, `symbolInfo` is null because the ETL doesn't contain RSDS events with PDB identity info. This caused `BinaryFormat` to return `Unspecified` (formerly `Unknown`), hitting the default case which skipped symbol resolution entirely.

**Symptoms:**
```
[Loading symbols for c:\windows\system32\ntoskrnl.exe]
LookupSymbolsForModule: Unknown binary format Unknown for c:\windows\system32\ntoskrnl.exe, skipping.
Could not find symbols.
```

**Fix:**
- Add a `case ModuleBinaryFormat.Unspecified` that falls back to PDB lookup when running on Windows. `OpenPdbForModuleFile` already handles missing PDB signatures gracefully by checking the local file on disk.
- Rename `ModuleBinaryFormat.Unknown` to `Unspecified` to better reflect that the format was not specified in the trace data rather than being truly unknown.

### DIFF
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -8970,10 +8970,34 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
                     }
                     break;
 
+                case ModuleBinaryFormat.Unspecified:
+                    {
+                        // For unmerged Windows traces, symbolInfo is null because the ETL
+                        // didn't contain RSDS events with PDB identity info.
+                        // Fall back to PDB lookup on Windows, which handles missing signatures
+                        // gracefully and returns null if the file isn't a PE binary.
+                        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                        {
+                            reader.m_log.WriteLine("LookupSymbolsForModule: Binary format unspecified for {0}, looking up PDB info from local file.", moduleFile.FilePath);
+                            NativeSymbolModule moduleReader = OpenPdbForModuleFile(reader, moduleFile) as NativeSymbolModule;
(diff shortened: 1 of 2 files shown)


src/TraceEvent/TraceLog.cs
```diff
@@ -8970,10 +8970,34 @@ private void LookupSymbolsForModule(SymbolReader reader, TraceModuleFile moduleF
                     }
                     break;
 
+                case ModuleBinaryFormat.Unspecified:
+                    {
+                        // For unmerged Windows traces, symbolInfo is null because the ETL
+                        // didn't contain RSDS events with PDB identity info.
+                        // Fall back to PDB lookup on Windows, which handles missing signatures
+                        // gracefully and returns null if the file isn't a PE binary.
+                        if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
leculver, line 8979: `IsOSPlatform` is the right function here?  I mean yes we will process ETL files on windows, but there's no "target we are analyzing" concept?

leculver (approved):
Non blocking question, otherwise looks good.

src/TraceEvent/TraceLog.cs
brianrob: Yes.  I was being intentional about making sure this happens specifically on a Windows machine because we use msdia underneath.

PR #2410, opened 2026-04-13
Enable Spectre mitigations and linker optimizations for EtwClrProfiler

danmoseley:
Add build hardening flags to resolve BinSkim findings now that EtwClrProfiler is included in the official scan (PR 727750).

**Untested** and not suggesting you should take any of these (AFAIK none are required by SDL currently). Feel free to pick any all or none.

Both x86 and x64 vcxproj:
- BA2024: Enable Spectre mitigations (/Qspectre) in Debug and Release
- BA6004: Enable COMDAT folding (/OPT:ICF) in Release
- BA6005: Enable Optimize References (/OPT:REF) in Release
- BA6006: Enable Link-Time Code Generation (/LTCG + /GL) in Release

### DIFF
--- a/src/EtwClrProfiler/ETWClrProfilerX64.vcxproj
+++ b/src/EtwClrProfiler/ETWClrProfilerX64.vcxproj
@@ -69,6 +69,7 @@
       <SDLCheck>true</SDLCheck>
       <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
       <AdditionalOptions>/Gs %(AdditionalOptions)</AdditionalOptions>
+      <SpectreM>Spectre</SpectreM>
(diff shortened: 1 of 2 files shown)


danmoseley:
did some validation building with this -- 
- ✅ BA6004, BA6005, BA6006 — all resolved (no findings for COMDAT/OptRef/LTCG)
- ⚠️ BA2024 — still fires because we built without Spectre libs (expected — the <SpectreM>Spectre</SpectreM> in the vcxproj is correct, just need Spectre libs at build time, which CI agents have)
 - ⚠️ BA2027 — SourceLink not addressed (as planned)

cincuranet (approved):

Issue #2394, opened 2026-03-23
Feat-Request: "Open Any Stacks" in Event viewer should work with multiple events


rauhs:
Feature request:

Currently I can select a single "Time MSec" column/row in the event viewer for an event with `HasStack=True` and presst "ALT+S" to open the stacktrace of that event ("Open Any Stacks" in the context menu).

But: I sometimes have thousands of events (e.g. filtered by "Text Filter" or whatever) and then I'd love to just display these events in the stack viewer all together.

Could we extend this functionality so the stack viewer can properly open up whatever is selected?

I might have another intern/student that could work on that. Copilot is welcome too :)

brianrob:
@rauhs seems very reasonable to look into this.  Thinking about the new ways that we're working, I think copilot can do a lot of the heavy lifting here.  I think the key would be to have a clear design before we let copilot do its thing.

I'm not sure how you/your company are working, but I'd be happy to delegate this kind of work to an intern as I feel this helps to train developers of the future.  If you have an intern that can put together a plan, I'm happy to review it, and then either they, or I can have copilot actually do the implementation and they can do the testing.

How does that sound?

rauhs:
@brianrob Sound fantastic!

I have a student who could work on that for 2 weeks at the End of April. I guess we'll review existing PR's by Copilot on how to approach "creating a plan for copilot".

> I'd be happy to delegate this kind of work to an intern

If you mean to and intern within Microsoft this would of course also be great.

=> If this ticket is still undone by end of April I will assign the student to it.

brianrob:
@rauhs, sounds great.  I did mean I'm happy to have you assign this to a student.  If you want to have them write code, I'm happy to have them do that too.  I can see this happening in a number of ways, including:

1. The student designs the feature and implements it themselves (we should still discuss design ahead of time).
2. The student designs the feature and uses AI on their own to implement it.
3. The student designs the feature, I trigger copilot to implement it, and they test it.

Or anything in-between.  Just let me know how you'd like to proceed.

rauhs:
Thanks Brian & Paul!

PR #2418, opened 2026-04-29
Add support for showing stacks from selected events

paulusakademius:
This pull request introduces the ability to show stacks for selected events with the 'Open Any Stacks' function, regardless of the column of the event selected. Previously, stacks were only accessible for either a single time or a time range and defaulted to the everything when some other selection happened. With this update:
- Stacks are now filtered by event ID in addition to time.
- Selecting any number of events and viewing their stacks is possible.
- Behavior for selecting time ranges remains unchanged.

Issue #2394

### DIFF
--- a/src/PerfView.Tests/EventViewer/EventWindowTests.cs
+++ b/src/PerfView.Tests/EventViewer/EventWindowTests.cs
@@ -1,0 +1,311 @@
+using EventSources;
+using Microsoft.Diagnostics.Tracing;
+using Microsoft.Diagnostics.Tracing.Stacks;
+using Microsoft.VisualStudio.Threading;
+using PerfView;
+using PerfView.TestUtilities;
+using PerfViewTests.Utilities;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.Serialization;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Controls;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace PerfViewTests.EventViewer
+{
+    public class EventWindowTests : PerfViewTestBase
+    {
+        public EventWindowTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        [WpfFact]
+        [UseCulture("en-US")]
+        public Task TestOpenStacksForOneSelectedCellAsync()
+        {
+            return TestOpenStacksAsync(SelectedCellScenario.OneCell);
+        }
+
+        [WpfFact]
+        [UseCulture("en-US")]
+        public Task TestOpenStacksForTwoSelectedCellsAsync()
+        {
+            return TestOpenStacksAsync(SelectedCellScenario.TwoCells);
+        }
+
+        [WpfFact]
+        [UseCulture("en-US")]
+        public Task TestOpenStacksForThreeSelectedCellsAsync()
+        {
+            return TestOpenStacksAsync(SelectedCellScenario.ThreeCells);
+        }
+
+        [WpfFact]
+        [UseCulture("en-US")]
+        public Task TestOpenStacksForSelectedTimeRangeAsync()
+        {
+            return TestOpenStacksAsync(SelectedCellScenario.TimeRange);
+        }
+
+        private Task TestOpenStacksAsync(SelectedCellScenario scenario)
+        {
+            Func<Task<EventWindow>> setupAsync = async () =>
+            {
+                await JoinableTaskFactory.SwitchToMainThreadAsync();
+
+                var eventData = new PerfViewEventSource(new SelectedStacksFile());
+                var opened = new TaskCompletionSource<bool>();
+                eventData.Open(GuiApp.MainWindow, GuiApp.MainWindow.StatusBar, () => opened.SetResult(true));
+                await opened.Task.ConfigureAwait(true);
+
+                var eventWindow = eventData.Viewer;
+                eventWindow.EventTypes.SelectAll();
+                eventWindow.Update();
+                await eventWindow.StatusBar.WaitForWorkCompleteAsync().ConfigureAwait(true);
+                return eventWindow;
+            };
+
+            Func<EventWindow, Task> cleanupAsync = async eventWindow =>
+            {
+                await JoinableTaskFactory.SwitchToMainThreadAsync();
+
+                foreach (var stackWindow in StackWindow.StackWindows.ToArray())
+                {
+                    stackWindow.Close();
+                }
+
+                eventWindow.Close();
+            };
+
+            Func<EventWindow, Task> testDriverAsync = async eventWindow =>
+            {
+                await JoinableTaskFactory.SwitchToMainThreadAsync();
+
+                var selectedCells = eventWindow.Grid.SelectedCells;
+                if (scenario == SelectedCellScenario.TimeRange)
+                {
+                    var timeColumn = eventWindow.Grid.Columns.Single(column => Equals(column.Header, "Time MSec"));
+                    selectedCells.Add(new DataGridCellInfo(eventWindow.Grid.Items[1], timeColumn));
+                    selectedCells.Add(new DataGridCellInfo(eventWindow.Grid.Items[2], timeColumn));
+                }
+                else
+                {
+                    selectedCells.Add(new DataGridCellInfo(eventWindow.Grid.Items[1], eventWindow.Grid.Columns[0]));
+                    if (scenario == SelectedCellScenario.TwoCells)
+                    {
+                        selectedCells.Add(new DataGridCellInfo(eventWindow.Grid.Items[2], eventWindow.Grid.Columns[0]));
+                    }
+                    else if (scenario == SelectedCellScenario.ThreeCells)
+                    {
+                        // Select two cells from the first event to verify that rows are de-duplicated.
+                        selectedCells.Add(new DataGridCellInfo(eventWindow.Grid.Items[1], eventWindow.Grid.Columns[1]));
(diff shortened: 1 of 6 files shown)


paulusakademius:
I thought this might be the most intuitive solution because it gives you the stacks for the events you selected. I kept the functionality for selecting a time range as it is, but when I noticed it only filters by time and it is possible for other events to happen at the same time, I thought maybe you could filter there for the event id too, like I did in the feature I implemented. What do you think about this, I think it might be more intuitive for the selection in the EventViewer. I am looking forward to hear your opinion on the idea and the code I already wrote.

rauhs:
@brianrob Hi Brian, have you had any chance to look at this? If there are any issues let us know.
Cheers

brianrob:
Thank you for pinging me on this one.  I had not looked yet - I am looking now.  With my initial testing, it looks great.  I'm reviewing the code now and will get back to you on that.

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -459,42 +462,39 @@ private void OpenStacks(string stackSourceName)
...
+                if (selectedCells.Count == 2)
                 {
                     string start = GetCellStringValue(selectedCells[0]);
                     double parsedStart;
                     if (!double.TryParse(start, out parsedStart))
                     {
-                        if (selectedCells.Count != 1)
-                        {
                             StatusBar.LogError("Could not parse " + start + " as a number.");
+                            OpenSelectedStacks(stackSourceName, selectedCells);
```
brianrob, line 472: Previous behavior was that if no cells were selected, the views would still open and populate. Now, they populate with no data.  It would be good to preserve this behavior.

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -557,6 +557,125 @@ private void OpenStacks(string stackSourceName)
...
+                        StatusBar.LogError("No event records found in selection.");
+                    });
+                    return;
+                }
+
+                StackSource aggregateSource;
+
+                // For ETW sources, filter by exact EventIndex so concurrent events on other
+                // threads at the same timestamp are excluded.
+                var etwRecords = uniqueRecords.OfType<ETWEventSource.ETWEventRecord>().ToList();
```
brianrob, line 608: Another potential type that could be held in `uniqueRecords` is `CsvEventRecord`.  In that case, this will produce an empty list and `Min` and `Max` will throw.  Will need to handle that here.

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -459,42 +462,39 @@ private void OpenStacks(string stackSourceName)
...
-                        {
-                            string end = GetCellStringValue(selectedCells[1]);
-                            if (!double.TryParse(end, out endTimeRelativeMSec))
-                            {
-                                StatusBar.LogError("Could not parse " + end + " as a number.");
-                                return;
-                            }
-                        }
+                        this.StatusBar.LogError("Could not parse " + end + " as a number.");
+                        this.OpenSelectedStacks(stackSourceName, selectedCells);
```
brianrob, line 481: The previous behavior was that if nothing was selected then the views would populate with all data.  Now, the views will come up but with no data.  We should preserve the previous behavior.

src/PerfView/PerfViewData.cs
```diff
@@ -4571,7 +4571,24 @@ public virtual StackSource GetStackSource(TextWriter log, double startRelativeMS
             return ret;
         }
 
-        // TODO not clear I want this method 
+        public virtual StackSource GetStackSource(TextWriter log, double startRelativeMSec, double endRelativeMSec, Predicate<TraceEvent> predicate)
+        {
+            StackSource ret = DataFile.OpenStackSourceImpl(SourceName, log, startRelativeMSec, endRelativeMSec, predicate);
```
brianrob, line 4576: If a predicate is specified, I think we ought to be using it.  It might toss all of the events, but if it does, then we shouldn't fall back.  Is there a specific reason for the fallback behavior?

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -557,6 +557,125 @@ private void OpenStacks(string stackSourceName)
...
+                    dataSource = dataFile.GetStackSource(stackSourceName);
+                }
+
+                if (dataSource == null)
+                {
+                    throw new ApplicationException("Could not find stack source " + stackSourceName);
+                }
+
+                // Collect unique event records (a row may have multiple selected cells).
+                var uniqueRecords = selectedCells
```
brianrob, line 589: `selectedCells` is a reference to `Grid.SelectedCells`.  You'll need to copy this collection before the call to `StatusBar.StartWork` to ensure that the collection doesn't change underneath you, since the UI is still fluid.

src/PerfView/EventViewer/EventWindow.xaml.cs
```diff
@@ -557,6 +557,125 @@ private void OpenStacks(string stackSourceName)
...
+                    aggregateSource = dataSource.GetStackSource(
+                        StatusBar.LogWriter,
+                        startTimeRelativeMSec - .001,
+                        endTimeRelativeMSec + .001,
+                        data => selectedIndices.Contains(data.EventIndex));
+                }
+                else
+                {
+                        // Fall back to per-event time windows for non-ETW sources.
+                        var sources = new List<StackSource>();
```
brianrob: I think a reasonable fallback here is to find the min and max and just make a single call that filters the data.  I worry that this could be relatively large set of selected fields and we'll be doing lots of work unintentionally.

brianrob (commented):
Thank you very much for contributing this feature.  I tried it out and it's slick.

I've added some comments and questions.  It would also be great to have some additional test coverage for this UI feature to make sure that we keep it functional as changes are made in the future.  The PerfView tests have a good pattern for implementing UI tests and AI agents are pretty good at picking it up.

paulusakademius:
Thanks for the review, @brianrob! I have addressed all your comments in my latest commit. Please take another look when you have a moment.

src/TraceEvent/Stacks/Stacks.cs
```diff
@@ -737,6 +737,30 @@ public InternStackSource(StackSource source, StackSourceStacks sourceStacks)
             Interner.DoneInterning();
         }
 
+        /// <summary>
+        /// Merge multiple stack sources into a single unified source by re-interning all frames by name.
+        /// Frames with the same name across sources (e.g. the same process frame) are automatically merged.
+        /// </summary>
+        public static InternStackSource Merge(IEnumerable<StackSource> sources)
```
brianrob: You should be able to remove this now.

src/PerfView/SupportFiles/UsersGuide.htm
```diff
@@ -3641,10 +3641,27 @@ <h3>
             (lower right corner).&nbsp;&nbsp;
         </li>
         <li>
-            <strong>Viewing Stacks</strong> If an event has a stack trace associated with it
-            the 'HasStack' field will be true. By selecting the time of that event and hitting
-            Alt-S you can see the stack for that time (which will include that events). You
-            can also select two times, and that region will be shown in the AnyStack view.
+            <strong>Viewing Stacks</strong> - An event has a stack when its 'HasStack' value is
```
brianrob, line 3644: Great idea to update the help.  Thanks!

src/PerfView/PerfViewData.cs
```diff
@@ -4571,7 +4571,19 @@ public virtual StackSource GetStackSource(TextWriter log, double startRelativeMS
...
+
+            if (ret == null)
+            {
+                throw new ApplicationException("Not a file type that supports the StackView.");
```
brianrob: This overload is used for every selected-event stack command (CPU, Any, Start/Stop, TaskTree, and Thread), so a null result is not specifically saying that the file type lacks StackView support or that Any Stacks is unsupported. It means the requested stack source could not produce a predicate-filtered result. Could we make that explicit, for example: `The '{SourceName}' stack source does not support filtering by selected events.`?

src/PerfView.Tests/EventViewer/EventWindowTests.cs
```diff
@@ -0,0 +1,311 @@
...
+                foreach (var eventRecord in m_events.Where(e => e.TimeStampRelatveMSec >= StartTimeRelativeMSec && e.TimeStampRelatveMSec <= EndTimeRelativeMSec))
+                {
+                    if (!callback(eventRecord))
+                    {
+                        break;
+                    }
+                }
+            }
+
+            private static ETWEventSource CreateRecordSource()
```
brianrob: Could we replace these reflection-based test doubles with the existing `net.4.5.2.x86.etl.zip` fixture? It contains real sampled-profile stacks, so linking it into `PerfView.Tests`, opening it with `PerfViewFile.Get`, and selecting known sampled-profile rows would remove the private-field reflection and synthetic ETW classes while providing stronger end-to-end coverage.

brianrob (commented):
Thanks for making these changes.  A few more comments.

paulusakademius:
Hi @brianrob , I have adjusted the code according to your comments. I've encountered some problems with the Azure pipeline, where it seemed like something blocked it while running the tests. I got it to work now but maybe you could have a look at that too.

brianrob:
> Hi @brianrob , I have adjusted the code according to your comments. I've encountered some problems with the Azure pipeline, where it seemed like something blocked it while running the tests. I got it to work now but maybe you could have a look at that too.

Thanks @paulusakademius.  I think that the hang you saw is not related to this change.  I have a change that I'm working on to hopefully address these hangs.

brianrob (approved):
This looks great.  Thank you for all your hard work on this!

paulusakademius:
I am happy to hear that. Thank you for reviewing and merging this :)

PR #2421, opened 2026-05-15
C entrypoint fixes

zachcmadsen:
A couple fixes related to https://github.com/microsoft/one-collect/pull/268:

- Recognize `<module>!_start` as a valid top frame
- Bump symbol length by one for dynamic symbols to catch addresses on the symbol end boundary. `_start` seems to be a pathological case

With the one-collect change and these fixes we get a complete stack:

```
+ Process32 workload (140228) Args: workload
 + Thread (140228) CPU=19984ms
  + workload!_start
  |+ libc.so.6!__libc_start_main
  | + libc.so.6!__libc_start_call_main
  |  + workload!main
  |   + workload!run_workload
  |    + workload!hot_loop
  |    + workload!warm_loop
  |    + workload!cold_loop
```

### DIFF
--- a/src/TraceEvent/TraceEventStacks.cs
+++ b/src/TraceEvent/TraceEventStacks.cs
@@ -673,6 +673,14 @@ namespace Microsoft.Diagnostics.Tracing.Stacks
                     return true;
                 }
 
+                // On Linux, <processName>!_start is an ELF entry point for C programs.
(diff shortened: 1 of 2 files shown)


zachcmadsen:
Cc @brianrob

brianrob (approved):
LGTM.  Thanks @zachcmadsen!

PR #2430, opened 2026-05-27
Stabilize XamlMessageBox UI-thread dispatch test

brianrob:
## Summary
- Convert `Show_AutoDispatchesToUIThreadFromBackgroundThread` to use xUnit's `[WpfFact]` dispatcher instead of manually creating and shutting down a WPF dispatcher thread.
- Await the background `XamlMessageBox.Show` call directly and keep a timeout assertion for clearer failure behavior.
- Preserve the existing regression coverage for background-thread auto-dispatch and UI-thread no-op dispatch.

## Validation
- Reproduced the intermittent local hang before the change.
- 50/50 filtered runs passed after the change.
- 20/20 full `PerfViewTests.dll` runs passed after the change, with 87/87 tests passing each run.

### DIFF
--- a/src/PerfView.Tests/Dialogs/XamlMessageBoxTests.cs
+++ b/src/PerfView.Tests/Dialogs/XamlMessageBoxTests.cs
@@ -1,14 +1,9 @@
 using System;
-using System.Threading;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Media;
-using System.Windows.Threading;
 using PerfView.Dialogs;
 using Xunit;
-
-#pragma warning disable VSTHRD001 // Use JoinableTaskFactory — we're explicitly testing WPF Dispatcher threading
-#pragma warning disable VSTHRD110 // Observe awaitable — fire-and-forget Task.Run is intentional in these tests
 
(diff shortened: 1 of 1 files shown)


src/PerfView.Tests/Dialogs/XamlMessageBoxTests.cs
```diff
@@ -25,97 +20,34 @@ public class XamlMessageBoxTests
...
-
-                    // Catch unhandled dispatcher exceptions so they don't silently hang.
-                    app.DispatcherUnhandledException += (s, args) =>
-                    {
-                        exception = args.Exception;
-                        args.Handled = true;
-                        Dispatcher.CurrentDispatcher.InvokeShutdown();
-                    };
+            Application app = Application.Current ?? new Application();
+            app.ShutdownMode = ShutdownMode.OnExplicitShutdown;
```
mconnew, line 29: If you have multiple WPF based tests, it might be worth using an IFixture to set these things up. But if it's the only test, not worth bothering with.

mconnew (approved):
:shipit:

PR #2434, opened 2026-06-03
Support embedded portable PDBs for managed symbol resolution

JeremyKuhne:
## Summary

Managed assemblies built with `<DebugType>embedded</DebugType>` carry their portable PDB inside the PE image rather than in a standalone `.pdb` file. Previously `SymbolReader` could not read these, so source/line lookup failed for such modules. This PR adds the ability to read embedded portable PDBs.

## Changes

- **`SymbolReader.OpenEmbeddedPortablePdb`** — reads the `EmbeddedPortablePdb` debug-directory entry from a module and returns a `PortableSymbolModule`. Results are cached under a key suffixed so they cannot collide with standalone-PDB cache entries.
- **`PortableSymbolModule`** — gains a constructor that takes a `MetadataReaderProvider` (which owns its own backing memory, so the module survives disposal of the `PEReader`/`FileStream` used to open it).
- **`TraceLog.OpenPdbForModuleFile`** — falls back to the module's embedded portable PDB when the on-disk module matches the trace and no standalone PDB exists.

## Tests

Adds an `EmbeddedPdbTestApp` fixture (built with `<DebugType>embedded</DebugType>`) and covers:

- the happy path (source/line resolution),
- readability after the `PEReader` is disposed,
- caching,
- not-embedded and missing-file cases,
- the end-to-end `TraceLog` fallback path.

All tests run cross-platform on **net462** and **net8.0** in Debug.

### DIFF
--- a/src/TraceEvent/Symbols/SymbolReader.cs
+++ b/src/TraceEvent/Symbols/SymbolReader.cs
@@ -4,6 +4,8 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
 using System.Text;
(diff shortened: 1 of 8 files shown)


JeremyKuhne:
Also tried the nupkg in a project where I need this support and it worked fine.

src/TraceEvent/Symbols/SymbolReader.cs
```diff
@@ -711,6 +713,115 @@ public NativeSymbolModule OpenNativeSymbolFile(string pdbFileName)
...
+        ///      <see cref="OpenSymbolFile(string)"/>.
+        ///   2. Failing that, a portable PDB embedded inside the module image itself
+        ///      (modules built with &lt;DebugType&gt;embedded&lt;/DebugType&gt;), via
+        ///      <see cref="OpenEmbeddedPortablePdb(string)"/>.
+        ///
+        /// Returns null if no symbols can be found for the module.
+        /// </summary>
+        /// <param name="moduleFilePath">The path to the managed/native module (.dll/.exe) to resolve symbols for.</param>
+        /// <returns>The symbol module for the given module, or null if none could be found.</returns>
+        public ManagedSymbolModule OpenSymbolFileForModuleFile(string moduleFilePath)
```
brianrob: It looks like this API isn't called from anywhere other than tests.  I don't think we actually need this, since you've already added embedded support into `TraceLog.OpenPdbForModule`.

src/TraceEvent/TraceLog.cs
```diff
@@ -9131,6 +9131,21 @@ private unsafe ManagedSymbolModule OpenPdbForModuleFile(SymbolReader symReader,
                 if (TraceModuleUnchanged(moduleFile, symReader.m_log))
                 {
                     pdbFileName = symReader.FindSymbolFilePathForModule(moduleFile.FilePath);
+
```
brianrob, line 9134: This will work for traces that are captured on the same machine before the binaries are removed or overwritten.  Should we be looking to capture the PDBs during trace merge so that anyone who captures a trace where binaries have embedded PDBs will have them extracted and zipped up with the etl (ala NGEN PDBs from .NET Framework)?

brianrob (commented):
I appreciate you putting this together.  A couple of comments below.

src/TraceEvent/Symbols/SymbolReader.cs
JeremyKuhne: Fair enough, this was just sugar.

src/TraceEvent/TraceLog.cs
JeremyKuhne: Yeah, I think so. I can do a follow up that explores that if you'd like. :)

brianrob (approved):
@JeremyKuhne thank you for contributing this.  Apologies - I had reviewed this, but forgot to approve/merge.

PR #2440, opened 2026-07-14
Fix GC dynamic ETLX replay

### LINKED ISSUE #2438: ArgumentOutOfRangeException opening nettrace GC stats and Heap analyzer
This is my first time using `dotnet-trace` and `PerfView`. I wanted to profile some memory problems in my application and the trace step seemed to work just fine. But then trying to open the GC stats page and the Heap Analyzer page in PerfView resulted in an exception and a blank UI.

I ran dotnet-trace 9.0.661903 like this
```shell
dotnet-trace collect --profile gc-verbose -- ./redacted.exe
```

and then ran perfview 3.2.4. Log is attached

[perfview redacted log.txt](https://github.com/user-attachments/files/29563427/perfview.redacted.log.txt)

I'm not sure what else to provide. But I can certainly try to get it if requested. I am new to both of these applications so I don't have much terminology to describe what I did.

brianrob:
PerfView 3.2.4 can throw `ArgumentOutOfRangeException` when GC Stats or Heap Analyzer replays `GC/CommittedUsage` events from ETLX. During conversion, `FixupData()` classifies the raw dynamic event and the ETLX persists the synthetic event ID, but the original dynamic payload envelope remains unchanged. On replay, the typed template therefore still needs to parse that envelope.

This change refreshes the payload layout in `EventPayload` only when processing a `TraceLog`. Raw ETW and EventPipe dispatch continue using the payload prepared by `FixupData()`, avoiding redundant parsing. Fixed-offset `CommittedUsage` fields also return safe defaults for truncated payloads.

Regression coverage includes valid ETLX replay without `FixupData()`, reused replay templates with distinct payloads, and malformed payload access through properties, `PayloadValues`, and `ToXml`.

Fixes #2438

### DIFF
--- a/src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/GCDynamicTraceEventParser.cs
@@ -127,29 +127,14 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.GCDynamic
         }
 
         /// <summary>
-        /// These are the raw payload fields of the underlying event.  TraceLog replay
-        /// reuses templates and does not call <see cref="FixupData"/>, so replayed events
-        /// decode the current layout on access.  Invalid layouts return type-appropriate
-        /// safe defaults.
+        /// These are the raw payload fields of the underlying event.  They read through
+        /// the cached <see cref="_payload"/> populated before the typed event is accessed.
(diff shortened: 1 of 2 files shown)


leculver (approved):

brianrob:
@microsoft-github-policy-service rerun

PR #2441, opened 2026-07-15
Enable W4 and warnings as errors for C++ projects

brianrob:
Enables /W4 and /WX for both ETW CLR profiler projects and fixes the resulting warnings.


src/EtwClrProfiler/CorProfilerTracer.cpp
```diff
@@ -561,7 +570,6 @@ STDMETHODIMP CorProfilerTracer::ObjectAllocated(ObjectID objectId, ClassID class
 
 			// We want to sample at a rate that ensures less 100 allocations per second per type.  
 			// However don't sample less than 1/1000, 
-			int oldSamplingRate = classInfo->SamplingRate;
```
mconnew, line 564: If you're removing this, should you remove the commented TODO below which would have used it?

mconnew (dismissed):
Just one small bit of potential cleanup related to one of the cleaned up lines.

src/EtwClrProfiler/CorProfilerTracer.cpp
brianrob: Good catch. Removed the stale commented debugging block.

mconnew (approved):

PR #2443, opened 2026-07-17
Support TraceLogging formatting hints

### LINKED ISSUE #2436: Support TraceLogging formatting hints like TraceLoggingHexUInt32
It would be nice if PerfView used the display formatting hints that the [C++ TraceLogging library](https://learn.microsoft.com/windows/win32/tracelogging/trace-logging-portal) lets you write. For example, there are hints to indicate that an integer value should hex formatted on display or that the value is a HRESULT. See [the TraceLogging Wrapper Macros docs](https://learn.microsoft.com/windows/win32/tracelogging/tracelogging-wrapper-macros) for the full list.

```
TraceLoggingWrite(
    g_hMyProvider,
    "MyEvent",
    TraceLoggingUInt32(value1ToLog),
    TraceLoggingHexUInt32(value2ToLog));

```

In PerfView today, this would be rendered as 

> value1ToLog="1" value2ToLog="32,769"

If the hints were used, this could be rendered like this:

> value1ToLog="1" value2ToLog="0x8001"

For things like flags this can be easier to read at a glance. There are also common values that are represented in hex, like PCI IDs.

As best I can tell, this metadata gets attached to the each event via the [EVENT_DATA_DESCRIPTOR struct](https://learn.microsoft.com/windows/win32/api/evntprov/ns-evntprov-event_data_descriptor) with a Type of EVENT_DATA_DESCRIPTOR_TYPE_EVENT_METADATA.

In the copy of TraceLoggingProvider.h in the 10.0.26100.0 Windows SDK (`\Include\10.0.26100.0\shared\TraceLoggingProvider.h`), it looks like there's a description of how this metadata is serialized starting at line 2421 (a comment that starts with "This comment defines the TraceLogging encoding format.")

WPA is able to make use of these hints already. Perhaps there's already parsing code in https://github.com/microsoft/microsoft-performance-toolkit-sdk or a similar code base that could be re-used.

chwarr:
Trace Data Helper lets you add display hints to fields in events. For example, there are hints to indicate that an integer value should hex formatted on display or that the value is a HRESULT. The commonly used [C++ TraceLogging library][0] supports these hints via its [TraceLogging Wrapper macros][1].

Add support for pretty printing the following _TDH_OUT_TYPE values:

* TDH_OUTTYPE_HEXINT8 - hex, width of the underlying type
* TDH_OUTTYPE_HEXINT16 - hex, width of the underlying type
* TDH_OUTTYPE_HEXINT32 - hex, width of the underlying type
* TDH_OUTTYPE_HEXINT64 - hex, width of the underlying type
* TDH_OUTTYPE_PID - decimal number without digit seperators
* TDH_OUTTYPE_TID - decimal number without digit seperators
* TDH_OUTTYPE_PORT - decimal number without digit seperators, converted from network byte order
* TDH_OUTTYPE_IPV4 - dotted quad
* TDH_OUTTYPE_IPV6 - colon-seperated, elision of runs of zeros
* TDH_OUTTYPE_SOCKETADDRESS - "address:port" for AF_INET and AF_INET6
* TDH_OUTTYPE_ERRORCODE - hex, no padding
* TDH_OUTTYPE_WIN32ERROR - hex, no padding
* TDH_OUTTYPE_NTSTATUS - hex, always 8 digits wide
* TDH_OUTTYPE_HRESULT - hex, always 8 digits wide
* TDH_OUTTYPE_CODE_POINTER - lowercase hex, minimum of 8 digits wide

Fixes #2436.

Given TraceLogging code like this:

```
TraceLoggingWrite(
    g_hMyProvider,
    "MyEvent",
    TraceLoggingUInt32(value1ToLog),
    TraceLoggingHexUInt32(value2ToLog));
```

In PerfView today, this would be rendered as

```
    value1ToLog="1" value2ToLog="32,769"
```

Now, it is rendered as

```
    value1ToLog="1" value2ToLog="0x00008001"
```

Visually, before:

screenshot of events before this change: no hex formatting

and after:

screenshot of events after this change: values are hex formatted, IP addresses are pretty-printed

For things like flags these format hints can make it easier to read values at a glance. There are also common values that are represented in hex, like PCI IDs.

When PerfView already had support for formatting a value, its existing formating style was followed. Otherwise, I matched WPA's formatting.

Changes made:

* Changed TraceEvent's representation of the output type from a ushort to the new TdhOutputType enum.
* Add TdhFormtter.cs that knows how to format a value with a given formatting hint. A formatting hint was chosen so that the PayloadFetch object didn't need to remember both its TdhInputType and its TdhOutputType.
* Reved the FastSerializable format to include the new formatting hint.
* Introduced a new FormatUtils helper for shared formatting code between TraceEvent.cs and DynamicTraceEventParser.cs
* Unit tests for the new formatting logic added.
* Updated the baseline files to capture the changes in formatting.

[0]: https://learn.microsoft.com/windows/win32/tracelogging/trace-logging-portal
[1]: https://learn.microsoft.com/windows/win32/tracelogging/tracelogging-wrapper-macros


src/TraceEvent/Utilities/FormatUtils.cs
```diff
@@ -0,0 +1,87 @@
...
+        public static string FormatSockaddr(ReadOnlySpan<byte> bytes)
+        {
+            if (bytes.Length < 16)
+            {
+                return null;
+            }
+
+            var family = (AddressFamily)(bytes[0] | (bytes[1] << 8));
+
+            Debug.Assert(BitConverter.IsLittleEndian, "Bit twiddling below assumes a little-endian host machine");
```
brianrob: There is work in-progress to support big-endian, so we should make sure to cover that here as well.

src/TraceEvent/DynamicTraceEventParser.cs
```diff
@@ -1734,6 +1827,9 @@ public PayloadFetchArrayInfo Array
 
             public Type Type;       // Currently null for arrays.  
 
+            // Display-formatting hint derived from the TraceLogging/TDH InType/OutType.
+            public TdhFormatter.FormatHint FormatHint;
```
brianrob: Please make this a property instead of a field.

src/TraceEvent/TdhFormatter.cs
```diff
@@ -0,0 +1,451 @@
...
+using System.Text;
+
+namespace Microsoft.Diagnostics.Tracing.Parsers
+{
+    /// <summary>
+    /// Formats values according to a formatting hint (see <see
+    /// cref="FormatHint"/>). The hints should be computed from a field's
+    /// TDH InType/OutType.
+    /// </summary>
+    public static class TdhFormatter
```
brianrob: Let's mark this class as internal to avoid new surface area.

brianrob (commented):
@chwarr, thank you for contributing this.  This is quite a bit of work, and overall looks great.  A few small requests below.

src/TraceEvent/DynamicTraceEventParser.cs
brianrob: I think you will also need to update this path as well to ensure that structs use their formatting hints as well.

src/TraceEvent/Utilities/FormatUtils.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                // Both the address and port are in network byte order in the sockaddr struct.
+                // IPEndPoint ctor wants the port in host order...
+                int port = (UInt16)(bytes[2] << 8 | bytes[3]);
+                // ... and the address in network byte order and a byte[16].
+                var ipv6 = new byte[16];
+                bytes.Slice(start: 8, length: 16).CopyTo(ipv6);
+                try
+                {
+                    return new IPEndPoint(new IPAddress(ipv6), port).ToString();
```
brianrob: From copilot:

> `SOCKADDR_IN6` includes a scope/interface ID in bytes 24–27. This code validates the full 28-byte structure but ignores that field, so scoped IPv6 addresses—especially link-local addresses—lose the interface identifier when formatted. Could we read `sin6_scope_id` and pass it to `new IPAddress(ipv6, scopeId)`?

src/TraceEvent/DynamicTraceEventParser.cs
chwarr: Added. There don't appear to be any structs in the baseline traces that this impacted. I validated with my own trace and it's working as expected. I also added unit tests for this code path.

Before:

before screenshot


After:

after screenshot

src/TraceEvent/Utilities/FormatUtils.cs
chwarr: This code was basically endian neutral before. I switched to the `IPAddress(byte[])` ctor for the IPv4 case to avoid a call to HostToNetworkLong and introduce two helper methods to extract UInt16 and UInt32 values from a `ReadOnlySpan<byte>` byte-by-byte. This is the same logic as before, but with better names. The byte-by-byte conversion avoid an endian issues.

chwarr:
I think I addressed all your feedback, @brianrob. When you have a moment to review, I'd appreciate another look. Thanks!

brianrob (approved):
LGTM.  Thank you!  This is a great addition.

PR #2445, opened 2026-07-20
Fix AccessViolation on malformed MethodILToNativeMap entry count

brianrob:
## Summary
Fixes an uncatchable **AccessViolationException (memory-safety / denial of service) on malformed EventPipe/ETW input** in the TraceEvent library.

`TraceLog.AddILMapping` trusts `CountOfMapEntries` — a 16-bit count read from fully untrusted `MethodILToNativeMap` (EventID 190) event payload — when reading the per-entry `ILOffset`/`NativeOffset` values. Because `NativeOffset(i)` is indexed *relative to* `CountOfMapEntries`, a corrupt or fuzzed stream that claims more entries than the payload actually contains makes the reads index past the end of the event buffer and access protected memory, crashing the process with an `AccessViolationException`.

Reachable by any caller that parses untrusted data (e.g. `TraceLog.CreateFromEventPipeDataFile`).

## How it was found
Discovered by the TraceEvent nettrace fuzzer.

## Fix
- Reject `MethodILToNativeMap` events whose payload is too small for the claimed entry count (a valid event has `EventDataLength >= CountOfMapEntries * 8 + 21`), and record the skip in the conversion log.
- Add regression test `MalformedILToNativeMapEntryCountDoesNotAccessViolation` that drives the full `TraceLog` conversion path with such a malformed event and asserts it is rejected (deterministic; fails without the fix).

### DIFF
--- a/src/TraceEvent/Parsers/ClrTraceEventParser.cs
+++ b/src/TraceEvent/Parsers/ClrTraceEventParser.cs
@@ -10635,6 +10635,7 @@ namespace Microsoft.Diagnostics.Tracing.Parsers.Clr
             : base(eventID, task, taskName, taskGuid, opcode, opcodeName, providerGuid, providerName)
         {
             Action = action;
+            NeedsFixup = true;
         }
         protected internal override void Dispatch()
         {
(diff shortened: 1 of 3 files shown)


src/TraceEvent/Parsers/ClrTraceEventParser.cs
```diff
@@ -10645,10 +10646,30 @@ protected internal override Delegate Target
             get { return Action; }
             set { Action = (Action<MethodILToNativeMapTraceData>)value; }
         }
+        internal override void FixupData()
```
mconnew, line 10649: As this is an override, do you need to call the method on base?

mconnew (approved):

src/TraceEvent/Parsers/ClrTraceEventParser.cs
brianrob: Nope - it's just an empty method body.

PR #2447, opened 2026-08-06
Add Accessible Label To Directory Search Field

brianrob:
The main view's directory search field lacks a persistent visible label, making its purpose harder to identify.

Add a visible **Path:** label beside the field and associate it through WPF's `Label.Target` and `AutomationProperties.LabeledBy`. Existing directory history, tooltip, and keyboard navigation behavior remain unchanged.

### DIFF
--- a/src/PerfView/MainWindow.xaml
+++ b/src/PerfView/MainWindow.xaml
@@ -178,7 +178,7 @@
                 <DockPanel DockPanel.Dock="Top">
                     <Label x:Name="DirectoryLabel" Content="Path:" Target="{Binding ElementName=Directory}"
                            VerticalAlignment="Center" Padding="5,0" />
-                    <controls:HistoryComboBox x:Name="Directory" AutomationProperties.Name="Path"
+                    <controls:HistoryComboBox x:Name="Directory" AutomationProperties.Name="Directory Search Bar"
                              AutomationProperties.LabeledBy="{Binding ElementName=DirectoryLabel}"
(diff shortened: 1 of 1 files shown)


src/PerfView/MainWindow.xaml
```diff
@@ -175,8 +175,13 @@
                 <ColumnDefinition Width="619*"/>
             </Grid.ColumnDefinitions>
             <DockPanel Background="{DynamicResource ContainerBackground}" Grid.Column="0">
-                <controls:HistoryComboBox x:Name="Directory" DockPanel.Dock="Top" AutomationProperties.Name="Directory Search Bar"
-                         ToolTip="This is the directory to search for performance data files." TextEntered="DoTextEnteredInDirectoryTextBox" />
+                <DockPanel DockPanel.Dock="Top">
+                    <Label x:Name="DirectoryLabel" Content="Path:" Target="{Binding ElementName=Directory}"
+                           VerticalAlignment="Center" Padding="5,0" />
+                    <controls:HistoryComboBox x:Name="Directory" AutomationProperties.Name="Path"
```
mconnew: This is changing the AutomationProperties.Name from "Directory Search Bar" to "Path", which is less descriptive than it was before. I don't know whether this is exposed as text spoken when using the accessibility feature that describes the selected item, but at the very least is someone has been automating this up to this point, this breaks their automation.

brianrob: Good catch. I restored `AutomationProperties.Name` to `Directory Search Bar` to preserve the existing spoken label and automation identifier while retaining the new visible `Path:` label and `LabeledBy` association.

mconnew (approved):

PR #2449, opened 2026-08-06
Fix Main Window Keyboard Focus Order

brianrob:
Keyboard users could move from the top Videos link directly to bottom status-bar controls because the main window's WPF logical order did not match its visual layout.

## Summary

- Align the main window's logical and visual layout order.
- Keep status-bar controls keyboard accessible after the main content.
- Add regression coverage that detects premature status-bar focus or a focus cycle.

### DIFF
--- a/src/PerfView.Tests/MainWindowTests.cs
+++ b/src/PerfView.Tests/MainWindowTests.cs
@@ -1,0 +1,81 @@
+using PerfView;
+using PerfViewTests.Utilities;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Documents;
+using System.Windows.Input;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace PerfViewTests
+{
+    public class MainWindowTests : PerfViewTestBase
+    {
+        public MainWindowTests(ITestOutputHelper testOutputHelper)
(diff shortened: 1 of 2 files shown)


mconnew (approved):
I did a quick deep dive on DockPanel and why it needed to be removed. It has a requirement that the element which fills the center of the panel is declared last, which interferes with the tab order. If you moved the StatusBar to the end, it would fill most of the panel instead, which isn't what you need. There are two possible fixes. You can either drop the DockPanel and switch to a Grid (and I couldn't find any functional difference, it's just a different way of placing elements), or you can specify the TabIndex attribute on each of the items to specify the precedence order when tabbing through them. Either is valid. I'm mentioning it so if it's later discovered there is a problem with the approach in this PR, you can try TabIndex instead.

brianrob:
> I did a quick deep dive on DockPanel and why it needed to be removed. It has a requirement that the element which fills the center of the panel is declared last, which interferes with the tab order. If you moved the StatusBar to the end, it would fill most of the panel instead, which isn't what you need. There are two possible fixes. You can either drop the DockPanel and switch to a Grid (and I couldn't find any functional difference, it's just a different way of placing elements), or you can specify the TabIndex attribute on each of the items to specify the precedence order when tabbing through them. Either is valid. I'm mentioning it so if it's later discovered there is a problem with the approach in this PR, you can try TabIndex instead.

Thanks @mconnew.  Much appreciated.

Issue #2451, opened 2026-08-09
[TraceEvent] TraceCallStacks interning tables grow without bound on the real-time/EventPipe path in long-lived processes (~0.6 GB/day to OOM)


therickfactr:
Opening this as a new issue rather than commenting on #1199, since that one was closed in 2022
because the original reporter had stopped seeing it (`"I haven't seen any alerts with this recently,
so go ahead and close"`) rather than because a cause was found or ruled out. Happy to have this
merged back into #1199 if you'd prefer it there.

Same array and same growth shape — `TraceCallStacks+CallStackInfo[]` growing without bound — but on
the **real-time EventPipe path inside a long-lived server process** rather than during offline trace
analysis. That difference seems to be the crux, so it may warrant a separate answer.

### Production symptom

A .NET identity service (ASP.NET Core, Linux, single instance) accumulating **~0.6 GB/day**
proportional to traffic, ending in an OOM kill. Convicted by an A/B gcdump comparison on the same
build under the same workload:

| | live heap | dominant types |
|---|--:|---|
| profiling enabled | **478 MB** (8.67M objects) | `Microsoft.Diagnostics.Tracing.*` |
| profiling disabled | **31 MB** (470k objects) | no TraceEvent types present, RSS falling |

The consumer is Sentry's .NET profiler, which starts a single continuous `TraceLog` real-time session
for the lifetime of the process. Tracked on their side at getsentry/sentry-dotnet#5469, where a
maintainer independently reproduced it in a minimal ASP.NET-free harness — bare EventPipe + TraceLog
plus synthetic stack-shape churn — and landed in the same place in the code.

### Mechanism, as far as we can tell

`TraceCallStacks.InternCallStackIndex` interns every distinct stack shape into `callStacks` and
appends to the parallel `callees` list. `TraceLog`'s real-time mode does trim — but only
`eventsToStacks`, `eventsToCodeAddresses` and `cswitchBlockingEventsToStacks`, which on the EventPipe
path are already cleared per event. The interning tables themselves (`callStacks`, `callees`,
`codeAddresses`) appear to have no eviction path at all.

Two observations that may be useful:

- **Growth tracks distinct stack _shapes_, not sample volume.** With a single fixed stack shape, 11×
  the samples grew the table only 24%. Real server traffic keeps minting new shapes — async state
  machines, generics, DI-heavy call paths — which is why an idle instance never reproduces it and a
  busy one does. It also means the rate depends on application shape more than on request count.
[...]

Issue #1199, opened 2020-06-08
[TraceEvent] Microsoft.Diagnostics.Tracing.Etlx.TraceCallStacks+CallStackInfo[] - High memory usage


suprak:
We've noticed that ever since update to the latest version of TraceEvent (from a much older one), we have been seeing much higher memory of TraceEvent. Mostly related to symbol resolution.

For instance from a dump (I can share internally),

```
0:000> !do 000001637fc31020
Name:        Microsoft.Diagnostics.Tracing.Etlx.TraceCallStacks+CallStackInfo[]
MethodTable: 00007ffe66678120
EEClass:     00007ffe666780a0
Size:        177629656(0xa9669d8) bytes
Array:       Rank 1, Number of elements 22203704, Type VALUETYPE (Print Array) 

0:000> !do 00000163b25e1020
Name:        Microsoft.Diagnostics.Tracing.Etlx.TraceCallStacks+CallStackInfo[]
MethodTable: 00007ffe66678120
EEClass:     00007ffe666780a0
Size:        134254640(0x8009030) bytes
Array:       Rank 1, Number of elements 16781827, Type VALUETYPE (Print Array)
```

That's 22million and 16milion entries per array.

Any plans or ideas on how TraceEvent memory usage can be improved?
We rely on TraceEvent as part of a production profiling workflow (can provide my context internally if you would like).

brianrob:
@suprak, do can you tell me which versions you're referring to here?  I suspect this might be caused by #1136.

suprak:
Sorry, should have mentioned that in my original post,

`<PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.49" />`

brianrob:
No worries.  Is 2.0.49 the new version that you're running or is that the old version?

suprak:
We upgraded from `2.0.2` back in February.
So `2.0.49` is what we are using currently.

brianrob:
@suprak, is this still an issue or should we close this?

suprak:
I haven't seen any alerts with this recently, so go ahead and close.

therickfactr:
For anyone finding this thread later: we're seeing the same unbounded `TraceCallStacks+CallStackInfo[]` growth, but on the **real-time/EventPipe path in a long-lived server process** rather than during offline analysis, with production evidence (~0.6 GB/day to OOM, convicted by an A/B gcdump).

[...]

PR #2452, opened 2026-08-12
Add TraceLog.TrimLiveSessionState to bound real time call stack growth

jamescrosswell:
## Background

The .NET runtime includes [EventPipe](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/eventpipe) which creates a *nettrace* stream for runtime events that are used for profiling. `TraceLog`, from the TraceEvent library in PerfView, turns those raw events into something symbolic — it tracks processes, threads, modules, methods and code addresses so that `TraceEvent.CallStack()` returns frames rather than hex addresses.

Call stacks are the bulk of that data and the most redundant: consecutive samples usually share almost every frame. So `TraceCallStacks` does not store stacks, it interns them into a prefix tree. Each node is `CallStackInfo { codeAddressIndex, callerIndex }`, a whole stack is identified by a single `CallStackIndex`. This structure is maintained in three arrays: `callStacks` (nodes), `callees` (each node's children, searched when interning) and `threads` (per-thread roots).

For a trace **file** this works really well. The input is finite, and while you are analysing it you want every stack it contains to stay resolvable, so the tree is deliberately a cache with no eviction.

## Real time use case

microsoft/perfview#1867 added streaming / in-memory EventPipe support, and microsoft/perfview#2169 added the rundown-provider API that lets frames resolve mid-session. Those made `TraceLog` usable against a live session rather than a file — which is what continuous profilers need, and what the Sentry .NET SDK is built on.

The data model came along unchanged. `InternCallStackIndex` still appends, and nothing is ever released. If you point that at a server process that runs for weeks then the tree grows for the life of the session (i.e. indefinitely).

Additionally:

- **It grows with stack *diversity*, not sample volume.** Repeated stacks are free — they resolve to   an existing node. Every new stack *shape* appends nodes. A fresh async state machine, a new code path, a JIT'd generic each add more, so a busy service keeps minting them indefinitely.
- **It grows whether or not anyone is consuming stacks.** The session interns everything it sees, so   a profiler that samples 10% of transactions pays the same growth as one that samples all of them.

## This is already known

- microsoft/perfview#1199 reported the same `CallStackInfo[]` growth in 2022. It was closed as not  planned because the original reporter stopped seeing it, not because it was resolved or ruled out. 
- microsoft/perfview#2451 was opened recently by a Sentry customer running into this issue. 
- `FlushRealTimeEvents` already carries deliberate memory management for real time sessions ([TraceLog.cs#L952-L955](https://github.com/microsoft/perfview/blob/af227405c475b7d8a04f2a4353bc28fccc379b3a/src/TraceEvent/TraceLog.cs#L952-L955)):

  > Try to keep our memory under control by removing old data.
  > Lots of data structures in TraceLog can grow over time.
  > However currently we only trim three, all CAN grow on every event (so they grow most quickly of
  > all data structures) and we know they are not needed after dispatched the events they are for.

The three it trims are the small event-to-stack maps. **This PR adds a fourth — the largest one**, roughly fourteen times the size of the next biggest structure in the dump above.

## The change

The addition of a new `TraceLog.TrimLiveSessionState()` method that discards the interned call stacks and returns the tree to its initial state.

**It is inert for every existing consumer.** It is opt in, nothing calls it, no existing behaviour changes, and it throws outside a real time session.

Two implementation details worth review attention:

1. It assigns **fresh** `GrowableArray`s rather than calling `GrowableArray.Clear()`. `Clear()` only resets the length and keeps the backing store — and for `callees` and `threads` that store holds a reference to every child `List<CallStackIndex>` ever created, which is the bulk of the memory being released.
2. It also clears `eventsToStacks` and `cswitchBlockingEventsToStacks`, which map events to the indexes just invalidated. Those hold plain structs, so a length reset genuinely is enough there — they are cleared for correctness, not for memory.

## Measured

Synthetic workload, 600 s, real time EventPipe session, .NET 9, Linux container:

| | without reset | with reset |
|---|--:|--:|
| interned call stacks at 600 s | 1,231,862, still climbing | bounded, cycling 919 – 7,121 |
| managed heap | 3.4 → 119.2 MiB | 3.5 → ~7 MiB, flat |
| RSS | 65.2 → 200.7 MiB | 65.6 → 81.4 MiB |
| samples processed | 307,418 | 340,987 |

After warm up the reset run moved **+2.1 MiB over 9 minutes and 174 resets** — about 0.012 MiB per reset. Sample throughput went *up*, because the baseline spends real time growing and collecting those tables. The workload is deliberately far more stack-diverse than a real service, to make the effect measurable in minutes; the shape matches the field report below, the rate is exaggerated.

From the field — getsentry/sentry-dotnet#5469, an ASP.NET Core service on Linux growing ~0.6 GB/day until the kernel OOM-killed it at ~8 days of uptime. `dotnet-gcdump` showed the two halves of this tree at **exactly** the same size:

```
35,087,256   CallStackInfo[]
35,087,256   List<CallStackIndex>[]
 2,548,664   MethodInfo[]
 2,548,536   CodeAddressInfo[]
```

Setting the profile sample rate to 0 — the only thing that stops a session being created at all — cured it completely, with every TraceEvent type disappearing from the heap.

## Why the other tables are left alone

`codeAddresses`, `methods` and `ILToNativeMap` also grow, and are deliberately untouched.

Clearing them would be **irreversible**. The rundown that populates method names for code JIT'd *before* the session started runs once, as a separate short-lived session at `TraceLog` creation, and cannot be re-run against a live session.

They are not *strictly* bounded — tiered compilation, expression trees, `RegexOptions.Compiled` and reflection emit all mint new methods over time — so a residual, much smaller growth remains for apps doing continuous runtime codegen.

Addressing this needs a different mechanism, and is left as separate work.

## The contract `TrimLiveSessionState` asks of callers

Universal preconditions:
1. No `CallStackIndex` obtained earlier may still be in use
2. It must be called from within an event callback, on the thread dispatching that event.
   This is now enforced rather than merely documented — `TrimLiveSessionState` calls
   `ThrowIfNotRunningOnDispatchThread()`, which validates against the dispatcher's thread ID.

What is *not* universal is when a given consumer can know its indexes are dead — a consumer that resolves synchronously in the callback can reset between any two events, while one that caches by index can only do so between cache lifetimes. That depends entirely on its own retention pattern, so the API states the invariant and leaves the timing to the caller.

This is also why we've added an explicit method rather than, say, a `MaxCallStackCount` property that resets itself inside the event pump. 

The dispatch-thread check is deliberately generic rather than named for this feature: `TraceLog` now tracks the thread ID of the in-progress dispatch on both real time paths (EventPipe, and the cloned-event ETW path), so other operations that mutate state the dispatcher is actively using can reuse the same guard.

## Test

`EventPipeParsing.TrimLiveSessionStateDiscardsInternedStacksAndKeepsInterning`, modelled on the existing
`SessionStreaming` test. It starts a real time session, lets stacks accumulate, resets from inside the sample callback, then asserts both halves of the contract: the tables are emptied, *and* stacks interned afterwards still resolve through to a method name.

It also walks the entire post-reset caller chain, range checking every link. That is deliberate — `GrowableArray`'s indexer only guards with `Debug.Assert`, so in Release a stale index reads the backing store and silently resolves to the wrong frame instead of throwing. Asserting one frame resolves is not enough to catch that; walking the chain is.

Runs in ~350 ms, and is skipped below .NET Core 3.0 like its neighbour.

`TrimLiveSessionStateThrowsWhenCalledOffTheDispatchThread` covers the enforcement from the other side: calling from the test thread while a dispatch is live throws. Together the two pin both halves — the guard admits the legitimate in-callback call and rejects the off-thread one.

**A known gap, flagged deliberately.** The test does not fail if `threads` is left uncleared — verified this by mutation. Stale per-thread roots resolve to `default(CallStackInfo)`, whose code address index is 0, which is almost never a valid/real address, so they sit inert and get appended to. The consequence is retained memory plus a rare correctness risk (a code-address-0 collision, or `IndexOutOfRangeException` once a stale index exceeds the new capacity) rather than deterministic corruption. A memory-based test would cover it but seems a poor trade for CI time. Suggestions welcome.

### DIFF
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -7864,12 +7864,8 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
         /// </summary>
         internal void Clear()
         {
-            // Deliberately assign fresh arrays rather than calling GrowableArray.Clear().  Clear()
-            // only resets the length, so the backing arrays would survive - and for 'callees' and
-            // 'threads' those arrays hold references to every List<CallStackIndex> ever created,
-            // which is the bulk of the memory we are trying to release.
-            // InternCallStackIndex reallocates callStacks/callees on its next call, and 'threads'
-            // grows on demand, so no explicit re-initialization is needed here.
+            // GrowableArray.Clear() only resets the length, so assign fresh arrays to actually
+            // release the memory - callees and threads reference every List ever created.
             callStacks = new GrowableArray<CallStackInfo>();
             callees = new GrowableArray<List<CallStackIndex>>();
             threads = new GrowableArray<List<CallStackIndex>>();
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -976,6 +976,43 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
         }
 
         /// <summary>
+        /// Discards every call stack interned so far, releasing the memory held by the call stack
+        /// interning tables (see <see cref="CallStacks"/>).
+        /// <para>
+        /// In a real time session those tables grow for the lifetime of the session: every distinct
+        /// call stack that is observed is interned and nothing is ever released.  For a long running
+        /// process with diverse stacks that growth is unbounded.  Trace files do not have this problem
+        /// because the session is finite, which is why this is restricted to real time sessions.
+        /// </para>
+        /// <para>
+        /// Every <see cref="CallStackIndex"/> obtained before this call is invalidated and must not be
+        /// used afterwards, so only call this when no such index is still live.  Call stacks observed
+        /// after the call are interned from scratch, so no information is lost going forward - only
+        /// the ability to resolve indexes handed out earlier.
+        /// </para>
+        /// <para>
+        /// Must be called from the thread that processes events (for example from within an event
+        /// callback).  Calling it concurrently with event processing races with call stack interning.
+        /// </para>
+        /// </summary>
+        /// <exception cref="InvalidOperationException">The TraceLog is not a real time session.</exception>
+        public void ResetCallStacks()
+        {
+            if (!IsRealTime)
+            {
+                throw new InvalidOperationException("ResetCallStacks is only supported for real time sessions.");
+            }
+
+            callStacks.Clear();
+
+            // These map events to the call stack indexes we have just invalidated, so they cannot be
+            // allowed to survive.  On the EventPipe real time path they are already cleared after every
+            // event; on the ETW path FlushRealTimeEvents only trims them, so clear them explicitly.
+            eventsToStacks.Clear();
+            cswitchBlockingEventsToStacks.Clear();
+        }
+
+        /// <summary>
         /// Given a process's virtual address 'address' and an event which acts as a
         /// context (determines which process and what time in that process), return
         /// a CodeAddressIndex (which represents a particular location in a particular
@@ -7819,6 +7856,24 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
         internal void SetSize(int origSize)
         {
             callStacks.RemoveRange(origSize, callStacks.Count - origSize);
+        }
+
+        /// <summary>
+        /// Discards every interned call stack, returning the interning tables to their initial state.
+        /// Only meaningful for a real time session, where these tables would otherwise grow for the
+        /// lifetime of the session.  See <see cref="TraceLog.ResetCallStacks"/>.
+        /// </summary>
+        internal void Clear()
+        {
+            // Deliberately assign fresh arrays rather than calling GrowableArray.Clear().  Clear()
+            // only resets the length, so the backing arrays would survive - and for 'callees' and
+            // 'threads' those arrays hold references to every List<CallStackIndex> ever created,
+            // which is the bulk of the memory we are trying to release.
+            // InternCallStackIndex reallocates callStacks/callees on its next call, and 'threads'
+            // grows on demand, so no explicit re-initialization is needed here.
+            callStacks = new GrowableArray<CallStackInfo>();
+            callees = new GrowableArray<List<CallStackIndex>>();
+            threads = new GrowableArray<List<CallStackIndex>>();
         }
 
         /// <summary>
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -7864,8 +7864,7 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
         /// </summary>
         internal void Clear()
         {
-            // GrowableArray.Clear() only resets the length, so assign fresh arrays to actually
-            // release the memory - callees and threads reference every List ever created.
+            // GrowableArray.Clear() only resets the length so we have to assign fresh arrays
             callStacks = new GrowableArray<CallStackInfo>();
             callees = new GrowableArray<List<CallStackIndex>>();
             threads = new GrowableArray<List<CallStackIndex>>();
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -976,8 +976,8 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
         }
 
         /// <summary>
-        /// Discards every call stack interned so far, releasing the memory held by the call stack
-        /// interning tables (see <see cref="CallStacks"/>).
+        /// Releases the state a live session can afford to discard.  Today that is the call stack
+        /// interning tables (see <see cref="CallStacks"/>); more may be trimmed here in future.
         /// <para>
         /// In a real time session those tables grow for the lifetime of the session: every distinct
         /// call stack that is observed is interned and nothing is ever released.  For a long running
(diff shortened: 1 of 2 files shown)


jamescrosswell:
@microsoft-github-policy-service agree company="Functional Software, Inc."

brianrob:
@jamescrosswell thank you for contributing this PR and for the detailed description of the scenario and change.  I agree that something along these lines is useful and necessary.  In fact, it's something that should exist for both ETW and EventPipe live sessions.  Ideally, we implement the same pattern for both, though it's possible that the underlying implementation may need to be slightly different.

Thinking about what you've written here and the code diff, I think it makes sense to have an opt-in API that users can call when they're ready to destroy existing state.  I had hoped that we could do something more automatic, but this just presents more complex lifetime trade-offs since we wouldn't have any way to know when it's truly safe to toss this state.

Question on your real scenario: Do you have a situation where you hold onto existing state (e.g. stack indices) and use them later to recall stacks, or would it technically be safe for you to call this on every callback?

I'm also thinking a bit about the API name - I think I'd like to name it something that can be more generic (e.g. `ClearSessionContext` or `TrimLiveSessionStructures`).  Do you have thoughts on this?  I'm kind of liking `TrimLiveSessionStructures`.

jamescrosswell:
> Question on your real scenario: Do you have a situation where you hold onto existing state (e.g. stack indices) and use them later to recall stacks

There was this (now shelved) PR building task/async profiling on ActivityComputer
- https://github.com/getsentry/sentry-dotnet/pull/3396#issue-2320667255

So having that historic data might have made it easier to solve some problems... but resolving a memory leak that is bringing down customer's servers whenever they have profiling enabled is obviously way more important, so we'll have to find another solution to that problem. 

> or would it technically be safe for you to call this on every callback?

Not quite safe, in our scenario.

Our sample handler resolves each stack synchronously inside the callback. It walks Caller() to the root, resolves, method names, and builds our own frame objects. 

However our builder keeps a Dictionary<int,int> keyed by CallStackIndex so a repeated stack isn't re-walked, and that cache lives for the duration of one profile (up to 30 s). After a reset, indexes are reissued starting from zero for different stacks, so a surviving cache entry would silently resolve to the wrong stack.

Our plan was to reset between profiles instead, where we'd also discard that cache.

So we don't retain indexes to recall stacks later, but we do use them as cache keys (same lifetime requirement).

> I'm also thinking a bit about the API name 

Happy to rename to something more generic, in case this grows to cover more structures.

No strong opinions there. Maybe `Trim` implies "make smaller" where `Clear` implies total clean out - so something like `ClearLiveSessionState` could work. I'm happy to go with TrimLiveSessionStructures if you think that's better though.

jamescrosswell:
> In fact, it's something that should exist for both ETW and EventPipe live sessions. Ideally, we implement the same pattern for both, though it's possible that the underlying implementation may need to be slightly different.

btw: @brianrob I did ask my LLM about this (it's a bit too deep for me with my knowledge of your code base, but might make sense to you):

`IncompleteStack` entries persist in `kernelStackKeyToInfo`/`userStackKeyToInfo` across events, and `EmitStackForEventIfReady` later passes the retained `UserModeStackIndex` as the caller index to `InternCallStackIndex` when splicing kernel frames on. A reset in between would intern onto a dead node silently — `GrowableArray`'s indexer only bounds-checks under `Debug`.

These would be challenges for a solution that worked with ETW as well...

brianrob:
Thanks for the context, @jamescrosswell.

Regarding the API name - Let's split the difference and go with `TrimLiveSessionState`.  I like the idea of going with trim since we're not clearing everything, we're only clearing what we can, and it allows us to trim in the future if we find we must keep more or keep some data.

On the ETW note - yes, the incomplete stack issue is real.  ETW emits kernel and user-mode stack fragments and we have to attach them.  If someone called the API in between them and we cleared these, then it would not behave properly.  We'd need to spend some time figuring out the right plan there, but still it would be opt-in.

One more thing - I saw in your PR comment that this new API is designed to be called from within a dispatched event.  I'm assuming this is to ensure that there isn't another thread mutating these structures.  I'm thinking we should have some enforcement of this.  We could do this by keeping track of the dispatcher thread ID and only allowing the call to succeed if the current thread ID matches.  This enforcement mechanism would likely be useful going forward, so we can build generically as opposed to naming it specific to this feature, and then just use it.  I think it would look something like:

m\_DispatchThreadID = Thread.CurrentThread.ManagedThreadId;

try

{

   realTimeSource.Dispatch...

}

finally

{

   m\_DispatchThreadId = -1;

}

Then, have an API ThrowIfNotRunningOnDispatchThread() that does the validation and call this from within the new Trim API.

src/TraceEvent/TraceLog.cs
```diff
@@ -975,6 +1006,47 @@ private void FlushRealTimeEventsNoLock(int minimumAgeMs)
...
+        /// <para>
+        /// Every <see cref="CallStackIndex"/> obtained before this call is invalidated and must not be
+        /// used afterwards, so only call this when no such index is still live.  Call stacks observed
+        /// after the call are interned from scratch, so no information is lost going forward - only
+        /// the ability to resolve indexes handed out earlier.
+        /// </para>
+        /// <para>
+        /// Must be called from within an event callback, on the thread dispatching that event.  Calling
+        /// it concurrently with event processing would race with call stack interning, so this is
+        /// enforced rather than merely documented.
```
jamescrosswell:
```suggestion
        /// it concurrently with event processing would race with call stack interning
```

jamescrosswell:
> Regarding the API name - Let's split the difference and go with `TrimLiveSessionState`.

Done in https://github.com/microsoft/perfview/pull/2452/commits/e4c85cc07ae9ca4c5a13c72084a77e137b73e075

> I saw in your PR comment that this new API is designed to be called from within a dispatched event. I'm assuming this is to ensure that there isn't another thread mutating these structures. 

Correct.

> I'm thinking we should have some enforcement of this. 

Done in https://github.com/microsoft/perfview/pull/2452/commits/17c6748db6319df2dc46c7c6026d6cb3c6801b55 - pretty much your code woven in.

brianrob (approved):
LGTM.  Thank you @jamescrosswell!

jamescrosswell:
Thanks heaps for your help with this @brianrob !

brianrob:
> Thanks heaps for your help with this Brian Robbins (@brianrob) !

Appreciate you bringing a well-written and documented PR.

PR #2455, opened 2026-08-17
Improve Stack Names For Dynamically Generated JIT Helpers

brianrob:
Updates ETLX conversion to recognize `JitHelper` method events from both live and rundown providers.

JIT helpers are associated with a process-specific synthetic module and displayed as:

```text
generatedruntimehelpers!<helper-name>
```

For example:

```text
generatedruntimehelpers!WriteBarrier
generatedruntimehelpers!VSD_ResolveStub
```

Non-helper method resolution remains unchanged.

### DIFF
--- a/src/TraceEvent/TraceLog.cs
+++ b/src/TraceEvent/TraceLog.cs
@@ -346,8 +346,8 @@ namespace Microsoft.Diagnostics.Tracing.Etlx
                 TraceProcess process = processes.GetOrCreateProcess(data.ProcessID, data.TimeStampQPC);
                 process.InsertJITTEDMethod(data.MethodStartAddress, data.MethodSize, delegate ()
                 {
-                    TraceManagedModule module = process.LoadedModules.GetOrCreateManagedModule(data.ModuleID, data.TimeStampQPC);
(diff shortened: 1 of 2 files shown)


src/TraceEvent/TraceLog.cs
```diff
@@ -4202,8 +4216,8 @@ int IFastSerializableVersion.MinimumReaderVersion
         private bool noStack;                               // This event should never have a stack associated with it, so skip them if we every try to attach a stack.
         private TraceThread thread;                         // cache of the TraceThread for the current event.
 
-        // TODO FIX NOW remove the jittedMethods ones.
-        private List<MethodLoadUnloadVerboseTraceData> jittedMethods;
+        // TODO FIX NOW remove the methodLoadEvents ones.
```
leculver, line 4219: Is this TODO FIX NOW still relevant?  If so, maybe a more in-depth comment about what's going here would be helpful.

leculver (approved):

src/TraceEvent/TraceLog.cs
brianrob: I don't think this is relevant anymore, but I need to check.  There are a lot of these, and I've been thinking about doing a pass over them to clean things up.  I'll plan to handle this as part of that pass.

PR #2468, opened 2026-09-15
Harden dynamic payload length parsing

msutara:
## Summary

Fix dynamic TraceEvent payload parsing for unsigned length prefixes, invalid offsets, and truncated fixed-count arrays. This prevents oversized counted strings and arrays from producing negative or out-of-bounds payload reads.

## Changes

- Preserve unsigned 16-bit and 32-bit string and array length prefixes.
- Use widened arithmetic before scaling UTF-16 lengths or narrowing array counts.
- Validate length prefixes, counted payloads, fixed-count arrays, and cached field offsets before unchecked reads.
- Add regression coverage for boundary values, malformed payloads, repeated field walks, and fixed/variable-length arrays.

## Testing

`dotnet test src\TraceEvent\TraceEvent.Tests\TraceEvent.Tests.csproj -c Debug --filter FullyQualifiedName~DynamicTraceEventParserVariableLengthFieldTests --no-restore`

Passed 143 tests on net462 and 143 tests on net8.0.


src/TraceEvent/TraceEvent.Tests/Parsing/DynamicTraceEventParserVariableLengthFieldTests.cs
```diff
@@ -0,0 +1,517 @@
...
+
+        private static int BytesPerCount(ushort size)
+        {
+            return (size & DynamicTraceEventData.IS_ANSI) == 0 && (size & DynamicTraceEventData.ELEM_COUNT) != 0 ? 2 : 1;
+        }
+
+        private static object ReadValue(DynamicTraceEventData traceEvent, DynamicTraceEventData.PayloadFetch fetch, int offset)
+        {
+            // Bypass the Debug-only pre-walk so it cannot mask a lookup-path regression.
+            return typeof(DynamicTraceEventData).GetMethod("GetPayloadValueAt", PrivateInstance)
```
brianrob: Do we need reflection here?  This sounds like it might be a similiar "avoid built-in validation" thing.

src/TraceEvent/TraceEvent.Tests/Parsing/DynamicTraceEventParserVariableLengthFieldTests.cs
```diff
@@ -0,0 +1,517 @@
...
+
+            WithEvent(payload, new[] { fetch }, traceEvent =>
+            {
+                Assert.Throws<ArgumentOutOfRangeException>(() => traceEvent.OffsetOfNextField(ref fetch, -31470, payload.Length));
+            });
+        }
+
+        [Fact]
+        public void PayloadValue_NegativeCachedOffset_ReturnsError()
+        {
```
brianrob: I suspect this test is not needed.  If I understand correctly, this is using reflection to test defense in-depth code specifically since it's no longer possible to get to this state anymore with your fix.  If so, let's remove this test, since I'd much rather not have reflection that can break in unexpected and unclear ways.

brianrob (commented):
Thank you very much for contributing this.  Overall, looks great.  A couple of questions specific to the tests.

src/TraceEvent/TraceEvent.Tests/Parsing/DynamicTraceEventParserVariableLengthFieldTests.cs
msutara: Good question. I removed reflection and kept `GetPayloadValueAt` private. To retain lookup-path coverage in Debug, the tests add a later fixed-offset field so the Debug pre-walk anchors there and `PayloadValue(0)` exercises the target lookup. Coverage confirms all 29 executable lines added by this PR and every changed condition are covered.

src/TraceEvent/TraceEvent.Tests/Parsing/DynamicTraceEventParserVariableLengthFieldTests.cs
msutara: Agreed. I removed the test that injected an unreachable negative cache state through private fields. The production negative-offset guard remains covered without private-state manipulation, and the full changed-line coverage remains complete.

src/TraceEvent/TraceEvent.Tests/Parsing/DynamicTraceEventParserVariableLengthFieldTests.cs
```diff
@@ -0,0 +1,497 @@
...
+            var element = new DynamicTraceEventData.PayloadFetch(0, 1, typeof(byte));
+            ushort size = (ushort)(LengthPrefixedArray | (widePrefix ? DynamicTraceEventData.BIT_32 : 0));
+            return DynamicTraceEventData.PayloadFetch.ArrayPayloadFetch(offset, element, size);
+        }
+
+        private static DynamicTraceEventData.PayloadFetch[] LookupFetches(DynamicTraceEventData.PayloadFetch target, int payloadLength)
+        {
+            Assert.InRange(payloadLength, 1, ushort.MaxValue);
+            // A later fixed offset lets Debug validation skip traversal of the target before its lookup is tested.
```
brianrob: This comment is hard to understand without knowing about `PayloadValue`'s Debug-only pre-walk. Please make the intent explicit, for example: `// Append a fixed-offset sentinel so PayloadValue's Debug-only full-payload validation does not traverse target; this ensures the test exercises target through GetPayloadValueAt.`

brianrob (commented):
Looks good.  One follow-up and then I think we're good.

src/TraceEvent/TraceEvent.Tests/Parsing/DynamicTraceEventParserVariableLengthFieldTests.cs
msutara: Updated the comment as suggested to explain the `PayloadValue` Debug-only pre-walk and why the fixed-offset sentinel ensures the test reaches `GetPayloadValueAt`.

brianrob (approved):
LGTM.  Thanks!

